PR #38, opened 2022-06-21
Add Platform.getSystemCharset() API

iloveeclipse:
With https://openjdk.java.net/jeps/400 implemented in Java 18,
"file.encoding" system property became meaningless and can't be used
anymore to determine system native encoding.

Unfortunately, that property was widely used in Eclipse API's and was
the standard way to get default system encoding. So both
org.eclipse.ui.WorkbenchEncoding.getWorkbenchDefaultEncoding() and
org.eclipse.core.resources.ResourcesPlugin.getEncoding() were using this
property and need now a proper replacement.

The new API tries first to get the value of the "native.encoding"
property (populated by Java 17+), and if not there, uses internal
"sun.jnu.encoding" property (used in all supported Java versions).
In case neither property is set, Charset.defaultCharset() is used as
fallback solution.

See https://github.com/eclipse-platform/eclipse.platform.resources/issues/154

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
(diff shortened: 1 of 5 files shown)


jukzi (approved):
LGTM

PR #54, opened 2022-07-18
Simplify set up and use of ConfigurationSessionTestSuite

HannesWell:
Provide a new method 'addMinimalBundleSet()' that adds all bundles from the `MINIMAL_BUNDLE_SET` to the suite. It obtains the bundles using a class from each bundle. This has the advantage that this method does not rely on fixed bundle names that might change in the future and therefore simplifies migration.
This has the second positive effect that the `o.e.core.tests.harness` plug-in requires all bundles from the `MINIMAL_BUNDLE_SET` and therefore automatically pulls them into the test-runtime. This avoids the need for users to add them 'manually', which also requires adjustments if a bundle name changes.

Additionally simplify `PlatformURLSessionTest` and `ConfigurationSessionTestSuite` itself.

If we use the new `addMinimalBundleSet()` everywhere the `MINIMAL_BUNDLE_SET` field could be removed.
Or is this Plug-in part of the public distribution and is not only internal?


akurtakov:
Or no one would have noticed that possibility and the change wouldn't have happened at all. Incremental improvements are the way to go!

HannesWell:
> Or no one would have noticed that possibility and the change wouldn't have happened at all. Incremental improvements are the way to go!

Sure. I meant if we would have noticed and changed before. But now we improve now. Not too bad either. :)

runtime/tests/org.eclipse.core.tests.harness/META-INF/MANIFEST.MF
```diff
@@ -8,11 +8,18 @@ Export-Package: org.eclipse.core.tests.harness,
  org.eclipse.core.tests.session,
  org.eclipse.core.tests.session.samples
 Require-Bundle: org.junit,
- org.eclipse.test.performance;resolution:=optional,
- org.eclipse.core.runtime
+ org.eclipse.test.performance,
```
laeubi, line 11: I'm missing a 3.13 version increase if new API is added here...

HannesWell: Done with https://github.com/eclipse-platform/eclipse.platform/pull/57.

PR #56, opened 2022-07-19
Try out the parallel build option

laeubi:


laeubi:
@iloveeclipse @sravanlakkimsetti parallel execution now work!

Non parallel execution time on GH before 13min versus 9min now, not that bad ...

laeubi:
Jenkins decreased from 14m to 11m we might want to ask Eclipse Infra for more processing power :-)

PR #61, opened 2022-07-24
Ensure build-timestamp is used as build-qualifier for o.e.sdk and o.e.platform

HannesWell:
In the `eclipse-platform-parent/pom.xml` the `format` configuration-property of the `tycho-packaging-plugin` is set to [`<format>${qualifier.format}</format>`](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/0288585a3dc80f4be45bfb4c2b29ed6977dc4912/eclipse-platform-parent/pom.xml#L478). Therefore `format` has already a value assigned and the property `tycho.buildqualifier.format`, which is new in Tycho 3.0.0 is not considered while populating the plugin's property values.

This re-enables to use `tycho.buildqualifier.format` in eclipse.platform build but I think it would be the cleaner way to replace the configuraiton-property `<format>${qualifier.format}</format>` in the platform-parent pom by a property `<tycho.buildqualifier.format>${qualifier.format}</tycho.buildqualifier.format>`.

Then we could simply just revert https://github.com/eclipse-platform/eclipse.platform/pull/60 and it should work.

To verify if this change works, one can check the build-logs and if the following output indicates a build-qulaifier for the o.e.sdk|platform plugins that is at the time when the build started (and not the time of the commit), it works:
```
[INFO] --- tycho-packaging-plugin:3.0.0-SNAPSHOT:build-qualifier (default-build-qualifier) @ org.eclipse.sdk ---
[INFO] The project's OSGi version is 4.25.0.v20220724-1049
```

But since I'm on vacation for two weeks from tomorrow on I don't want to start a change now that maybe could cause some more trouble that has to be fixed quickly.


mickaelistria (dismissed):
> build-qulaifier for the o.e.sdk|platform plugins that is at the time when the build started (and not the time of the commit),

It's actually not what is desired. What is desired is that the qualifier is the highest of latest commit that affected the feature and of all other qualifiers of included versions/plugins.
See https://wiki.eclipse.org/Tycho/Reproducible_Version_Qualifiers#Feature_version_qualifiers

iloveeclipse:
> It's actually not what is desired. 

Sure it is. Every new SDK build should have proper build id.

> What is desired is that the qualifier is the highest of latest commit that affected the feature and of all other qualifiers of included versions/plugins.

For other bundles, yes, but not for the bundle that contains the build mapping properties

HannesWell:
> I think it would be the cleaner way to replace the configuraiton-property `<format>${qualifier.format}</format>` in the platform-parent pom by a property `<tycho.buildqualifier.format>${qualifier.format}</tycho.buildqualifier.format>`.

Suggestion for that is created with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/448.

Alternatively we can keep the platform-parent pom unchanged and simply use the qualifier.format property here. I have done that with the second commit. If we agree to apply the change in the mentioned aggr-PR that second commit can be removed.

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/448 the fix of https://github.com/eclipse-platform/eclipse.platform/pull/60 can be just reverted.

The build logs indicate that now the build-timestamp is used as qualifier for the special projects as indented, but I will check tomorrows I-build for it.
```
[INFO] --- tycho-packaging-plugin:3.0.0-SNAPSHOT:build-qualifier (default-build-qualifier) @ org.eclipse.sdk ---
[INFO] The project's OSGi version is 4.25.0.v20220809-1530
```

Nevertheless thank you @iloveeclipse for the intermediate fix!

HannesWell:
The buildId in the about dialog looks fine. But I noticed that the `buildId` is not displayed on the splash-screen. Is this intentional?

iloveeclipse:
The splash before workspace selection doesn't have it, the progress after does. I haven't verified if that was the case before, if yes, we are fine.

HannesWell:
> The splash before workspace selection doesn't have it, the progress after does. I haven't verified if that was the case before, if yes, we are fine.

Indeed. I oversaw this. Just checked an older I-Build and that was the behavior before as well. So everything is fine.

Issue #62, opened 2022-07-28
NPE in InternalPlatform.getLog(Bundle) on startup


iloveeclipse:
Saw this error reported two times on startup of the IDE, I assume bad race conditions
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.equinox.log.ExtendedLogReaderService.addLogListener(org.osgi.service.log.LogListener, org.eclipse.equinox.log.LogFilter)" because "logReader" is null
	at org.eclipse.core.internal.runtime.InternalPlatform.lambda$11(InternalPlatform.java:386)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
	at org.eclipse.core.internal.runtime.InternalPlatform.getLog(InternalPlatform.java:381)
	at org.eclipse.core.runtime.Plugin.getLog(Plugin.java:239)
	at de.tobject.findbugs.FindbugsPlugin.logMessage(FindbugsPlugin.java:591)
	at de.tobject.findbugs.FindbugsPlugin.logException(FindbugsPlugin.java:549)
	at de.tobject.findbugs.builder.FindBugsWorker.updateBugCollection(FindBugsWorker.java:373)
	at de.tobject.findbugs.builder.FindBugsWorker.work(FindBugsWorker.java:228)
	at de.tobject.findbugs.builder.FindBugsBuilder$StartedFromBuilderJob.runWithProgress(FindBugsBuilder.java:219)
	at de.tobject.findbugs.FindBugsJob.run(FindBugsJob.java:142)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

Looks like the regressed code was last changed in https://bugs.eclipse.org/bugs/show_bug.cgi?id=573396 and it was supposed to fix race conditions :-).

laeubi:
Well the problem seems that the log service itself is not available at the call of that method, so any synchronization is a bit out of luck, thats the price you pay for using static methods that they might suffer from startup order ;-)

No idea if it would be suitable to simply skip the registration in such a case?

iloveeclipse:
Haven't analyzed the code yet, just one thing to mention : this was the first time I saw that error, and that in our "big" workspace after moved from 4.21 to 4.25 as IDE.

Do we have different log service init code now? How a log service can be *not* started at this point? Shouldn't the code wait for service to start?

laeubi:
This is the code:

[...]

PR #63, opened 2022-07-28
If platform is not running, always return an anonymous logger

laeubi:
Currently a logger is returned (and cached) regardless of the platform
is (still) running, that is the OSGi trackers are open.

This now guards the access to loggers, check for null in cases where
still the service is not running anymore, clears the loggers on shutdown
and return an anonymous logger in cases where platform is shut down.

Fix https://github.com/eclipse-platform/eclipse.platform/issues/62

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.net.URL;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import org.eclipse.core.internal.preferences.exchange.ILegacyPreferences;
 import org.eclipse.core.internal.preferences.exchange.IProductPreferencesService;
 import org.eclipse.core.internal.preferences.legacy.InitLegacyPreferences;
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@laeubi : i'm not an OSGI or service expert, so don't expect any qualified review on this PR.
It sounds as it would work :-)
I home @tjwatson can review that.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -745,6 +752,13 @@ private DebugOptions getDebugOptions() {
 	}
 
 	private void closeOSGITrackers() {
+		ExtendedLogReaderService logReader = logReaderTracker.getService();
+		logs.values().removeIf(log -> {
```
tjwatson, line 756: Probably shouldn't even do this stream if `logReader` is null.

A stream with `removeIf` seems like an expensive way to do this.  Why not do this?
```java
ExtendedLogReaderService logReader = logReaderTracker.getService();
if (logReader != null) {
  logs.values().forEach((log) -> logReader.removeLogListener(log));
}
logs.clear();
```

laeubi: > A stream with `removeIf` seems like an expensive way to do this. Why not do this?

Not sure if we really safe anything here:

1. logReader should be not null in almost all cases
2. iterate first and then clear seem not much less "expensive" as we are shut down and there are not millions of bundles having a logger anyways...

tjwatson: This is clearly a style discussion.  I think only doing a single null check is better on its own, but I also think the 5 lines of code is more simple and doing a single `clear()` call is more optimal than removing each element individually.

Is it going to make any measurable difference, I doubt seriously it.  But I still think it is better.

laeubi: Done!

tjwatson (approved):
LGTM

iloveeclipse (approved):

PR #69, opened 2022-08-03
Re-enable building team examples.

akurtakov:

PR #70, opened 2022-08-03
Enable the teams examples

laeubi:
FYI @akurtakov @vogella @iloveeclipse 

Actually here are some warnings disabled (I used that from the original team pom xml) but the build still complains if I urn this locally: https://github.com/eclipse-platform/eclipse.platform/blob/564174870d375d2dc773d31b68bf40c04602dc97/team/pom.xml#L22


akurtakov:
Done via https://github.com/eclipse-platform/eclipse.platform/pull/69

PR #76, opened 2022-08-11
Merge resources repository

laeubi:


laeubi:
@vogella Jenkins CI succeeds, but we need to disable the branchprotected until this is merged can you please:

- [ ] Disable branch protection
- [ ] Merge this PR if you are fine with it
- [ ] Enable branch protection

iloveeclipse:
Considering previous experience, could you please try aggregator build locally **before** merge?

How long will be the tests afterwards?

laeubi:
> Considering previous experience, could you please try aggregator build locally **before** merge?

Aggregator build requires that this is merged before it can be adjusted because it always consumes code from the current master branch.

> How long will be the tests afterwards?

You mean CI build? This build took `18min 56s` based on the single run, current master build took between `13min 16s` and `14min 52s` based on the last four runs.

iloveeclipse:
> Aggregator build requires that this is merged before it can be adjusted because it always consumes code from the current master branch

Not sure I understand this. The git repo is local. You don't have to **clone** on every aggregator build.



> Any specific problem you are concerned about?

SDK build will fail and again multiple commits annd a day will be spent trying to get it working again.

laeubi:
I have now created a PR here https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/471 that should point to my forks on github for the aggregator build

vogella:
Merge commits are now allowed, I currently not available to test / review but if you are confident please go ahead

laeubi:
At least I can build the aggregator when I manually replace the directories, no idea why it does not work with the submodules replacing by other references, so I think we can go ahaead...

laeubi:
@vogella merge commits can be disabled now

akurtakov:
Combined repo has 21 mins build time, compared to 14 mins prior to that. Still manageable time.

laeubi:
@merks can you adjust the setup file for this?

laeubi:
@merks I can see your commit here: https://github.com/eclipse-platform/eclipse.platform/commit/7f8f96dd4122ea324e8b9ea650979258f634d268 that seems from the resources repo, for some reason it seems the move into the sub-folder was not recorded by git :-\

iloveeclipse:
We have a big problem: I don't know how that merge was done, but there is NO GIT HISTORY MORE at all, on any resource related code.

iloveeclipse:
Ideally this should be reverted and re-done again with *preserved* history.

laeubi:
> Ideally this should be reverted and re-done again with _preserved_ history.

If anyone knows how.. according to https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history it "should" work automatically and if it do not work one can do little about it, so if there is any git expert around (I'm not) to help here that would be good.

I will go back to the last merge and see if I did something different hat could be applied here too...

iloveeclipse:
What I do to confirm "right"  merge is to check if git history for a specific file shows more than one commit.
This seem to be the case for merged platform runtime code,  but not for resources. There is only one entry in git history about merge commit.

laeubi:
> This seem to be the case for merged platform runtime code

If I look at https://github.com/eclipse-platform/eclipse.platform/commits/master/runtime/bundles/org.eclipse.core.contenttype/META-INF there also the history just starts with the move to a new location.

Git has a `--follow` option https://git-scm.com/docs/git-log#Documentation/git-log.txt---follow don't know if EGit allows enabling this?

iloveeclipse:
> there also the history just starts with the move to a new location.

This is a directory. Please check same for the MANIFEST.MF.

Note: same issue is with the "team".

So in sum:
1) runtime code is OK (for single files only, but that's OK) - was done by @vogella 
2) team + resources lost all history - was done by @laeubi 

@vogella : you seem to use **different** command to merge repositories as @laeubi, or uses different git version.
Please share exact way you used.

I think we should open a new issue for that.

PR #110, opened 2022-08-13
Remove unnecessary aggregator .project and .gitignore

HannesWell:
The `resources/examples/.project` file is not necessary because the contained projects are imported one by one.

Furthermore remove .gitignore files whose entries are already contained in a parent git-ignore.


merks:
This .project too required a workaround in the setup, but you are all-knowing.  🧙

HannesWell:
> This .project too required a workaround in the setup, but you are all-knowing. 🧙

Yes my Crystal ball told look into this. 🔮

But now I have to know why the build fails on Jenkins while resolving the TP.
The master is failing too, but it just worked in the other PR. And GH works as well. Can this be an Eclipse-Infra issue?

HannesWell:
> 
> But now I have to know why the build fails on Jenkins while resolving the TP. The master is failing too, but it just worked in the other PR. And GH works as well. Can this be an Eclipse-Infra issue?

It indeed was. Jenkins is now running, but since GH already succeeded we don't have to wait for that.

akurtakov:
It is quite likely to be caused by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/9ff5c43f463d67a7769c2c1ca2db475aaed08098 and automated deploy of target. This week we have 2 builds per day so hopefully will be resolved once the running one is done.

PR #113, opened 2022-08-17
Create README.md

jukzi:


mickaelistria:
This would require to update the file and maintain it as we do move content into this repository. Can we try to get some more abstract and more general content that wouldn't require updates?
I'm actually not sure a README is so helpful in general for this repo. Maybe just linking to organization README file would be enough?

jukzi:
the main pain is that there is no Readme at all currently. while there have been readmes in the repos moved. for every repo moved the readme in the subfolder should be deleted and integrated in the main readme.

akurtakov:
I'll merge this one and enhance it further to link to the organization one too.

Issue #111, opened 2022-08-16
slow BuildManager DeltaDataTree computations


jukzi:
I sampled my eclipse while i was programming in a big workspace with many projects and see half the time during autobuild is spend below org.eclipse.core.internal.events.BuildManager$DeltaCache.computeIfAbsent(). That feels wrong: the actual building should take longer then computing what to build.
![image](https://user-images.githubusercontent.com/51790620/184820766-b32914e9-bb88-4ec3-bfd6-1c16cefd0cd9.png)

Hotspots are in AbstractDataTreeNode, DeltaDataTree:
![image](https://user-images.githubusercontent.com/51790620/184822438-3c4aa88d-a326-499b-b00a-277905113e20.png)
example calltree:
![image](https://user-images.githubusercontent.com/51790620/184822971-b42f35b6-6cee-4a3c-9267-34ec59bf29c4.png)
![image](https://user-images.githubusercontent.com/51790620/184823178-519ee927-f60a-4d32-852b-70852b0a54da.png)

This is just a reminder that there may be possible optimizations: To compute if a single project has a delta it is wasted time to compute the delta for other projects too. Also to compute if a projects needsBuild() it could exit early as soon as any delta is found instead of actually computing the whole delta.
![image](https://user-images.githubusercontent.com/51790620/184822563-05a3f06d-f8f0-4944-9f19-d2f4d8ff796d.png)

vogella:
Moved to platform, as resources has been merged with eclipse.platform repo.

jukzi:
> Moved to platform, as resources has been merged with eclipse.platform repo.

Please disable issues and pull requests in the /settings of the old repos.

vogella:
> > Moved to platform, as resources has been merged with eclipse.platform repo.
> 
> Please disable issues and pull requests in the /settings of the old repos.

The repo may still be used for maintenance work (IIRC some IBM committers do this sometimes) so issues and pr needs to stay activated.

PR #114, opened 2022-08-19
BuildManager: compute delta only for relevant projects #111

jukzi:
Previously the delta for the whole workspace was calculated even though
it was only checked for few projects. That caused a large overhead for
Workspaces with a lot projects.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
@@ -23,8 +23,6 @@ import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import org.eclipse.core.internal.dtree.AbstractDataTreeNode;
-import org.eclipse.core.internal.dtree.DeltaDataTree;
 import org.eclipse.core.internal.resources.*;
 import org.eclipse.core.internal.resources.ComputeProjectOrder.Digraph;
 import org.eclipse.core.internal.utils.Messages;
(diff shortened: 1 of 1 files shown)


jukzi:
@jhonnen 
The computed delta of the whole workspace was never relay used - only for the relevant projects.
What's the (big) difference between DeltaDataTree.compareWith and DeltaDataTree.forwardDeltaWith? is it relevant?
I did not test this at all yet - just to get a idea and a CI test result

jhonnen:
> What's the (big) difference between DeltaDataTree.compareWith and DeltaDataTree.forwardDeltaWith? is it relevant?

It probably has something to do with that part

```
	 * There is no requirement that a and b be related, although it is usually more
	 * efficient if they are. The node keys are used as the basis of correlation
	 * between trees.
	 *
	 * Note that if b is already represented as a delta over a,
	 * then c will have the same internal structure as b.
	 * Thus the very common case of previous forwardDeltaWith: current
	 * is actually very fast when current is a modification of previous.
```
... but I don't know what that means exactly, I'm not familiar with the delta computation.

jukzi:
unrelated OrderedLockTest.testLockRequestDisappearence failed

jukzi:
The current "solution" is sometimes slower then the previous code.

jukzi:
There is something wrong in DeltaDataTree.basicCompare: the basicCompare loops over all tree.parent and calls searchNodeAt which also loops over the same trees. As the trees grows this is an O(n^2) for n trees:
image
The outer loop should continue at the tree where the inner loop found a result to avoid handling the same node over and over.

jukzi:
... or the inner loop should not loop over the trees too :-)

jukzi:
Performance for my workspace with ~500 projects:
Each needsBuild() did cost about  ~00ms which is reduced to ~1ms.
The total full autobuild (upon changing the manifest) is reduced from 550sec to 420sec.
before:
image
after:
image

It has a slight overhead if there is no delta, but that is irrelevant as both implementations are extremely quick (<1ms) when there is no delta.

In total it is a significant performance gain for autobuild with many projects in workspace after https://github.com/eclipse-platform/eclipse.platform/pull/186 was fixed

iloveeclipse:
Please check https://github.com/eclipse-platform/eclipse.platform/issues/243 for a possible regression.

PR #125, opened 2022-08-26
[win] ignore "Problem occurred in auto-refresh native code: 5." #124

jukzi:
WaitForMultipleObjects -> WAIT_FAILED -> GetLastError can return 5
(ERROR_ACCESS_DENIED) as described in:
https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects
https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-

Happens sometimes and can flood log without giving any value.

PR #127, opened 2022-08-30
fix random failing OrderedLockTest

jukzi:
* do not wait random time but use random order (faster)
* log thread dump on timeout (better diagnosis)
* execute test multiple times for (easier finding random failings)
* fixed waiting conditions.


jukzi:
The test failed two times in https://github.com/eclipse-platform/eclipse.platform/pull/125#issuecomment-1228547021

iloveeclipse:
This causes regression on all platforms, see https://github.com/eclipse-platform/eclipse.platform/issues/145.

Issue #128, opened 2022-08-30
failing TestWorkspaceEncodingExistingWorkspace.testExpectedEncoding1


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20220829-1800/testresults/html/org.eclipse.core.tests.resources_ep425I-unit-win32-java11_win32.win32.x86_64_11.html

```
junit.framework.ComparisonFailure: expected:<[Cp]1252> but was:<[windows-]1252>
at org.eclipse.core.tests.resources.session.TestWorkspaceEncodingExistingWorkspace.testExpectedEncoding1(TestWorkspaceEncodingExistingWorkspace.java:53)

```

jukzi:
can reproduce locally on windows.

HeikoKlare:
Test is still/again randomly failing, e.g., in this build: https://github.com/eclipse-platform/eclipse.platform/pull/1268/checks?check_run_id=23148172761

```
junit.framework.AssertionFailedError: expected:<UTF-8> but was:<windows-1252>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.Assert.assertEquals(Assert.java:86)
	at junit.framework.TestCase.assertEquals(TestCase.java:246)
	at org.eclipse.core.tests.resources.session.TestWorkspaceEncodingExistingWorkspace.testExpectedEncoding1(TestWorkspaceEncodingExistingWorkspace.java:55)
```

laeubi:
Could it be a timing issue so maybe just check in a loop if the encoding is like desired for some (milli) seconds e.g. 5 times with a wait of 200ms?

laeubi:
Java 21 has UTF-8 as a default file encoding on all platforms (and Since Java 17 there is a java.native.encoding instead). @merks and @iloveeclipse have worked on this in the past to mitigate the problem, but the original report mentions `<[Cp]1252> but was:<[windows-]1252>` (what are both possible windows (native) system encodings)... so probably the test itself is testing something that only holds true on some very specific configuration that can not be assumed in general.

merks:
Yes, the file.encoding is always UTF-8 by default for Java 21 whereas `native.encoding` has the native encoding.  There is this method now to get the org.eclipse.core.resources.ResourcesPlugin.getEncoding() method to get the native encoding.

[...]

PR #129, opened 2022-08-30
fix TestWorkspaceEncodingExistingWorkspace.testExpectedEncoding1() #128

jukzi:
expected:<[Cp]1252> but was:<[windows-]1252>
The Charset is the same, but not it's ambigous name.


jukzi:
@iloveeclipse thats your test.
I wonder if that failed test only recently.

akurtakov:
Looks good to me.

Issue #107, opened 2022-04-13
random failing org.eclipse.core.tests.resources.LinkedResourceTest testMoveFolder


jukzi:
invocation 1 failed with org.eclipse.core.internal.resources.ResourceException: Resource '/OtherExistingProject/nonExistingFolderInOtherExistingProject' already exists.

java.lang.AssertionError: invocation 1 failed with org.eclipse.core.internal.resources.ResourceException: Resource '/OtherExistingProject/nonExistingFolderInOtherExistingProject' already exists.
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:101)
at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:114)
at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:114)
at org.eclipse.core.tests.resources.TestPerformer.performTest(TestPerformer.java:58)
at org.eclipse.core.tests.resources.LinkedResourceTest.testMoveFolder(LinkedResourceTest.java:1684)

https://download.eclipse.org/eclipse/downloads/drops4/I20220412-1800/testresults/html/org.eclipse.core.tests.resources_ep424I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

jukzi:
still happens -  (why should it change if nobody fixes it?)

https://download.eclipse.org/eclipse/downloads/drops4/I20220829-1800/testresults/html/org.eclipse.core.tests.resources_ep425I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

laeubi:
> still happens - (why should it change if nobody fixes it?)

https://geekandpoke.typepad.com/geekandpoke/2009/07/the-art-of-bugfixing-chapter-2.html

mickaelistria:
Closing #130 did close this issue, but do I understand it right that #130 is more a patch to get additional debug info and is not a fix, so this issue should be reopen?

mickaelistria:
I suspect there is some similarity, and maybe even root causes with #95 , as it's also involving linked resource and project move...

PR #130, opened 2022-08-30
TestPerformer: append exception cause to error message - for #107

jukzi:
also
* fixed hiding warning.
* removed unused name


jukzi:
@akurtakov thanks for helping with all those technical stuff.

akurtakov:
> @akurtakov thanks for helping with all those technical stuff.

You're welcome.

PR #133, opened 2022-08-30
WIP Semaphore release check.

jukzi:
Semaphore may wait endlessly in acquire() when released multiple times

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/Semaphore.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/Semaphore.java
@@ -31,6 +31,12 @@ public class Semaphore {
 	 * and false otherwise.
 	 */
 	public synchronized boolean acquire(long delay) throws InterruptedException {
+		if (notifications == 0) {
+			// constructor already acquires Semaphore
+			// has to be released before requiring again or gets stuck in wait.
+			System.err.println("Semaphore already acquired"); //$NON-NLS-1$
+			throw new IllegalStateException("Semaphore already acquired"); //$NON-NLS-1$
+		}
 		if (Thread.interrupted())
 			throw new InterruptedException();
 		long start = System.nanoTime();
@@ -76,6 +82,11 @@ public class Semaphore {
 	}
 
 	public synchronized void release() {
+		if (notifications > 0) {
+			System.err.println("Semaphore already released"); //$NON-NLS-1$
+			throw new IllegalStateException("Semaphore already released"); //$NON-NLS-1$
+			// TODO just do not increment any further?
+		}
 		notifications++;
 		notifyAll();
 	}
--- a/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.jobs; singleton:=true
-Bundle-Version: 3.13.100.qualifier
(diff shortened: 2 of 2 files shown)


laeubi:
> Semaphore may wait endlessly in acquire() when released multiple times

Is it possible that a Lock is actually desired on those places than a semaphore?

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/Semaphore.java
```diff
@@ -31,6 +31,12 @@ public Semaphore(Runnable runnable) {
 	 * and false otherwise.
 	 */
 	public synchronized boolean acquire(long delay) throws InterruptedException {
+		if (notifications == 0) {
```
laeubi, line 34: This does not make sense, that way the process will **never** wait... because there are only two cases:

1. notifications = 0 (semaphore is blocked) --> throws exception
2. notifications = 1 and will return immediately

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/Semaphore.java
```diff
@@ -76,6 +82,11 @@ public int hashCode() {
 	}
 
 	public synchronized void release() {
+		if (notifications > 0) {
+			System.err.println("Semaphore already released"); //$NON-NLS-1$
```
laeubi, line 86: Actually a Semaphore holds a positive number of permits and thus can even has notifications = 5, as there is no way for the code to init the initial number of permits it seems that the intend is here to

1. construct the semaphore
2. call `release` the number of permits one likes to allow (e.g. 5 times)
3. now there should always be symmetric calls to aquire/release

laeubi (changes requested):
This change actually forces the `Sempahore` to work like a `Lock` (that is a semaphore with a single permit).

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/Semaphore.java
jukzi: thanks for the explanation, would be good if some javadoc explains something.

i just had a deadlocked VM and a thread was stuck in an endless wait in acquire with notifications=0 and i know nothing  how this is meant to work. so i just wanted to know what the CI says.

laeubi: I don't know *why* this class is there and *where* it is used, so I can only guess from the name, but actually java has also a sempahore class (it says since 1.5): https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html that has a rather good description about Sempahores...

So usually an endless wait on `notifications=0` means no one likes to release the permit (or the semaphore never has had any what would be quite strange).

jukzi: There is only two usages:
# OrderedLock.operations- as a Lock - like java.util.concurrent.locks.ReentrantLock(true)?
# JobManager - as a barrier  - like java.util.concurrent.CyclicBarrier ?

Back when eclipse was written those java classes may have not been existing in the jdk yet. But since there are standard classes now, i would favor the use of those.

laeubi: Sure standard java classes are often more complete, better tested and better integrate with other parts of the system and are likely better understood by other developers.

So as this is an internal class, as soon as all references are replaced, the class can be deleted.

jukzi:
I figured out my root problem was the debug mode - when i suspend the VM via breakpoint and afterwards try to resume the VM sometime not *all* Threads start again. There was a Threadpool worker thread not running (jstack showed the thread without stacktrace), where the other threads where waiting on. Sometimes the problem can be worked around by suspend+resume again multiple times.

Issue #95, opened 2022-04-29
random failing org.eclipse.core.tests.internal.alias.BasicAliasTest.testBug256837


jukzi:

jukzi:
can be locally reproduced with 

Thread.sleep(200) 
 at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(String, Throwable)

jukzi:
It is some multi threading problem: if the whole VM is suspended at org.eclipse.core.internal.resources.ResourceTree.standardMoveProject(IProject, IProjectDescription, int, IProgressMonitor)
 and only main thread is resumed, then the test passes.
if otherwise only the main thread is suspended there then the test fails.

jukzi:
I don't know how long it failed. It might only became obvious as hardware or software became faster or slower.

mickaelistria:
So indeed, there is a RefreshJob running at some point in another thread, and the success or failure depends on whether this job has completed before next instruction.
Here is the stack that triggers the RefreshJob (from main thread)
```
Thread [main] (Suspended (entry into method refresh in RefreshJob))	
	owns: RunnableLock  (id=115)	
	RefreshJob.refresh(IResource) line: 199	
	RefreshManager.refresh(IResource) line: 98	
	FileSystemResourceManager.asyncRefresh(IResource) line: 148	
	FileSystemResourceManager.read(IFile, boolean, IProgressMonitor) line: 833	
	File.getContents(boolean) line: 277	
	File.getContents() line: 268	
	File.getLineSeparator(boolean) line: 448	
	FileSystemResourceManager.internalWrite(IProject, IProjectDescription, int, boolean, boolean) line: 640	
	Project.writeDescription(IProjectDescription, int, boolean, boolean) line: 1421	
	Project.writeDescription(int) line: 1402	
	ResourceTree.movedProjectSubtree(IProject, IProjectDescription) line: 718	
	ResourceTree.standardMoveProject(IProject, IProjectDescription, int, IProgressMonitor) line: 1108	
	Project.move(IProjectDescription, int, IProgressMonitor) line: 993	
	BasicAliasTest.replaceProject(IProject, URI) line: 304	
	BasicAliasTest.testBug256837() line: 411	
[...]
```
In an ideal world, the "move" opreation wouldn't require to refresh the content, as we do already have some pointers to this content. It would be great to investigate that particular case: if we can get rid of the need to `refresh()` then it would fix this tests (and maybe other that are of the same kind).

[...]

PR #135, opened 2022-08-31
fix random failing BasicAliasTest.testBug256837 #95

jukzi:


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/alias/BasicAliasTest.java
```diff
@@ -410,6 +410,8 @@ public void testBug256837() throws CoreException {
 		// change the location of p2 project to the temp folder
 		replaceProject(p2, tempStore.toURI());
 
+		waitForRefresh(); // takes some milliseconds until aliasManager is notified
```
mickaelistria, line 413: What is the contract of AliasManager? Is it supposed to be async?
I'm afraid this test failure shows we actually have something broken in the synchronization.

jukzi: I don't know. If the AliasManager has any contract it is for sure broken multiple times: It does not even try to be thread safe and is NOT synchronized anywhere. There is totally no guarantee from JDK that it works.

what happens is that AliasManager.resourceChanged(IResourceChangeEvent) is called asynchronously and marks AliasManager.changedProjects = true . With a bit luck of having x86 memory architecture the main thread will see the change ;-)

This "fix" is the same as already  done for 
testBug232426, testBug358830, testCreateFolderInBackground, BasicAliasTest.testDeepLink, ...
which all suffer from the same problem.

I'd be happy if someone makes AliasManager thread safe + synchron.

jukzi:
unrelated
```
 junit.framework.AssertionFailedError: 1.2: 
/testBuildDeltaUsingRelaxedRuleBug343256[~]: {} expected:<1> but was:<0>
org.eclipse.core.tests.internal.builders.RelaxedSchedRuleBuilderTest.testBuilderDeltaUsingRelaxedRuleBug343256(RelaxedSchedRuleBuilderTest.java:353)

```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/alias/BasicAliasTest.java
mickaelistria: > what happens is that AliasManager.resourceChanged(IResourceChangeEvent) is called asynchronously

That actually surprises me, I was kind of expecting POST_CHANGE resource listeners to happen synchronously. But maybe I build wrong assumption based on lucky side-effects all those years.
Looking at the code, I do not see where the resourceChanged() invocation becomes async. I see

Project.move() directly calls
Workspace.endOperation(), direcly calls
Workspace.broadcastPostChange(), directly calls
NotificationManager.notificationManager.broadcastChanges(tree, event, true), directly calls
NotificationManager.notify(), directly calls
IResourceChangeListener.resourceChanged(), and for AliasManager, directly changes the value of `changedProjects`, which is then used by AliasManager.computeAliases()

So can you please enlighten as to why this AliasManager would react asynchronously here? I'm missing something.

jukzi: I do not know WHY ... i am already happy that after much debugging i found that it does - if and only if the test fails  - which is was hard to reproduce (delay first ResourceException.provideStackTrace in main thread).

Something runs in parallel and its a race condition

laeubi: First thing I do is usually put `Thread.dumpStack()` into method where I' curious why and from where it is called

```
synchronized(Object.class) {
  Sysout(Thread.currentThread().getName());
  Thread.dumpStack();
}
```

jukzi: BTW you do not need to change the source for such but can put it as condition in a breakpoint.

akurtakov:
@mickaelistria Can you help with making constantly failing test showing the bug?

mickaelistria:
> Can you help with making constantly failing test showing the bug?

I don't have an idea how to achieve that without adding artificial test only configuration to AliasManager.
However, I think the test should work even with the auto-refresh stuff, the aliasManager most likely needs some synchronization guards.

mickaelistria:
@jukzi I am spending more time on AliasManager, and I have some hope that if we do flag `changedProjects` as volatile, we might get help to reduce the problem here. What do you think? Would there be any risk or negative impact to check if we change that?

jukzi:
typically volatile would be a good step in the right direction. However in this particular case it's not enough . the  "if (changedProjects){ ...changedProjects=false}" needs to be atomic operation and the datastructures must no be iterated while changed. 
- all those improvements however do not relate to the problem in this test.

mickaelistria:
> all those improvements however do not relate to the problem in this test.

There is no problem in this test, there is a problem in AliasManager not being thread safe, and the test reveals the problem.

PR #146, opened 2022-09-05
Some code improvements in AliasManager

mickaelistria:
+ Stronger typing
+ reuse Java API classes
+ leverage streams

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
@@ -20,6 +20,7 @@ package org.eclipse.core.internal.resources;
 
 import java.net.URI;
 import java.util.*;
+import java.util.function.Consumer;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileStore;
 import org.eclipse.core.internal.events.ILifecycleListener;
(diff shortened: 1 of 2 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -55,29 +57,18 @@
...
-		void doit(IResource resource);
-	}
-
-	class FindAliasesDoit implements Doit {
+	class FindAliasesDoit implements Consumer<IResource> {
 		private int aliasType;
 		private IPath searchPath;
 
+		public FindAliasesDoit(IResource aliasResource) {
+			this.aliasType = aliasResource.getType();
```
jukzi, line 66: thats good, than you can declare the members final!

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -129,32 +113,17 @@ class LocationMap {
 		/**
 		 * Map of FileStore-&gt;IResource OR FileStore-&gt;ArrayList of (IResource)
 		 */
-		private final SortedMap<IFileStore, Object> map = new TreeMap<>(IFileStore::compareTo);
+		private final SortedMap<IFileStore, ArrayList<IResource>> map = new TreeMap<>(IFileStore::compareTo);
```
jukzi, line 116: The idea behind this Object was to either use single IResource or ArrayList<IResource> to safe memory in the case that the list is only 1 entry long. if you change that you should verify that the additional memory is neglectable.

mickaelistria: I removed this part and reduced the change to get tests passing.

PR #147, opened 2022-09-05
Refactor Dtree: rootkey, ElementNotFoundException

jukzi:
* AbstractDataTree.rootKey() is a constant
 => ROOT_PATHS is constant too
 => getter/setter for DeltaDataTree rootData
 => move detailed toString from ElementTree to DeltaDataTree
* inline elementNotFound to avoid "// can't get here"

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/AbstractDataTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/AbstractDataTree.java
@@ -213,8 +213,18 @@ public abstract class AbstractDataTree {
 	/**
 	 * Returns the key of the root node.
 	 */
-	public IPath rootKey() {
+	public static final IPath rootKey() {
 		return Path.ROOT;
+	}
+
+	public boolean isRoot(IPath key) {
+		return key == rootKey();
+	}
+
+	private static final IPath[] ROOT_PATHS = new IPath[] { rootKey() };
+
+	public IPath[] rootPaths() {
+		return ROOT_PATHS;
 	}
 
 	/**
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
@@ -13,9 +13,11 @@
  *******************************************************************************/
 package org.eclipse.core.internal.dtree;
 
+import java.util.Objects;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.StringPool;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.IPath;
 
 /**
  * Externally, a <code>DeltaDataTree</code> appears to have the same content as
@@ -56,11 +58,13 @@ public class DeltaDataTree extends AbstractDataTree {
 	 *	root node of new tree.
 	 */
 	public DeltaDataTree(AbstractDataTreeNode rootNode) {
+		Objects.requireNonNull(rootNode);
 		this.rootNode = rootNode;
 		this.parent = null;
 	}
 
 	protected DeltaDataTree(AbstractDataTreeNode rootNode, DeltaDataTree parent) {
+		Objects.requireNonNull(rootNode);
 		this.rootNode = rootNode;
 		this.parent = parent;
 	}
@@ -166,7 +170,7 @@ public class DeltaDataTree extends AbstractDataTree {
 		DeltaDataTree newTree;
 		if (this == other) {
 			newTree = new DeltaDataTree();
-			newTree.setData(Path.ROOT, new NodeComparison(null, null, 0, 0));
+			newTree.setRootData(new NodeComparison(null, null, 0, 0));
 		} else if (other.hasAncestor(this)) {
 			AbstractDataTreeNode assembled = other.searchNodeAt(path);
 			DeltaDataTree tree = other;
@@ -241,7 +245,7 @@ public class DeltaDataTree extends AbstractDataTree {
 		DeltaDataTree newTree;
 		if (this == other) {
 			newTree = new DeltaDataTree();
-			newTree.setData(Path.ROOT, new NodeComparison(null, null, 0, 0));
+			newTree.setRootData(new NodeComparison(null, null, 0, 0));
 		} else if (other.hasAncestor(this)) {
 
 			AbstractDataTreeNode assembled = other.getRootNode();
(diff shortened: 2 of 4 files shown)


mickaelistria:
There are some good changes in here (eg I like `element.setRootData()` which is less error-prone.
However, the fact that the root is statically constant seems more like an implementation detail. I believe one can easily imagine seeing 2 trees having a different root (although it doesn't seem useful in practice at the moment). What benefit do we earn in making this "detail" surface in the signatures? What is the concrete gain with this refactoring?

jukzi:
The only improvement is that ROOT_PATHS is final so it does not need temp memory.
In theory a subclass could overload getRoot() but since there is no subclass and never will be there is no use for having that configurable and it would probably not work to change the root anyway - since there are many other references to Path.ROOT that would be wrong then.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/AbstractDataTree.java
```diff
@@ -213,10 +213,20 @@ public boolean isImmutable() {
...
 	}
 
+	public boolean isRoot(IPath key) {
+		return key == rootKey();
+	}
+
+	private static final IPath[] ROOT_PATHS = new IPath[] { rootKey() };
+
+	public IPath[] rootPaths() {
+		return ROOT_PATHS;
```
iloveeclipse, line 227: Since array content is modifiable, any caller could per mistake change the array.
Not sure if this is a good idea to provide this method in the parent class.
I see it is only used in `org.eclipse.core.internal.watson.ElementTree.getChildIDs(IPath)`.
Why not to add that there, as private method and init it to the instance (not static) array?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
```diff
@@ -951,8 +963,17 @@ public void storeStrings(StringPool set) {
 	/** for debugging purposes only */
 	@Override
 	public String toString() {
-		return this.getClass().getSimpleName() + " " + Integer.toHexString(System.identityHashCode(this)) + " parent=" //$NON-NLS-1$ //$NON-NLS-2$
-				+ (parent == null ? null : Integer.toHexString(System.identityHashCode(parent))) + " rootNode=" //$NON-NLS-1$
-				+ rootNode.toShortString();
+		DeltaDataTree root = getParent();
```
iloveeclipse, line 966: Could you please provide before / after state? Why is this needed, what is better now?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/watson/ElementTree.java
```diff
@@ -206,7 +206,7 @@ public synchronized void createElement(IPath key, Object data) {
 		try {
 			tree.createChild(parent, key.lastSegment(), data);
 		} catch (ObjectNotFoundException e) {
-			elementNotFound(parent);
+			throw createElementNotFoundException(parent);
```
iloveeclipse, line 209: I wonder why don't we supply original exception as root cause?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
```diff
@@ -909,6 +913,14 @@ public void setData(IPath key, Object data) {
 		assembleNode(key, new DataDeltaNode(key.lastSegment(), data));
 	}
 
+	public void setRootData(Object data) {
+		setData(rootKey(), data);
+	}
+
+	public Object getRootData() {
```
iloveeclipse, line 920: Is this really an Object? I see below explicit cast to ElementTree

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/watson/ElementTree.java
```diff
@@ -733,14 +725,10 @@ public String toDebugString() {
...
+	/** for debugging purposes only */
 	@Override
 	public String toString() {
-		ElementTree root = getParent();
-		while (root != null && root.getParent() != null) {
-			root = root.getParent();
-		}
-		return "ElementTree(" + treeStamp + "->" + (getParent() == null ? null : getParent().treeStamp) + "..." //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-				+ (root == null ? null : root.treeStamp) + ")"; //$NON-NLS-1$
+		return "ElementTree(" + treeStamp + ")"; //$NON-NLS-1$ //$NON-NLS-2$
```
iloveeclipse, line 731: Why don't we need parent timestamp now?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/AbstractDataTree.java
jukzi: It's a general problem/pattern in eclipse code that the caller can modify arrays. the best solution would be to not use arrays at all. In this case if the caller modifies the error he would have a problem with ElementTree.ChildIDsCache.childPaths anyway. it's not the scope of this change to fix that

laeubi: I think the question is why a `public IPath[] rootPaths()` is needed at all...

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
jukzi: somewhere in BuildManager.needsBuild() i wanted information about the depth- which according to the ElementTree javadoc may be the root error to my problems - if it grows. But where to write the depth? ElementTree  or DeltaDataTree? as i found that that tree is actually stored in the DeltaDataTree i think the information belongs there.
before:
image

after:
image

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/watson/ElementTree.java
jukzi: as it never happens anway?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
jukzi: the ElementTree only uses it as ElementTree, but there are other Trees that use it differently :-(
org.eclipse.core.internal.dtree.DeltaDataTree.basicCompare(DeltaDataTree, IComparator, IPath) and compareWith set NodeComparison as rootData

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/watson/ElementTree.java
jukzi: (see screenshot) that tree information is now available in it's DeltaDataTree (ElementTree.tree)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/AbstractDataTree.java
laeubi: It seems the current state of the file do not contain any "rootPaths()" method so what has this to do with java 9? Even pre java 9 there is no one forcing you to return arrays ;-)

PR #150, opened 2022-09-05
574883fix

jukzi:


jukzi:
@iloveeclipse this commit is to show that extending the lock solves the problem - i hope you have a idea for a better implementation

PR #153, opened 2022-09-07
synchronize Job Notification to fix join() (fixes #148)

### LINKED ISSUE #148: Bug 574883 - Job.getJobManager().join(family) doesn't wait for a re-scheduled job
@iloveeclipse 
the problem is that join() uses non-synchronized jobListeners :

JobManager.endJob(InternalJob, IStatus, boolean) notifies 
1. "done" and then 
2. "scheduled".

Between the two calls it looks to the listener like there is no more job scheduled.

A way to fix this might be to extend the "synchronized (lock)" block in endJob() around "done" and "schedule".  
However somehow it sounds also dangerous/unwanted to synchronize around 3rd party calls.
JobManager.schedule() and JobManager.nextJob() also synchronize on lock so that those would wait till listeners returned.

jukzi:
Bug 574883 - Job.getJobManager().join(family) doesn't wait for a
re-scheduled job

* Add and use a Job specific lock to make sure the notification that a
job stopped is send before the job is reported to be started again
(in another thread).

* JobManager.schedule() - when called from endJob() - is called within
the synchronized(lock)) instead of using a second (non atomic)
synchronized(lock) afterwards. The notification is still send outside
that synchronized(lock).

Previously it was possible that the Job notifications was send and
processed in different threads and - since not synchronized - in
arbitrary order. Especially a job
could restart in another worker thread before the notification was
processed that previous job execution was done.
Still the events are processed in different threads, but the
notification order is fixed.

=> The same Job will not start again before all listeners noticed that
the Job was finished. That is also wanted for other Listeneres then the
join() Listener since the Listeners should be able to decide if the Job
should start again (see JobTest.testCancelFromAboutToRun())

The benefit is proven by JUnit Test Bug_574883.testJoinLambdaOften().
Basic functionality is tested by JobTest.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.jobs; singleton:=true
-Bundle-Version: 3.13.100.qualifier
+Bundle-Version: 3.13.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.core.internal.jobs;x-internal:=true,
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -124,7 +124,7 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	 * or -1 if the job should not be rescheduled.
 	 * @GuardedBy("manager.lock")
 	 */
-	private long startTime;
+	private volatile long startTime;
 
 	/**
 	 * Stamp added when a job is added to the wait queue. Used to ensure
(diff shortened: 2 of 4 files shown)


jukzi:
damn now notification can deadlock. as happens in org.eclipse.core.tests.runtime.jobs.JobGroupTest.testShouldCancel_4() where endJob notifies JObGroupTest which tries to cancel. 
image

jukzi:
> Don't know if such a thing exits, but for me it seems it would be useful to have a Flow-Chart

I would prefer if the events happen in a intuitive order instead of this &/%&%" - as in the example the same Job is ending two times in parallel. Need to find a way that a job is not started again before all its events have been processed.

laeubi:
> I would prefer if the events happen in a intuitive order instead of this &/%&%"

But what is intuitive order? That's what I mean, should one really expect some "order" of events, or are these actually only notifications?

> as in the example the same Job is ending two times in parallel. Need to find a way that a job is not started again before all its events have been processed.

That's another point that needs to be decided is it *valid* to schedule a job twice? If not one could simply skip the schedule if the job is marked for scheduling already...

iloveeclipse:
> That's another point that needs to be decided is it _valid_ to schedule a job twice

No need to be decided anything, it **is** valid.

laeubi:
> > That's another point that needs to be decided is it _valid_ to schedule a job twice
> 
> No need to be decided anything, it **is** valid.

Sorry I wanted to write 'describe' instead, but in that case it is valid what was tried to be fixed here and actually only the client code might decide if it should schedule the job?

jukzi:
Two hours ago the test did not fail on my local computer. now they reproducible fail. its driving me nuts.
org.eclipse.core.internal.jobs.JobManager.now()
 "can only be used to compare it with another value returned from this function,"
but what it is compared against?
InternalJob.T_NONE,
delay in JobManager.schedule(I)
org.eclipse.core.internal.jobs.InternalJob.T_INFINITE



AAAAAAAAAAAAAAAAAAAAAAAAA

jukzi:
Please review. Bug_574883.testJoinLambdaOften still fails but in average after much more iterations on my computer

laeubi (approved):
Looks good to me.

PR #154, opened 2022-09-07
Apply autoformatting to AliasManager

mickaelistria:


mickaelistria:
Build failure is not related, and seems caused by a transtive dependency (org.eclipse.equinox.feature) update. Let's merge.

PR #156, opened 2022-09-07
Make AliasManager more thread-safe

mickaelistria:
* Use a semaphore to avoid concurrent modification of the internal model
* Factorize the check/update of internal model in a single entry point
* Use thread-safe structures to accumulate data about what needs to be updated

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
@@ -21,6 +21,7 @@ package org.eclipse.core.internal.resources;
 import java.net.URI;
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileStore;
@@ -282,17 +283,17 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 	 * invalidate the locations map or aliased projects set.  These will be
 	 * updated incrementally on the next alias request.
 	 */
-	private final Set<IResource> changedLinks = new HashSet<>();
+	private final Set<IResource> changedLinks = ConcurrentHashMap.newKeySet();
 
 	/**
 	 * This flag is true when projects have been created or deleted and the
 	 * location map has not been updated accordingly.
 	 */
-	private boolean changedProjects = false;
+	private volatile boolean changedProjects = false;
 
 	/**
-	 * This maps IFileStore -&gt; IResource, associating a file system location
-	 * with the projects and/or linked resources that are rooted at that location.
+	 * This maps IFileStore -&gt; IResource, associating a file system location with
+	 * the projects and/or linked resources that are rooted at that location.
 	 */
 	protected final LocationMap locationsMap = new LocationMap();
 	/**
@@ -473,11 +474,7 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 					removeFromLocationsMap(link, link.getStore());
 				//fall through
 			case LifecycleEvent.PRE_FILTER_ADD :
-				changedLinks.add(event.resource);
-				break;
 			case LifecycleEvent.PRE_FILTER_REMOVE :
-				changedLinks.add(event.resource);
-				break;
 			case LifecycleEvent.PRE_LINK_CREATE :
 				changedLinks.add(event.resource);
 				break;
@@ -504,18 +501,10 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 		boolean noAliases = !aliasedProjects.contains(project);
 
 		//now update any structure changes and check again if an update is needed
-		if (hasStructureChanges()) {
-			updateStructureChanges();
+		if (checkStructuralChanges()) {
 			noAliases &= nonDefaultResourceCount <= 0 || !aliasedProjects.contains(project);
 		}
 		return noAliases;
-	}
-
-	/**
-	 * Returns whether there are any structure changes that we have not yet processed.
-	 */
-	private boolean hasStructureChanges() {
-		return changedProjects || !changedLinks.isEmpty();
 	}
 
 	/**
@@ -552,12 +541,20 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 
 	@Override
 	public void resourceChanged(IResourceChangeEvent event) {
+		if (changedProjects) {
+			// no need to evaluate delta, we already know projects have changed
+			// and recomputation is necessary.
+			return;
+		}
 		final IResourceDelta delta = event.getDelta();
 		if (delta == null)
 			return;
 		//invalidate location map if there are added or removed projects.
-		if (delta.getAffectedChildren(IResourceDelta.ADDED | IResourceDelta.REMOVED, IContainer.INCLUDE_HIDDEN).length > 0)
+		if (delta.getAffectedChildren(IResourceDelta.ADDED | IResourceDelta.REMOVED,
+				IContainer.INCLUDE_HIDDEN).length > 0) {
 			changedProjects = true;
+			return;
+		}
 
 		// invalidate location map if any project has the description changed
 		// or was closed/opened
@@ -565,7 +562,7 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 		for (IResourceDelta element : changed) {
 			if ((element.getFlags() & IResourceDelta.DESCRIPTION) == IResourceDelta.DESCRIPTION || (element.getFlags() & IResourceDelta.OPEN) == IResourceDelta.OPEN) {
 				changedProjects = true;
-				break;
+				return;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


mickaelistria:
@jukzi would you like to review this?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -552,19 +548,24 @@ private void removeFromLocationsMap(IResource link, IFileStore location) {
 
 	@Override
 	public void resourceChanged(IResourceChangeEvent event) {
+		if (changedProjects.get()) {
```
jukzi, line 551: nice idea

jukzi (changes requested):
Without having testing it i bet it will not solve the issues in the BasicAliasTest.

Overall this looks like a half hearted attempt for thread-safety. It would be bette to first make a overall decision how to establish thread-safety.
a) full mutual exclusion (synchronize)
b) read/write lock
c) concurrent data structures.
If the final solution would require a) or b) it does not make sense to partly introduce concurrent data-structures c).
Typically a) is the easiest to archive the goal. if that is not performant enough one could think of b), But c) will be more challenging to adopt the multiple data-structures used into a single concurrent data structure. so i'd say this change is a step in the wrong direction.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +625,41 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
 			hadChanges = true;
 			buildLocationsMap();
-		} else {
-			//incrementally update location map for changed links
-			for (IResource resource : changedLinks) {
+		}
+		// incrementally update location map for changed links
+		while (!changedLinks.isEmpty()) {
+			Collection<IResource> changedLinksSnapshots = new HashSet<>(changedLinks);
+			changedLinks.removeAll(changedLinksSnapshots);
```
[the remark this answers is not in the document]
jukzi: but will probably not work together with hasStructureChanges() - As the isEmpty() is not synchronized with this code

laeubi:
Why do we use/need a semaphore here? That sounds rather strange here, it would be good to explain what kind of concurrency should be avoided here so one might choose better synchronization primitives here... I also wonder if it is really needed to have concurrent access here, it is not enough to synchronize on the critical section?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +625,41 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
 					addToLocationsMap(resource, ((Resource) resource).getStore());
 			}
 		}
-		changedLinks.clear();
-		if (hadChanges)
+		if (hadChanges) {
 			buildAliasedProjectsSet();
-		changedProjects = false;
+		}
+		updating.release();
```
laeubi, line 662: Just a note: If **anything** goes wrong, e.g. a runtime exception is thrown in any of the action before the semaphore will never be released at all it also seems that a semaphore is used here without real need for such and a simple synchronized and/or lock will be enough.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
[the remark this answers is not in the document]
mickaelistria: `hasStructuredChanges` was removed, it was merged in with `updateStructuredChanges` into `checkStructuredChanges` `as all calls to `hasStructuredChanges` were followed `updateStructuredChanges`, this allows to synchronize only a single method.

mickaelistria:
thanks for the feedback, I'm processing it. The semaphore is indeed not interesting with this patch, it's coming from a former experiment I made, but a synchronized seems indeed good enough and better for the checkStructuredChanges method.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +625,41 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
+		try {
+			updating.acquire();
+		} catch (InterruptedException e) {
+			ResourcesPlugin.getPlugin().getLog()
+					.log(new Status(IStatus.ERROR, ResourcesPlugin.PI_RESOURCES, e.getMessage(), e));
+			return false;
+		}
 		boolean hadChanges = false;
-		if (changedProjects) {
+		while (changedProjects.get()) {
```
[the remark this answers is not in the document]
mickaelistria: buildLocationsMap() is not supposed to run in parallel. It's only 2 callers are startup() and checkStructuredChanges(). The former is supposed to happen very early before workspace notification are set (before the state can be invalidated), the later is synchronized.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -552,19 +548,24 @@ private void removeFromLocationsMap(IResource link, IFileStore location) {
...
+			// and recomputation is necessary.
+			return;
+		}
 		final IResourceDelta delta = event.getDelta();
 		if (delta == null)
 			return;
 		//invalidate location map if there are added or removed projects.
 		if (delta.getAffectedChildren(IResourceDelta.ADDED | IResourceDelta.REMOVED, IContainer.INCLUDE_HIDDEN).length > 0)
-			changedProjects = true;
+			changedProjects.set(true);
```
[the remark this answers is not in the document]
mickaelistria: OK, a volatile boolean was my initial idea. I'd be fine using it instead of atomic boolean.

mickaelistria:
A new patch was pushed and should improve the code according to suggestions.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +621,40 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
+	 * @return <code>true</code> if some structural changes where processed,
+	 *         <code>false</code> if not structural changes were detected
 	 */
-	private void updateStructureChanges() {
+	/*
+	 * This method is synchronized as it calls
+	 * buildLocationsMap/addToLocationsMap/addToLocationsMap which are not meant to
+	 * run in parallel. Incoming events/notifications that cause a structural change
+	 * can come from different threads, but we only want a single thread to process
+	 * them as a time.
```
jukzi, line 633:
```suggestion
	 * them at a time.
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +621,40 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
 	/**
 	 * Process any structural changes that have occurred since the last alias
 	 * request.
+	 *
+	 * @return <code>true</code> if some structural changes where processed,
+	 *         <code>false</code> if not structural changes were detected
```
jukzi, line 626:
```suggestion
	 *         <code>false</code> if no structural changes were detected
```

jukzi (commented):
looks more natural.
There are still members which are not synchronized (or i fail to see how):
AliasManager.aliasedProjects, aliases, locationsMap, nonDefaultResourceCount, suffix

mickaelistria:
Indeed, those members are not synchronized. I was focused on guarding the bits that modify the state and didn't implement a good strategy for reading them. I'll prepare a new patch that use concurrent data structures.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +621,40 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
 			//if a project is added or removed, just recompute the whole world
 			changedProjects = false;
 			hadChanges = true;
 			buildLocationsMap();
-		} else {
-			//incrementally update location map for changed links
-			for (IResource resource : changedLinks) {
+		}
+		// incrementally update location map for changed links
+		while (!changedLinks.isEmpty()) {
```
jukzi, line 644: Don't use a while loop here otherwise addToLocationsMap could be called for same resource again.  which contradicts the use of a Set.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -624,27 +621,40 @@ public void updateAliases(IResource resource, IFileStore location, int depth, IP
...
+	 * This method is synchronized as it calls
+	 * buildLocationsMap/addToLocationsMap/addToLocationsMap which are not meant to
+	 * run in parallel. Incoming events/notifications that cause a structural change
+	 * can come from different threads, but we only want a single thread to process
+	 * them at a time.
+	 */
+	private synchronized boolean checkStructuralChanges() {
 		boolean hadChanges = false;
-		if (changedProjects) {
+		while (changedProjects) {
```
jukzi, line 637: previously it was either buildLocationsMap or addToLocationsMap now both can happen and add the same project - which is also a resource? That may increase nonDefaultResourceCount one time too much?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
mickaelistria: It wouldn't really contradict: if a link is changed 2 times, and was already processed once in addLocationsToMap, and in the meantime it's changed again, then it needs to call addLocationsToMap again. The set is not really the set of IResource in general, but the Set of IResource than needs an update. It can happen that a resource needs several updates if it was changed while it was being processed.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
mickaelistria: I need to investigate. Do you have a clearer idea of how things can be wrong? I hope we can make a test case for it.

jukzi: already the current implementation looks buggy or tricky as hell: 
only handleEvent(LifecycleEvent)  does nonDefaultResourceCount--
but both  handleEvent(LifecycleEvent) and resourceChanged(IResourceChangeEvent event)
could lead to nonDefaultResourceCount++

i guess the Trick was that buildLocationsMap() resets it to 0 - which does not hold anymore in your patch

laeubi:
I'm not very familiar with the `AliasManger` and what kind of synchronization problem one needs to solve here but from the changes here and the codeflow I get the feeling that a [ConcurrentLinkedQueue<IResource>](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html) might be a more appropriate data-structure here, then any thread can insert either "plain" resources or projects (that are also resources) and forget about the distinction in the `resourceChanged` and the optimization to skip further events.

Then in the update method, one can drain the queue into a List of resources, if it encounters a project can set a boolean that a full refresh is required and clear the queue, otherwise process the item in the List.

That way two things could happen:

1. new items are inserted _while_ we drain the queue and are consumed
2. new items are inserted _after_ we have drained the queue and will be processed next time

jukzi:
> forget about the distinction

I like the idea. if it is enough to fulfill all the contracts needs deeper investigation. in special i hope such thing would also remove the boolean notification which then would simply isEmpty check.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
mickaelistria: You're right, the while and removing the else make it more fragile while mot much improving/fixing anything.
I've restored to something that's closer from initial state and added comments to keep that in mind for further work.

mickaelistria:
I won't be able to spend much more time on this topic. If you agree this is already an improvement over current state, I'd like to merge it as it and see whether it improves at least the stability of BasicAliasTest before spending more time there.

jukzi:
> I won't be able to spend much more time on this topic. If you agree this is already an improvement over current state, I'd like to merge it as it and see whether it improves at least the stability of BasicAliasTest before spending more time there.

OK LGTM

jukzi (approved):

mickaelistria:
Thanks @jukzi and @laeubi , it was a good technical talk!

Issue #158, opened 2022-09-08
Race condition in AutoBuildJob sometimes prevents an AutoBuildJob to be scheduled() even though buildNeeded is true


xpomul:
It can happen that the AutoBuildJob's `buildNeeded` flag is `true`, but the AutoBuildJob is not scheduled and not executed until the next workspace change.

This one seems to have been in the code since forever and obviously was not noticed before (similar to [Bug 575486](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575486) - AutoBuildJob booleans need volatile modifier maybe?).

We have detected it in a custom IDE in which we run (RCPTT) UI tests that changes a file, saves that file, and waits for the result of the (auto) build to show in the (custom) UI. In the UI test case, this change-save-verify cycle is multiple times in quick succession. It fails randomly, and the analysis shows the following:

Since the AutoBuildJob is scheduled even if `buildNeeded` is false, there is a good chance that it is currently running when we save a file in the test case (as at least the test case will simulate user interaction very quickly. Most likely in daily use, the users won't experience or notice this behavior, because it should occur so rarely in the case of human interaction...).

Since saving a file is done in a ISchedulingRule, and since the build is also locking the workspace, there is some synchronization happening (which makes the phenomenon even more random most likely) but the following can still happen:

The AutoBuildJob, executing in the "Building..." thread, reaches [line 176](https://github.com/eclipse-platform/eclipse.platform/blob/master/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java#L176) which releases the workspace lock. But the Job is still in state RUNNING.

Now at this point, the main Thread is saving the file and at the end, it calls `AutoBuildJob.build(true)` from `org.eclipse.core.internal.events.BuildManager.endTopLevel()`. In the `build()` method, the `buildNeeded` flag is set to `true` and then the method checks the state of the AutoBuildJob and finds that it is currently (still) in the RUNNING state, so it does not schedule it again (it falls through the switch in [line 98](https://github.com/eclipse-platform/eclipse.platform/blob/master/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java#L98)).

In this situation, the AutoBuild is not scheduled even though it is marked as being in need of a build. 
So until the next workspace change triggering a build, this situation won't change and the AutoBuildJob will not be executed.

[...]

PR #159, opened 2022-09-08
Race condition in AutoBuildJob #158

xpomul:
* Let build() reschedule the AutoBuildJob even if it is currently running.
* Remove additional reset of buildNeeded flag in doBuild(), so it does not interfere with other threads concurrently requesting a build.

see https://github.com/eclipse-platform/eclipse.platform/issues/158


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -166,7 +167,6 @@ private void doBuild(IProgressMonitor monitor) throws CoreException, OperationCa
 			if (!result.isOK()) {
 				throw new ResourceException(result);
 			}
-			buildNeeded = false;
```
iloveeclipse, line 169: This means, `buildNeeded` is now never set to false? Why that?

jukzi: hidden in shouldBuild
forceBuild = avoidBuild = buildNeeded = false;

jukzi (approved):
LGTM

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -99,6 +99,7 @@ synchronized void build(boolean needsBuild) {
 			case Job.SLEEPING :
 				wakeUp(delay);
 				break;
+			case RUNNING :
```
jukzi, line 102: or even better: call schedule() independent of job state, as the job state is not synchronized and may change in parallel.

xpomul: think I have found a caveat.

The originally missing RUNNING case did prevent a constant rescheduling of the AutoBuildJob.
doBuild() ends in a call to Workspace.endOperation() -> BuildManager.endTopLevel(). 
The latter will call build(false). And if case RUNNABLE schedules the build. -> endless loop.

So I think I need to distinguish cases and only schedule the build if `state != RUNNING || buildNeeded`.

What about the SLEEPING case? Should I even schedule in that case or leave the wakeUp()? WDYT?

Will update the PR tomorrow...

iloveeclipse: Is the current change fixing your issue? If yes, I would consider not touch the code anymore, as long as you don't have a case to verify there is still some problem.

jukzi: it should only schedule if needsBuild

xpomul: @iloveeclipse the problem is that I need to run our UITest 10-20 times repeatedly to reproduce the issue. This makes it difficult to tell whether the race is completely fixed or not. But I think I'll invest the additional time and check if I can run this over the weekend to get it stable with a minimal change.

jukzi: if  not shouldBuild() it should not even prepare/begin/endOperation but should exit early as the event was already consumed

xpomul: @jukzi I am a bit worried about regressions. 
This code is so old and untouched for years and I fear no one is left who knows why it seems so complex :-(
For example, there is this comment in line [106](https://github.com/eclipse-platform/eclipse.platform/blob/master/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java#L106) about sending events even if the autobuild is off. 
If we add an early exit where before there was none couldn’t that also have unwanted side-effects?

xpomul: After thinking about this once more on Friday, I decided to go for the least intrusive solution:
The scheduling of the AutoBuildJob is unchanged in all states except RUNNING.
In the RUNNING state, it is rescheduled, if autobuild is on and buildNeeded is true.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
xpomul: I have added a comment before the if to make this clearer for the future reader.

xpomul:
Ok, thanks.
I’ll be on vacation for two weeks starting tomorrow. I’ll have another look when I am back.

xpomul:
Hi @jukzi,

I have slightly reworked my code to cover more cases regarding the state, added a condition for avoidBuild, and added a testcase as you have proposed.

Sadly, since our case is a race condition that occurs unpredictably, I have cheated in the test case by using a spoiled Scheduling Rule to get a deterministic behavior. The test case fails with the current master, but is green with my changes.
Unfortunately, I haven’t found a way to check the cancel scenario. What happens in the test case is that `dirty(file)` in the other thread will actually set the interrupt flag in the AutoBuildJob, so even if I cancel the Job via its monitor, it will come to `avoidBuildIfNotInterrupted` which does not set avoidBuild, because the job has already been interrupted.

This means, of course, that in case of the scenario handled here, if the user cancels the build, it will be reschedule nonetheless. But this behavior is there already in the current master branch, because a cancelled AutoBuildJob will rebuild on the next change anyways. In our case, this next change just happens more or less at the same time the user cancels the build. I think we cannot deterministically handle this case.

Also, regarding your doubt regarding the concurrent state change: also this behavior is there in the current master branch. To fix this, we would need to make the `run()` method synchronized. And I am not sure this is what we should do.

One question about the failing JUnit tests in the github-results. Are they expected to fail or are they actually side-effects caused by my changes?

Should I un-draft this PR or do you have any concrete ideas what I could improve to get this towards a state that can be merged?

jukzi:
`java.lang.OutOfMemoryError: Java heap space
`
is probably a side effect of your changes. the unit tests are not meant to fail. only sometimes 1 fails randomly.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
```diff
@@ -0,0 +1,200 @@
...
+				dirty(file);
+				// Register our jobChangeListener so we can record when the AutoBuildJob is
+				// rescheduled and done.
+				Job.getJobManager().addJobChangeListener(jobChangeListener);
+				// Signal to the builder that it can continue now
+				barrier.setStatus(AUTOBUILD_CONTINUE);
+				return Status.OK_STATUS;
+			}
+		};
+		job.setRule(new SpoiledSchedulingRule());
```
jukzi, line 190: I do not understand that rule. what should it do other then just no rule? and why do you schedule and join the job instead of having no job at all and just runinworkspace?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
```diff
@@ -0,0 +1,200 @@
...
+			desc.setBuildSpec(new ICommand[] { createCommand(desc, EmptyDeltaBuilder.BUILDER_NAME, "ProjectBuild") });
+			project.setDescription(desc, getMonitor());
+			project.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, getMonitor());
+
+			// Perform the first build. This instantiates the EmptyDeltaBuilder and makes
+			// its instance accessible.
+			setAutoBuilding(true);
+
+			// Configure the behavior of the builder
+			EmptyDeltaBuilder.getInstance().setRuleCallback(new BuilderRuleCallback() {
```
jukzi, line 135: shouldn't that happen before autobuild?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
```diff
@@ -0,0 +1,200 @@
...
+					}
+					return new IProject[] { project };
+				}
+			});
+
+			// Wait until the Content Description Update job has finished, because it also
+			// might schedule a build and we don't want that interfere with our test
+			// process.
+			waitForContentDescriptionUpdate();
+		} catch (CoreException e) {
```
jukzi, line 157: setUp() throws Exception
why convert the exception?

jukzi:
@xpomul will you continue to work on this? you need to fix the commit to succeed tests.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
```diff
@@ -0,0 +1,200 @@
...
+	 */
+	private IFile file;
+
+	/**
+	 * A thread-safe flag to record the rescheduling of the AutoBuildJob
+	 */
+	private AtomicBoolean rescheduled = new AtomicBoolean(false);
+
+	/**
+	 * A somewhat illegal {@link ISchedulingRule} which satisfies the sanity checks
```
mickaelistria, line 60: If this is illegal as per contract, we won't add support for it, but instead `JobManager.validateRule` should be improved if possible to reject it.

xpomul: I am not fixing broken scheduling rules. The original issue is created by a race condition in a scenario that uses correct scheduling rules. I was asked to write a test case to reproduce the issue, but the nature of a race condition is that you can’t reproduce it. So I found this hole in the scheduling rule check which makes it possible to have a test case.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
xpomul: For the job/runnable itself, there is no difference. But dirty(file) internally will create a rule and will block on the AutoBuildJob. The `SpoiledSchedulingRule` behaves like a null rule for the job/runnable but still contains the rule for `file` which causes the dirty operation to not block on the AutoBuildJob.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
xpomul: Actually doesn’t matter as we had an explicit incremental build before (to let the system instantiate EmptyDataBuilder and make its instance available). Therefore, setting the autobuild to true here will not invoke the builder just yet.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
xpomul: And about the `Workspace.run()` you are right, I was thinking too complex here...

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobRescheduleTest.java
xpomul: the other tests sometimes did that, so this got in when I based my code on the other tests. But you are correct, it makes no sense here.

xpomul:
> @xpomul will you continue to work on this? you need to fix the commit to succeed tests.

@jukzi yes I still plan to get the tests fixed. But I don’t have much time at the moment. This issue has taken more effort than I had hoped. Also since @mickaelistria has expressed doubts, I’m a bit reluctant to invest more effort just to be turned down in the end. This would not be my first contribution attempt that is rejected after spending days of effort...

jukzi:
hint: IResourceChangeListenerTest.test_1GDK9OG() has an endless loop with your change in AutoBuildJob

iloveeclipse:
> > @xpomul will you continue to work on this? you need to fix the commit to succeed tests.
> 
> @jukzi yes I still plan to get the tests fixed. But I don’t have much time at the moment. This issue has taken more effort than I had hoped. Also since @mickaelistria has expressed doubts, I’m a bit reluctant to invest more effort just to be turned down in the end. This would not be my first contribution attempt that is rejected after spending days of effort...

Yea, creating reliable tests for autobuild racing condition is hard.
Can we split this into two PR's then?

I would like to see the patch "as is" on master, as it obviously makes sense, and the test ... Well, if you have no time it is pity, but should not stop the fix from merge.

So I would propose following:

- rebase on latest master
- remove or disable test that is not finished 
- bump version for org.eclipse.core.resources MANIFEST.MF from 3.18.100 to 3.18.200 **in a second commit**
- merge.

iloveeclipse:
> So I would propose following:
> 
> * rebase on latest master
> * remove or disable test that is not finished
> * bump version for org.eclipse.core.resources MANIFEST.MF from 3.18.100 to 3.18.200 **in a second commit**
> * merge.

Done in https://github.com/eclipse-platform/eclipse.platform/pull/297

You can keep this PR for improving the test or close it, as you like.

iloveeclipse:
I've also squashed all commits to one and updated commit message to explain the problem / fix.

iloveeclipse:
> hint: IResourceChangeListenerTest.test_1GDK9OG() has an endless loop with your change in AutoBuildJob

This makes the patch "AS IS" not acceptable.

I see it now: the AutoBuildJob **always** schedules itself if some POST_BUILD listener touches anything in the workspace.
Fix is: https://github.com/eclipse-platform/eclipse.platform/compare/7cbc99a8f683e99f318fea4f3e2750824188a45f..a9ba826f5fe3c47a16a8cc0867afdc713da612d3

Beside this, the test case passes with and without the patch, so test isn't reproducing the problem. 
I doubt you can write such a test, because you have to trigger build exactly between doBuild() and "return" in AutobuildJob.run().

Issue #167, opened 2022-09-13
Bug_574883.testJoinLambdaOften failing often on CI


mickaelistria:
Lately, most of the PRs have pushed received a negative vote, often because of [Bug_574883.testJoinLambdaOften](https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-166/1/testReport/junit/org.eclipse.core.tests.runtime.jobs/Bug_574883/testJoinLambdaOften/) failing.
It's important to get CI capable of sending positive feedback reliably in the normal state, so such failures must be fixed.

PR #169, opened 2022-09-13
Fix Job Notification order to fix JobManager.join() #167

jukzi:
Fixes left over part from
Bug 574883 - Job.getJobManager().join(family) doesn't wait for a re-scheduled job

Previously it could happen that a Job was scheduled and then executed but did not yet notify that it was finished. Then it could be scheduled again in another thread which would notify the scheduling before the done() event was handled. A join() would see the Job as running but would receive an obsolete done() event which terminates the join() even though the current job execution was not done yet.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -155,7 +155,7 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	/**
 	 * This signal is used to synchronize Job listener notification
 	 */
-	volatile boolean waitForNotificationFinsished;
+	volatile boolean waitForNotificationFinished;
 
 	private static synchronized int getNextJobNumber() {
 		return nextJobNumber++;
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
@@ -701,7 +701,7 @@ public class JobManager implements IJobManager, DebugOptionsListener {
 		final boolean scheduled;
 		try {
 			// do not restart job before notification finished:
-			job.waitForNotificationFinsished = true;
+			job.waitForNotificationFinished = true;
 			synchronized (lock) {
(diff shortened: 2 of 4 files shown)


jukzi:
@akurtakov what is this build error?  `Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:3.0.0-SNAPSHOT:test (default-test) on project org.eclipse.jsch.tests: No tests found. `

iloveeclipse:
This PR causes severe debugger regression, see https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/122

iloveeclipse:
This PR creates deadlock
see 
`org.eclipse.jdt.internal.launching.SocketListenConnectorProcess.waitForConnection(ListeningConnector, Map<String, Argument>)` which does
```
fWaitForConnectionJob.addJobChangeListener(new JobChangeAdapter(){
			@Override
			public void running(IJobChangeEvent event) {
				fireReadyToAcceptEvent();
			}
			@Override
			public void done(IJobChangeEvent event) {
				if (event.getResult().isOK() && continueListening()) {
					fWaitForConnectionJob.schedule();
				} else {
					try{
						terminate();
					} catch (DebugException e){}
				}
			}
		});
		fWaitForConnectionJob.schedule();
```
and that leads to hang in
```
Thread [Worker-3: Waiting for vm to connect at port 46605...] (Suspended (breakpoint at line 1357 in JobManager))	
	JobManager.schedule(InternalJob, long, boolean) line: 1357	
	SocketListenConnectorProcess$WaitForConnectionJob(InternalJob).schedule(long) line: 390	
	SocketListenConnectorProcess$WaitForConnectionJob(Job).schedule() line: 654	
	SocketListenConnectorProcess$1.done(IJobChangeEvent) line: 132	
	1147591614.notify(IJobChangeListener, IJobChangeEvent) line: not available	
	JobListeners.doNotify(JobListeners$IListenerDoit, IJobChangeEvent) line: 80	
	JobListeners.done(Job, IStatus, boolean) line: 118	
	JobManager.endJob(InternalJob, IStatus, boolean) line: 730	
	WorkerPool.endJob(InternalJob, IStatus) line: 117	
	Worker.run() line: 83	
```

There is no restriction that listeners on job events can't schedule new jobs while job notification.

I would revert this commit now to unblock jdt.debug tests from hang.

iloveeclipse:
Revert: https://github.com/eclipse-platform/eclipse.platform/pull/175

PR #166, opened 2022-09-13
Use org.eclipse.platform as groupId

mickaelistria:
to map more naturally to Maven coordinates

PR #172, opened 2022-09-15
Revert "Use org.eclipse.platform as groupId"

jukzi:
Reverts eclipse-platform/eclipse.platform#166

JSCH test failures

### DIFF
--- a/.settings/org.eclipse.core.resources.prefs
+++ b/.settings/org.eclipse.core.resources.prefs
@@ -1,2 +1,0 @@
-eclipse.preferences.version=1
-encoding/<project>=UTF-8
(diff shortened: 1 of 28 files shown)


jukzi:
let's test if a revert would make CI work again

mickaelistria:
OK, I now see the cause of the error and it's indeed the patch causing it. I didn't notice it in previous build as it was reporting another test failure as well.
Please don't merge the revert, I'll submit a more focused fix (basically replacing the pom.xml that was removed for this bundle so it's not detected as a test).

mickaelistria:
#174 is supposed to fix it.

mickaelistria:
Thanks for the heads-up. This is now fixed with #174

PR #175, opened 2022-09-16
Revert "Fix Job Notification order to fix JobManager.join() #167"

iloveeclipse:
This reverts commit cc397e8bca80e88c4adf5012b445a32f95a84e14.

Reason for revert: this change causes deadlocks if new jobs are scheduled from job listeners.

See https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/122


jukzi:
if you can wait ~ an hour i would prefer to submit a fix :-)

PR #176, opened 2022-09-16
Fix deadlock when reschedule from the JobChangeAdapter done()

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/122

It is however not recommended to reschedule from done() but better from within the job as it is undefined what other JobChangeAdapter will see and thus if a join() would wait for the rescheduled job or not.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -155,7 +155,7 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	/**
 	 * This signal is used to synchronize Job listener notification
 	 */
-	volatile boolean waitForNotificationFinished;
+	volatile Thread notifyPendingThread;
 
 	private static synchronized int getNextJobNumber() {
 		return nextJobNumber++;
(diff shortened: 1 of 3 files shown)


jukzi:
have not seen those failing previously, and coudl not reproduce locally:
```
java.lang.AssertionError: Too many worker threads active: 511, must be <= 50
	at org.eclipse.core.tests.runtime.jobs.WorkerPoolTest.testIdleWorkerCap(WorkerPoolTest.java:68)

junit.framework.AssertionFailedError: 3.0
	at org.eclipse.core.tests.runtime.jobs.JobTest.testJoinWithTimeout(JobTest.java:926)
	
junit.framework.AssertionFailedError: 2.0
	at org.eclipse.core.tests.runtime.jobs.JobGroupTest.testJoinWithTimeout(JobGroupTest.java:541)
```

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
```diff
@@ -155,7 +155,7 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	/**
 	 * This signal is used to synchronize Job listener notification
 	 */
-	volatile boolean waitForNotificationFinished;
+	volatile Thread notifyDoneThread;
```
iloveeclipse, line 158: I would rename it to "notifyPendingThread", to indicate that notification is not yet done.

jukzi: ok

Issue #178, opened 2022-09-19
[TEAM] (Local History view) Get Contents of the current revision zeroes out the file


TIBCOeddie:
Eclipse 4.24

We saw this in our product, however I then realized it seems to be a platform issue.
I reproduced as follows:
Have eclipse platform + JDT
Create a java class
Find java file, ask for Local History
in the history view, right click the bold entry (the latest/current), select Get Contents
observe the file got zeroed out (contents are blanked)

mickaelistria:
Reproduced with latest I-Build, but I think it's a particular case of Java file creation. In other circumstances, Get Contents works fine with me.

TIBCOeddie:
What's interesting is how our custom file type/editor also suffered from this - ending up with a blank file.

mickaelistria:
My bad, you're probably right. I tried more carefully and could also reproduce it with other files. I most likely didn't test correctly earlier.
Are you willing to investigate this issue and maybe submit a fix?

TIBCOeddie:
OK, basic workspace/java stuff. Not sure when (if?) this has changed in some ... past version.

Note, I'm on MacOS, running Eclipse here under Java 11.

The history view is doing simple things: file.setContents using an input stream against the same file it's going to write to. We could circumvent the bug here; do not try to do anything when the current revision/store is just the pointer to the actual file. Fix in team.

Still, maybe worth pursuing over in core land - org.eclipse.core.internal.utils.FileUtil.transferStreams(InputStream, OutputStream, String, IProgressMonitor) , basically if the input stream and the output stream point to the same file, the FileInputStream read will return -1 immediately (since the FileOutputStream is read first, truncating the file contents). Thoughts on pursuing anything there? (at least javadoc in IFile.setContents ? In case callers could assume the new contents will be read first?))

mickaelistria:
Looks like `transferStreams` and `setContents` do what's expected from them and that we cannot reliably add checks to prevent from overriding while reading/writing (because we don't really know anything about the streams at that point). I believe that `getContentsAction` should simply be disabled or do nothing in case it's on the current revision.

TIBCOeddie:
PR #183

PR #183, opened 2022-09-19
#178 do not show Get Contents for current-state local file revisions

TIBCOeddie:
Issue 178, do not show Get Contents when the selected revision represent a local, current state

### DIFF
--- a/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
+++ b/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
@@ -567,8 +567,19 @@ public class LocalHistoryPage extends HistoryPage implements IHistoryCompareAdap
 					if (sel instanceof IStructuredSelection) {
 						IStructuredSelection tempSelection = (IStructuredSelection) sel;
 						if (tempSelection.size() == 1) {
-							manager.add(new Separator("getContents")); //$NON-NLS-1$
-							manager.add(getContentsAction);
+							boolean showGetContentsAction = true;
+							// Issue 178 avoid presenting this action against the current
+							// revision which would result in an undesired truncation
+							if (tempSelection.getFirstElement() instanceof LocalFileRevision) {
+								LocalFileRevision lfr = (LocalFileRevision) tempSelection.getFirstElement();
+								if (lfr.isCurrentState()) {
+									showGetContentsAction = false;
+								}
+							}
+							if (showGetContentsAction) {
+								manager.add(new Separator("getContents")); //$NON-NLS-1$
+								manager.add(getContentsAction);
+							}
 						}
 					}
 				}
--- a/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
+++ b/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
@@ -562,11 +562,9 @@ public class LocalHistoryPage extends HistoryPage implements IHistoryCompareAdap
 			}
 			if (compareAction != null)
 				manager.add(compareAction);
-			if (getContentsAction != null) {
-				if (showGetContentsAction(treeViewer.getStructuredSelection())) {
-					manager.add(new Separator("getContents")); //$NON-NLS-1$
-					manager.add(getContentsAction);
-				}
+			if (getContentsAction != null && showGetContentsAction(treeViewer.getStructuredSelection())) {
+				manager.add(new Separator("getContents")); //$NON-NLS-1$
+				manager.add(getContentsAction);
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
```diff
@@ -567,8 +567,19 @@ protected void fillTableMenu(IMenuManager manager) {
 					if (sel instanceof IStructuredSelection) {
 						IStructuredSelection tempSelection = (IStructuredSelection) sel;
 						if (tempSelection.size() == 1) {
-							manager.add(new Separator("getContents")); //$NON-NLS-1$
-							manager.add(getContentsAction);
+							boolean showGetContentsAction = true;
```
mickaelistria, line 570: If only we were using a more recent Java, this would be a 1-liner.

```java
if (!(tempSelection.getFirstElement() instanceof LocalFileRevision lfr && lfr.isCurrentState())) {
  managed.add(...);
}
```

cc @akurtakov , in case you need some example convincing people that Java 17 would really make Ecipse code better.

laeubi: Less "lines" (you can always write this code already as one line) does not makes better code. While it seems "smarter" it makes the code hard to debug and understand.

For this particular case, a method that returns a boolean like 

```
if (canGetContent(sel)) { 
   managed.add(...); 
}
```

is (independent of the java version) much more readable, better understandable and easier to debug even though it might result in even more lines.

laeubi: And just for the record, the code can already be simplified as since 3.11 JFaces offers a 

`org.eclipse.jface.viewers.AbstractTreeViewer.getStructuredSelection()`

so cast in L567 can be avoided.

mickaelistria:
Can you please add another commit which increase the version of the modified bundle by +0.0.100, in the MANIFEST.MF, and -if existing- in the pom.xml file?

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
TIBCOeddie: I actually was wondering how far I can take the code to modern style ; I would have even used a var!  
Will trim things down a bit. @laeubi do you think this warrants a method extraction? Only use, can't even predict another usage...

TIBCOeddie:
Just committed new code style, should be nice and clear/easy to read, with a method extract, it ended up well.
also incremented service version.

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
```diff
@@ -562,18 +563,35 @@ protected void fillTableMenu(IMenuManager manager) {
...
-				ISelection sel = treeViewer.getSelection();
-				if (!sel.isEmpty()) {
-					if (sel instanceof IStructuredSelection) {
-						IStructuredSelection tempSelection = (IStructuredSelection) sel;
-						if (tempSelection.size() == 1) {
-							manager.add(new Separator("getContents")); //$NON-NLS-1$
-							manager.add(getContentsAction);
-						}
-					}
+				if (showGetContentsAction(treeViewer.getStructuredSelection())) {
```
laeubi, line 566: if statements can be merged.

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
```diff
@@ -562,18 +563,35 @@ protected void fillTableMenu(IMenuManager manager) {
...
+	 * @param sel selection to work against
+	 * @return whether the Get Contents action should be displayed to the user
+	 */
+	private boolean showGetContentsAction(ITreeSelection sel) {
+		if (sel.size() == 1) {
+			Object selectedElem = sel.getFirstElement();
+			// Issue #178 avoid presenting this action against the current
+			// revision which results in an undesired file truncation
+			if (selectedElem instanceof LocalFileRevision) {
+				if (((LocalFileRevision) selectedElem).isCurrentState()) {
```
laeubi, line 586: you can directly return the boolean expression instead of using another if

laeubi (approved):
Looks good to me, just some minor comments for possible polishing.

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
TIBCOeddie: done

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/history/LocalHistoryPage.java
TIBCOeddie: this one I believe is clearer left legible as ("if current state, return false") vs having the negation inlined which can be missed with `return !(..quite a few characters later..).isCurrentState() `

mickaelistria:
Thank you!

PR #182, opened 2022-09-19
DeadlockDetector shutup. fixes #145

jukzi:
Logging in DeadlockDetector.reportDeadlock(Deadlock) may timeout test if filesystem is busy.

PR #184, opened 2022-09-20
Fix API tools errors complaining without reason

iloveeclipse:
Should fix https://github.com/eclipse-platform/eclipse.platform/pull/182 **unrelated** build issues

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,6 @@ pipeline {
 					sh """
 					mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 						-Pbuild-individual-bundles -Pbree-libs -Papi-check \
-						-Dcompare-version-with-baselines.skip=false \
 						-Dmaven.compiler.failOnWarning=true -Dproject.build.sourceEncoding=UTF-8 -T1C
 					"""
 				}


mickaelistria (changes requested):
Please don't merge it and remain a bit patient. A fix is being built in Tycho and will be available shortly. In general, please do not assume that disabling a quality gate is something that can be done lightly. If this change gets merged and is not reverted later (anyone can forget), then we'll have tons of other erroneous case of comparator errors to deal with. Overall, we'd loose even more time.

iloveeclipse:
> please do not assume that disabling a quality gate is something that can be done lightly

Same fix in jdt debug was proposed by @sravanlakkimsetti , see https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/124
So I assumed removing `-Dcompare-version-with-baselines.skip=false` is correct thing to do.

laeubi:
> Please don't merge it and remain a bit patient. A fix is being built in Tycho and will be available shortly.

Or even maybe a bit **more patient** because INFRA is low priority for EF and thus we are suffering from big build backlogs, estimated ETA is 2hrs at the moment for a fix already merged and about 6hrs for other fixes:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1955

iloveeclipse:
So once again: 
Is removing `-Dcompare-version-with-baselines.skip=false` is correct thing to do or not?
I can't see any meaningful explanation what the default value for this property is in 
https://www.eclipse.org/tycho/sitedocs/tycho-extras/tycho-p2-extras-plugin/compare-version-with-baselines-mojo.html

If the default is `true`, we have to revert https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/124

laeubi:
> Is removing `-Dcompare-version-with-baselines.skip=false` is correct thing to do or not?

As @mickaelistria most probably not. Whether or not you will break the I-Build can not be determined, it might pass because the check would have passed for your PR anyways, it might fail the build.

mickaelistria:
> Is removing -Dcompare-version-with-baselines.skip=false is correct thing to do or not?

No, this will cause more issues later. The correct thing to do is to wait for Tycho to provide a fixed version.

mickaelistria:
> If the default is true, we have to revert https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/124

The default is `false` in Platform parent pom because we check versions on a later step (not as part of the Maven build). But for CI, we want to fail the build in case of missing version bump to reduce the risk of having such issues later. That's what this flag in Jenkinsfile enables.

iloveeclipse:
> I can't see any meaningful explanation what the default value for this property is in
> https://www.eclipse.org/tycho/sitedocs/tycho-extras/tycho-p2-extras-plugin/compare-version-with-baselines-mojo.html

So is the default to **skip** baseline check if `-Dcompare-version-with-baselines.skip` is **not set**?

PR #186, opened 2022-09-23
fix DeltaDataTree.basicCompare() to not iterate over O(n*n) parents

jukzi:
If there is a long chain of parents this gives a huge performance advantage for ResourceDeltaFactory.computeDelta() - used during workspace save and build.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/DeltaDataTree.java
@@ -178,7 +178,7 @@ public class DeltaDataTree extends AbstractDataTree {
 			/* Iterate through the receiver's ancestors until the receiver is reached */
 			while ((tree = tree.getParent()) != this) {
 				//ancestor may not contain the given path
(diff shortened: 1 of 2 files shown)


mickaelistria:
While I understand that it's a performance gain, I think this change introduce a behavior change: imagine we're in this loop with the following tree A(parent of)<B<C<D, and we're running the method with `other`=C<D and `this`=A<B<C<D, and we have a value for `path` in B. Then with this change, it seems the value for `path` is now not copied, while it used to be.
I'm not certain what is the expected behavior here.

jukzi:
> While I understand that it's a performance gain, I think this change introduce a behavior change: imagine we're in this loop with the following tree A(parent of)<B<C<D, and we're running the method with `other`=C<D and `this`=A<B<C<D, and we have a value for `path` in B. Then with this change, it seems the value for `path` is now not copied, while it used to be.
> I'm not certain what is the expected behavior here.

That's interesting. i'l try to get a junit

jukzi:
@mickaelistria i added junit tests you may play around with.
In your case "DeltaDataTreeTest.testCompareWithPath2()" other.hasAncestor(this) returns false and we won't go into the changed branch.

PR #188, opened 2022-09-27
Full deterministic JobEvent order.

jukzi:
Events are queued inside the synchronize(JobManager.lock) blocks and send in the queued order after the block.
There is however no guarantee in which Thread the event is delivered - only the order is guaranteed to be the same order as the events happened.
Events will never be delivered in parallel - for the same job, but can be delivered in parallel for distinct jobs.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -18,6 +18,8 @@
 package org.eclipse.core.internal.jobs;
 
 import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.*;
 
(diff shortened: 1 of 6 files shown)


jukzi:
@iloveeclipse 
I do not get the reason for https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/137 but a clean solution for all Events might help.

iloveeclipse:
@jukzi : I haven't reviewed the code, but it should be the fix for #193, so please do following:

1) Rebase on master to get the `org.eclipse.core.tests.runtime.jobs.GithubBug_193` test in your branch.
2) Add this test to the `org.eclipse.core.tests.runtime.jobs.AllTests` testsuite. It should not fail with your patch.
3) Update commit message to add link to the https://github.com/eclipse-platform/eclipse.platform/issues/193 issue.

iloveeclipse:
> 1. Rebase on master to get the `org.eclipse.core.tests.runtime.jobs.GithubBug_193` test in your branch.
> 2. Add this test to the `org.eclipse.core.tests.runtime.jobs.AllTests` testsuite. It should not fail with your patch.
> 3. Update commit message to add link to the [JobManager.join(family) or related code is broken in 4.26 M1 #193](https://github.com/eclipse-platform/eclipse.platform/issues/193) issue.

Jörg, could you please do 2) and 3) too?

jukzi:
> Jörg, could you please do 2) and 3) too?

done

Issue #136, opened 2022-08-31
random failing RelaxedSchedRuleBuilderTest.testBuilderDeltaUsingRelaxedRuleBug343256


jukzi:
https://github.com/eclipse-platform/eclipse.platform/pull/135/checks?check_run_id=8113529011

```
junit.framework.AssertionFailedError: 
Error observed: junit.framework.AssertionFailedError: 1.2: 
/testBuildDeltaUsingRelaxedRuleBug343256[~]: {} expected:<1> but was:<0>
	at org.eclipse.core.tests.harness.CoreTest.fail(CoreTest.java:74)
	at org.eclipse.core.tests.internal.builders.RelaxedSchedRuleBuilderTest.testBuilderDeltaUsingRelaxedRuleBug343256(RelaxedSchedRuleBuilderTest.java:353)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:90)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
[...]

PR #190, opened 2022-09-28
reduce log message for Bug 352867 #83

jukzi:
The log message can get so long that it freezes the UI on hover.


jukzi:
unrelated testBuilderDeltaUsingRelaxedRuleBug343256  https://github.com/eclipse-platform/eclipse.platform/issues/136

PR #191, opened 2022-09-29
correctly fail testRefreshOtherProjectDuringRefresh #368

jukzi:
fail was only logged in other thread


jukzi:
The error was probably introduced with 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=29624#c14 
even tough
https://bugs.eclipse.org/bugs/show_bug.cgi?id=29624#c12
stated listeners should run into error .. but they run in the same thread.
The test was added years later and silently failed.
don't know how to proceed.

mickaelistria:
So the test has constantly been (silently) failing for all those years?
If so, I suggest we `@Ignore` it and comment that it's a good test case for an existing bug that doesn't have a fix at the moment and it should be re-enabled whenever a fix is being worked on.

HeikoKlare:
This one should have been resolved by #366, so I think it can be closed.
I didn't see that there was already a PR for the according issue when I provided #366, sorry for that.

PR #195, opened 2022-09-30
Regression test case for broken job event order (#193)

iloveeclipse:
This test relies on proper job event order reported to job listeners to succeed. Listener expects that for any job, "scheduled" event is always sent before "done". If that is not the case, test will fail.

See https://github.com/eclipse-platform/eclipse.platform/issues/193

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/GithubBug_193.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/GithubBug_193.java
@@ -1,0 +1,270 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Andrey Loskutov <loskutov@gmx.de> and others
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Andrey Loskutov <loskutov@gmx.de> - initial API and implementation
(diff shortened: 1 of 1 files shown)


jukzi:
thanks, Andrey, The test fails on Windows too (after 60s):
```
java.lang.AssertionError: Jobs delivered in wrong order for [job 390 family 3, job 585 family 3] expected:<0> but was:<1>
	at org.eclipse.core.tests.runtime.jobs.JobWatcher.waitUntilJobsAreDone(GithubBug_193.java:251)
	at org.eclipse.core.tests.runtime.jobs.GithubBug_193.testDoneEventReceivedBeforeSchedule(GithubBug_193.java:51)
```
Would be good to have a bit faster fail.
when i change the sleep to `Thread.sleep(0, jobId % 10);` the first fail "!!! 'done' before 'schedule' for job 338 family 3" come after a few ms and the test could end as soon as the first missorder happend.

iloveeclipse:
@jukzi : feel free to improve the test in https://github.com/eclipse-platform/eclipse.platform/pull/188

mickaelistria:
> Would be good to have a bit faster fail.

As the test is supposed to soon turn green forever, it doesn't seem worth the extra effort of optimizing the failure case.

jukzi:
> As the test is supposed to soon turn green forever, it doesn't seem worth the extra effort of optimizing the failure case.

agreed. just it can also finish faster :-)

HannesWell:
Looks like this PR the tests on the master branch:
https://ci.eclipse.org/platform/job/eclipse.platform/job/master/112/testReport/

Is this intended, known or unrelated?

iloveeclipse:
> Is this intended, known or unrelated?

Unrelated. PR adds a test that shouldn't be executed by default because it is not added to the test suite.

HannesWell:
Indeed. Looks like this was fixed with #188, which was merged after I asked.

PR #197, opened 2022-10-01
Remove now unused elements of ConfigurationSessionTestSuite

HannesWell:
As mentioned in https://github.com/eclipse-platform/eclipse.platform/pull/54#issuecomment-1188239725 the `ConfigurationSessionTestSuite` can be now cleaned up since the rest of the platform does not use the old arrays/methods.

Furthermore the `EclipseTestHarnessApplication` and `TestBarrier` seems to be unused. The latter is also deprecated.
Or is `EclipseTestHarnessApplication` used in some of the I-build tests?


HannesWell:
> Furthermore the `EclipseTestHarnessApplication` and `TestBarrier` seems to be unused. The latter is also deprecated. Or is `EclipseTestHarnessApplication` used in some of the I-build tests?

I could not find any other occurrence of `EclipseTestHarnessApplication` in any file in the Eclipse-SDK and not in the Jenkins jobs for the `Automated Tests`. So I assume it is save to remove that class.

PR #198, opened 2022-10-01
Simplify Platform.getBundles() and tracker open/closing

HannesWell:


HannesWell:
@iloveeclipse or somebody else want to have a look at this?

iloveeclipse:
@laeubi might be interested, I'm not.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -654,78 +649,37 @@ public void stop(BundleContext bundleContext) {
...
-		}
-		userLocation = new ServiceTracker<>(context, filter, null);
-		userLocation.open();
-
-		try {
-			filter = context.createFilter(Location.CONFIGURATION_FILTER);
+			filter = context.createFilter(filterStr);
 		} catch (InvalidSyntaxException e) {
-			// ignore this.  It should never happen as we have tested the above format.
+			// ignore this. It should never happen as we have tested the above format.
```
laeubi, line 672: Better throw assertionError then here...

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -654,78 +649,37 @@ public void stop(BundleContext bundleContext) {
...
-			filter = context.createFilter(Location.USER_FILTER);
-		} catch (InvalidSyntaxException e) {
-			// ignore this.  It should never happen as we have tested the above format.
-		}
-		userLocation = new ServiceTracker<>(context, filter, null);
-		userLocation.open();
-
-		try {
-			filter = context.createFilter(Location.CONFIGURATION_FILTER);
+			filter = context.createFilter(filterStr);
```
laeubi, line 670: Maybe use FrameworkUtil.createFilter instead

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -803,13 +737,8 @@ private void closeOSGITrackers() {
...
 	 */
 	public static void message(String message) {
-		StringBuilder buffer = new StringBuilder();
-		buffer.append(new Date(System.currentTimeMillis()));
-		buffer.append(" - ["); //$NON-NLS-1$
-		buffer.append(Thread.currentThread().getName());
-		buffer.append("] "); //$NON-NLS-1$
-		buffer.append(message);
-		System.out.println(buffer.toString());
+		System.out.println(
```
laeubi, line 740: Using String.format would make this more readable

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
HannesWell: Good Point.
Done.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
HannesWell: I can't see a real improvement over context.createFilter() and the drawback would be that the containers configuration DEBUG_FILTER setting is not considered anymore.
I don't think that is a very important feature since those filters are actually testsed during development, but because FrameworkUtil.create() does not have a clear advantage, I would leave it as it is.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
HannesWell: Personally I find String.format more readable.
I understand and agree that splitting the string with `+arg+` also has its difficulties to read it, but 'manually' identifing the markers in the format-String and matching it to the appended args is also something I don't consider very readable.
And unfortunately [String templates](https://openjdk.org/jeps/430) are not (yet) available in Java.

laeubi: For me its hard to understand how the final message is supposed to looks like. Having placeholders is very common concept for example in SLF4J API, also eclipses `//$NON-NLS-1$` don't make it better here

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
laeubi: FrameworkUtil.createFilter do not need a bundlecontext and can be called anywhere, even creating a constant. If its assumed you never change the filter, what DEBUG_FILTER is good for?

HannesWell: > FrameworkUtil.createFilter do not need a bundlecontext and can be called anywhere, even creating a constant. If its assumed you never change the filter, what DEBUG_FILTER is good for?

Probably for nothing. :) But since the context is needed in this method anyway we don't gain much in using FrameworkUtil.
However, I changed it to use `FrameworkUtil`.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
HannesWell: > For me its hard to understand how the final message is supposed to looks like. Having placeholders is very common concept for example in SLF4J API, also eclipses `//$NON-NLS-1$` don't make it better here

Yes, that's all true, but at least if there are many arguments I personally find that approach not very readable as well.
Anyways, since there are only three arguments I agree with you and changed it.

HannesWell:
@laeubi do you have any more remarks?

HannesWell:
> @laeubi do you have any more remarks?

I consider the silence as No. :)
Thanks for the other remarks.

PR #203, opened 2022-10-04
Do not synchronize while calling IJobChangeListener client code #193

jukzi:
Because some clients just do not progress.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -20,6 +20,7 @@ package org.eclipse.core.internal.jobs;
 import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.locks.ReentrantLock;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.*;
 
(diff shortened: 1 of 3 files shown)


jukzi:
@jonahgraham should solve the CDT deadlock

jukzi:
https://github.com/eclipse-cdt/cdt/issues/81

jukzi:
```
junit.framework.AssertionFailedError: Job was not joined 1/100000 times. example: after 80092 tries: executionsAfterJoin: 1/2 expected:<0> but was:<1>
	at org.eclipse.core.tests.runtime.jobs.Bug_574883Join.testJoinLambdaQuick(Bug_574883Join.java:129)
```

jukzi:
> Job was not joined 1/100000 times

Seem there are still problems left with join. Some things that come into mind but seem unrelated: 
* there is no guarantee in which order the different listeners are called. and the join() listener could be called before a test listener
* while the order of events per job is deterministic the order of events per jobfamily is unclear
* ...?

jukzi:
I can reproduce locally "Job was not joined 37/100000000 times"  - only with this patch - without it always joins.

jukzi:
This PR was created as an experiment if it is possible to return from schedule() before scheduled() returns. However it leads to other indeterministic behaviour. 
As the javadoc clearly specifies that `all job state changes, but whether the state change occurs before, during, or after listeners are notified is unspecified.` it is totally fine from specification that schedule() does not return before scheduled() does. The only bad thing about it is:
schedule() does not return before IJobChangeEvent.scheduled() and
scheduled() does not happen before IJobChangeEvent.done().
=> schedule() also waits for IJobChangeEvent.done() to return. And this changed in this eclipse version. It may result in deadlock if schedule() synchronizes with done().

PR #207, opened 2022-10-06
rename all "AllTests" to specific Suites

jukzi:
confusing to have so many same classnames


akurtakov:
Not a problem in this case but as this pattern is used in many places and is in need of sanitization for years, this should help everyone deciding to help with tests:
Keep an eye on test.xml files in the bundle and if you rename the suite used as entry point like in https://github.com/eclipse-platform/eclipse.platform/blob/master/runtime/tests/org.eclipse.core.tests.runtime/test.xml#L48 , change it in the test.xml as otherwise it will no longer be executed in I-build tests.

jukzi:
> Keep an eye on test.xml

fair point. i missed https://github.com/eclipse-platform/eclipse.platform/blob/master/runtime/tests/org.eclipse.core.tests.runtime/test.xml#L56

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/IContentTypeManagerTest.java
```diff
@@ -80,7 +80,7 @@ public void reset() {
 	private final static String XML_ROOT_ELEMENT_EXTERNAL_ENTITY2 = "<?xml version=\"1.0\"?><!DOCTYPE org.eclipse.core.resources.tests.root-element PUBLIC \"org.eclipse.core.resources.tests.root-elementId\" \"org.eclipse.core.resources.tests.root-element.dtd\" ><org.eclipse.core.resources.tests.root-element/>";
 	private final static String XML_ROOT_ELEMENT_ISO_8859_1 = "<?xml version=\"1.0\" encoding=\"ISO-8859-1\"?><org.eclipse.core.resources.tests.root-element/>";
 	private final static String XML_ROOT_ELEMENT_NO_DECL = "<org.eclipse.core.resources.tests.root-element/>";
-	private final static String XML_US_ASCII_INVALID = "<?xml version='1.0' encoding='us-ascii'?><!-- Non-ASCII chars: ����� --><org.eclipse.core.resources.tests.root/>";
+	private final static String XML_US_ASCII_INVALID = "<?xml version='1.0' encoding='us-ascii'?><!-- Non-ASCII chars: ����� --><org.eclipse.core.resources.tests.root/>";
```
jukzi, line 83: what's going on here? eclipse compare does not show a difference here

jukzi:
random failed 
https://github.com/eclipse-platform/eclipse.platform/issues/107
https://github.com/eclipse-platform/eclipse.platform/issues/87

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/IContentTypeManagerTest.java
mickaelistria: We can just ignore that and merge with it.

jukzi: ok.

jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform/issues/210

akurtakov:
Thank you so much! These "small" improvements on tests are what makes huge difference in maintainability.

jukzi:
cheers

laeubi:
> change it in the test.xml as otherwise it will no longer be executed in I-build tests.

Can you explain who is the consumer of this "test.xml"? I digged around a bit in the aggregator repo but can't find how this is used? They seem al a bit like copy & paste and just carry the "classname" as a difference? Are these directly executed as ant-runs? e.g. BND uses a special header to mention the test-classes to execute, amybe we can think about something similar here as well?

akurtakov:
It's what drives https://wiki.eclipse.org/Platform-releng/Eclipse_Test_Framework , which itself is used to run the test suites for different os/jvm combinations for each I-build.

jukzi:
i verified that still the same number of tests are executed. However i did not find 
contenttype.tests
update.configurator.tests
in either https://download.eclipse.org/eclipse/downloads/drops4/I20221010-1840/testResults.php
nor https://download.eclipse.org/eclipse/downloads/drops4/I20220912-1800/testResults.php

(why) are they not executed? or to which test do they contribute?

akurtakov:
Let's handle in new issue. https://github.com/eclipse-platform/eclipse.platform/issues/219

PR #216, opened 2022-10-09
[win] Fix diffs direction for MS Word 2007 and above (#65)

### LINKED ISSUE #65: Compare with Microsoft Word 2007 and above diffs in direction opposite to Word 2003

pbodnar:
Fixes #65.

There was a clear, even if not documented, change since Word version 2007: the document on which we call the "Compare" method needs to be the base, not the revised (typically newer), document. Changes get presented in an opposite (illogical) direction otherwise.

## Testing

Here is a sample Git project to reproduce the issue: https://github.com/pbodnar/eclipse-platform-65-word-diff-sample. So from Eclipse, one can easily run diffs either on "doc.\*" files history (this PR is/was primarily about fixing this diff), or on 2 selected files ("v1.\*" vs. "v2.\*"). Of course, a desired version of MS Word needs to be installed on the testing computer as well. 

Text files are also included in the sample project, so one can compare the diffing behavior.

### DIFF
--- a/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordComparison.java
+++ b/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordComparison.java
@@ -37,6 +37,8 @@ import org.eclipse.ui.IWorkbenchWindow;
  */
 public class WordComparison {
 
+	private static final int OFFICE_VER_2007 = 12;
+
 	private final OleFrame frame;
 	private OleClientSite site;
 	private boolean inplace;
@@ -198,10 +200,20 @@ public class WordComparison {
 		resetSite(null);
 		OleAutomation application = createApplication();
 		try {
-			OleAutomation document = openDocument(application, revisedDocument);
+			final String primaryDocument;
+			final String secondDocument;
+			if (getWordVersion(application) >= OFFICE_VER_2007) {
+				primaryDocument = baseDocument;
+				secondDocument = revisedDocument;
+			} else {
+				primaryDocument = revisedDocument;
+				secondDocument = baseDocument;
+			}
+
+			OleAutomation document = openDocument(application, primaryDocument);
 			try {
 				setDocumentVisible(document,false);
-				compareDocument(document, baseDocument, revisedDocument);
+				compareDocument(document, secondDocument, primaryDocument);
 				OleAutomation activeDocument = getActiveDocument(application);
 				try {
 					Variant varResult = invoke(activeDocument, document, "SaveAs", workingCopy); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordComparison.java
```diff
@@ -198,10 +200,20 @@ public void createWorkingCopy(String baseDocument, String revisedDocument, Strin
 		resetSite(null);
 		OleAutomation application = createApplication();
 		try {
-			OleAutomation document = openDocument(application, revisedDocument);
+			final String primaryDocument;
```
pbodnar, line 203: Note: While I could just switch the 2 variables here, I think it is an anti-pattern to modify references to method arguments (Kotlin even won't let you do so), so I better introduced new variables here.

team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordComparison.java
```diff
@@ -304,6 +316,19 @@ private OleAutomation getActiveDocument(OleAutomation application) {
 		return getAutomationProperty(application, "ActiveDocument"); //$NON-NLS-1$
 	}
 
+	private int getWordVersion(OleAutomation application) {
```
pbodnar, line 319: Not sure if it is worth it adding some error handling for the case that parsing of Word (Office) version would fail here. What do you think?

pbodnar:
So here comes the preview of the fix of #65. It is basically ready for review.

Just one related problem I found when testing: When comparing 2 files, like e.g. `v1.docx` and `v2.docx`, changes are presented **opposite** to when comparing file versions _from history_, which I consider is a mistake. I think that it _might_ be that historically this Word compare feature was tested with Office 2007 already, but just on this `v1.docx` vs. `v2.docx` scenario. Here is the issue description / problematic part in the form of a diff:

```diff
diff --git a/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordMergeViewer.java b/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordMergeViewer.java
index 7a21bd98f4..2876973ec0 100644
--- a/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordMergeViewer.java
+++ b/team/bundles/org.eclipse.compare.win32/src/org/eclipse/compare/internal/win32/WordMergeViewer.java
@@ -331,6 +331,9 @@ public class WordMergeViewer extends AbstractMergeViewer implements IFlushable,
                                        File result = getResultFile();
                                        int direction = getCompareInput().getKind() & Differencer.DIRECTION_MASK;
                                        File base, revised;
+                                       // TODO PBO It is not clear why base and revised get swapped here - opposite to
+                                       // natural order when "v2.docx" (right) should be treated as revised and
+                                       // "v1.docx" (left) as base
                                        if (direction == Differencer.RIGHT) {
                                                base = left;
                                                revised = right;
```

So what do you think? Is it safe to swap the logic here to fix this inconsistency? Or is it possible I have overlooked something? I'm not sure if I have enough time to study another pieces of code in order to grasp what exactly is going on here... :P

pbodnar:
Also, if somebody could please test this with Word 2007 and Word 2003, it would be great. I only have Office 365 available on my machine and it looks like it is not quite easy to install other versions of Office, even if just for testing...

jukzi:
@pbodnar I understand your comment https://github.com/eclipse-platform/eclipse.platform/issues/65#issuecomment-1430282920 that this PR won't fix the issue. Can we close it?

pbodnar:
@jukzi, yeah, I think you can close it. Although it _does_ fix the comparison logic in the `WordComparison` class, unfortunately as described, there are other problems in the Eclipse code base which render this fix "useless" in various scenarios. And there doesn't seem to be enough drive from the community to fix those...

jukzi:
> drive from the community to fix those...

Feel free to contribute a fix for those, personally i never use word compare.

PR #229, opened 2022-10-15
Improve JobManager IJobChangeListener Notification #193

jukzi:
* Fall back to asynchronous events if IJobChangeListener timeout detected (possible deadlock).

* Make sure all IJobChangeListener calls returned before a state change happens to avoid stale Events to happen after the state change.

* Fix order of scheduled() and done() in case of reschedule: Some IJobChangeListener like ProgressManager.createChangeListener() expect a scheduled() to happen after done() even if a rescheduling schedule() was called before done() happened.

* Fix cancel of InternalJobGroup (did synchronize wrong lock during cancel events)

* avoid nested "synchronized (lock)"

https://github.com/eclipse-platform/eclipse.platform/issues/193


jukzi:
fixes https://github.com/eclipse-platform/eclipse.platform.ui/pull/413

PR #230, opened 2022-10-17
testNestedSchedule #193

jukzi:


jukzi:
@iloveeclipse please improve test to fail, if possible

iloveeclipse:
> I will check what's the difference to our application code is.

OK, they do both specify exclusive rule (same on both jobs) and additionally do some calculation who should run or not in "shouldSchedule()"... I'm trying to separate a test case.

Issue #237, opened 2022-10-20
random failing JobTest.testSetThread


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20221019-1800/testresults/html/org.eclipse.core.tests.runtime_ep426I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
```
10.0

junit.framework.AssertionFailedError: 10.0
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.assertTrue(Assert.java:22)
at junit.framework.Assert.assertNotNull(Assert.java:256)
at junit.framework.TestCase.assertNotNull(TestCase.java:399)
```

jukzi:
Locally test passes when i execute the whole testclass but reproducible fails when only executing that single test with
```
java.lang.AssertionError: 2.0. Actual: Thread[main,5,main]
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failEquals(Assert.java:187)
	at org.junit.Assert.assertNotEquals(Assert.java:163)
	at org.eclipse.core.tests.runtime.jobs.JobTest.testSetThread(JobTest.java:1652)

```

jukzi:
InternalJob.ABOUT_TO_RUN decodes to Job.RUNNING in InternalJob.getState()
ABOUT_TO_RUN is set in JobManager.nextJob() without setting a thread.
The thread is only set in JobManager.startJob(Worker) when the internal state ABOUT_TO_RUN is changed to RUNNING.

Issue #99, opened 2022-04-25
random failing HiddenResourceTest.testDeltas


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20220424-1800/testresults/html/org.eclipse.core.tests.resources_ep424I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
Verifier has not yet been given a resource delta

java.lang.AssertionError: Verifier has not yet been given a resource delta
at org.junit.Assert.fail(Assert.java:89)
at org.eclipse.core.tests.resources.ResourceDeltaVerifier.getMessage(ResourceDeltaVerifier.java:466)
at org.eclipse.core.tests.resources.HiddenResourceTest.testDeltas(HiddenResourceTest.java:618)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at junit.framework.TestCase.runTest(TestCase.java:177)
at junit.framework.TestCase.runBare(TestCase.java:142)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at junit.framework.TestSuite.run(TestSuite.java:236)
at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:90)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
[...]

PR #238, opened 2022-10-20
fix random failing JobTest.testSetThread #237

jukzi:
use internalGetState instead of getState() to avoid ABOUT_TO_RUN


jukzi:
ignoring https://github.com/eclipse-platform/eclipse.platform/issues/99

PR #242, opened 2022-10-21
JobManager: do not send notifications while holding lock #193

jukzi:


jukzi:
ignoring freeze. "M2 is a lightweight release so there will be no freeze. "

Issue #243, opened 2022-10-21
ObjectNotFoundException: Tree element '/' not found


iloveeclipse:
We observe this error with I20221020-1800 & our application tests.
This looks like regression from https://github.com/eclipse-platform/eclipse.platform/pull/114.

```
!ENTRY org.eclipse.core.resources 4 2 2022-10-21 09:24:38.127
!MESSAGE Problems occurred when invoking code from plug-in: "org.eclipse.core.resources".
!STACK 0
org.eclipse.core.internal.dtree.ObjectNotFoundException: Tree element '/' not found.
	at org.eclipse.core.internal.dtree.AbstractDataTree.handleNotFound(AbstractDataTree.java:178)
	at org.eclipse.core.internal.dtree.DeltaDataTree.getData(DeltaDataTree.java:572)
	at org.eclipse.core.internal.events.ResourceDeltaFactory.createDelta(ResourceDeltaFactory.java:137)
	at org.eclipse.core.internal.events.ResourceDeltaFactory.computeDelta(ResourceDeltaFactory.java:64)
	at org.eclipse.core.internal.events.BuildManager.lambda$7(BuildManager.java:964)
	at org.eclipse.core.internal.events.BuildManager$DeltaCache.lambda$0(BuildManager.java:67)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1133)
	at org.eclipse.core.internal.events.BuildManager$DeltaCache.computeIfAbsent(BuildManager.java:67)
	at org.eclipse.core.internal.events.BuildManager.getDeltaCached(BuildManager.java:958)
	at org.eclipse.core.internal.events.BuildManager.hasDelta(BuildManager.java:1296)
	at org.eclipse.core.internal.events.BuildManager.needsBuild(BuildManager.java:1288)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:216)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:303)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:392)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:395)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:506)
[...]

PR #245, opened 2022-10-25
Fix for "ObjectNotFoundException: Tree element '/' not found"

iloveeclipse:
Don't try to compute delta for projects neither existing in the workspace nor in any of the build trees.
ResourceDeltaFactory.computeDelta() can't deal with that and will fail with ObjectNotFoundException, breaking build.

Application use case: importing an Xtext based project that references another (not yet imported) project via .project file. Xtext builder reports it is interested in the project that isn't there (but is referenced in the .project file).

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/243


jukzi:
Thanks for fixing this, Andrey!

PR #249, opened 2022-10-28
Synchronize IJobChangeListener with Job State change #193

jukzi:
Possible deadlock in IJobChangeListener
---------------------------------------

Attention: The JobManager implementation was changed! The old JobMangager implementation notified IJobChangeListener about various IJobChangeEvent without strict order in various Threads. For example scheduled() was called in a Thread that did call schedule() on a Job while done() was normally called within a Worker Thread. In case of a repeated schedule() both Notifications could run in parallel. That was a race condition. The Listener could not deterministically know if the Job was still scheduled or already done. As a consequence a join() could have returned too early or the Progress View did not show Jobs that did still run. And probably many other things went wrong totally unnoticed.
Even multiple notifications could have happened in parallel in various Worker Threads, as there is no guarantee that the next execution is done in the same Worker.

To fix this indeterministic behavior all notifications for a Job will now be done one after the other. It is still not defined in which Thread the notification happens, but the new implementation will not call any IJobChangeListener until all pending events for the jobs are processed by all registered IJobChangeListener. The new implementation will also wait with any job state change until all calls to IJobChangeListener for that job returned.

The IJobChangeListener javadoc clearly specifies "whether the state change occurs before, during, or after listeners are notified is unspecified." - and the implementation changed within that broad specification. Unfortunately some IJobChangeListener around rely on the old implementation. They may now deadlock. For example the following snippet can now deadlock:

```
 job.schedule(); // may result in done();
 synchronized (lock){
  boolean schedule = ...; // lock needed for reasons
  if (schedule ) job.schedule(); // schedule() will notify scheduled()
  // - which may not return before previous Notifications return!
 }

 public void done(IJobChangeEvent event) {
  // can not enter while other Thread holds lock:
  synchronized (lock) {// possible deadlock
  }
 }
```

Instead use:

```
 job.schedule();
 boolean schedule;
 synchronized (lock) {
  schedule=...; // lock needed for reasons
 }
 if (schedule) job.schedule();

 public void done(IJobChangeEvent event) {
  synchronized (lock) {// OK
  }
 }
```

The same problem may occur on all IJobChangeListener methods and all methods that result in IJobChangeEvent being sent: Job.schedule(), cancel(), wakeUp(), sleep() and JobManager shutdown. Make sure all IJobChangeListener implementations do not block and return promptly.

Due to the extreme risk the new Implementation tries to identify non-conforming IJobChangeListener and do a fall back: When a IJobChangeEvent is not handled within 3 seconds a Timeout error is logged with stacktraces of the competing Threads and the JobManager will no longer wait - until JVM is restarted. Further calls to IJobChangeListener may occur in random order and join(family) can return too soon! However there may also be some deadlocks in Clients code due to the changed JobManager implementation that may be undetected by JobManager. **So better check your IJobChangeListener!**

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -21,7 +21,6 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.*;
@@ -161,7 +160,7 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	 * Used to synchronize Job listener notification
 	 */
 	final Queue<JobChangeEvent> eventQueue = new ConcurrentLinkedQueue<>();
-	final Lock eventQueueLock = new ReentrantLock();
+	final ReentrantLock eventQueueLock = new ReentrantLock();
 	final AtomicReference<Thread> eventQueueThread = new AtomicReference<>();
 
 	private static synchronized int getNextJobNumber() {
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobListeners.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobListeners.java
@@ -52,7 +52,13 @@ public class JobListeners {
 	protected final ListenerList<IJobChangeListener> global = new ListenerList<>(ListenerList.IDENTITY);
 
 	/** Send=true should not be used during a lock */
-	void waitAndSendEvents(InternalJob job, boolean send) {
+	void waitAndSendEvents(InternalJob job, boolean shouldSend) {
+		// Instead of just waiting this threads may also need to help to send
+		// to make sure there is progress.
+		// For example during a Job.cancel() within a IJobChangeListener.scheduled()
+		// this thread already has a lock for sending so it can not wait for another
+		// Thread. See JobTest.testCancelAboutToSchedule() for an example.
+		boolean send = shouldSend || job.eventQueueLock.isHeldByCurrentThread();
 		// Synchronize eventQueue to get a stable order of events across Threads.
 		// There is however no guarantee in which Thread the event is delivered.
 		while (!job.eventQueue.isEmpty()) {
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
@@ -26,6 +26,7 @@ package org.eclipse.core.internal.jobs;
 
 import java.util.*;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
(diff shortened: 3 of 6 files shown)


jukzi:
@iloveeclipse please review - especially the commit message as template for sending an email to cross project

iloveeclipse:
I've briefly read through the PR message, I can't follow on the "Instead use:" case, which doesn't seem to do anything meaningful in the first synchronized block anymore.

> The IJobChangeListener javadoc clearly specifies "whether the state change occurs before, during, or after listeners are notified is unspecified." - and that detail changed. 

If so, we should update javadoc to match the current state and clearly describe what is to expect.

> will not call any IJobChangeListener before the last IJobChangeEvent was processed by all relevant IJobChangeListener

Not clear here what "relevant" is.

Was it meant: "no new events for same job will be sent until all pending events for this job are processed by listeners" ?

>  The new implementation will also not do any job state change before all calls to IJobChangeListener for that job returned.

Does it mean: "The new implementation *will block* on job state change if any of the IJobChangeListener callback on same job is still in progress"?

> and all calls to Job that causes a IJobChangeEvent.

"and all Job state changes that result in IJobChangeEvent being sent?"

Would be good to explicitly list which Job methods are changing Job state & now dangerous to use because sending events.
schedule(), cancel(), sleep(), wakeUp() - what else?


Reading that above, I still can't understand the rationale to wait with (or block on) job state changes until all previously created job events are sent. If the events are queued anyway, why can't events sending be decoupled from job state changes?

If sending was done in parallel before, and one could change job state while some listener still processed some intermediate events, and not this is not possible - it sounds like a breaking behavior change now.

jukzi:
> I've briefly read through the PR message, I can't follow on the "Instead use:" case, which doesn't seem to do anything meaningful in the first synchronized block anymore.

updated. better?

> If so, we should update javadoc to match the current state and clearly describe what is to expect.

updated: "the implementation changed within that broad specification."

It is pretty good that the specification was so unspecific in this case. I don't want to expose the current implementation to the API. So that if somebody comes of with a better solution it can be used.

> Not clear here what "relevant" is.

"registered"

> Was it meant: "no new events for same job will be sent until all pending events for this job are processed by listeners" ?

yes

> > The new implementation will also not do any job state change before all calls to IJobChangeListener for that job returned.
> 
> Does it mean: "The new implementation _will block_ on job state change if any of the IJobChangeListener callback on same job is still in progress"?

"The new implementation will also wait with any job state change until all calls to IJobChangeListener for that job returned."

> > and all calls to Job that causes a IJobChangeEvent.
> 
> "and all Job state changes that result in IJobChangeEvent being sent?"
> 
> Would be good to explicitly list which Job methods are changing Job state & now dangerous to use because sending events. schedule(), cancel(), sleep(), wakeUp() - what else?

" all methods that result in IJobChangeEvent being sent: Job.schedule(), cancel(), wakeUp(), sleep() and JobManager shutdown."

> Reading that above, I still can't understand the rationale to wait with (or block on) job state changes until all previously created job events are sent. If the events are queued anyway, why can't events sending be decoupled from job state changes?
> 

Excellent question. While it does not affect IJobChangeListener the JobManager internal methods join() and join(family) synchronize on job state change and then expect to receive meaningful IJobChangeEvents afterwards. I.e. they assume that any event that is received after the lock was queued after the lock. That would not be the case if an notification happens parallel to job state change. Outdated events could let join() return too early.
Another implementation could be that those joins in a first step would wait till all events for the relevant jobs are processed before the read lock. But that is hard (or impossible?) for the family join: it is unclear in advance which jobs will belong to the family.

> If sending was done in parallel before, and one could change job state while some listener still processed some intermediate events, and not this is not possible - it sounds like a breaking behavior change now.

It's not breaking: "whether the state change occurs before, during, or after listeners are notified is unspecified." - now it's just never "during".

iloveeclipse:
@jukzi : I'm on vacation this week, if you are confident in PR, feel free to merge. M3 is soon, so will be better to have some "final" state of the Job changes merged & tested before M3.

Issue #252, opened 2022-11-02
random failing JobGroupTest.testJobManagerPublishesJobGroupResults


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20221028-1800/testresults/html/org.eclipse.core.tests.runtime_ep426I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20221029-1800/testresults/html/org.eclipse.core.tests.runtime_ep426I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20221030-1800/testresults/html/org.eclipse.core.tests.runtime_ep426I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

but not on day 26,27,31
~ every 2nd day, only on macos

```
junit.framework.AssertionFailedError: Should have seen as many job completion events as the count of jobs in the job group. expected:<3> but was:<2>
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.failNotEquals(Assert.java:329)
at junit.framework.Assert.assertEquals(Assert.java:78)
at junit.framework.Assert.assertEquals(Assert.java:234)
at junit.framework.TestCase.assertEquals(TestCase.java:377)
at org.eclipse.core.tests.runtime.jobs.JobGroupTest.testJobManagerPublishesJobGroupResults(JobGroupTest.java:1456)
```

PR #254, opened 2022-11-02
fix testJobManagerPublishesJobGroupResults #252

jukzi:


jukzi:
ignoring https://github.com/eclipse-platform/eclipse.platform/issues/136

PR #261, opened 2022-11-08
Test improvements

akurtakov:


jukzi:
@akurtakov the commit "Remove useless assert/fail methods from CoreTest" broke a lot of tests on windows. 
(see https://github.com/eclipse-platform/eclipse.platform/issues/265)
reproducible locally for example: org.eclipse.core.tests.resources.LinkedResourceTest.testBlockedFolder()
`org.eclipse.core.internal.resources.ResourceException: Problems encountered while deleting resources.
`
please fix or revert

jukzi:
Please note that the removed CoreTest.assertClose() did not only asserted but was used to close the files

akurtakov:
All assertClose should be in try-with-resources with this change so supposedly not an issue unless I've missed some. I'm looking into it.

jukzi:
The org.eclipse.core.resources.IFile.getContents() then also needs to be used as try-with-resources

akurtakov:
I have opened https://github.com/eclipse-platform/eclipse.platform/pull/274 with the initial issue you pointed to to see windows failures proper and will build on top of that to get them fixed. Now that there is verification job showing the issue things should be easier.

PR #268, opened 2022-11-14
Remove unused require bundle (that was even marked optional)

laeubi:
Seems not used anyways...


laeubi:
From the history, this was added to use DS annotations ...

laeubi:
@akurtakov can this be merged or is there any freeze period? But I think this can also wait so I'll leave this up to you :-)

akurtakov (approved):
I'm fine with merging this one for RC1

akurtakov:
In a way it's releng improvement :) .

Issue #265, opened 2022-11-10
win: problems with deleting files during core.tests.resources since 20221108


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20221108-1800/testResults.php
`The process cannot access the file because it is being used by another process.`
My best guess is something like a virusscanner is scanning all newly created files.

tomaswolf:
Note that this Windows error message is misleading. It might also be that the file is open within the same process.

PR #272, opened 2022-11-18
Add windowsPR GH action verification

akurtakov:
Pre-req to investigate
https://github.com/eclipse-platform/eclipse.platform/issues/265

### DIFF
--- a/.github/workflows/windowsPR.yml
+++ b/.github/workflows/windowsPR.yml
@@ -1,0 +1,18 @@
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 17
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov you better can use a matrix build here, see for example the Tycho workflow.

laeubi:
We might even add mac as well ...

akurtakov:
Do you want to enhanche this PR ? My GH action-foo is still not good enough but I try ;) .

akurtakov:
Right now I would probably keep the matrix to the minimum and extend when needed to not delay win investigation with possible mac issues.

laeubi:
I'll propose a patch!

akurtakov:
> I'll propose a patch!

Thanks a lot!

akurtakov:
Close in favor of https://github.com/eclipse-platform/eclipse.platform/pull/273 as @laeubi teaches me :)

PR #274, opened 2022-11-18
Fix failing tests on windows

akurtakov:
Ensure stream is closed.


akurtakov:
Reverted the suspicious patch so this one will no longer be investigated.

PR #277, opened 2022-11-19
Test if the smartbuilder already improves things

laeubi:
Just for testing ...

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -5,4 +5,9 @@
 		<artifactId>tycho-build</artifactId>
 		<version>3.0.0</version>
 	</extension>
+	<extension>
+		<groupId>io.takari.maven</groupId>
+		<artifactId>takari-smart-builder</artifactId>
+		<version>0.6.1</version>
+	</extension>
 </extensions>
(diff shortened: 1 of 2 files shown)


laeubi:
Result with smartbuilder:

**Jenkins** Segment walltime 983 s, segment projects service time 1948 s, effective/maximum degree of concurrency 1.98/2 (Job Time 21 minutes)

laeubi:
With 4 CPUs we get

 Segment walltime 639 s, segment projects service time 1843 s, effective/maximum degree of concurrency 2.88/4

and a total build time of 13 minutes compared to 27 minutes average with two CPUs and without smart builder.

laeubi:
@akurtakov Smartbuilder is now enabled by default in Tycho, so this change is not required anymore, but my measures seem to indicate that we can improve the build time by having runners with 3 CPUs, just in case you like to adjust the `Jenkinsfile` if it is appropriate.

akurtakov:
Infra is not providing such builders by default AFAIK.

laeubi:
Yes, but one can configure this in the Jenkinsfile (if desired):
https://wiki.eclipse.org/Jenkins#What_is_killing_my_build.3F_I.27m_using_custom_containers.21

I just heard that people complaining about build-times, so if we can get more CPU in the agents, and possible even some more spare resource-packs build times can be improved (here: cut by ~50%)

Issue #282, opened 2022-11-23
TextMergeViewer deadlock 


jukzi:
My eclipse IDE just stuck. I had a git pull which reported a change in "build.properties" and "MANIFEST.MF". I opened compare viewer and used "replace with" in Git stating view for both files.
Swt Thread "main" waits for a worker thread that contrary waits that something in SWT finished = deadlock:

2022-11-23 11:53:26
Full thread dump OpenJDK 64-Bit Server VM (17.0.5+8 mixed mode, sharing):

```
"main" #1 prio=6 os_prio=0 cpu=76359.38ms elapsed=3218.47s tid=0x000001f661ecc910 nid=0x590c waiting for monitor entry  [0x0000003e654fd000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.eclipse.core.internal.filebuffers.SynchronizableDocument.getModificationStamp(SynchronizableDocument.java:151)
        - waiting to lock <0x000000075e6b4548> (a java.lang.Object)
        at org.eclipse.text.undo.DocumentUndoManager$UndoableTextChange.canUndo(DocumentUndoManager.java:163)
        at org.eclipse.core.commands.operations.DefaultOperationHistory.canUndo(DefaultOperationHistory.java:269)
        at org.eclipse.text.undo.DocumentUndoManager.undoable(DocumentUndoManager.java:835)
        at org.eclipse.jface.text.TextViewerUndoManager.undoable(TextViewerUndoManager.java:348)
        at org.eclipse.jface.text.TextViewer.canDoOperation(TextViewer.java:3886)
        at org.eclipse.jface.text.source.SourceViewer.canDoOperation(SourceViewer.java:816)
        at org.eclipse.compare.internal.MergeSourceViewer$TextOperationAction.isEnabled(MergeSourceViewer.java:139)
        at org.eclipse.compare.internal.MergeSourceViewer$TextOperationAction.update(MergeSourceViewer.java:144)
        at org.eclipse.compare.internal.MergeSourceViewer.updateContentDependantActions(MergeSourceViewer.java:818)
        at org.eclipse.compare.internal.MergeSourceViewer$$Lambda$2457/0x00000008023456f0.run(Unknown Source)
        at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
        at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
[...]

PR #283, opened 2022-11-23
Compare: asyncExec propertyChange() to avoid Deadlock #282

jukzi:
SynchronizableDocument holds a lock that is used by IPropertyChangeListener.  This deadlocked if the document was updated in background thread as the lock is only reentrant inside a thread.

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/Utilities.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/Utilities.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -178,11 +178,7 @@ public class Utilities {
 				SafeRunner.run(() -> listener.propertyChange(event));
 			}
 		};
-		if (Display.getCurrent() == null) {
-			Display.getDefault().syncExec(runnable);
-		} else {
-			runnable.run();
-		}
+		Display.getDefault().execute(runnable);
 	}
 
 	public static boolean okToUse(Widget widget) {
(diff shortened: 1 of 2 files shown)


vogella:
Can you add the version update, maybe as separate commit before the functional change?

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/Utilities.java
```diff
@@ -179,7 +179,7 @@ public static void firePropertyChange(final ListenerList<IPropertyChangeListener
 			}
 		};
 		if (Display.getCurrent() == null) {
-			Display.getDefault().syncExec(runnable);
+			Display.getDefault().asyncExec(runnable);
 		} else {
 			runnable.run();
 		}
```
mickaelistria, line 185: What about leveraging the new `Display.getDefault().execute(runnable)`?

jukzi:
> Can you add the version update, maybe as separate commit before the functional change?

interactive rebase for the win. my first time.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/Utilities.java
jukzi: ok

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/Utilities.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2021 IBM Corporation and others.
```
szarnekow, line 2: 2022 since I last checked ;)

vogella:
@mickaelistria WDYT? Something for RC2 or better to wait for the next release cycle?

mickaelistria:
As far as I know, this is not a regression, so not worth RC2.

vogella:
@jukzi Time to merge this?

jukzi:
@vogella i did not have time to test it yet

jukzi:
ignoring random failing https://github.com/eclipse-platform/eclipse.platform/issues/460

PR #286, opened 2022-11-30
Updating spash screen for 2023-03

SDawley:
Tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/716


akurtakov:
Let's keep the png instead of moving to jpg.

PR #289, opened 2022-12-02
The auto-builder shouldn't be interrupted by its own workspace changes.

pzi-dsa:
If a worker job is created from within the auto-build job, performing
workspace changes, it is no longer possible to gracefully cancel the
job.

Those changes cause the auto-builder to reschedule itself over and over
again, until the build finished successfully or the auto-builder is
disabled.

This unit-test should illustrate the problem described above.

**Expectation**:

The auto-builder should start due to the workspace modification, respond to the user cancellation and then switch into an idle state.

**Result**:

The auto-builder re-schedules itself indefinitely, resulting in an infinite loop until the automatic build is deactivatied.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2015 IBM Corporation and others.
+ * Copyright (c) 2003, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -201,8 +201,7 @@ class AutoBuildJob extends Job implements Preferences.IPropertyChangeListener {
 				setInterrupted(!sleep());
 				break;
 			case RUNNING :
-				//make sure autobuild doesn't interrupt itself
-				if (Job.getJobManager().currentJob() == this)
+				if (!shouldInterrupt())
 					return;
 				setInterrupted(true);
 				break;
@@ -296,6 +295,18 @@ class AutoBuildJob extends Job implements Preferences.IPropertyChangeListener {
 			//regardless of the result, clear the build flags for next time
 			forceBuild = avoidBuild = buildNeeded = false;
 		}
+	}
+
+	/**
+	 * Returns {@code true}, if the auto-build job should be interrupted. The job
+	 * should always be interrupted, unless this method is called by a job within
+	 * the {@link ResourcesPlugin#FAMILY_AUTO_BUILD} family.
+	 */
+	private synchronized boolean shouldInterrupt() {
+		Job currentJob = Job.getJobManager().currentJob();
+
+		// make sure autobuild doesn't interrupt itself
+		return currentJob == null || !currentJob.belongsTo(ResourcesPlugin.FAMILY_AUTO_BUILD);
 	}
 
 	/**
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilder.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilder.java
@@ -1,0 +1,115 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.tests.internal.builders;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.eclipse.core.resources.*;
+import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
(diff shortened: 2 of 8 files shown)


mickaelistria:
Using and checking the build family seems like a good idea, no new API and very little prone to regression (and after all, would that even be regression as it's not specified how the family is used).

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilderTest.java
```diff
@@ -0,0 +1,102 @@
...
+	private IFile file;
+
+	public InterruptingBuilderTest() {
+		super("Interrupting Builder");
+	}
+
+	@Override
+	public void setUp() throws Exception {
+		super.setUp();
+		setAutoBuilding(false);
```
jukzi, line 39: Typically you have to wait till autobuild job finished in junit tests after disabling it

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
+
+/**
+ * This class delegates "performance-heavy" computation to a worker thread.
+ */
+public class InterruptingBuilder extends IncrementalProjectBuilder {
+
+	public static final String BUILDER_NAME = "org.eclipse.core.tests.resources.interruptingBuilder";
+	private static CountDownLatch LATCH = new CountDownLatch(1);
```
jukzi, line 31: We use all-capitals only for constants. (final immutable)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+
+		public WorkerJob() {
+			super("Worker");
+		}
+
+		@Override
+		protected IStatus run(IProgressMonitor monitor) {
+			try {
+				// Raises the interrupt flag of the auto-builder
+				getProject().getFile(".project").touch(monitor);
```
jukzi, line 87: I don't get that use-case: a Builder should touch the workspace only inside the build thread. do you try to emulate a user pressed "cancel" here?

jukzi (changes requested):
You complain that the  autobuild is reschedule over and over. But instead fixing the reschedule your change that it is not (self-)interrupted anymore?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -298,6 +297,18 @@ private synchronized boolean shouldBuild() {
...
+	/**
+	 * Returns {@code true}, if the auto-build job should be interrupted. The job
+	 * should always be interrupted, unless this method is called by a job within
+	 * the {@link ResourcesPlugin#FAMILY_AUTO_BUILD} family.
+	 */
+	private synchronized boolean shouldInterrupt() {
+		Job currentJob = Job.getJobManager().currentJob();
+
+		// make sure autobuild doesn't interrupt itself
+		return currentJob == null || !currentJob.belongsTo(ResourcesPlugin.FAMILY_AUTO_BUILD);
```
iloveeclipse, line 309: note, that currentJob == null check is basically an invite to modify resources from **any** non-job thread, and the following family check works not only for the actual autobuild job but for **any** job that decides to be in same family.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/InterruptingBuilder.java
pzi-dsa: > a Builder should touch the workspace only inside the build thread.

That's right. If the touch is done from within the auto-build job, the interruption is ignored. But not if the change is done from within any other job. Regardless of whether it was scheduled by the auto-build job.

> do you try to emulate a user pressed "cancel" here?

Yes. The interrupt flag is cleared, when the build finishes succesfully. Which means I have to avoid this by canceling the job.

I wonder if it's sufficient to just call monitor.cancel() instead. In any case, I will clarify the commint.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
pzi-dsa: > note, that currentJob == null check is basically an invite to modify resources from any non-job thread

I don't think that's the case. If `currentJob == null`, then the interrupt flag is set.
Which is how it's currently working with the `Job.getJobManager().currentJob() == this` check.

>  the following  family check works not only for the actual autobuild job but for any job that decides to be in same family.

That's the general idea. But going over the family is also how other frameworks like XText or the Debug UI plugin observe the state of the auto-builder, given that the BuildManager class is only an internal class.

Obviously it should only be used when necessary, but I think this is an appropriate use case.

jukzi:
I still don't think it's a good usecase. Aou would be better move writes to builder thread. But i guess the change won't break anything either. Or is there any build out there that may intentional use that pattern to rebuild - i don't know. @szarnekow ?

szarnekow:
> Or is there any build out there that may intentional use that pattern to rebuild

Code that wants to trigger a rebuild needs to request it explicitly.

I'm not sure about relaxing the sandbox of the AutoBuilder. The only thread that's supposed to write while a resource tree is locked, is the thread holding that lock. Not sure why other threads should be able to modify the resource tree mid-flight.

To me, it's much more helpful to follow the proven practice of other builders (e.g. JDT) and process in parallel but write sequentially. Everything else will eventually lead to trouble with the consistency of the resource tree, I guess.

pzi-dsa:
If there is this much concern regarding the auto-build, then I don't think it makes much sense to pursue this pull request any further.

Thank you all for the feedback!

PR #297, opened 2022-12-13
Race condition in AutoBuildJob.build() (#158)

iloveeclipse:
It can happen that the AutoBuildJob's buildNeeded flag is true, but the
AutoBuildJob is not scheduled and not executed until the next workspace
change.

The race condition happens when the AutoBuildJob in doBuild() method
reaches last line which releases the workspace lock, but the job is
still in RUNNING state.

At same time client code calls AutoBuildJob.build(true), the buildNeeded
flag is set to true and then the method checks the state of the
AutoBuildJob and finds that it is currently (still) in the RUNNING
state, so it does *not* schedule it again (it falls through the
switch(state).

In this situation, the AutoBuild is not scheduled even though it is
marked as being in need of a build, so until the next workspace change
triggering a build, this situation won't change and the AutoBuildJob
will not be executed.

- Let build() reschedule the AutoBuildJob even if it is currently
running.
- Remove additional reset of buildNeeded flag in doBuild(), so it does
not interfere with other threads concurrently requesting a build.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/158

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
@@ -88,20 +88,26 @@ class AutoBuildJob extends Job implements Preferences.IPropertyChangeListener {
 		buildNeeded |= needsBuild;
 		long delay = computeScheduleDelay();
 		int state = getState();
-		if (Policy.DEBUG_BUILD_NEEDED)
-			Policy.debug("Auto-Build requested, needsBuild: " + needsBuild + " state: " + state + " delay: " + delay); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-		if (needsBuild && Policy.DEBUG_BUILD_NEEDED_STACK && state != Job.RUNNING)
-			Policy.debug(new RuntimeException("Build needed")); //$NON-NLS-1$
+		if (Policy.DEBUG_BUILD_NEEDED) {
+			Policy.debug("build requested, needsBuild: " + needsBuild + " state: " + state + ", delay: " + delay); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+		}
+
 		//don't mess with the interrupt flag if the job is still running
 		if (state != Job.RUNNING)
 			setInterrupted(false);
 
 		switch (state) {
 			case Job.SLEEPING :
+				if (Policy.DEBUG_BUILD_INVOKING) {
+					traceMessageOrFullStack("wakeup, needsBuild: " + needsBuild + ", delay: " + delay); //$NON-NLS-1$ //$NON-NLS-2$
+				}
 				wakeUp(delay);
 				break;
 			case NONE :
 				if (isAutoBuilding) {
+					if (Policy.DEBUG_BUILD_INVOKING) {
+						traceMessageOrFullStack("scheduled, needsBuild: " + needsBuild + ", delay: " + delay); //$NON-NLS-1$//$NON-NLS-2$
+					}
 					schedule(delay);
 				} else {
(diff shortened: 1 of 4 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -95,6 +95,7 @@ synchronized void build(boolean needsBuild) {
 		//don't mess with the interrupt flag if the job is still running
 		if (state != Job.RUNNING)
 			setInterrupted(false);
```
jukzi, line 97: It feels wrong, that this simple check is unrelated too all those more complicated conditions below.  is it because not running implies current thread is not the autobuild?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -111,6 +112,17 @@ synchronized void build(boolean needsBuild) {
...
 				break;
+			case RUNNING:
+				// In rare cases, we can end up in a situation where some other thread calls
+				// WorkManager.setBuild(true) while the AutoBuildJob is already running.
+				// In this case the AutoBuildJob should be rescheduled so we are certain that
+				// the new build request is eventually processed and does not get lost.
+				// Therefore, if a build is needed and the autobuild is enabled, then reschedule
+				// the job
+				if (isAutoBuilding && buildNeeded && !avoidBuild && Job.getJobManager().currentJob() != this) {
+					schedule(delay);
```
jukzi, line 123: what's the difference between wakeUp() in line 101 and schedule() in 123? wouldn't a schedule imply a wakeup?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -111,6 +112,17 @@ synchronized void build(boolean needsBuild) {
...
 				}
 				break;
+			case RUNNING:
+				// In rare cases, we can end up in a situation where some other thread calls
+				// WorkManager.setBuild(true) while the AutoBuildJob is already running.
+				// In this case the AutoBuildJob should be rescheduled so we are certain that
+				// the new build request is eventually processed and does not get lost.
+				// Therefore, if a build is needed and the autobuild is enabled, then reschedule
+				// the job
+				if (isAutoBuilding && buildNeeded && !avoidBuild && Job.getJobManager().currentJob() != this) {
```
jukzi, line 122: it feels wrong that the conditions here are different to that in line 93. at least the logging in line 94 is probably wrong.

jukzi:
I would like to see a test that if a build was explicitly cancelled by user that it does not automatically start again (Bug 538789).

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
iloveeclipse: I will update logging in an extra commit

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
iloveeclipse: no, see javadoc

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
iloveeclipse: See bug 164496

Issue #300, opened 2022-12-19
RefreshProvider shoudl have a priority


laeubi:
Currently providers are simply unsorted and have random order, but we might want to give more special ones precedence over simpler ones.

Issue #299, opened 2022-12-16
Create a RefreshProvider base on the NIO API and enable autorefresh for all operating systems except MacOS


laeubi:
Currently Eclipse do not refresh resources by default, and even if enabled, [only Win32 has support for native events](https://github.com/eclipse-platform/eclipse.platform/blob/master/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/refresh/win32/Win32Monitor.java).

Java contains an own [platform independent API to watch local files systems](https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchService.html). We should reuse this whenever possible as it is likely better maintained, optimized and improved than everything we can offer ourself.

Given that [all operating systems except MacOS seem to support native events](https://github.com/eclipse-m2e/m2e-core/issues/1013#issuecomment-1302764470), we should even enable this by default (except for MacOS), one can even think about simply checking if `PollingWatchService` is used (e.g. by create a dummy one and check the classname).

mickaelistria:
Big +1 here.

vogella:
Big +1. Just to consider, IIRC a running Maven build did render the ide useless as the notifications were triggering many updates in the ide (e.g.builds etc)

laeubi:
> Just to consider, IIRC a running Maven build did render the ide useless as the notifications were triggering many updates in the ide (e.g.builds etc)

I think this should be handled on a different level, strategies might include to cancel running builds if the become useless, not triggering (automatic) builds more often or to condense repeating updates. I think it is even fair to collect an resource-refresh during a build and trigger the refresh afterwards.

iloveeclipse:
Not sure if this is changed, but in Java 8 we've evaluated that for our internal use and found a big black hole in the API: WatchService doesn't monitor files in sub directories of monitored directory, and it is supposed to explicitly add watch service **recursively to every sub directory** in order to monitor really entire tree.

Not sure how this would work for monitoring any decent project, not saying about multiple hundreds of projects with lot of files, because in order to watch them, the entire forest need to be traversed at startup and for every subdirectory registration needed, and of course all subtree creation/move etc need another traversal with unregister/register etc. I'm pretty sure this can't scale for bigger workspaces.

Here is the example implementation from Oracle: https://docs.oracle.com/javase/tutorial/essential/io/examples/WatchDir.java
[...]

PR #303, opened 2022-12-20
RefreshProvider shoud have a priority

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform/issues/300

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/schema/refreshProviders.exsd
+++ b/resources/bundles/org.eclipse.core.resources/schema/refreshProviders.exsd
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!-- Schema file written by PDE -->
-<schema targetNamespace="org.eclipse.core.resources">
+<schema targetNamespace="org.eclipse.core.resources" xmlns="http://www.w3.org/2001/XMLSchema">
 <annotation>
       <appInfo>
          <meta.schema plugin="org.eclipse.core.resources" id="refreshProviders" name="Auto-refresh providers"/>
@@ -11,6 +11,11 @@
    </annotation>
 
    <element name="extension">
+      <annotation>
+         <appInfo>
+            <meta.element />
+         </appInfo>
+      </annotation>
       <complexType>
          <sequence>
             <element ref="refreshProvider"/>
@@ -62,6 +67,14 @@
                <appInfo>
                   <meta.attribute kind="java" basedOn="org.eclipse.core.resources.refresh.RefreshProvider"/>
                </appInfo>
+            </annotation>
+         </attribute>
+         <attribute name="priority" type="string">
+            <annotation>
+               <documentation>
+                  If more than one provider is aviable, providers with a higher priority will be asked first.
+The value must be parsable as an integer, otherwhise 0 is assumed as default.
+               </documentation>
             </annotation>
          </attribute>
       </complexType>
--- a/resources/tests/org.eclipse.core.tests.resources/plugin.xml
+++ b/resources/tests/org.eclipse.core.tests.resources/plugin.xml
@@ -534,18 +534,33 @@
       <content-type id="org.eclipse.core.tests.resources.nature_associated_2"/>       
 </extension>
 
+<!-- ORDER MATTER HERE START -->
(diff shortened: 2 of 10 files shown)


resources/bundles/org.eclipse.core.resources/schema/refreshProviders.exsd
```diff
@@ -64,6 +69,13 @@
                </appInfo>
             </annotation>
          </attribute>
+         <attribute name="priority" type="string">
```
iloveeclipse, line 72: Not sure about type, if possible, integer would be better.
In any case please give some hints in the documentation about allowed range, e.g is "-1" or `Long.MAX_LONG` are OK or not etc.

iloveeclipse (changes requested):
Ideally new API should have tests, especially such basic functionality like selecting "right" provider.

resources/bundles/org.eclipse.core.resources/schema/refreshProviders.exsd
laeubi: Only string, class or boolean are allowed :-(

laeubi: I have now clarified the description.

laeubi (commented):
Thanks for review this, I have now clarified the documentation and added assertions to the existing provider test.

iloveeclipse (changes requested):
I think a different approach should be evaluated first, see comment in code.

So far I don't see this change provides a good base for a proper fix of https://github.com/eclipse-platform/eclipse.platform/issues/299

resources/tests/org.eclipse.core.tests.resources/plugin.xml
```diff
@@ -534,18 +534,33 @@
...
+         priority="5"
+         />
 </extension>
 <extension
       point="org.eclipse.core.resources.refreshProviders">
    <refreshProvider
          class="org.eclipse.core.tests.resources.refresh.Bug316182RefreshProvider"
-         name="org.eclipse.core.tests.resources.bug316182RefreshProvider"/>
+         name="org.eclipse.core.tests.resources.bug316182RefreshProvider"
+         priority="10"
```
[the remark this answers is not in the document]
iloveeclipse: > Shouldn't one create new "supplemental" attribute for additional optional providers that could be contributed?

Thinking a bit longer, this also doesn't make sense. Either the provider can refresh or not, two or more of different providers on a project wouldn't make things better and just do same work multiple times.

What you actually want in https://github.com/eclipse-platform/eclipse.platform/issues/299 is a possibility for 3rd party code to **selectively** switch auto-refresh on/off for some specific (m2e) projects.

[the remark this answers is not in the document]
laeubi: > if the "real" one will win during tests on Windows?

Only if the windows fragment is included in the test (I talked with @HannesWell that he want to test this as well under "real" Windows)

> Missing explanation in extension point what happens if two providers have same priority.

That is true, I just wondering how far one want to go here, I don not expect many providers are implemented, actually we only have one for windows ...

> Should we disallow that & throw an exception if there is a second provider with same prio defined?
> If we don't do that, "random" one will be selected, based on extension point reading order by the OSGI.

Actually this could be a valid case, if we have one provider that only supports local resource and one that only supports remote ones, still I think we will most probably define a total order here.

Current behavior is to just call *all* providers... that seems not what one wants but again, as there is only one in reality...

> Also looking on the RefreshProviderTest, I wonder how that can work if we want both test refresh provider and real one (like Win32RefreshProvider) to be tested.

Real testing is actually not possible .. the tests mostly check that some methods are called, and one provider is even misused to check a completely different thing :-\

> For me it looks like original extension point was implemented only with single production implementation in mind, so it assumed there will be no multiple "production" refresh providers contributed. Now with introducing priority we actually create expectation that it is allowed to contribute multiple but at the end only the "highest" will be used.

probably, at laest the code is not very well prepared currently... but having a priority and using the highest that can handle the resource seems mostly resemble the "only one provider" mindset?!?

> What you actually want in https://github.com/eclipse-platform/eclipse.platform/issues/299 is a possibility for 3rd party code to selectively switch auto-refresh on/off for some specific (m2e) projects.

Actually I want that no one has to think about it or configure anything :-)
But the API is somehow prepared for such a case, but in the moment it is only all on/off ..

PR #307, opened 2023-01-04
Enable publish of code metrics to github

laeubi:
FYI @akurtakov


laeubi:
Just a build-configuration change so code-freeze should not apply.

laeubi:
> What's with the many Maven warnings like "Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet ..." ?

Seems something in P2 is not happy with the contract requirement, but I haven't analyzed that.

Issue #313, opened 2023-01-17
IJobChangeListener with lambda for done(IJobChangeEvent event) ?


marcushoepfner:
Is it worth to have method like

`static IJobChangeListener jobDoneAdapter(Consumer<IJobChangedEvent> c) {
       return new JobjChangedAdapter() {
                 public void done(IJobChangedEvent e) {
                      c.accept(e);
                 }
       };
}`
???

AFAICS the done method in a listener is most implemented one and having that as lambda would make coding shorter (though I have to admit with the JobChangeAdapter class it is quite short already).

But the question is, where do we stop. At the end there is a lambda for every method of IJobChangeListener. And later even for mixtures.

laeubi:
> But the question is, where do we stop. At the end there is a lambda for every method of IJobChangeListener. And later even for mixtures.

For general discussions/ideas you can also use the https://github.com/eclipse-platform/eclipse.platform/discussions in this repository.

marcushoepfner:
Done in https://github.com/eclipse-platform/eclipse.platform/pull/314.

PR #314, opened 2023-01-18
IJobChangeListener with lambda for done #313

marcushoepfner:
Adds a static method to IJobChangeListener for the done method which accepts a consumer.

JobManager has been adapted to use this new method.

Signed-off-by: Marcus Hoepfner <marcus.hoepfner@sap.com>

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
@@ -134,7 +134,7 @@ public interface IJobChangeListener {
 	 *
 	 * @param c the consumer of the event
 	 * @return IJobChangeAdapter
-	 * @since 3.14
+	 * @since 3.15
 	 */
 	static IJobChangeListener onDone(Consumer<IJobChangeEvent> c) {
 		return new JobChangeAdapter() {
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
@@ -13,23 +13,30 @@
  *******************************************************************************/
 package org.eclipse.core.runtime.jobs;
 
+import java.util.function.Consumer;
+
 /**
- * Callback interface for clients interested in being notified when jobs change state.
+ * Callback interface for clients interested in being notified when jobs change
+ * state.
  * <p>
- * A single job listener instance can be added either to the job manager, for notification
- * of all scheduled jobs, or to any set of individual jobs.  A single listener instance should
- * not be added to both the job manager, and to individual jobs (such a listener may
- * receive duplicate notifications).
- * </p><p>
+ * A single job listener instance can be added either to the job manager, for
+ * notification of all scheduled jobs, or to any set of individual jobs. A
+ * single listener instance should not be added to both the job manager, and to
+ * individual jobs (such a listener may receive duplicate notifications).
+ * </p>
+ * <p>
  * Clients should not rely on the result of the <code>Job#getState()</code>
  * method on jobs for which notification is occurring. Listeners are notified of
  * all job state changes, but whether the state change occurs before, during, or
  * after listeners are notified is unspecified.
- * </p><p>
+ * </p>
+ * <p>
  * It is undefined in which Thread the notification occurs.
- * </p><p>
+ * </p>
+ * <p>
  * Clients may implement this interface.
  * </p>
+ *
  * @see JobChangeAdapter
  * @see IJobManager#addJobChangeListener(IJobChangeListener)
  * @see IJobManager#removeJobChangeListener(IJobChangeListener)
@@ -119,4 +126,22 @@ public interface IJobChangeListener {
 	 * @param event the event details
 	 */
 	void sleeping(IJobChangeEvent event);
+
+	/**
+	 * Static helper method to create an <code>IJobChangeListener</code> for the
+	 * {@link #done(IJobChangeEvent)}) method, given a lambda expression or a method
+	 * reference.
+	 *
+	 * @param c the consumer of the event
+	 * @return IJobChangeAdapter
+	 * @since 3.14
+	 */
+	static IJobChangeListener onDone(Consumer<IJobChangeEvent> c) {
+		return new JobChangeAdapter() {
+			@Override
+			public void done(IJobChangeEvent event) {
+				c.accept(event);
+			}
+		};
+	}
 }
--- a/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.jobs; singleton:=true
-Bundle-Version: 3.14.100.qualifier
+Bundle-Version: 3.15.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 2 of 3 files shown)


laeubi:
@marcushoepfner won't it be much more useful to have the Job API providing a Completable Future here? That is much more generic and powerfull something like:

`CompletableFuture<IJobChangeEvent> IJob.whenDone()`

szarnekow:
Jobs can be scheduled multiple times thus the job listener would be notified multiple times about the job being done. Futures are not a good match for that usecase.

laeubi:
> Jobs can be scheduled multiple times thus the job listener would be notified multiple times about the job being done. Futures are not a good match for that usecase.

Yes they *can* but there is no reason why such a listener won't be removed after the first notification, if code is still interested it can simply create a new feature, what gives a much better programming model (not counting the endless bugs recently with wrong/incomplete/out of order/ ... notifications).

szarnekow:
I don't see *any* point in providing an alternative way to being notified about job states given there's already the job listener API. Having two ways of doing this is prone to confusion. Client code that wants to remove the listener after the first notification can just do that. Client code that prefers working with futures can just do that based on a thin adapter based on the job listener. Does that mean it needs to become part of the core job API? No, it would just be confusing.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
```diff
@@ -119,4 +126,22 @@ public interface IJobChangeListener {
...
+	/**
+	 * Static helper method to create an <code>IJobChangeListener</code> for the
+	 * {@link #done(IJobChangeEvent)}) method, given a lambda expression or a method
+	 * reference.
+	 *
+	 * @param c the consumer of the event
+	 * @return IJobChangeAdapter
+	 * @since 3.14
+	 */
+	static IJobChangeListener jobDoneAdapter(Consumer<IJobChangeEvent> c) {
```
laeubi, line 139: `onDone` as a method name would be enough, I think there is no purpose in exposing the internal implementation in the method name.

laeubi:
> I don't see _any_ point in providing an alternative way to being notified about job states given there's already the job listener API. Having two ways of doing this is prone to confusion. Client code that wants to remove the listener after the first notification can just do that. Client code that prefers working with futures can just do that based on a thin adapter based on the job listener. Does that mean it needs to become part of the core job API? No, it would just be confusing.

Well.. so why add another "convenience" method then at all if it is all that cool and client code can do anything itself? I really thing this "confusing" argument becomes more and more an equivalent to "I don't like the idea" and I would prefer people would just speak that out clear instead of finding weird arguments why common code patterns "confuse" people... the world around java is moving every day and it seems no one is confused... except "usual Eclipse Developers".

szarnekow:
It doesn't matter if I like the idea or not. It's just not a good match for the existing abstraction.

iloveeclipse:
>  I really thing this "confusing" argument becomes more and more an equivalent to "I don't like the idea" and I would prefer people would just speak that out clear instead of finding weird arguments why common code patterns "confuse" people... the world around java is moving every day and it seems no one is confused... except "usual Eclipse Developers".

@laeubi I personally find such comments violate https://www.eclipse.org/org/documents/Community_Code_of_Conduct.php .

This is offending & disrespectful.

I see these points violated here:

> Using welcoming and inclusive language;
> Being respectful of differing viewpoints and experiences;
> Gracefully accepting constructive criticism;
> Showing empathy towards other community members;

Please consider to avoid such comments in the future.

Beside this I agree with @szarnekow arguments.

laeubi:
@iloveeclipse if you feel offended I'm sorry for that but I'm just expressed a personal feeling, but well, statements like

> I don't see _any_ point in providing an alternative way 
> it would just be confusing.
> It doesn't matter 

could be interpreted as not being

> Using welcoming and inclusive language;
> Being respectful of differing viewpoints and experiences;
> Gracefully accepting constructive criticism;
> Showing empathy towards other community members;

It just contains no way to understand the argumentation why there is no point, or why it should be confusing or why it doesn't matter, so at least I can hardly tell what are the arguments you agree with  ... Again just my personal opinion no offense, keep it as a form of constructive criticism on how probably others read such statement.

Also I have reviewed the code, and I hardly see that this change simplifies the code, so I proposed a possible alternative where I can imagine such lambda constructs might benefit, if I really like to have some code executed repeatedly and often the current Adapter seems much more suitable to me and this only adds another layer over an existing layer that even requires another layer (lamdas), so we now have IJobChangeListener > Adapter > LamdaWrapper > Lamda and I need to pass this to a Job.

So even if it was claimed to be confusing just for example:

```
IJob job ...
job.done().thenConsume(e-> ... do something short when job is done ...);
job.scedule();
... now I don't care about the job anymore
```

is nothing that would confuse me in anyway.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobChangeListener.java
marcushoepfner: done

laeubi:
> Please understand that you are not the only user of the API you proposed.

Agreed, but that's just true for every change (including the proposed new method) so if one never thinks about alternatives one never make progress, java itself offers many ways to archive the same thing because how people use java changes and the same should be valid for eclipse-platform.

So every time a new convenience method is added, one should think about if the current API is simply not (any longer) fulfill users demands and modern coding styles.

> There are more considerations to adding another concurrency primitive to a library/framework that does already offer comparable means to achieve the same.

The current approach is simply not comparable to what e.g. `CompletableFuture` offers as it will require a lot of boilerplate, so if you think that's fine as indicated here and claimed it would be confusing (`CompletableFuture` is there since Java 1.8 so one can assume it well known to java developers since many years):

> Client code that prefers working with futures can just do that based on a thin adapter based on the job listener. Does that mean it needs to become part of the core job API? No, it would just be confusing.

So why is this new method acceptable? Won't it be possible for client that prefer that way to "just do that based on a thin adapter based on the job listener" ... there even `is` an Adapter already, so now there is Adapter of Adapter ... should this not equally confusing?

Still I don't claim that the proposal of @marcushoepfner are bad or useless, just to think about possible new and more flexible ways, as I can't think that the example shown here really *simplifies* things and there are probably other cases and this was just the first example found to show how the new method *might* be used.

szarnekow:
@laeubi Please keep in mind that I also explicitly stated why this would be confusing. It's not an unsolicited claim.

Jobs can run multiple times, be rescheduled, or cancelled. That's a fact and won't change anytime soon. Having a future as a done-listener for one-time notifications - there would be no way to use that API for repeated notifications - is not a great fit for the eclipse job framework.

My line of thinking:
A future that's obtained from a method `done(): CompletableFuture<..>` would either need to be cancelled if the job got cancelled or will be forever unfulfilled. It would also raise the question, if the Job would be cancelled if said future would be cancelled. There are likely more questions that need to be answered before such an API should be introduced.

Do I think it would be nice to have an ExecutorService that's backed by the Job infra? Yes, that might be useful. Would it carry its own weight? Maybe, maybe not. Not my call to make without having seen a few clients that immediately benefit from something like that.

Do I think the proposed sugar for the JobAdapter adds value? To be honest: I'm not terribly sure. Does it cause harm? No, probably not. Is it in line with other, similar APIs in Eclipse? Yes, it was (prior to the rename to `onDone()` - see  `org.eclipse.swt.browser.ProgressListener.completedAdapter(Consumer<ProgressEvent>)` as one example that come to my mind. Is it therefore acceptable for me personally? Yes, it is. It does not leave any question unanswered.

laeubi:
Thanks for the detailed explanation and concerns seem valid to me and would need some more thoughts.
If you think the previous naming was better, I won't veto it, it just feels odd if the name tells about the implementation.

szarnekow:
> If you think the previous naming was better, I won't veto it, it just feels odd if the name tells about the implementation.

Personally, I don't have a strong opinion on the name. I'd leave it up to @marcushoepfner and the committer who is going to review this. Either is fine for me.

marcushoepfner:
Locally testJoinInterruptNonUIThread (org.eclipse.core.tests.runtime.jobs.JobTest) is green.
So I reverted JobManager to check if build still fails.

BeckerWdf (approved):

PR #323, opened 2023-01-29
Balance text compare changes

### LINKED ISSUE #66: Make Text Compare more smart
I recently noticed that text-compare seems to present a undesirable changeset if the following happen:

1. You have some repeated code block  with common start (e.g. `<dependency>` in a pom.xml or `@Test` in a Unit test) and ending (e.g. `</dependency>` in a pom.xml or `}` in a Unit test
2. You insert a new block in between
3. Now the change set spans into the next block

grafik

**What would be expected is that the whole block is marked as an insertion (here line 74 - 78)**

Github for example displays this more nice

grafik

Bananeweizen:
The current diff algorithm optimizes for longest identical subsequences between the changes. Change this to balance identical start and end parts of a change.

Imagine the ranges ...ABCDABEF... where LCS has led to marking CDAB as difference (because it tries to maximize the sequence before that change). E.g. AB[CDAB]EF is the marked change. If the other side is an _empty_ change, then humans typically prefer BCDA as change, e.g. A[BCDA]BEF. In fact, anything from [ABCD]ABEF to AB[CDAB]EF is possible, and choosing the median of all those possibilities should lead to better results if the input is for instance structured text (with different indentation levels).

This fixes #66, but only for changes where one side is empty. I would still have to think about making this work for changes where both left and right are not empty.

This is what a change looks like. Compare with the screenshot from #66: 
grafik

### DIFF
--- a/team/bundles/org.eclipse.compare.core/src/org/eclipse/compare/rangedifferencer/RangeDifferencer.java
+++ b/team/bundles/org.eclipse.compare.core/src/org/eclipse/compare/rangedifferencer/RangeDifferencer.java
@@ -97,7 +97,64 @@ public final class RangeDifferencer {
 	 * @since org.eclipse.compare.core 3.5
 	 */
 	public static RangeDifference[] findDifferences(AbstractRangeDifferenceFactory factory, IProgressMonitor pm, IRangeComparator left, IRangeComparator right) {
-		return RangeComparatorLCS.findDifferences(factory, pm, left, right);
+		return balanceDifferences(RangeComparatorLCS.findDifferences(factory, pm, left, right), factory, left, right);
+	}
+
+	/**
+	 * Balance differences. The LCS computation tries to produce identical parts in
+	 * the comparison as large as possible. That is not necessarily what humans
+	 * expect.
+	 * <p>
+	 * Imagine the ranges ...ABCDABEF... where LCS has led to marking CDAB as
+	 * difference (because it tries to maximize the sequence before that change).
+	 * E.g. AB[CDAB]EF is the marked change. If the other side is an _empty_ change,
+	 * then humans typically prefer BCDA as change, e.g. A[BCDA]BEF. In fact,
+	 * anything from [ABCD]ABEF to AB[CDAB]EF is possible, and choosing the median
+	 * of all those possibilities should lead to better results if the input is for
+	 * instance structured text (with different indentation levels).
+	 *
+	 * @param diff    if not <code>null</code> used to report progress
+	 * @param factory factory to instantiate new {@link RangeDifference}s
+	 * @param left    the left range comparator
+	 * @param right   the right range comparator
+	 * @return an array of range differences, or an empty array if no differences
+	 *         were found
+	 */
+	private static RangeDifference[] balanceDifferences(RangeDifference[] diffs, AbstractRangeDifferenceFactory factory, IRangeComparator left,
+			IRangeComparator right) {
+		ArrayList<RangeDifference> result = new ArrayList<>();
+		int previousEnd = 0;
+		for (RangeDifference diff : diffs) {
+			RangeDifference optimized = diff;
+			if (diff.rightLength() == 0) {
+				int identical = 0;
+				while (identical < diff.leftStart // don't underflow
+						&& identical < diff.leftLength // cannot move more than diff size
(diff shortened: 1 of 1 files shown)


jukzi:
please update bundle version.
I tested it:
image
Works well for multilines. but unfortunately inserting [b] between [a] and [c] shows insertion of "][b"

Bananeweizen:
> but unfortunately inserting [b] between [a] and [c] shows insertion of "][b"

Just checked that in the debugger. That intra-line diffing is driven by a tokenizer, which splits tokens for that line as "[", "a", "][", "b", .. and which can only highlight full tokens. My change made the LCS return index 3 instead of index 4 in that line (which is what we want), but the tokenizer then decided to not start coloring at "b" anymore, but rather at the beginning of token "][", since that contains the new index 3. So the index calculation is better, but the tokenizer can't create highlighting in the middle of a token.

Given that knowledge, I currently tend to not modify existing intraline diffs (of characters/tokens), but to restrict my change to line comparisons of documents. That would mean moving the changed method to DocumentMerger instead (i.e. 4 methods up in the stack trace, compared to current debugging). Any comments?

laeubi:
Yes this looks much better in the diff (and less confusing), I'm just not sure if @jukzi case was already addressed?

jukzi:
Even though it does not handle all cases, this commit still is an improvement. Submit?

vogella:
>Submit?

+1,  after the verification build on Linux worked.

team/bundles/org.eclipse.compare.core/src/org/eclipse/compare/rangedifferencer/RangeDifferencer.java
```diff
@@ -97,7 +97,64 @@ public static RangeDifference[] findDifferences(IProgressMonitor pm, IRangeCompa
...
+	 * @param factory factory to instantiate new {@link RangeDifference}s
+	 * @param left    the left range comparator
+	 * @param right   the right range comparator
+	 * @return an array of range differences, or an empty array if no differences
+	 *         were found
+	 */
+	private static RangeDifference[] balanceDifferences(RangeDifference[] diffs, AbstractRangeDifferenceFactory factory, IRangeComparator left,
+			IRangeComparator right) {
+		ArrayList<RangeDifference> result = new ArrayList<>();
+		int previousEnd = 0;
```
sratz, line 126: `previousEnd` is never incremented, which I think, causes overlapping `RangeDifference`s:

We have a failing unit, which compares two sources and applies all differences:

```java
		MultiTextEdit result = new MultiTextEdit();

		TextRangeComparator left = new TextRangeComparator(fromText, compareMode);
		TextRangeComparator right = new TextRangeComparator(toText, compareMode);

		RangeDifference[] differences = RangeDifferencer.findDifferences(left, right);
		for (RangeDifference difference : differences) {
			TextRange leftRange = left.getRange(difference.leftStart(), difference.leftLength());
			TextRange rightRange = right.getRange(difference.rightStart(), difference.rightLength());

			ReplaceEdit replaceEdit = new ReplaceEdit(leftRange.offset, leftRange.length, rightRange.text);
			result.addChild(replaceEdit);
		}
		result.apply(document);

```
which is failing since this change:
```
org.eclipse.text.edits.MalformedTreeException: Overlapping text edits
	at org.eclipse.text.edits.TextEdit$InsertionComparator.compare(TextEdit.java:126)
	at org.eclipse.text.edits.TextEdit$InsertionComparator.compare(TextEdit.java:1)
	at java.base/java.util.Collections.indexedBinarySearch(Collections.java:336)
	at java.base/java.util.Collections.binarySearch(Collections.java:324)
	at org.eclipse.text.edits.TextEdit.computeInsertionIndex(TextEdit.java:799)
	at org.eclipse.text.edits.TextEdit.internalAdd(TextEdit.java:784)
	at org.eclipse.text.edits.TextEdit.addChild(TextEdit.java:340)
	...
```

sratz: https://github.com/eclipse-platform/eclipse.platform/issues/458

I'll provide a fix proposal.

sratz: https://github.com/eclipse-platform/eclipse.platform/pull/459

PR #326, opened 2023-02-06
Interupt the thread on job cancel

laeubi:
currently the default impl of (Internal)Job#canceling do nothing, but this could mean that jobs possibly hang forever (and even prevent Eclipse from shutting down) if a Job is hung for example in an I/O operation or a wait condition.

This performs Thread cancelation if not overriden by an implementor.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
@@ -200,7 +200,10 @@ public abstract class InternalJob extends PlatformObject implements Comparable<I
 	}
 
 	protected void canceling() {
-		//default implementation does nothing
+		Thread t = getThread();
+		if (t != null) {
+			t.interrupt();
+		}
 	}
 
 	@Override
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
@@ -301,6 +301,7 @@ public abstract class Job extends InternalJob {
 	@Override
 	protected void canceling() {
 		//default implementation does nothing
+		super.canceling();
 	}
 
 	/**


iloveeclipse (changes requested):
Clear "no". This change will interrupt a (client) thread middle in some (unknown to us) operation.

This clearly violates the API contract via behavior change and will for sure cause unwanted side effects.

See `org.eclipse.core.runtime.jobs.Job.cancel()` Javadoc: 
> If the job is currently executing, it will be asked to stop but there is no guarantee that it will do so.

laeubi:
> Clear "no". This change will interrupt a (client) thread middle in some (unknown to us) operation.

Thread interruption do not interrupt threads "in the middle" of somewhere, it just interrupts specific methods where otherwise Threads normally can't escape see https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html

>> An interrupt is an indication to a thread that it should stop what it is doing and do something else. It's up to the programmer to decide exactly how a thread responds to an interrupt

So it is just a _signal_ every implementation should handle (and even the Job API *does* handle the fact that Threads might be interrupted.

> This clearly violates the API contract via behavior change and will for sure cause unwanted side effects.

There is still no guarantee that a job actually stops, or did I get it wrong and the API contract guarantees that Eclipse is hung forever and I need to kill the process?

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
```diff
@@ -200,7 +200,10 @@ protected boolean cancel() {
 	}
 
 	protected void canceling() {
-		//default implementation does nothing
+		Thread t = getThread();
+		if (t != null) {
+			t.interrupt();
```
jukzi, line 205: don't like it because there is no guarantee that that thread is currently running the desired job. The worker threadpool can already be executing another job. The Api is designed such that the Job should activly listen for the monitor.cancel flag. if that is not enough during a IO operation the jjob should spawn a dedicated thread and interrupt that thread when monitor is canceled.

jukzi (changes requested):

jukzi:
The job can implement a custom org.eclipse.core.runtime.jobs.Job.canceling() instead.

mickaelistria (changes requested):
I agree with @iloveeclipse that this doesn't seem like a safe change in general. The purpose of job cancelling is to let job producer decide of how to handle cancellation in a good way for this particular job. By killing the thread, this may cause more issues: for example, if this is a UIJob, then the thread is the UI Thread, and SWT may just stop when the UI Thread receives the signal, shutting down the application.
As mentioned by @jukzi , there is no 1-1 mapping between jobs and threads that make it safe-ish to cancel a thread when cancelling a job. It's possible that the assigned thread is to be reused later and killing it becomes an issue for the job orchestration. Looking at CompletableFuture, which is somehow similar to Eclipse Jobs  API, I don't see cancelling having any effect on the running thread; it's even specified that "mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing." which is also what we expect here.
I would agree that this might have been a decent implementation in the beginning of Jobs API, but changing it now after 20 years of usage of that API is likely to just cause much bigger issues than the ones it's intended to solve.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
laeubi: The Thread is currently assigned to this job, and the job API is using that thread for different purpose, so I don't understand why there should be an unrelated thread set here?

laeubi:
Interruption is **not killing a thread** and ANY code can interrupt the thread so this is really nothing fancy or new, while having to **kill my eclipse process** because of hanging jobs that do not respond to Thread interruption is really bad.

The general contract for any Thread Pooling is that the interrupted flag has to be cleared before returning a Thread to the pool as otherwise all kind of spurious interrupts can happen, e.g by calling:
https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#interrupted--

**So to summarize:** If Thread interruption currently is a problem, it should be fixed, instead of hoping no one will ever call interrupt on a worker thread...

> The job can implement a custom org.eclipse.core.runtime.jobs.Job.canceling() instead.

Theoretically yes, but practically I didn't found any Job extension in Platform that actually overrides the method... and as application code is often not in control of the job calling it it is just not possible most of the time..

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/InternalJob.java
merks: Is it possible that between line 203 and line 205 the job ends and the thread is then used by different job?  I don't see anything that would prevent that from happening...

jukzi:
@laubi Can we close this PR? Sounds like there is no consent to change cancel() behavior.

Issue #320, opened 2023-01-25
random failing JobTest.testJoinInterruptNonUIThread


jukzi:
```
Thread not interrupted expected:<Status CANCEL: unknown code=1 > but was:<Status OK: unknown code=0 OK>
junit.framework.AssertionFailedError: Thread not interrupted expected:<Status CANCEL: unknown code=1 > but was:<Status OK: unknown code=0 OK>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.TestCase.assertEquals(TestCase.java:238)
	at org.eclipse.core.tests.runtime.jobs.JobTest.testJoinInterruptNonUIThread(JobTest.java:1019)
```

in https://github.com/eclipse-platform/eclipse.platform/pull/318/checks?check_run_id=10878070950
and https://github.com/eclipse-platform/eclipse.platform/pull/317#issuecomment-1403299916

jukzi:
can be reproduced with new TestJob("job", **0**, 100);
i.e. probably a timing issue

HeikoKlare:
Test is still failing. See: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4734529994/jobs/8403581311

Issue #294, opened 2022-12-12
SaveManager: ForkJoinPool used for parallel workspace save does not play well with potentially installed SecurityManager


sratz:
In https://bugs.eclipse.org/bugs/show_bug.cgi?id=579065 parallel workspace save was introduced by means of a ForkJoinPool.

However, this has a a side effect in case a `SecurityManager` is installed in the Java runtime (even if it permits all!).

**Steps to reproduce:**

1. Have any Eclipse-plugin that installs a `SecurityManager`. Use attached demo plugin [foo.bar.securitymanager.zip](https://github.com/eclipse-platform/eclipse.platform/files/10208024/foo.bar.securitymanager.zip) (Main Menu `SecurityManager -> Install SecurityManager`) or the following snippet:
```java
Policy.getPolicy(); // init
Policy.setPolicy(new Policy() {
    @Override
    public boolean implies(ProtectionDomain domain, Permission permission) {
        return true;
    }
});
System.setSecurityManager(new SecurityManager());
```
2. Open any project and some file
3. Close Eclipse

**Result:**
![image](https://user-images.githubusercontent.com/14908423/207048595-841162c6-c245-4534-811a-27554f551a26.png)
```
Problems occurred while trying to save the state of the workbench.
  Problems occurred during save.
  java.security.AccessControlException: access denied ("java.io.FilePermission" "C:\eclipse_workspaces\runtime-EclipseContribute\.metadata\.plugins\org.eclipse.core.resources\.projects\A\.markers" "read")
```

**Reason:**
The reason is the way `ForkJoinPool` works in recent versions of Java:

- https://bugs.openjdk.org/browse/JDK-8237117 (CVE-2020-14556: Better ForkJoinPool behavior)
- https://bugs.openjdk.org/browse/JDK-8249846
- https://github.com/openjdk/jdk17u/commit/ecabcca746407c52c3587920f74647cbf7defaab#diff-fb8afbdfd38bab6e834dd44c7bd5d7e6deeab455c4a82aba96339e564d7ece85R156-R158

As a side effect of this change, the threads spawned in the `ForkJoinPool` run without any permissions and are thus able to do the needed i/o operations during workspace save.

**Potential workaround:**
1. Remember the caller security context and restore inside the threads:
```patch
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java b/resources/bundles/org.eclipse.core.resources/src/o
[...]

PR #327, opened 2023-02-06
Support executing tasks with the Job Framework using CompletableFuture

laeubi:
The JobFramework has a nice integration in eclipse but using them (especially for smaller tasks) can be a bit cumbersome.

This adds two ne methods (execute and compute) that connect the Job API with the `CompletableFuture` which allows for a uniform programming model that integrates nicely with many other APIs.

I'm creating this as a draft, to show possible API and initiate discussion about making the Job API more convenient to use especially for cases where smaller background work is desired (e.g. to not do it inside the UI thread) while post processing the data or trigger other actions with an event based chaining (e.g. to update the UI when the computation is done).


runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -194,6 +196,82 @@ protected IStatus run(IProgressMonitor monitor) {
...
+	 * <ul>
+	 * <li>If the {@link CompletableFuture} is canceled, best effort is made to also
+	 * cancel the job</li>
+	 * <li>If the backing {@link Job} is canceled, best effort it made to also
+	 * cancel the {@link CompletableFuture}</li>
+	 * </ul>
+	 *
+	 * @since 3.14
+	 */
+	public static CompletableFuture<?> execute(String name, IJobRunnable runnable) {
```
mickaelistria, line 211: Why a new IJobRunnable type instead of a `Consumer<IProgressMonitor>`?
Also the return type should be `CompletableFuture<Void>`.
Shouldn't we call it `runAsync()` like CompletableFuture API? or `schedule`? The current `name` fails at carrying the async nature IMO.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -194,6 +196,82 @@ protected IStatus run(IProgressMonitor monitor) {
...
+	 * <ul>
+	 * <li>If the {@link CompletableFuture} is canceled, best effort is made to also
+	 * cancel the job</li>
+	 * <li>If the backing {@link Job} is canceled, best effort it made to also
+	 * cancel the {@link CompletableFuture}</li>
+	 * </ul>
+	 *
+	 * @since 3.14
+	 */
+	public static <T> CompletableFuture<T> compute(String name, IJobCallable<T> callable) {
```
mickaelistria, line 230: Same as above: what about `Function<IProgressMonitor, T>`? and what about `supplyAsync` or `scheduleCompute` ?

mickaelistria (commented):
I think such API is a great and necessary enhancement for the Job API. I've put some matter for discussion inline.
However, I think one thing we'll miss is some migration path for existing jobs: much code does use Job directly; it would be nice for those to have a `CompletableFuture` view of the job (eg `job.asCompletableFuture()` or `Job.asCompletableFuture(Job)` to allow more incremental adoption.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
laeubi: > Why a new IJobRunnable type instead of a Consumer<IProgressMonitor>?

Mostly to allow for checked Exceptions.

> Also the return type should be CompletableFuture<Void>

What is the advantage of this?

> Shouldn't we call it runAsync() like CompletableFuture API? or schedule? The current name fails at carrying the async nature IMO.

For me the primary think here is to execute something but other names would be fine.

laeubi:
> it would be nice for those to have a CompletableFuture view of the job (eg job.asCompletableFuture() or Job.asCompletableFuture(Job) to allow more incremental adoption.

For consistent behavior, I think only something like `job.sceduleAsFuture()` would work, that can be used as a replacement for the "classic" `scedule()` call, because if one allows adaption of any job in any state it will porbabbly become inconsitent as we never know if the job will ever be sceduled (and thus the future might never complete), or what should happen in case of resceduling.

mickaelistria:
>  only something like job.sceduleAsFuture() would work

OK.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -194,6 +196,82 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		CompletableFuture<T> future = new CompletableFuture<>();
```
szarnekow, line 231: I don't think a cancellation request on the future is propagated to the job yet, is it? Please add a test that illustrates this.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobRunnable.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface IJobRunnable {
+
+	/**
+	 * Method to run inside a job
+	 *
+	 * @param monitor the monitor, never <code>null</code>.
+	 * @throws Exception in case of error
+	 */
+	void run(IProgressMonitor monitor) throws Exception;
```
szarnekow, line 32: It would be great if the new interfaces had (default) methods that allow specifying a few of the attributes that a job has, e.g. the scheduling rule, family, etc.

laeubi:
```
future.exceptionally(ex -> {
	if (ex instanceof CancellationException) {
		job.cancel();
	}
	return null;
});
```
this do the cancellation of the job in case a feature was canceled (what is the same as complete it with a `CancellationException`)

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
szarnekow: A, my bad: I oversaw the future.exceptionally(..) below.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobRunnable.java
laeubi: I just wanted to complicate this too much, if someone wants full Job power a Job is probably better (combined with what @mickaelistria  suggests to have it "convert" to a future...

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
jukzi: > Mostly to allow for checked Exceptions.

CompletableFuture.get() does not throw an checked exception. and ExecutorService.submit does not accept a checked exception.

laeubi: > CompletableFuture.get() does not throw an checked exception

[Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) says different:

> In case of exceptional completion with a CompletionException, methods [get()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get--) and [get(long, TimeUnit)](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) throw an [ExecutionException](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutionException.html) with the same cause as held in the corresponding CompletionException.

>  ExecutorService.submit does not accept a checked exception.

No idea how this relates here?

Beside what other API does, it is here for the user to not being forced to wrap checked exceptions into runtime exceptions, so the true exception can be accessed for example in 
- https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#exceptionally-java.util.function.Function-
- https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#handle-java.util.function.BiFunction-
- ... and its async variants ...

jukzi: CompletableFuture is meant to be used together with an Executor. All its async methods are defined to "are performed using the ForkJoinPool.commonPool()" - not what i expect them to do when generated by a eclipse. 
CancellationException/OperationCanceledException may behave other then the handling of exception handling defined in the vice versa API .
CompletableFuture .cancel(boolean mayInterruptIfRunning) will probably behave other then defined.
I am not convinced that it is a good idea to mix java.util.concurrent with eclipse stuff. Both have their  pros and cons, but it will be hard or impossible to get them together so that they fullfill both contracts for all edge cases.

laeubi: > CompletableFuture is meant to be used together with an Executor.

Only the async methods allow to specify a separate executor, but this has nothing to do with exception handling or how the _initial value_ is computed (and where exceptions are happening), because all later stages only operate on a _result_ and this might be processed where ever (even in another job if desired).

> not what i expect them to do when generated by a eclipse

Eclipse uses the common poll al time long if it is using streams, so I don't really see how this relates here, this is all only about computing the inital result, the rest is plain java and is not affected at all by where it runs...

> CompletableFuture .cancel(boolean mayInterruptIfRunning) will probably behave other then defined.

Please explain or give concrete example "probably" is nothing one can act upon.

> but it will be hard or impossible to get them together so that they fullfill both contracts for all edge cases.

There is essentially no need to fulfill *all* contracts (beside "Eclipse" and "java.util.concurrent" are not mutual exclusive in any way).

jukzi: See for example https://github.com/eclipse-platform/eclipse.platform/issues/294 for problems with commonPool() in eclipse. Eclipse normally does NOT use stream().parallel.
If the job throws CancellationException or OperationCanceledException what should happen? what should isCancelled() return? what is the IStatus?
mayInterruptIfRunning: see https://github.com/eclipse-platform/eclipse.platform/pull/326#discussion_r1097353063

laeubi:
@mickaelistria I now added a method `org.eclipse.core.runtime.jobs.Job.future()` where you can convert a traditional job scheduling into a completable future to observe the result.

vogella:
@jukzi @HeikoKlare looks like you guys almost stabilized the tests. GH actions are all green (thanks for that), only one tests failing (https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-327/5/testReport/junit/org.eclipse.core.tests.runtime.jobs/JobTest/testJoinInterruptNonUIThread/) on Jenkins and I already see https://github.com/eclipse-platform/eclipse.platform/pull/412 for this.

Really looking forward to have reliable verification builds on all platforms.

HeikoKlare:
Just as a note regarding tests: The failing test belongs to `JobTest`, whereas #412 concerns `JobGroupTest`. I have reopened #320 for the failing test and will have another look at it.

szarnekow:
There was some effort in platform to avoid the use of the commonPool: https://github.com/eclipse-platform/eclipse.platform.text/pull/190

Might be worthwhile to keep that in mind when work on this task might or might not continue.

laeubi:
> There was some effort in platform to avoid the use of the commonPool:

Can you explain why/where the commonPool is used here? The (initial) execution is actually performed by the Job framework and even though CompletableFuture can be further used with the commonPool ist is not done here as far as I rembmer.

mickaelistria:
I like this `Job.future()` a lot and already see some opportunities for using it. Mainly I prefer to create Jobs over Future in Eclipse because the default pool used by futures may cause some work to run in the UI Thread (which is somehow included in the default pool); Jobs are now easy to create with 1-liners and they don't run in UI Thread by default, so they're better.
However, they easily become annoying to chain and a typical case is chaining a Job (doing some computation) with a UI Job (doing some rendering). With `job.future()`, it would then become possible to do things like
```
Job job = Job.createSystem(monitor -> doWork());
job.thenAcceptAsync(status -> uiRendering(), Display.getDefault());
job.schedule();
```
This would save a dozen of lines over typical listener approach.

szarnekow:
> > There was some effort in platform to avoid the use of the commonPool:
> 
> Can you explain why/where the commonPool is used here? The (initial) execution is actually performed by the Job framework and even though CompletableFuture can be further used with the commonPool ist is not done here as far as I rembmer.

You're right, it's not being used here explicitly as of now. I wrote the comment on the phone and I should have taken the time to share my thoughts better.

If we try to move the Job framework and the CompletableFuture API closer together (which I think is a great idea), we should have an offering that makes it *hard* to us wrongly. Given that there was some effort going on the avoid implicit or explicit usage of the common pool, we could consider overwriting `defaultExecutor` and `newIncompleteFuture` on the offered CompletableFutures such that a single shared, privileged ForkJoinPool can be used. That one might be made available similar to `ForkJoinPool.commonPool()` as `EclipseForkJoinPool.instance()` or similar.

laeubi:
@szarnekow I think it might be a good idea what you have suggested here, e.g. "defaultExecutor and newIncompleteFuture" might also use the Job APi tro spin new threads, but I would preffere that we just finish the basics here and the work on such thinks on top of the inital change (as it is already lingering around for a while already...)

So I'll mark this as "ready for review" now, please let me know what are the changes to make this a minimal-viable-increment so we don't miss another release here.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
laeubi: Can we probably do this on top as further enhancements to this feature?

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -654,6 +732,72 @@ public final void schedule() {
...
+	 * </ul>
+	 *
+	 * @see #schedule()
+	 * @since 3.14
+	 */
+	public final CompletableFuture<IStatus> future() {
+		CompletableFuture<IStatus> future = new CompletableFuture<>();
+		JobChangeAdapter listener = new JobChangeAdapter() {
+
+			volatile boolean sceduled;
```
szarnekow, line 755: Typo in field name: Should be `scheduled`

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -654,6 +732,72 @@ public final void schedule() {
...
+	 * <li>When the job completes with an ERROR status the future will complete
+	 * exceptionally with a {@link CoreException}</li>
+	 * <li>In all other cases the future is completed with the job execution
+	 * status.</li>
+	 * </ul>
+	 *
+	 * @see #schedule()
+	 * @since 3.14
+	 */
+	public final CompletableFuture<IStatus> future() {
```
szarnekow, line 751: I see that the JavaDoc says that `future` will also `schedule` but I don't feel like `future` as a method name conveys that very well. 

Something else:
Assuming I've an existing Job that was already running for some reason:
- future() will reschedule successfully
- the scheduled jobChangeEvent will be received
- the already running job will finish eventually
- the listener will be notified and the future will be completed

Yet this was not the run that was scheduled and cancelling the future would have cancelled to job that was already running and not the execution that was scheduled. In other words: `Job.future().cancel()` does not imply that the Job was cancelled.
Should we guard against this and only allow `future()` to be called if the job wasn't scheduled yet?

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
```diff
@@ -654,6 +732,72 @@ public final void schedule() {
...
+				if (event.getJob() == Job.this) {
+					sceduled = true;
+				}
+			}
+		};
+		addJobChangeListener(listener);
+		if (trySchedule(0)) {
+			future.exceptionally(ex -> {
+				if (ex instanceof CancellationException) {
+					cancel();
```
szarnekow, line 789: This might cancel a subsequently scheduled run of the same job.

- Job created
- Future obtained
- Job running
- Job being cancelled via Job.cancel
- Job being rescheduled via Job.schedule
- Running Job checks the progress monitors and does indeed cancel now
- Listener is notified, future is cancelled
- Rescheduled Job is now being cancelled from here

Again this *might* not be a real problem but *if* someone is bitten by this, it'll be very hard to find, understand, and fix.

szarnekow (changes requested):
Please add tests for the future API and the interaction with the Job framework.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
laeubi: we can name is `scheduleFuture()` if that's better, but I often prefer shorter names as it makes things easier to read.

About the other notes: Yes, because of how the Job API works there is no way to give hard guarantees, but that's already with the "traditional" Job usage, so the "owner" (the one how creates the job) is somewhat responsible how the job is used (e.g. do not share the instance with other threads that do perform concurrent schedules).

I previously have had tried to add a guard that one can only call `future()` if job is somehow "idle" state but found that even then there is no safe way to do so as the job might be scheduled right after this check... or just complete a microsecond afterwards and so on... The crucial part is really the adding of the listener and the scheduling of the job, so what one actually want is some kind of

`Job.trySchedule(0, myListener)`

and that listener is then always called only for that scheduled execution ... but this seems not be possible right now without major efforts.

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/Job.java
laeubi: Actually there is no guarantee today that a cancel cancel "that" execution, it just cancels whatever is currently running, so we sadly can't do anything better here. So cancel7rescedule is already hard to debug today, one reason for that I best would always call `Job.create().future()` and leave out of band cancel/schedule alone :-)

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/runtime/jobs/IJobRunnable.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.eclipse.core.runtime.IProgressMonitor;
+
+/**
+ * A pice of coude that runs inside a job and has access to a progressmonitor
+ *
+ * @since 3.14
+ *
+ */
+public interface IJobRunnable {
```
HannesWell, line 24: Can't we reuse the existing `ICoreRunnable` instead of this new interface with very similar signature and only different exception that may be thrown? In general it's bit unfortunate that Eclipse has invented it's own 'root' exception, this make integration with standard Java methods often cumbersome.

We could then just add only a new `ICoreCallable`? There we could also throw just `Exception`.
Then there would certainly be a asymmetry between `ICoreRunnable` and `ICoreCallable`, but in my opinion the big value of this change is the new `Job.compute()` method, so I assume it's used much more often.

PR #328, opened 2023-02-06
Add some missing implicit dependencies

laeubi:
As discovered by new Tycho resolver...

### DIFF
--- a/ant/org.eclipse.ant.tests.core/build.properties
+++ b/ant/org.eclipse.ant.tests.core/build.properties
@@ -38,3 +38,9 @@ pom.model.property.testClass = org.eclipse.ant.tests.core.AutomatedAntSuite
 pom.model.property.tycho.surefire.testClassesDirectory = target/anttestscore.jar-classes
 pom.model.property.tycho.surefire.useUIHarness = true
 pom.model.property.tycho.surefire.useUIThread = true
+additional.bundles = org.eclipse.e4.core.contexts,\
+                     org.eclipse.e4.core.services,\
+                     org.eclipse.e4.core.di.extensions,\
+                     org.eclipse.e4.core.di.extensions.supplier,\
+                     org.eclipse.team.ui, \
+                     org.eclipse.core.net
--- a/ant/org.eclipse.ant.tests.ui/build.properties
+++ b/ant/org.eclipse.ant.tests.ui/build.properties
@@ -26,6 +26,15 @@ source.anttestsui.jar = test plugin/,\
                         Ant Debug Tests/,\
                         External Tools/,\
                         Ant Tests/
+additional.bundles = org.eclipse.e4.core.di,\
+                     org.eclipse.e4.core.contexts,\
+                     org.eclipse.e4.core.services,\
+                     org.eclipse.e4.core.di.annotations,\
+                     org.eclipse.e4.core.di.extensions,\
+                     org.eclipse.e4.core.di.extensions.supplier,\
+                     org.eclipse.compare,\
+                     org.eclipse.team.ui,\
+                     org.eclipse.core.net
 						
 source.lib/antUITestsSupport.jar = test support/
 jars.compile.order=anttestsui.jar,lib/antUITestsSupport.jar
--- a/resources/bundles/org.eclipse.core.tools.resources/build.properties
+++ b/resources/bundles/org.eclipse.core.tools.resources/build.properties
@@ -15,3 +15,4 @@ source.. = src/
 src.includes=*.html
 bin.includes=plugin.xml,icons/,doc/,*.html,.options, META-INF/,.
 qualifier=context
+additional.bundles = org.eclipse.team.ui
--- a/runtime/tests/org.eclipse.core.expressions.tests/build.properties
+++ b/runtime/tests/org.eclipse.core.expressions.tests/build.properties
@@ -26,3 +26,14 @@ src.includes = about.html
 pom.model.property.testClass = org.eclipse.core.internal.expressions.tests.AllExpressionTests
 pom.model.property.tycho.surefire.useUIHarness = true
 pom.model.property.tycho.surefire.useUIThread = true
+additional.bundles = org.eclipse.core.filesystem, \
+                     org.eclipse.e4.core.di,\
+                     org.eclipse.e4.core.contexts,\
+                     org.eclipse.e4.core.services,\
+                     org.eclipse.e4.core.di.annotations,\
+                     org.eclipse.e4.core.di.extensions,\
+                     org.eclipse.e4.core.di.extensions.supplier,\
+                     org.eclipse.core.resources,\
+                     org.eclipse.core.net,\
+                     org.eclipse.compare, \
+                     org.eclipse.team.core
--- a/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.compare; singleton:=true
-Bundle-Version: 3.8.500.qualifier
+Bundle-Version: 3.8.600.qualifier
 Bundle-Activator: org.eclipse.compare.internal.CompareUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/team/bundles/org.eclipse.compare/build.properties
+++ b/team/bundles/org.eclipse.compare/build.properties
@@ -20,3 +20,9 @@ bin.includes = icons/,\
                META-INF/
 src.includes = about.html,\
                schema/
+additional.bundles = org.eclipse.e4.core.di,\
+                     org.eclipse.e4.core.contexts,\
+                     org.eclipse.e4.core.services,\
+                     org.eclipse.e4.core.di.annotations,\
+                     org.eclipse.e4.core.di.extensions,\
+                     org.eclipse.e4.core.di.extensions.supplier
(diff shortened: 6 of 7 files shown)


iloveeclipse:
> Add some missing implicit dependencies

Missing for **what**?  Could you please explain the change?

mickaelistria:
How come that everything has been working so far if they are actually missing? Isn't it the Tycho resolver failing at something if it now requires things that were not required so far?

laeubi:
Currently [eclipse.platform](https://github.com/eclipse-platform/eclipse.platform) depends on items from [eclipse.platform.ui](https://github.com/eclipse-platform/eclipse.platform.ui) that in turn depend on projects in [eclipse.platform](https://github.com/eclipse-platform/eclipse.platform).
Because of this cyclic dependencies, builds that only build this repository (like the Jenkins / Github verification builds) are inherently inconsistent if they do not declare the dependencies from the reactor (eclipse.platform).

One very simple chain is:
eclipse-plugin:org.eclipse.compare:3.8.500.qualifier (reactor project) 
--> eclipse-plugin:org.eclipse.ui.ide:3.20.100.v20230203-1344 (target dependency)  (part of https://github.com/eclipse-platform/eclipse.platform.ui/tree/master/bundles/org.eclipse.ui.ide)
--> eclipse-plugin:org.eclipse.e4.core.di:1.8.300.qualifier (reactor project)

mickaelistria:
>  builds that only build this repository (like the Jenkins / Github verification builds) are inherently inconsistent if they do not declare the dependencies from the reactor (eclipse.platform)

Those dependencies do come from the target-platform, the previous build is listed as repository to resolve deps.

> Enhanced Tycho resolver is more strict on this.

Here, I don't see any benefit, so it actually seems more bogus. How come that with everything avaiable in reactor+target, the same build is now failing while it has been working for years?

laeubi:
> Those dependencies do come from the target-platform, the previous build is listed as repository to resolve deps.

Repositories are just an alternative way for input what Tycho calls a Target platform (other sources are target files or pom declared dependencies)

> Here, I don't see any benefit, so it actually seems more bogus.

Yep, it is really bogus that we won't be able to build eclipse-platform anymore from scratch :-)

> How come that with everything avaiable in reactor+target, the same build is now failing

Everything is still available, but some projects are missing implied dependencies to reactor projects.

> while it has been working for years?

Because previously such inconsistencies are just ignored or "bumped-away" when noticed, this won't be required if proper dependency management would be used or builds are no split into parts (see aggregatorbuild where everything is build together and thus everything works).

That's why I suggested to merge platform.ui + platform a while back

mickaelistria:
Sure, things are inconsistent, but OSGi doesn't require those extra dependencies and Tycho has managed to deal with OSGi-metadata only (deps in MANIFEST.MF) for a while as far as I know. So Tycho used to understand OSGi properly and now it doesn't any longer and some more build metadata is required to compensate it.
This still seems like a regression to me. Please try convincing me of what the new resolver does better and that makes it worth this extra burden.

laeubi:
This has nothing to do with OSGi and Tycho can resolve the projects without a problem (it just prints an **ERROR** now), but platform relies on the fact that if `A` (currently compiled) depends on `B` (prebuild artifact) what depends on `C `(not build yet) will work, what it does not unless you ignore the fact that if `C` changes `B` will be incompatible without being noticing it unless you run into a `java.lang.IncompatibleClassChangeError`, and someone (most likely @akurtakov) gracefully bump the version of `B` and triggers a new ibuild...

akurtakov:
I don't recall IncompatibleClassChangeError or anything like that nor bumping version for anything similar.

mickaelistria:
> it just prints an ERROR now

But this is not an error, this is totally legal.

> A (currently compiled) depends on B (prebuild artifact) what depends on C (not build yet) will work

Indeed, because A does not care that B requires C (unless A requires C or C is reexported). So this case is not a problem.
However, I wouldn't mind seeing the build order modified to prefer C over A automatically, but having such extra build metadata for something that was not a problem for a while is a sign that a new problem was introduced...

laeubi:
> But this is not an error, this is totally legal.

Not from the point of a build-tool (maven) because you can never build such graph of projects.

> but having such extra build metadata for something that was not a problem for a while is a sign that a new problem was introduced...

It is a problem long time ignored by eclipse that is cyclic build dependencies between git-repositories, as said, the aggregator build works all fine because all code is available in the reactor.

mickaelistria:
> Not from the point of a build-tool (maven) because you can never build such graph of projects.

So because Maven is weaker than OSGi, we should reduce the quality of Tycho resolver and add a ton of extra dependencies to fix a problem that was not existing so far (or a problem for which Tycho was providing a solution) ?

laeubi:
Everything comes at a price, and Tycho can't take the burden of the complexity of Platform not able to manage its dependencies. And the requirement of platform to build impossible build-graphs just for organizational reasons is not necessarily a goal of Tycho, it supports it as good as possible (with some extra configuration, not sure if plain maven would allow it).

This has nothing to do with weakening Tycho, it is the other way round good shaped projects where always punished for being slower and less flexible just for the sake of platform not willing to clean up their build-cycles.

akurtakov:
I still miss the part about what the benefits are from the new tycho resolver.

mickaelistria:
> Everything comes at a price

what is "everything" here? I see the price, I don't see the value.

akurtakov:
As someone that maintained Maven rpms in Fedora/RHEL I have to say that the build complexity and cycles that Eclipse Platform has is not bigger or worse than what Maven itself has/had. 
Comparing Eclipse platform and plugins is like Maven and its plugins and in both cases they have nothing in common in terms of complexity and actually this allows the plugins to be that much simpler to do because both platforms takes a big part of that complexity on their side. 
With that said, there are tons of things to improve on Eclipse side but this one exactly makes things worse not better (that said with my current knowledge of the new resolver).

laeubi:
> I still miss the part about what the benefits are from the new tycho resolver.

Much faster in the setup phase, advance parallelization, much more flexible (e.g. supports mixed reactors in almost every form, no need for needing all data before-the-fact), support for `-am` / `-amd` and `-fae` /  `-fn` and many more advanced features that will come in the future :-)

> I see the price, I don't see the value.

Sure for platform there is little value (beside that it makes cyclic dependencies visible now so we hopefully do not introduce new ones), but for Tycho this will allow to get rid of a lot burden that where carried around in the past.

> I have to say that the build complexity and cycles that Eclipse Platform has is not bigger or worse than what Maven itself has/had.

The main difference is that each maven-plugin is release independent, so e.g. maven-jar plugin can only use new features from maven core one there is a new maven core release. This has some drawbacks as well, but ensures that each plugin can always build independent and the is no relation between a change in plugin and core. With platform, if not released as a whole big thing that would work equally and you won't see this kind of inconsistency.

e.g platform has a (hidden) indirect dependencies to ECF, but ECF is not part of the aggregator build, that means if ever ECF would require a change from platform that crosses that path it can't release before the next platform release and platform can't build before ECF do a release --> mated ...

Anyways one can always enable the "old" way for this repository if it is valuable to hide those problems instead of trying to get it working.

akurtakov:
> The main difference is that each maven-plugin is release independent, so e.g. maven-jar plugin can only use new features from maven core one there is a new maven core release. This has some drawbacks as well, but ensures that each plugin can always build independent and the is no relation between a change in plugin and core. With platform, if not released as a whole big thing that would work equally and you won't see this kind of inconsistency.
> 
> e.g platform has a (hidden) indirect dependencies to ECF, but ECF is not part of the aggregator build, that means if ever ECF would require a change from platform that crosses that path it can't release before the next platform release and platform can't build before ECF do a release --> mated ...

For the sake of pure information : Maven <-> Wagon is exactly the same cycle from what I remember but that's not the point. 

> 
> Anyways one can always enable the "old" way for this repository if it is valuable to hide those problems instead of trying to get it working.

I'm afraid that this will spread like fire over equinox/p2/jdt/pde repositories and will become a nightmare to manage. Can you make the "old" way enabled when "build-individual-bundles" profile is enabled? That might be simplest for both sides.

laeubi:
> I'm afraid that this will spread like fire over equinox/p2/jdt/pde repositories and will become a nightmare to manage. Can you make the "old" way enabled when "build-individual-bundles" profile is enabled? That might be simplest for both sides.

Due to how maven works I don't think one can enable it with a profile but in `maven.config`, so we have a "backup" but other repos are not that much affected as these do usually only have a dependency form Project > Platform (what are not a problem) in contrast to how `platform` and `platform.ui` relate to each other.

I even tried to decouple things in the past, e.g. for p2 and equinox but that was nothing one can find consent, so i better add some additional items to some config files to fix that up instead of trying to "convince" people...

mickaelistria:
Is there some documentation available about how to stick to the resolver that is capable of handling those cases?

laeubi:
> Is there some documentation available about how to stick to the resolver that is capable of handling those cases?

See https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-tycho-resolver-is-the-default-now how to switch to old mode, can be configured for example in `.mvn/jvm.config` or  `.mvn/maven.config` but I have learned that the first one is preferable for system properties. Alternatively can be specified on commandline directly.

laeubi:
In parallel, I'll try to fall back to the Targetplatfrom in such case (then it could be a warning) and only fail if it is not found there... I'm not convinced this is actually a good idea, but that resembles more what is requested from OSGi POV...

laeubi:
I found a problem that can not be explained by cyclic dependencies (or at least they are not detected and reported what is equally bad):
- https://github.com/eclipse-tycho/tycho/issues/2059

From this discussion that a system-property is not really well suited for what I want to archive here and it must be controllable through the targetplatform configuration. Another finding seems that the `-Pbuild-individual-bundles` use-case is not currently covered well by the Tycho-Checks, I therefore will enhance the checks by a build of the `eclipse.platform` build as it seems the most challenging (for Tycho).

I therefore will put this one on-hold and *disable* the new default for now until the issues are addressed and more suitable configuration is available.

laeubi:
Closing this as no further work is planned here.

PR #332, opened 2023-02-07
Port team.ui away of deprecated ResourceSorter

akurtakov:
It has been announced for removal in 2019 and notification period ended in 2021 so it's high time to get rid of it in 2023. In order to do it proper a good number of places had to be modified from deprecated ViewerSorter to ViewerComparator
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=549953 for actual deprecation.


iloveeclipse:
Briefly checked in browser, seem to be OK.

akurtakov:
In that case merging.

PR #333, opened 2023-02-08
Move core.tests.resources to Java 17 and use features from it

akurtakov:


akurtakov:
@laeubi If you look at the windows failure here it already uses Maven 3.9.0 which makes tycho fail. What is our strategy here? Keep it that way and move Tycho ASAP? Downgrade (how?) maven to 3.8.7 and stay on it until someone notices it in the future?

laeubi:
I think one should configure an explicit Maven version, using the "latest and greatest" seem not a good strategy as it shows here that 3.9.x contains some breaking changes it is likely people want to use 3.8.x for a while. 

https://github.com/eclipse-tycho/tycho/blob/acdf49a23d04bde88184dd6364160b08b728ec43/.github/workflows/maven.yml#L43-L46

Issue #301, opened 2022-12-19
Leverage "Foreign Function & Memory API"


laeubi:
Currently especially the resource-api uses some specialized smaller JNI access e.g. for setting file-access (linux/mac) or watch directories (windows).

The upcoming [Project Panama](https://openjdk.java.net/projects/panama/) aims at a close integration of java and native code as an alternative for hand-crafted JNI code.

One very interesting part is the [jextract](https://github.com/openjdk/panama-foreign/blob/foreign-jextract/doc/panama_jextract.md) tool as part of JDK19.

This is based on the [JEP 412: Foreign Function & Memory API](https://openjdk.java.net/jeps/412) already part of Java-17.

Even though I don't know if Java 17 is yet an option (@akurtakov ?) and java 19 even a bit far away I'd like to bring the topic to platform right now so we can investigate and maybe improve and give feedback (e.g. can all relevant native APIs be parsed by jextract?), so hopefully in a glory future we no longer need to build native libraries and can use a pure java interface what will greatly improve the circle of people that could work in that area.

PR #346, opened 2023-03-06
Add a fragment that supports linux native polling using panama

laeubi:
This is a draft to investigate the possible options to use the [Project Panama](https://openjdk.org/projects/panama/) in [Eclipse platform](https://github.com/eclipse-platform/eclipse.platform/issues/301) code.

The goal is to wrap the Linux [inotify API](https://man7.org/linux/man-pages/man7/inotify.7.html) to watch folders with native hooks but providing graceful degradation in case of resource exhaustion so we need access to the native error codes, and explicitly only watching folders.

Beside that, this is to investigate the usage of [Multi-Release JAR Files](https://openjdk.org/jeps/238) to adopt to changes between java versions, something that is currently not used (and not yet well supported [by the IDE itself](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265)) but will become prominent in the future of Java APIs evolving.

So this can also be used to discuss or being used as a starting point for own experiments (e.g. any mac users might try the native mac API as well that can be seen here https://github.com/openjdk/jdk/pull/10140)


akurtakov:
Is this still of interest? While I have high hopes for Project Panama I'm not eager to have both jni and panama implementations around.

laeubi:
@akurtakov I need to rework this to align with final FFM API for Java 21, as there is no JNI implementation for this feature I don't see any risk for overlap.

PR #350, opened 2023-03-07
Only consider depth of content types when they are related

### LINKED ISSUE #151: By default hpp files are not opened with CDT editor anymore if tm4e language pack is installed
Initially reported at tm4e: https://github.com/eclipse/tm4e/issues/437

1. Get 4.25 SDK
2. Create simple project and empty hello.hpp file
3. Install CDT SDK into it, restart
4. Double click on hello.hpp - it is opened in default C++ editor from CDT. 
5. Right click in explorer -> "Open with" shows C++ editor in the proposed editors list
6. Close C++ editor.
7. Install org.eclipse.tm4e (TextMate support in Eclipse IDE, 0.5.1.202207061559) and  org.eclipse.tm4e.language_pack (TextMate Language Pack, 0.1.0.202206131926)
8. Restart Eclipse
9. Double click on hello.hpp - it is opened in generic text editor, NOT with CDT C++ editor.
10. Right click in explorer -> "Open with" does NOT show C++ editor in the proposed editors list

Both CDT and tm4e define content types for hpp files, CDT with high prio, tm4e with low:
https://github.com/eclipse/tm4e/blob/972061af2add6231c5667e31f83d393c102bdf9b/org.eclipse.tm4e.language_pack/plugin.xml#L69

Commenting line above in tm4e plugin.xml fixes the issue.

This seem to be a bug in the platform. CDT contribution priority is high, tm4e is low. tm4e generic editor shouldn't ever be opened by default if CDT is installed.

However, interesting fact is that cpp files can still be opened with CDT C++ editor (even if tm4e also defines the content type for them, in the same line with hpp files).

May be related code: https://github.com/eclipse-platform/eclipse.platform/blob/c198507304892befd104b82dc7af20a0dedb4494/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java#L47

jonahgraham:
In the case that two content types match a given input, the depth was being used to select between them. This was done top make sure more specific types for related types were selected as appropriate.

The way content types are selected is by first sorting all the candidate types, and then choosing the first one.

This change makes the sort only consider the depth of the content type if the content types have a descendent-ancestor relationship. If they are not in the same lineage, the depth is not a criteria for sorting the content types.

New content type tests have been added to cover the case from #351, specifically, conflict4 matches the conflict for *.hpp files between the tm4e and cdt plug-ins.

The comment update on conflict2 test and the new conflict2a test should probably have been done when Bug 86915 was completed because in 9cbda7e4a42e2fd94e50d8fe700c0ad1574fb5a9 the order of content types was changed, but the comment was not changed, and it wasn't apparent to me if there was an explicit test that covered the conflict case of described content types.

Fixes #151

### DIFF
--- a/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java
+++ b/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java
@@ -41,16 +41,30 @@ public final class ContentTypeCatalog {
 	private ContentTypeManager manager;
 
 	/**
-	 * A sorting policy where the more generic content type wins. Lexicographical comparison is done
-	 * as a last resort when all other criteria fail.
+	 * Return true if type1 is an ancestor of type2 or if type2 is an ancestor of
+	 * type1
+	 *
+	 * @param type1
+	 * @param type2
+	 * @return true type1 is ancestor or type2, or vice versa. false otherwise
+	 */
+	private static boolean isAncestor(ContentType type1, ContentType type2) {
+		return type1.isKindOf(type2) || type2.isKindOf(type1);
+	}
+
+	/**
+	 * A sorting policy where the more generic content type wins. Lexicographical
+	 * comparison is done as a last resort when all other criteria fail.
 	 */
 	private final Comparator<IContentType> policyConstantGeneralIsBetter = (IContentType o1, IContentType o2) -> {
 		ContentType type1 = (ContentType) o1;
 		ContentType type2 = (ContentType) o2;
-		// first criteria: depth - the lower, the better
-		int depthCriteria = type1.getDepth() - type2.getDepth();
-		if (depthCriteria != 0)
-			return depthCriteria;
+		if (isAncestor(type1, type2)) {
+			// first criteria: depth - the lower, the better
+			int depthCriteria = type1.getDepth() - type2.getDepth();
+			if (depthCriteria != 0)
+				return depthCriteria;
+		}
 		// second criteria: priority - the higher, the better
 		int priorityCriteria = type1.getPriority() - type2.getPriority();
 		if (priorityCriteria != 0)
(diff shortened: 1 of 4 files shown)


mickaelistria:
I think the priority should not matter when 1 contnet-type is a super/sub-type of the other: in such case, the most specialized content-type needs to be returned as it encompass the higher priority one as well.
I would recommend that we write good tests in ContentTypeManagerTest to ensure we first agree on the expectations.

merks:
@jonahgraham Maybe you have a concrete case/example where the old algorithm produces a result different than the one you would like?

jonahgraham:
I have created the issue in this repo now, see #351

mickaelistria (changes requested):
(copied from https://github.com/eclipse-platform/eclipse.platform/issues/351#issuecomment-1461618797 )

> As far as I can tell, depth is not documented as an important factor

Indeed depth is not an important factor per se, the important factor is to return "most specialized" content-type in a lineage, ie basicContentType(priority=high) being a parent of specializedContentType(priority=low), then querying a file which has both content-type should return specializedContentType independently of the priority. That is what enables the most value.
Depth was an implementation choice to achieve that, but it doesn't make sense to use it when the content-types we compare are not of the same lineage: basicContentType has 2 independent children specializedContentType1(priority=high) and specializedContentType2(priority=normal) and that one has a child surSpecializedContentType2(priority=normal), then the expectation is that specializedContentType1 is returned first because depth isn't an interesting criterion to compare specializedContentType1 and surSpecializedContentType2.

An implication of both is this trickier case:
basicContentType
  * specializedContentType1(priority=high)
    * surSpecializedContentType1(priority=low)
  * specializedContentType2(priority=normal)
    * surSpecializedContentType2(priority=normal)
 
Then we would expect that the best content-type here is surSpecializedContentType1. Basically, to fulfill both requirements the priority of a content-type is the max of its lineage.

jonahgraham:
> [...] should return specializedContentType independently of the priority.

This comment caused me lots of confusion as in practice that isn't what happens unless the content of the file is also analyzed, which seems mostly to only happen when dealing with binary or xml files in practice. I added a new test case, conflict2a and fixed a comment on conflict2 that covers this case. The best summary of the behaviour came when it was last written in [Bug 86915](https://bugs.eclipse.org/bugs/show_bug.cgi?id=86915) back in 2005.

> An implication of both is this trickier case:

My PR does not handle this case. 

Once the CI confirms the tests behave the same as on my machine I will mark this ready for review. I updated the description + summary of the PR to properly reflect the current state.

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java
```diff
@@ -41,16 +41,58 @@ public final class ContentTypeCatalog {
...
-	 * A sorting policy where the more generic content type wins. Lexicographical comparison is done
-	 * as a last resort when all other criteria fail.
+	 * Return true iff type1 is an ancestor of type2 or if type2 is an ancestor of
+	 * type1
+	 *
+	 * @param type1
+	 * @param type2
+	 * @return true type1 is ancestor or type2, or vice versa. false otherwise
+	 */
+	private static boolean isAncestor(ContentType type1, ContentType type2) {
```
mickaelistria, line 51: Why not `type1.isKindOf(type2) || type2.isKindOf(type1)`?

mickaelistria:
Thank you!

PR #352, opened 2023-03-08
Implement removeFileSpec for pattern style content types

jonahgraham:
Commit 85b23928b9d76e3b2ed9aeca2df0c520a79040ca for Bug 263316 introduced filename patterns for content type selection, but at the time the remove wasn't implemented.

This patch adds the implementation for removeFileSpec. The test was extended to remove the new content type and verify that the findContentTypeFor returns to the initial state.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/IContentTypeManagerTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/IContentTypeManagerTest.java
@@ -970,8 +970,14 @@ public class IContentTypeManagerTest extends ContentTypeTest {
 		assertNull("File pattern unknown at that point", single);
 
 		textContentType.addFileSpec("*Text*", IContentType.FILE_PATTERN_SPEC);
+		try {
+			single = finder.findContentTypeFor(getInputStream("Just a test"), "someText.unknown");
+			assertEquals("Text content should now match *Text* files", textContentType, single);
+		} finally {
+			textContentType.removeFileSpec("*Text*", IContentType.FILE_PATTERN_SPEC);
+		}
 		single = finder.findContentTypeFor(getInputStream("Just a test"), "someText.unknown");
-		assertEquals("Text content should now match *Text* files", textContentType, single);
+		assertNull("File pattern unknown at that point", single);
 	}
 
 	@Test
--- a/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java
+++ b/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentTypeCatalog.java
@@ -277,12 +277,22 @@ public final class ContentTypeCatalog {
 	}
 
 	synchronized void dissociate(ContentType contentType, String text, int type) {
-		Map<String, Set<ContentType>> fileSpecMap = ((type & IContentType.FILE_NAME_SPEC) != 0) ? fileNames : fileExtensions;
-		String mappingKey = FileSpec.getMappingKeyFor(text);
-		Set<ContentType> existing = fileSpecMap.get(mappingKey);
-		if (existing == null)
-			return;
-		existing.remove(contentType);
+		Map<String, Set<ContentType>> fileSpecMap = null;
+		if ((type & IContentType.FILE_NAME_SPEC) != 0) {
+			fileSpecMap = fileNames;
+		} else if ((type & IContentType.FILE_EXTENSION_SPEC) != 0) {
+			fileSpecMap = fileExtensions;
+		}
+		if (fileSpecMap != null) {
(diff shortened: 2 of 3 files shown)


mickaelistria:
Can you please add a commit that bumps the version? Is there any particular reason why this is marked as a draft?

jonahgraham:
It's a draft because it wasn't ready for review, in part because it needs the requested version bump.

vogella:
@jonahgraham can you finish this one?

jonahgraham:
AFAICT all the failing checks are invalid and that this change is ready to go.

laeubi:
> AFAICT all the failing checks are invalid and that this change is ready to go.

Why should the be invalid? Some test are failing on some platforms, but this mean we ned to investigate / fix them and not that checks are invalid and should be ignored per-se.

merks:
I think the point is that the failing tests are not failing because of the changes of this PR, but rather there's a problem with the tests themselves on specific platforms for unrelated topics:

image

So the wording that the "checks are valid" is inaccurate.  That said, you're not suggesting this change shouldn't be merged until the unrelated test failures are fixed are you?  It seems a common practice, however bad, to commit things with failing tests...

laeubi:
The checks indicate that this change adds one new testfailure:

grafik

So I would expect at least to document this failure, if we are sure it is only a flapping test, instead of assuming that **any failing test** is fine by claiming checks are invalid, even a failing check is valid when it shows a test failure we ignore but only if it was specifically verified that this is the case

Beside that, currently the **Jenkins Check** also fails, what I think is the absolute minimum for a PR to be merged.

merks:
Doesn't this mean that only Windows and Mac are failing?

image

And only the ones here:

https://github.com/eclipse-platform/eclipse.platform/pull/352#issuecomment-1500805940

Which to look to me look totally randomly unrelated...

What do you mean by the "Jenkins Check"?

merks:
FYI, I tried to kick off all the builds again to see if they fail the same way consistently.    I do normally expect a Jenkins build for a PR, but I don't actually see one.  Maybe I'm not looking correctly or at the correct place?

laeubi:
> What do you mean by the "Jenkins Check"?
> I do normally expect a Jenkins build for a PR, but I don't actually see one

grafik

> Doesn't this mean that only Windows and Mac are failing?

Why should it be okay to have **more failing test** on mac/windows? This usually means that either the code or the test is platform specific and will likely not work on other platforms we support as we expect.

> Which to look to me look totally randomly unrelated...

Everything on Platform is related somehow ;-)
As said, if we think it is unrelated then one should create an issue to fix that unrelated failure and explain *why* it is unrelated, e.g. it fails already at master, but a `+1` is always something look deeper instead of simply accept it.

iloveeclipse:
> Unable to parse maven.config file options: Unrecognized option: --threads 1C

Could that be related to recent maven update?

laeubi:
> Could that be related to recent maven update?

Yes, it needs to be

```
 --threads
1C
```

(with a line break)

laeubi:
See:
- https://github.com/eclipse-platform/eclipse.platform/pull/407

jonahgraham:
The invalid term was because of the invalid freeze. Ie we shouldn't be in a freeze after the build has been released right?

As for macos and Windows tests, they appeared invalid to me as nearly every Github actions ci build on those platforms have test failures since it was turned on a couple of weeks ago. I assume the expansion to running on gha for macos and Windows is still under development. Therefore those reported failures seemed invalid to me. Even the rerun of the jobs seems worse as windows has hung(?). I see @HeikoKlare has been doing lots of test stability improvements recently, so I look forward to this situation being better in the future.

Therefore I stated this is ready to go.

iloveeclipse:
I've rebased on master to get maven arguments fix for jenkins build.

laeubi:
> The invalid term was because of the invalid freeze

If you think that anything is wrong please report an issue so it can be fixed instead of assuming it could be ignored.

> Therefore those reported failures seemed invalid to me.

There is nothing "under development", every error is a failure that must be fixed at best everyone tries to help with that instead of waiting for "someone" to possibly fix things, so from releng / build perspective everything is stable now in that area.

HeikoKlare:
> I assume the expansion to running on gha for macos and Windows is still under development. Therefore those reported failures seemed invalid to me.

Just for clarification: The failures are valid in the sense that some tests (and implicit dependencies between tests) are erroneous or assume specific concurrent timing behavior that is not fulfilled when running on the GitHub Actions infrastructure. Thus, the GitHub Actions build only revelead some problems that must have already existed before.

With respect to this PR: The failing tests in the latest build do not seem to be related to the PR but are independent, random fails, which are already documented. One of them would even be fixed when merging #391.

In addition, the PR contents look good to me.

mickaelistria (approved):

mickaelistria:
PR looks correct to me, and also to @jonahgraham and @HeikoKlare ; let's merge.

mickaelistria:
Thanks @jonahgraham

PR #356, opened 2023-03-16
JavaSE-17 update with SecurityManager

jukzi:
.classpath:
 JRE_CONTAINER
*.launch:
 JRE_CONTAINER
org.eclipse.jdt.core.prefs:
 compiler.compliance
 codegen.targetPlatform
 compiler.source
MANIFEST.MF:
 Bundle-RequiredExecutionEnvironment
 Bundle-Version (minor version update)
pom.xml
 (minor version update)


sratz:
> ```
> 11:24:30.880 [INFO] Using Declarative Service specification version 1.3.0 to generate component definitions
> 11:24:30.903 [WARNING] /home/jenkins/agent/workspace/eclipse.platform_PR-356/runtime/bundles/org.eclipse.e4.core.services/src/org/eclipse/e4/core/internal/services/MessageFactoryServiceImpl.java:[23] 
> 	import java.security.AccessController;
> 	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> The type AccessController has been deprecated since version 17 and marked for removal
> 11:24:30.903 [WARNING] /home/jenkins/agent/workspace/eclipse.platform_PR-356/runtime/bundles/org.eclipse.e4.core.services/src/org/eclipse/e4/core/internal/services/MessageFactoryServiceImpl.java:[106] 
> 	if (System.getSecurityManager() == null) {
> 	           ^^^^^^^^^^^^^^^^^^^^
> The method getSecurityManager() from the type System has been deprecated since version 17 and marked for removal
> ```

For Java 17 there is nothing to do here. It's deprecated but Java 17 still allows it, so we should be prepared for its presence. If we move to BREE 18+ then we can remove this condition (then, `System.getSecurityManager()` will always return `null`).

> ```
> Warning: 400 [WARNING] /home/runner/work/eclipse.platform/eclipse.platform/ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/AntSecurityManager.java:[35] 
> 	public class AntSecurityManager extends SecurityManager {
> 	                                        ^^^^^^^^^^^^^^^
> The type SecurityManager has been deprecated since version 17 and marked for removal
> Warning: 400 [WARNING] 
> error: warnings found and -failOnWarning specified
> ```

Here I am not sure. I am not familiar with eh `AntSecurityManager` and its intentions. But as long as Java 17 is BREE, we can opt to simply ignore this warning. Once me move to BREE18+ this probably has to be removed (replaced?) completely.



> @sratz you need it?

No, we don't. I think you are referring to #295? That was just a side effect that occurred if some plugin is active that does install a SecurityManager.

merks:
I think the primary purpose of that is to prevent exit:

```
	@Override
	public void checkExit(int status) {
		// no exit allowed from the restricted thread...System.exit is being called
		// by some ant task...do not want Eclipse to exit if
		// in the same VM.
		if (Thread.currentThread() == fRestrictedThread) {
			throw new AntSecurityException();
		}
		if (fSecurityManager != null) {
			fSecurityManager.checkExit(status);
		}
	}
```

jukzi:
My guess is that we we wil never build an eclipse that has BREE 18, because 18 is not LTS and there should be an eclipse that runs within 17 and 18 as well. So it woould be good if we could already remove all code marked for removal.

merks:
Maybe the newer version of Ant has a different/better solution.   One was being contributed to Orbit, but I don't know if that was actually finished.

There are lots of uses of this as well and it will go away too:

image

sratz:
> My guess is that we we wil never build an eclipse that has BREE 18, because 18 is not LTS and there should be an eclipse that runs within 17 and 18 as well. So it woould be good if we could already remove all code marked for removal.

As long as we support Java 17 we will have to deal with SecurityManager being present, so the coding in `MessageFactoryServiceImpl` should probably stay. Then with the next BREE bump to 21 LTS we can remove it.

As for `AntSecurityManager`, it's probably a good idea to get rid of it immediately, if possible.

sratz:
> There are lots of uses of this as well and it will go away too:
> 
> image

Hopefully, these are also just `AccessController.doPrivileged()` calls, that are just dealing with a SecurityManager someone else might have installed. They also need to be kept as long as we support Java 17.

The deprecation/removal plan (https://openjdk.org/jeps/411) is quite defensive, so the code will not suddenly break with newer Java versions but semantically many of those operations will simply become no-ops.

Deferring removal of all these calls until BREE 21 should therefore be fine.

laeubi:
Just a note, [BREE is deprecated since many years now](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.bree), and one currently should uses [EE capability](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.namespaces.html#framework.namespaces.osgi.ee) instead  as BREE in its current form this is only a minimum!

So if one really wants to get safe, (and these methods will be removed in Java 21) there are two choices:

1. Using a Multi-Release Jar that has variants for 17 (with security manager) + 21 (without security manager)
2. One is using EE capability and limit the supported range to `(17, 21]` so the bundle does not resolve on a higher EE

akurtakov:
FWIW, I've just replaced one usage of SecurityManager with StackWalker in https://github.com/eclipse-platform/eclipse.platform.text/commit/9a47ea6220c2fcb4fe2bd93c95d3814e288bc342 . Hopefully that helps someone.

jukzi:
I am not sure what the best way to ignore the warnings is: 
either remove "failOnWarning" in the build jobs (as in platform.ui) or adding `@SuppressWarnings` in the code.

laeubi:
> adding `@SuppressWarnings` in the code.

Please do that probably  with a comment why it was suppressed.

iloveeclipse:
@jukzi : what is the status of this PR? Anything is still pending? Can it be merged?

Asking because this is probably needed to fix https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/484

jukzi:
From my point of view it could technically be merged. However it only contains a lot of suppress warnings which could cause serious(!) problems in future JDKs: if the securitymanager becomes a NO-OP the check if ant is running in eclipse will silently fail and probably just terminate the IDE. It would be better if serious solutions could be found! (I currently dont't have the time for that work.)

PR #369, opened 2023-03-23
Test Build Issue

laeubi:
This is only to investigate / track possible build issues with the new setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+xxxx
+
 # eclipse.platform
 This repository contains the basis for the [Eclipse IDE](https://www.eclipse.org/eclipseide/) and a sub-repository of the [eclipse-platform](https://github.com/eclipse-platform) organization. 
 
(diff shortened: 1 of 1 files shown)


laeubi:
One of the issues on mac seems related to this fix in Tycho and should hopefully be fixed then:
- https://github.com/eclipse-tycho/tycho/pull/2280

the other is
> Unbound classpath container

what seem to indicate that the macos fragment for jdt is not resolved, I remember we have had a similar issue at m2eclipse ( @kwin @HannesWell ), I'll check if this can maybe fixed at JDT itself similar to how SWT requires a fragment conditionally.

HannesWell:
> what seem to indicate that the macos fragment for jdt is not resolved, I remember we have had a similar issue at m2eclipse ( @kwin @HannesWell ), I'll check if this can maybe fixed at JDT itself similar to how SWT requires a fragment conditionally.

Sorry, I cannot remember that.

laeubi:
Everything works now as expected.

Issue #319, opened 2023-01-25
random failing TeamPrivateMemberTest.testDeltas


jukzi:
```
Verifier has not yet been given a resource delta
java.lang.AssertionError: Verifier has not yet been given a resource delta
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.core.tests.resources.ResourceDeltaVerifier.getMessage(ResourceDeltaVerifier.java:466)
	at org.eclipse.core.tests.resources.TeamPrivateMemberTest.testDeltas(TeamPrivateMemberTest.java:462)
```
in https://github.com/eclipse-platform/eclipse.platform/pull/318/checks?check_run_id=10878070950

like https://github.com/eclipse-platform/eclipse.platform/issues/99

Issue #103, opened 2022-04-14
random failing org.eclipse.core.tests.internal.localstore.FileSystemResourceManagerTest.testWriteFile


jukzi:
`java.lang.IllegalStateException: No ResourceInfo for: L/Project0/testWriteFile`
https://ci.eclipse.org/platform/job/eclipse.platform.resources/job/PR-46/4/testReport/junit/org.eclipse.core.tests.internal.localstore/AllTests/testWriteFile/

```!ENTRY org.eclipse.core.resources 4 0 2022-04-14 08:08:42.439
!MESSAGE exception occured during write of file: L/Project0/testWriteFile
!STACK 1
org.eclipse.core.internal.resources.ResourceException(/Project0/testWriteFile)[274]: java.lang.Exception: Resource is out of sync with the file system: '/Project0/testWriteFile'.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.write(FileSystemResourceManager.java:1133)
	at org.eclipse.core.tests.internal.localstore.FileSystemResourceManagerTest$WriteFileContents.run(FileSystemResourceManagerTest.java:536)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2319)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2339)
	at org.eclipse.core.tests.internal.localstore.FileSystemResourceManagerTest.write(FileSystemResourceManagerTest.java:497)
	at org.eclipse.core.tests.internal.localstore.FileSystemResourceManagerTest.testWriteFile(FileSystemResourceManagerTest.java:341)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
[...]

PR #371, opened 2023-03-24
Fix missing workspace cleanups after executing tests in oe.core.tests.resources

HeikoKlare:
Several test cases/classes in `org.eclipse.core.tests.resources` do not perform a proper cleanup after their execution. In consequence, the workspace is not in a clean state afterwards, e.g., in terms of remaining projects/files, changed workspace settings, and still registered or missing listeners.
This can lead to failures of subsequent test executions that assume a clean workspace state, depending on the execution order of tests. This particularly occurs when executing a bunch of test classes within Eclipse in a single run, but may also affect the Maven build.

Since tests should be isolated to be executable in any order without affecting other tests, I propose the following changes to cleanup several places I found when investigating test failures:
* Add missing removal of workspace listeners 
* Properly reset workspace properties such as the `autobuild` property in `teardown`
* Add missing calls to `teardown` because of unnecessarily overwritten methods, missing calls to the super method or missing extensions of an according test class ensuring cleanup (i.e. `ResourceTest`)
* Add removal of all files places in the workspace directory but not contained in projects

This missing isolation of test executions may also be a reason for the randomly occuring `Verifier has not yet been given a resource delta` test failures, such as #99 and #319. At least during local execution I had random occurrences of that failure because of interfering workspace listeners still registered from previously executed tests.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -16,13 +16,16 @@ package org.eclipse.core.tests.internal.resources;
 
 import java.util.*;
 import junit.framework.ComparisonFailure;
+import junit.framework.Test;
 import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.resources.WorkspacePreferences;
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.*;
-import org.eclipse.core.tests.resources.ResourceTest;
+import org.eclipse.core.tests.resources.AutomatedResourceTests;
+import org.eclipse.core.tests.resources.WorkspaceSessionTest;
+import org.eclipse.core.tests.session.WorkspaceSessionTestSuite;
 
-public class WorkspacePreferencesTest extends ResourceTest {
+public class WorkspacePreferencesTest extends WorkspaceSessionTest {
 	private IWorkspace workspace;
 	private Preferences preferences;
 
@@ -300,4 +303,9 @@ public class WorkspacePreferencesTest extends ResourceTest {
 		assertEquals(message + " - 9", description1.getMaxBuildIterations(), description2.getMaxBuildIterations());
 		assertEquals(message + " -10", description1.isKeepDerivedState(), description2.isKeepDerivedState());
 	}
+
+	public static Test suite() {
+		return new WorkspaceSessionTestSuite(AutomatedResourceTests.PI_RESOURCES_TESTS, WorkspacePreferencesTest.class);
+	}
+
 }
--- a/resources/tests/org.eclipse.core.tests.resources/META-INF/MANIFEST.MF
+++ b/resources/tests/org.eclipse.core.tests.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Core Tests Resources
 Bundle-SymbolicName: org.eclipse.core.tests.resources; singleton:=true
-Bundle-Version: 3.11.0.qualifier
+Bundle-Version: 3.11.100.qualifier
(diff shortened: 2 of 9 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
```diff
@@ -16,13 +16,16 @@
...
 import org.eclipse.core.internal.resources.WorkspacePreferences;
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.*;
-import org.eclipse.core.tests.resources.ResourceTest;
+import org.eclipse.core.tests.resources.AutomatedResourceTests;
+import org.eclipse.core.tests.resources.WorkspaceSessionTest;
+import org.eclipse.core.tests.session.WorkspaceSessionTestSuite;
 
-public class WorkspacePreferencesTest extends ResourceTest {
+public class WorkspacePreferencesTest extends WorkspaceSessionTest {
```
HeikoKlare, line 28: This should be a `WorkspaceSessionTest` to have the test run with a dedicated workspace setup. Otherwise, the import/export test for preferences leads to a complete reinitialization of the preferences, which miss some of the previously attached listeners. Subsequent tests can fail due to those missing listeners. Having a dedicated workspace setup seems to be the option for test isolation from my point of view.

jukzi:
Hi Haiko, for eclipse we typically squash our changes before review, please squash to a single commit to keep history small or create separate pull requests when changes are totally unrelated. 
The setAutobuild() may cause unwanted side affects when it happens during an autobuild. Enabling and disabling it may cause a new autobuild run. So should consider to wait for autobuild to finish (see org.eclipse.core.internal.events.BuildManager.waitForAutoBuild() / waitForAutoBuildOff).

HeikoKlare:
@jukzi Thank you for your instant feedback!
I have squashed the commits and updated the pull request with the single resulting commit. I also improved the restoration of the autobuild setting by properly waiting for the autobuild job to finish.

jukzi:
Please solve all failed tests. If any fail is unrelated to the change make sure there is a bug open for the random(?) failed test.

HeikoKlare:
Both failures are unrelated to these changes. An issue for one of them was already open (https://github.com/eclipse-platform/eclipse.platform/issues/103). I created another one for the second failure, which also occurs in unrelated pull requests (https://github.com/eclipse-platform/eclipse.platform/issues/373).

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -406,8 +407,38 @@ protected void cleanup() throws CoreException {
 			}
 		}, null);
 		getWorkspace().save(true, null);
-		//don't leak builder jobs, since they may affect subsequent tests
+		// don't leak builder jobs, since they may affect subsequent tests
 		waitForBuild();
+		deleteFilesInWorkspace();
+	}
+
+	private void deleteFilesInWorkspace() {
```
jukzi, line 415: this deletes files in the Filesystem - not in the workspace. It would be better to remove them with workspace API. Otherwise the workspace will be out of sync with filesystem. Probably there is already helpercode for that somewher.e (isn't that line 402??!)

HeikoKlare: Yes, the code deletes files in the filesystem and not in the workspace. This was my intention, which was obviously not properly reflected by the naming. The reason for that proposal is improper cleanup by different test classes, which lead to the Eclipse workspace being cleaned (by deleting the root), but files still remaining in the file system at the workspace location. In subsequent tests, this could lead to failures due to unexpected files being present.
My proposal was kind of a fallback solution for tests that somehow only clean the workspace but not all files in the workspace directory. However, I agree with you that this may not be the best option. I will figure out _why_ files are still present. One reason I saw is that tests delete the workspace root without forcing files in closed projects to be deleted, but there may also be other cases.

However, I will separate that from this pull request and make a distinct issue/pull request for the file cleanup topic. I have removed the according changes (i.e. the `deleteFilesInWorkspace` method in `ResourceTest`) from the pull request.

vogella:
@jukzi can this be merged? Looks like it improves the test setup and does not introduce new errors.

jukzi:
ok, looks like random fails only.
@HeikoKlare please keep an eye on https://download.eclipse.org/eclipse/downloads/index.html in the next days o see if it causes additional errors.

jukzi:
@HeikoKlare 
we now have TestBug294854.testDelete_02 failing on all platforms. Since i have seen this first time with this commit i ask you to investigate.
https://download.eclipse.org/eclipse/downloads/drops4/I20230329-1800/testResults.php

Issue #373, opened 2023-03-24
Random failing org.eclipse.core.tests.runtime.jobs.JobEventTest.testDeadlockRecovery


HeikoKlare:
```
timeout (probably deadlock)
```

e.g. in 
https://github.com/eclipse-platform/eclipse.platform/pull/371/checks?check_run_id=12255479559
https://github.com/eclipse-platform/eclipse.platform/pull/349/checks?check_run_id=12120746113

Does not occur when locally executing Maven build on Windows and Linux system.

HeikoKlare:
The test fails again:
https://download.eclipse.org/eclipse/downloads/drops4/R-4.27-202303020300/testresults/html/org.eclipse.core.tests.runtime_ep427I-unit-win32-java17_win32.win32.x86_64_17.html

Issue is the same as mentioned in #373: The deadlocking thread is still dumping stack traces when the timeout for the thread to avoid endless execution is already reached, as can be seen [in the log](https://download.eclipse.org/eclipse/downloads/drops4/R-4.27-202303020300/testresults/ep427I-unit-mac64-java17_macosx.cocoa.x86_64_17/org.eclipse.core.tests.runtime.AutomatedRuntimeTests.txt).

So timeout value is still too low on some infrastructures. I will provide a further cleanup of the test to use the existing barrier implementation with proper, generic timeout behavior instead of using manually implemented CountDownLatches with another timeout thread.

PR #378, opened 2023-03-27
Close project when failing to restore metadata in SaveManager (#377)

### LINKED ISSUE #377: SaveManager does not close project when reading metadata fails while restoring workspace
When a workspace is not properly saved while/after performing resource changes, inconsistent metadata can lead to a faulty state when restoring the workspace during subsequent loads.

In detail: The test `TestBug294854.testDelete` randomly fails (see https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-371/6/testReport/ and particularly the `FileNotFoundException` in the according log https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-371/6/consoleFull).
The test deletes a projects and afterwards calls `System.exit`. After a project deletion, the snapshot job is scheduled to store the workspace modification. However, there is a race condition because the snapshot job may properly write data before `System.exit` is executed or after it is executed. For that reason the other tests in `TestBug294854` try to wait for the snapshot job being finished, as has been discussed here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=294854
When the snapshot job does not finish before calling `System.exit`, the subsequent workspace restoration by the `SaveManager` tries to load the deleted project and throws a `FileNotFoundException`. Although the exception is caught, the project is not closed as intended when an exception is thrown due to a missing try-finally:
https://github.com/eclipse-platform/eclipse.platform/blob/8f9546dbcdbef9ad236c79821c030cad39ec250a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java#L914-L916

In summary, I suggest the following changes to fix the issues:
* `SaveManager` should properly close the project when an exception is thrown while writing metadata on workspace restoration
* `TestBug294854.testDelete` has to be fixed and should be split into two test cases
  1. Wait for the snapshot job to be finished and expect the project to be properly removed afterwards
  1. Call `System.exit` deterministically between deleting the project and executing the snapshot job and expect the project to be closed when restoring the workspace

HeikoKlare:
As explained in #377, the `SaveManager` does not properly close a project when failing to restore the workspace with inconsistent metadata (e.g. after a process kill). The randomly failing test case `TestBug294854.testDelete` revealed that behavior.

The proposed changes ensure that a project is properly closed in such a case (as intended by the existing code). In addition, the changes fix the indeterministic behavior of `TestBug294854.testDelete` by splitting it up into the cases which randomly occurred before and of which only one was only successful:
1. Kill the process before metadata has been stored in the snapshot job
2. Kill the process after metadata has been stored in the snapshot job

In addition, I propose to replace the `sleep` calls, which try to wait for the snapshot job to be finished, with an according utility that properly waits for it.

Fixes #377

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -910,10 +910,13 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 		}
 		project.internalSetDescription(description, false);
 		if (failure != null) {
-			// write the project tree ...
-			writeTree(project, IResource.DEPTH_INFINITE);
-			// ... and close the project
-			project.internalClose(monitor);
+			try {
+				// write the project tree ...
+				writeTree(project, IResource.DEPTH_INFINITE);
+			} finally {
+				// ... and close the project
+				project.internalClose(monitor);
+			}
 			throw failure;
 		}
 		if (Policy.DEBUG_RESTORE_METAINFO)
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Failing test is already documented (https://github.com/eclipse-platform/eclipse.platform/issues/373).

vogella:
@jukzi could you also review this one?

jukzi (changes requested):
little formal detail: Please update the copyright year for SaveManager.

HeikoKlare:
Thanks for the feedback! I have updated the copyright year and pushed the according change.

iloveeclipse:
> little formal detail: Please update the copyright year for SaveManager.

I believe copyright years aren't supposed to be updated anymore.

But this was fixed now, so @jukzi can you please approve / disapprove again?

akurtakov:
It is not required as per https://www.eclipse.org/lists/eclipse-dev/msg10183.html but committers can still do so if they want.

jukzi (approved):
LGTM

iloveeclipse:
Thanks everyone.

Issue #381, opened 2023-03-28
Copy project to external location logs an Error and creates aduplicate project in workspace.


raghucssit:
Steps to reproduce:
1. Create a `General Project` and apply any `Resource Filter` on the project.
2. Right click on the project and select `Copy`. Right click on the `Package Explorer` and `Paste`.
3. Now you will be prompted with `Copy Project` dialog, Uncheck the `Use default location` button. Browse any location outside the workspace and click `Copy` button.

Expected:
It has to copy the project to external location without errors.
Actual:
1. Logs an `Error` in `Error Log`.
2. Creates a duplicate project in the workspace.
Error stacktrace:
`java.lang.Exception: File not found: ext-loc/dest-proj/.project.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.read(FileSystemResourceManager.java:835)
	at org.eclipse.core.internal.resources.File.getContents(File.java:277)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.descriptionChanged(FileSystemResourceManager.java:382)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.internalWrite(FileSystemResourceManager.java:652)
	at org.eclipse.core.internal.resources.Project.writeDescription(Project.java:1421)
	at org.eclipse.core.internal.resources.Project.writeDescription(Project.java:1402)
	at org.eclipse.core.internal.resources.Project.internalCopy(Project.java:697)
	at org.eclipse.core.internal.resources.Project.copy(Project.java:242)
	at org.eclipse.ui.ide.undo.CopyProjectOperation.copyProject(CopyProjectOperation.java:215)
	at org.eclipse.ui.ide.undo.CopyProjectOperation.doExecute(CopyProjectOperation.java:124)
	at org.eclipse.ui.ide.undo.AbstractWorkspaceOperation.lambda$0(AbstractWorkspaceOperation.java:200)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2380)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2405)
	at org.eclipse.ui.ide.undo.AbstractWorkspaceOperation.execute(AbstractWorkspaceOperation.java:200)
[...]

PR #382, opened 2023-03-28
Fix : Copy project with resource filter to external location logs Error and create a duplicate project in workspaces (Issue #381)

raghucssit:
If we try to copy a project which has resource filter applied on it to an external location, It logs Error and creates a duplicate project within workspace. For project without resource filter it works as expected.

see https://github.com/eclipse-platform/eclipse.platform/issues/381


iloveeclipse:
@raghucssit : please 
- sign ECA if not yet signed
- make sure the mail you use in the commit matches the one which you've used to sign ECA.

raghucssit:
Will re-create PR from different branch.

PR #387, opened 2023-03-30
Get rid of WorkspaceSessionTest

akurtakov:
It gives nothing over using ResourceTest directly but complicates the class hierarchy and thus makes moving to newer than 3.x JUnit even less likely.


iloveeclipse:
> It gives nothing over using ResourceTest directly

Not sure what you want to achive finally, but session tests are supposed to run each in a dedicated freshly started Eclipse instance, and validate things where workspace lifecycle need to be involved. So they can't be just used as plain unit tests.

laeubi:
> from WorkspaceSessionTest which is just empty class

It seem to suppress the teardown step at least, but a rule sounds good for such kind of behavior control, or some kind of `@RunWith(....)`

akurtakov:
> > from WorkspaceSessionTest which is just empty class
> 
> It seem to suppress the teardown step at least,

Which makes even less sense as super.setUp could create resources but is suppressed to not clean it up!?!

> but a rule sounds good for such kind of behavior control, or some kind of `@RunWith(....)`

akurtakov:
I'm giving up on this one as this would happen to be more time than I can spend but I still believe these tests need some work to make them easier to grasp.

iloveeclipse:
> [dec1d51](https://github.com/eclipse-platform/eclipse.platform/commit/dec1d51c290cad841d67febd6a723474dd95e68c) points out an example of some further improvements to the tests needed so they are better contained before other improvements can even be considered.

This example is wrong, as far as I can see. The "improved" test doesn't test what the original one did: it does **NOT** validate workspace save **and** successful restore, because all assertions after save run in **same**  workspace instance!

Issue #375, opened 2023-03-27
WeakGroupedListenerList: super slow cleanup / Eclipse process does not terminate after closing all windows


jukzi:
Sometimes eclipse.exe does not terminate within reasonable amount of time for minutes. 
Example:

OpenJDK 64-Bit Server VM (17.0.6+10 mixed mode, sharing)
```
"Framework stop - Equinox Container: 8638e99d-ac0d-4962-a6f4-4a2f053ab08e" #6849 prio=6 os_prio=0 cpu=33802,49ms elapsed=34,95s allocated=11658M defined_classes=4 tid=0x00007ff3a0259600 nid=0x5f95 runnable  [0x00007ff3bdd3d000]
   java.lang.Thread.State: RUNNABLE
	at java.util.HashMap$HashIterator.nextNode(java.base@17.0.6/HashMap.java:1601)
	at java.util.HashMap$KeyIterator.next(java.base@17.0.6/HashMap.java:1620)
	at org.eclipse.e4.core.internal.contexts.WeakGroupedListenerList.cleanup(WeakGroupedListenerList.java:130)
	- locked <0x000000069dad6af8> (a org.eclipse.e4.core.internal.contexts.WeakGroupedListenerList)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.cleanup(EclipseContext.java:786)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.cleanup(EclipseContext.java:784)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:208)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.dispose(EclipseContext.java:164)
	at org.eclipse.e4.core.internal.contexts.osgi.EclipseContextOSGi.dispose(EclipseContextOSGi.java:102)
	at org.eclipse.e4.core.internal.contexts.osgi.EclipseContextOSGi.bundleChanged(EclipseContextOSGi.java:144)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.dispatchEvent(BundleContextImpl.java:949)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:234)
	at org.eclipse.osgi.framework.eventmgr.ListenerQueue.dispatchEventSynchronous(ListenerQueue.java:151)
[...]

Issue #374, opened 2023-03-26
org.eclipse.ant.tests.ui.BuildTests fails on mac


laeubi:
```
Cannot invoke "org.eclipse.jdt.launching.IVMInstall.getName()" because "vm" is null
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.launching.IVMInstall.getName()" because "vm" is null
	at org.eclipse.ant.tests.ui.testplugin.ProjectHelper.setVM(ProjectHelper.java:202)
	at org.eclipse.ant.tests.ui.testplugin.ProjectHelper.createLaunchConfigurationForSeparateVM(ProjectHelper.java:187)
	at org.eclipse.ant.tests.ui.testplugin.ProjectHelper.createLaunchConfigurationForBoth(ProjectHelper.java:163)
	at org.eclipse.ant.tests.ui.testplugin.AbstractAntUITest.assertProject(AbstractAntUITest.java:201)
	at org.eclipse.ant.tests.ui.testplugin.AbstractAntUITest.setUp(AbstractAntUITest.java:149)
	at junit.framework.TestCase.runBare(TestCase.java:140)
	at org.eclipse.ant.tests.ui.testplugin.AbstractAntUITest.runBare(AbstractAntUITest.java:133)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at org.eclipse.ant.tests.ui.AbstractAntUIBuildTest.access$0(AbstractAntUIBuildTest.java:1)
	at org.eclipse.ant.tests.ui.AbstractAntUIBuildTest.lambda$0(AbstractAntUIBuildTest.java:46)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

HeikoKlare:
Has been fixed with #390.

PR #388, opened 2023-03-30
EclipseContext: allow GC for undisposed contexts #375

jukzi:
* WeakGroupedListenerList: Use JDK's WeakHashMap (with a cleaner) to not leak WeakComputationReference

* InjectorImpl: Use WeakHashMap to avoid strong references to injected Contexts to avoid leaked EclipseContext in the key of InjectorImpl.injectedObjects

* Fix modifiers and their order

Fixes Eclipse-IDE process stuck in EclipseContext.cleanup during shutdown. Happened if EclipseContext could be garbage collected but is not explicitly disposed.

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/internal/contexts/WeakGroupedListenerList.java
+++ b/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/internal/contexts/WeakGroupedListenerList.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2015 IBM Corporation and others.
+ * Copyright (c) 2012, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


jukzi:
@laeubi i don't get the current CI build: where can i see which tests failed?

laeubi:
grafik

jukzi:
1. that link was not available when Windows+Mac already failed and Linux hung around
2. that 66 "fails" refer to the 66 WARNINGS?
image

jukzi:
@laeubi thanks for the explanation. It is very confusing, that in the results the term "fail" vs "error" vs "warn" is not used consequently:
image

jukzi:
@laeubi where can i see on which VM the junit errors occured? My interpretatioN: Since the test on linux passes  and windows has a random faill al those errors are only from macOS
https://github.com/eclipse-platform/eclipse.platform/issues/374. And the change can actually be merged?

laeubi:
If you rebase you should get a lot of less failures now, yeas there is a little discrepancy because of what Github annotations support (error/warning) verus the junit namings.

jukzi:
ok, only mac fails

PR #389, opened 2023-03-30
Remove ExampleWorkspaceTest

HeikoKlare:
The `ExampleWorkspaceTest` uses the `IProject` in a way in which it does not work (anymore) or makes assumption on the extended `ResourceTest` that do not hold anymore (`testProject` has to be created before test execution). The class is not executed in the Maven build and all the tests fail when executed. The `IProject`, `IFolder` and `IFile`, which are partly tested in this class, are covered by dedicated test classes. Thus this class does not do anything meaningful.

If I did not miss any reasons for keeping that test class, I propose to remove it.


HeikoKlare:
> Ideally, you create a commit with the proper commit message first and if you do it right, github will simply take over commit title and message to the PR title and message, so you can save your and others time.

@iloveeclipse Totally makes sense, thank you for the hint!
I have added the missing commit message to document the reasons for removal.

iloveeclipse (approved):
LGTM

PR #390, opened 2023-03-30
Use p2.inf to pull in depedency to org.eclipse.jdt.launching.macosx

laeubi:
jdt.launcher requires special treatment to find jvms on MacOS that is
handeled with an additional bundle and the ant test depend on the
launching framework.

This adds a p2.inf that pulls in this additional depenedency using a
p2.inf for the test framework resoloution.

### DIFF
--- a/ant/org.eclipse.ant.tests.ui/META-INF/p2.inf
+++ b/ant/org.eclipse.ant.tests.ui/META-INF/p2.inf
@@ -1,0 +1,3 @@
+requires.0.namespace = org.eclipse.equinox.p2.iu
+requires.0.name = org.eclipse.jdt.launching.macosx
+requires.0.filter = (osgi.os=macosx)


iloveeclipse (changes requested):
Please provide commit message **why** this change is needed and which problem it is supposed to solve.

laeubi:
> > I can only guess and test here ...
> 
> And the later code maintainer will also "guess and test" only to answer a simple question that could be resolved by providing a proper commit comment "why the hack this dependency was added?"

When I know if / how to resolve the issue I can describe why it was done that way, at the moment I first need to evaluate possible options to resolve the issue...

laeubi:
This now fixes 60 out of 64 of the failing test.
@iloveeclipse please review and then merge.

iloveeclipse (approved):

PR #391, opened 2023-03-30
Make ResourceTest restore workspace description #406

### LINKED ISSUE #408: Random failing org.eclipse.core.tests.internal.builders.BuilderTest.testCopyProject
The test `BuilderTest.testCopyProject` randomly fails.
E.g., https://github.com/eclipse-platform/eclipse.platform/pull/352/checks?check_run_id=12610055665
When executed locally and in isolation, the test case always fails on my system.

Test will be fixed when #391 is merged.

```
4.4 expected:<[SetInitializationData, StartupOnInitialize, Build0]> but was:<[SetInitializationData, StartupOnInitialize, Build0, Build0]>
java.lang.AssertionError: 4.4 expected:<[SetInitializationData, StartupOnInitialize, Build0]> but was:<[SetInitializationData, StartupOnInitialize, Build0, Build0]>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.core.tests.internal.builders.TestBuilder.assertLifecycleEvents(TestBuilder.java:98)
	at org.eclipse.core.tests.internal.builders.BuilderTest.testCopyProject(BuilderTest.java:751)
```

HeikoKlare:
Several realizations of a `ResourceTest` perform changes to the workspace description, e.g., in terms of changing the auto-build settings, without resetting them afterwards. And even if they do, the functionality is redundantly present at different places within the source code.
This induces dependencies between test executions, as the behavior of a subsequent test case execution depends on the modifications of the workspace description by previous ones.

The proposed changes do the following: 
- Store the workspace description before executing a `ResourceTest` and restore it during teardown
- Provide only one central method enabling and disabling autobuild and setting the build order and remove all redundant methods or in-place realizations
- Remove manual restorations of autobuild settings after test execution, as this is subsumed by `ResourceTest` performing `setup` and `teardown`
- Properly wait for background jobs completion in `BuilderTest` to avoid race conditions because of changed runtime behavior due to resetting workspace description after executing other test cases

I am not sure whether the missing restoration of the workspace settings after a test run leads to build failures right now (might be a cause for some of the random failures). But locally executing certain extracts of available tests, such as a single class or package, can currently lead to failures due to missing test isolation.

Contributes to #406.
Fixes #408.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AbstractBuilderTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AbstractBuilderTest.java
@@ -15,7 +15,6 @@
 package org.eclipse.core.tests.internal.builders;
 
 import java.util.Map;
-import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.tests.resources.ResourceTest;
@@ -24,9 +23,6 @@ import org.eclipse.core.tests.resources.ResourceTest;
  * This class does not define any tests, just convenience methods for other builder tests.
  */
 public abstract class AbstractBuilderTest extends ResourceTest {
-
-	private boolean autoBuilding;
-	private int simultaneousBuilds;
 
 	public AbstractBuilderTest(String name) {
 		super(name);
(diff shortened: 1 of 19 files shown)


jukzi:
@laubi can you please review the workflow changes?

laeubi:
> @Laubi can you please review the workflow changes?

I would assume they are only for testing purpose and should be reverted before merging this.

jukzi (commented):
the changes in *.java LGTM

jukzi:
> I would assume they are only for testing purpose and should be reverted before merging this.

@HeikoKlare please revert or explain the workflow changes.

HeikoKlare:
I rebased on the wrong branch, which pulled in the workflow change. Sorry for that!
I have removed the commit with the workflow changes. The other changes are the same after force push.

PR #392, opened 2023-03-31
Improve JobEventTest.testDeadlockRecovery, avoid random failure (#373)

### LINKED ISSUE #373: Random failing org.eclipse.core.tests.runtime.jobs.JobEventTest.testDeadlockRecovery

HeikoKlare:
The `testDeadlockRecovery` randomly failed because the deadlock recovery mechanism in `JobListeners.waitAndSendEvents` was still processing the recovery in terms of providing a log entry while the test already failed due to timeout.

Since there is no mechanism to listen to deadlock recovery being finished (and potentially should not be as it may easily lead to a deadlock again), this change increases the timeout (which is not a problem as a timeout is exceptional behavior anyway) and restructures the test to easy comprehension.

The reason for the test failing due to still processing deadlock recovery while running into the test timeout can be seen in the log of this run: https://github.com/eclipse-platform/eclipse.platform/pull/391/checks?check_run_id=12409259744

Fixes #373.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobEventTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobEventTest.java
@@ -397,71 +397,76 @@ public class JobEventTest {
 	@Test
 	public void testDeadlockRecovery() throws Exception {
 		Object deadlock = new Object();
-		CountDownLatch testDoneSignal1 = new CountDownLatch(1);
-		CountDownLatch testDoneSignal2 = new CountDownLatch(1);
+		CountDownLatch threadHavingLockSignal = new CountDownLatch(1);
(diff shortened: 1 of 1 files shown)


jukzi:
increasing timeout is a good way to resolve random fails on slow servers

HeikoKlare:
Failures in `IJobManagerTest` in Windows and MacOS build also occur in current CI build of `master` branch, e.g.: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4573426969/jobs/8073852961

jukzi (approved):
LGTM. but please next time reduce refactoring to a minimum to make review easier.

HeikoKlare:
Thanks, @jukzi, and sorry for the effort! I will reduce refactoring in future pull requests. Could you merge this one?

Issue #397, opened 2023-04-04
Random failing org.eclipse.core.tests.runtime.jobs.IJobManagerTest.* tests


HeikoKlare:
Different test methods in `IJobManagerTest` randomly fail since about a week in the matrix build for different operating systems, e.g.:
`testJobFamilyJoinWhenSuspended_1`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4573426969
`testJobFamilyJoinWhenSuspended_3`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4555499253
`testSleep`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4550620965
`testOrder`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4562563603

PR #398, opened 2023-04-04
Improve timing behavior of IJobManagerTest for slow servers #397

HeikoKlare:
On a slow server some of the `IJobManagerTest` test cases (`testSleep`, `testOrder`, `testJobFamilyJoinWhenSuspended_1`/`2`/`3`) may fail because the main thread is executed so slowly that the jobs are already finished while the main thread still assumes them running. 

Increasing the according values for the time until termination will reduce the chance for random failures of these test cases. However, this should be a temporary solution to be replaced by properly checking a condition as mentioned in https://github.com/eclipse-platform/eclipse.platform/pull/398#issuecomment-1496940439.

Contributes to #397 as a temporary improvement, but does not resolve it.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
@@ -1202,7 +1202,7 @@ public class IJobManagerTest extends AbstractJobManagerTest {
 		final int[] familyJobsCount = new int[] {-1};
 		final TestBarrier2 barrier = new TestBarrier2();
 		final Job waiting = new FamilyTestJob("waiting job", 1000000, 10, TestJobFamily.TYPE_ONE);
-		final Job running = new FamilyTestJob("running job", 2, 1, TestJobFamily.TYPE_ONE);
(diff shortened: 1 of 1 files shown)


jukzi:
by increasing the sleep time the tests will run slower on fast servers too :-(
would be better when the testjobs would wait on a condition instead of just burning time.

HeikoKlare:
True. Just thought it could be a quick initial fix since one of these tests fails in almost every build. But it should quite easily be possible to have a proper condition instead of assuming a specific timing behavior in these tests. I will have a look at that.

iloveeclipse:
> will run slower on fast servers too

Where have you seen fast servers? Surely not on Eclipse.org. :)

I think if the change doesn't add minutes to test time, it would be beneficial for platform, to stop wasting our time investigating random fails.

HeikoKlare:
That's also true. But checking a proper condition should really be easy and will deterministically solve the problem (even if servers become slower again ;-) ). There is still the chance that tests randomly fail (but maybe less often) with the "fix" in the current commit on this PR.

If you like, you can merge this one, so there will hopefully be less random fails for now. But I will also provide a "proper" fix in the next days that replaces the temporary improvement.

iloveeclipse:
> If you like, you can merge this one, so there will hopefully be less random fails for now. 

I would do that. Every "false positive" test fail is one too much.

> But I will also provide a "proper" fix in the next days that replaces the temporary improvement.

Perfect.

iloveeclipse (approved):

PR #404, opened 2023-04-06
Fix random failing JobTest.* tests #402

### LINKED ISSUE #402: Random failing org.eclipse.core.tests.runtime.jobs.JobTest.* tests
Several test methods in `JobTest` are randomly failing. In particular, the ones I found are:
* `testJoinInterruptNonUIThread`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4615917148/jobs/8160303205
  * This was already found and improved by increasing timeout before: https://github.com/eclipse-platform/eclipse.platform/issues/320
* `testAsynchJobConflict`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4572138024/jobs/8071083283
* `testJoinWithTimeout`: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4616213598/jobs/8160940708

HeikoKlare:
Several test cases in `JobTest` fail because of relying on non-deterministic timing behavior.
- Increase too low `TestBarrier2` timeout while waiting for status change to fix `testAsynchJobConflict`
- Add proper barrier in `testJoinWithTimeout`
- Make `testJoinInterruptNonUIThread` validate execution time in proper thread

Fixes #402.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
@@ -23,7 +23,6 @@ import static org.junit.Assert.assertNotEquals;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.Arrays;
 import java.util.concurrent.atomic.*;
 import org.eclipse.core.internal.jobs.*;
 import org.eclipse.core.runtime.*;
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Failing test is already documented.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
```diff
@@ -1002,18 +991,20 @@ public void testJoinWithCancelingMonitor() {
 	}
 
 	public void testJoinInterruptNonUIThread() throws InterruptedException {
+		TestBarrier2 barrier = new TestBarrier2();
 		final Job job = new TestJob("job", 50000, 100);
 		Thread t = new Thread(() -> {
 			job.schedule();
 			try {
+				barrier.setStatus(Job.RUNNING);
```
jukzi, line 999: (i have not looked at the whole commit)
this change does totally change the test and probably render it meaningless: scheduling a job does not mean the job is already started.

jukzi (changes requested):

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
HeikoKlare: The comment is correct w.r.t. naming, i.e., the barrier actually ensures the job is scheduled and not necessarily running. Thanks for pointing that out.
However, being scheduled in enough to keep the semantics of the test. The test is about interrupting a blocked join operation, which is still given when the job is scheduled and not necessarily running. Thus the tests remains meaningful and in particular validates the same functionality as before, but now with deterministic behavior.

I will update the PR with proper naming for the barrier.

HeikoKlare: I have updated the commit with proper naming for the barrier status and the according comment

jukzi (approved):

PR #410, opened 2023-04-12
Measure AntUtil performance via thread instead of system time #403

### LINKED ISSUE #403: Random failing org.eclipse.ant.tests.ui.AntUtilTests.testGetIncludeTargetsPerformance on MacOS
`AntUtilTests.testGetIncludeTargetsPerformance` takes more than the expected maximum of 10 seconds to finish in some of the recent MacOS builds. As far as I can see there were no recent changes in the productive code, so failures may be due to too slow build infrastructure?

```
java.lang.AssertionError: Expecting a duration < 10000, but we have 12125ms
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.ant.tests.ui.AntUtilTests.testGetIncludeTargetsPerformance(AntUtilTests.java:221)
```

See:
https://github.com/eclipse-platform/eclipse.platform/actions/runs/4621006483/jobs/8171898101
https://github.com/eclipse-platform/eclipse.platform/actions/runs/4620696572/jobs/8171161377
https://github.com/eclipse-platform/eclipse.platform/actions/runs/4616360586/jobs/8161276288

HeikoKlare:
`AntUtilTests.testGetIncludeTargetsPerformance` measures the performance of AntUtil using the system time. With multiple threads potentially being executed, in particular in terms of a parallel Maven build, the thread executing the performance test may not be dispatched for longer times leading to a failure, even though the execution time of the test itself has been low.
With this change, the thread CPU time of the performance test is used rather than the system time to achieve more accurate and reproducible results.

Fixes #403.

### DIFF
--- a/ant/org.eclipse.ant.tests.ui/Ant Tests/org/eclipse/ant/tests/ui/AntUtilTests.java
+++ b/ant/org.eclipse.ant.tests.ui/Ant Tests/org/eclipse/ant/tests/ui/AntUtilTests.java
@@ -15,6 +15,7 @@
 package org.eclipse.ant.tests.ui;
 
 import java.io.File;
+import java.lang.management.ManagementFactory;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -200,11 +201,11 @@ public class AntUtilTests extends AbstractAntUITest {
 		 * is not set), but every include-file now includes the "big build.xml"
 		 */
 		String buildFileName = "bug412809/performance/buildFileHierarchical"; //$NON-NLS-1$
-		long startTime = System.currentTimeMillis();
+		long startTimeInNanoseconds = ManagementFactory.getThreadMXBean().getCurrentThreadCpuTime();
 		File buildFile = getBuildFile(buildFileName + ".xml"); //$NON-NLS-1$
 		IAntModel model = AntUtil.getAntModel(buildFile.getAbsolutePath(), false, true, true);
 		AntProjectNode project = model.getProjectNode();
-		long endTime = System.currentTimeMillis();
+		long endTimeInNanoseconds = ManagementFactory.getThreadMXBean().getCurrentThreadCpuTime();
 
 		Assert.assertNotNull(project);
 		/*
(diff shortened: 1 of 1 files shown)


jukzi:
Good idea to measure resource time instead of clock time. However measuring the time of a single thread could be as well wrong when multiple threads are involved. Also other Resources then CPU like filesystem or network could be involved in performancetests. The testframework to use parallel build for performancetests sounds broken.

HeikoKlare:
That's true, both that there may be multiple threads involved as well as that this kind of performance test does not work well with parallel builds. Some further thoughts on that:
1. **Test environment**: Performance tests using clock time always depend on a specific environment, not only in terms of the execution times you can expect, but also in terms of having the same conditions regarding the infrastructure on every run. Even if the Maven build is not run in parallel, the test infrastructure has to ensure that no other processes are running (or with the same timing behavior on every build) to get reproducible results. Thus, performance tests actually need to be separated from functional tests and run on a dedicated infrastructure for which these conditions hold.
2. **Measurement subject**: In general, filesystem and network access may have an impact on the overall execution time of an operation but not on the resource time of the executing thread because it does not consume resources while waiting for IO. In this specific test case the execution time is dominated by parsing the file already read from the file system and no (at least no significant) amount of work is outsourced to other threads, according to my measurements.

Since there is no simple and fast solution for (1) but the test is still valuable, I propose to elimiate as much effects on measured execution time as possible to get reasonable results. With (2) I would argue for the proposed solution to be valuable (at least more valuable than measuring time using system clock), but I see that restricting measurement to the thread executing the test  case may be too inaccurate.
But what about simply measuring the resource times of all threads in the same VM? That would at least eliminate effects of all other processes running in the test infrastructure (in particular also test cases executed in parallel by the same Maven build)?

jukzi:
> But what about simply measuring the resource times of all threads in the same VM?

Don't know how that could be done. We could keep it simple and just increase timeout.

jukzi:
let's try with your change. i tested, that on my computer both ManagementFactory.getThreadMXBean().getCurrentThreadCpuTime() and System.nanotime() report ~3600ms for that test.

PR #412, opened 2023-04-17
Fix random failing JobGroupTest.testJoinWithTimeout #30

### LINKED ISSUE #30: random failing org.eclipse.core.tests.runtime.jobs.JobGroupTest.testJoinWithTimeout
https://download.eclipse.org/eclipse/downloads/drops4/I20220522-0600/testresults/html/org.eclipse.core.tests.runtime_ep424I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

```
junit.framework.AssertionFailedError: 2.0
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.TestCase.assertTrue(TestCase.java:192)
at org.eclipse.core.tests.runtime.jobs.JobGroupTest.testJoinWithTimeout(JobGroupTest.java:540)
```

HeikoKlare:
The test assumed a very specific timing behavior between main thread and another thread calling `JobGroup.join`. This change replaces it with a proper barrier and the expectation of proper timing behavior only with respect to the timeout for the join operation in the thread executing it.
The test is similar to `Job.testJoinWithTimeout`, which has been fixed in the same way (see #404).

Fixes #30.


runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -515,31 +516,30 @@ public void testJoinWithTimeout() {
...
+
+		barrier.waitForStatus(TestBarrier2.STATUS_DONE);
+		// Verify that thread is done
+		assertNotEquals("Waiting for join to time out failed", -1, duration[0]);
+		assertEquals("Group with long running jobs should still be active after join timed out", JobGroup.ACTIVE,
+				firstJobGroup.getState());
+		// Verify that join call is blocked for at least the duration of given timeout
+		String durationAndTimoutMessage = "duration: " + duration[0] + " timeout: " + timeout;
+		assertTrue("Join did not run into timeout; " + durationAndTimoutMessage, duration[0] >= timeout);
+		assertTrue("Join took significantly longer than timeout; " + durationAndTimoutMessage,
```
jukzi, line 541: when the build hangs for a random second due to other things this can randomly fail. better avoid this check

jukzi (changes requested):

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
HeikoKlare: Yes, thanks! In case of a regression regarding timeout functionality, the test will fail anyway as the barrier waiting for a status change times out (but then in a reasonable amount of time rather than the 400ms stated here).

I have removed the check. We should then remove the analogous check in `JobTest` as well, as it may randomly fail for the same reasons. I'll provide a follow-up PR for that.

HeikoKlare: @jukzi are you okay with the removal of the check?

jukzi (approved):

PR #414, opened 2023-04-19
Fix random failing JobTest.testJoinInterruptNonUIThread #320

HeikoKlare:
The test can randomly fail for two reasons:
1) `cancel` may be called on a job that has been scheduled but was not running yet. Then the result status is null instead of CANCEL. 
2) `interrupt` may be called before `join` is called, then the interrupt is not processed

The changes ensure that `interrupt` is called after `join` blocked and that the job is terminated after the interrupt occurred, no matter in which state it was before.

To reproduce the two issues:
1) add a delay to the `schedule` call
2) add a `sleep` between setting the barrier value and calling `join`


laeubi:
> cancel may be called on a job that has been scheduled but was not running yet. Then the result status is null instead of CANCEL.

@jukzi as you recently overhauled the status port, do you think it should actually have CANCEL set in such a case?

jukzi:
> @jukzi as you recently overhauled the status port, do you think it should actually have CANCEL set in such a case?

not necessarily. If the existing API uses CANCEL only for jobs that already started i would not touch that without a strong reason.

HeikoKlare:
I agree with @jukzi. From what the API states, I think it is correct that a job that is cancelled before running has no result. The documentation of `getResult` says:
```
`null` if this job has never finished running.
```
And indeed it has not finished in that case. Even when calling `cancel` the resultof the job may not be `Cancel`, because it is up to the job to handle a cancel event or not: 
```
If the job is currently executing, it will be asked to stop but there is no guarantee that it will do so.
```

jukzi (approved):

iloveeclipse:
Please link to issues next time: https://github.com/eclipse-platform/eclipse.platform/issues/402

HeikoKlare:
Thanks for linking. I forgot to add the link when marking the PR ready to review.
I also added the link to #320, which is the specific issue that this PR is supposed to fix (as indicated in the commit message / PR title)

Issue #416, opened 2023-04-19
LocalFileNativesManager initialisation is not flexible enough to avoid DLL Hijacking AND avoid performance degradation


feilimb:
When the LocalFileNativesManager class is classloaded, initialisation code is invoked, which by default attempts to load a library named "_unixfile_1_0_0.dll_".  

This was previously documented as a security vulnerability due to the potential for a DLL Hijacking type attack, and a workaround / advisory was published here: https://www.ibm.com/support/pages/default-eclipse-settings-can-cause-potential-vulnerability 
.. and with a protected Bugzilla discussion here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=558017

The issue with the workaround detailed in the above advisory is that whilst usage of the system property "_eclipse.filesystem.useNatives=false_" does indeed prevent the offending code (_UnixFileNatives.isUsingNatives()_) from being invoked, it also **_prevents_** (on a Windows system) native implementations from being benefitted from / used by the various File IO operations within the _LocalFileNativesManager_ class API itself.

This performance degradation is observed in the tool I work on, when the system property above is set, and renders the performance to be twice as slow on some operations observed.  Also not that the class description itself within _LocalFileNativesManager_ states:
```
 * Please notice that the native implementation is significantly faster than the non-native one.
 * The BenchFileStore test runs 3.1 times faster on Linux with the native code than without it.
```

I would suggest that further flexibility should be offered to avoid the invokation of _UnixFileNatives.isUsingNatives()_ but in a way that *does* still allow (on Windows) for the native implementations to be used for the File IO operations within _LocalFileNativesManager_. As far as I can see the workaround detailed in the above advisory is piggy-backing on a pre-existing flag which was incorporated relating to the testing of the code on Unix platforms.

If a new additional system property(ies) were added, it should be possible to both (a) avoid _UnixFileNatives.isUsingNatives()_ from being invoked on a Windows system, _and_ (b) allow native implementations to continue to be used for the File IO operations.

I can create a proposal for a fix if that is useful.

jukzi:
This issue sound almost like a duplicate of https://github.com/eclipse-platform/eclipse.platform/issues/168 ?
Please contribute a fix.
As far as i understand a clean solution would be to load the library(s) from an absolute location.

jukzi:
see also
[...]

PR #418, opened 2023-04-20
LocalFileNativesManager initialisation is not flexible enough #416

feilimb:
LocalFileNativesManager initialisation is not flexible enough to avoid DLL Hijacking #416

If the use of native libraries is enabled (ie. the default), the default priority order for the loading of associated JNI library is Unix followed by Windows. With this proposed change, this order can be reversed to give priority to attempt to load Windows library first by use of the system property _-Declipse.filesystem.favourWindowsNative=true_

This avoids the potential for DLL Hijack type attack on the (non-existent) file named "_unixfile_1_0_0.dll_" on Windows systems.

See Issue #416 for further detail.

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
@@ -21,17 +21,18 @@ import org.eclipse.core.filesystem.provider.FileInfo;
 import org.eclipse.core.internal.filesystem.local.nio.*;
 import org.eclipse.core.internal.filesystem.local.unix.UnixFileHandler;
 import org.eclipse.core.internal.filesystem.local.unix.UnixFileNatives;
+import org.eclipse.osgi.service.environment.Constants;
 
 /**
- * Dispatches methods backed by native code to the appropriate platform specific
+ * <p>Dispatches methods backed by native code to the appropriate platform specific
  * implementation depending on a library provided by a fragment. Failing this it tries
- * to use Java 7 NIO/2 API's.
- * <p>
- * Use of native libraries can be disabled by adding -Declipse.filesystem.useNatives=false to VM
- * arguments.
- * <p>
- * Please notice that the native implementation is significantly faster than the non-native one.
- * The BenchFileStore test runs 3.1 times faster on Linux with the native code than without it.
+ * to use Java 7 NIO/2 API's.</p>
+ * 
+ * <p>Use of native libraries can be disabled by adding -Declipse.filesystem.useNatives=false 
+ * to VM arguments.<p>
+ * 
+ * <p>Please notice that the native implementation is significantly faster than the non-native one.
+ * The BenchFileStore test runs 3.1 times faster on Linux with the native code than without it.</p>
  */
 public class LocalFileNativesManager {
 	public static final boolean PROPERTY_USE_NATIVE_DEFAULT = true;
@@ -55,10 +56,12 @@ public class LocalFileNativesManager {
 	 */
 	public static boolean setUsingNative(boolean useNatives) {
 		boolean nativesAreUsed;
-		if (useNatives && UnixFileNatives.isUsingNatives()) {
+		boolean isWindowsOS = Constants.OS_WIN32.equals(LocalFileSystem.getOS());
+
+		if (useNatives && !isWindowsOS && UnixFileNatives.isUsingNatives()) {
 			HANDLER = new UnixFileHandler();
 			nativesAreUsed = true;
-		} else if (useNatives && LocalFileNatives.isUsingNatives()) {
+		} else if (useNatives && isWindowsOS && LocalFileNatives.isUsingNatives()) {
 			HANDLER = new LocalFileHandler();
 			nativesAreUsed = true;
 		} else {


jukzi:
i understand this PR tries only to change the order unix/windows. I don't know why there even is an order: on windows it could try the windows dll and on unix like it could try the unix only. java knows the OS.
Also this fix does not fix the issue on windows :-(

feilimb:
There is a difference because there is different handler logic depending on the OS, it is not just the library which is loaded which differs, see:
image

This fix allows configuration on Windows based tools to avoid the unnecessary attempt to load "_unixfile_1_0_0.dll_" _which will never be found_.

On Windows the ultimate loading of the "_localfile_1_0_0.dll_" library is found in the configuration area (of which the location itself is configurable via _-Dosgi.configuration.area_) - and this occurs _without_ attempts to load from all locations in the user's path first, and other locations in Windows DLL search order.

This fix relates to allowing an avoidance of the unnecessary attempt to load "_unixfile_1_0_0.dll_" on Windows OS, which as per Windows behaviour attempts to load it from _all_ of the following locations:

- the directory the application is being loaded from
- the current directory
- the system directory
- the windows directory
- the directories located in the PATH environment

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
```diff
@@ -46,16 +52,31 @@ public class LocalFileNativesManager {
...
+	 * loading of a native Windows library <i>ahead of</i> a Unix native library
+	 * @return <code>true</code> if natives are used as result of this call <code>false</code> otherwise
 	 */
-	public static boolean setUsingNative(boolean useNatives) {
+	public static boolean setUsingNative(boolean useNatives, boolean favourWindowsNative) {
 		boolean nativesAreUsed;
-		if (useNatives && UnixFileNatives.isUsingNatives()) {
+
+		if (useNatives && favourWindowsNative && LocalFileNatives.isUsingNatives()) {
+			// avoids an invocation of System.loadLibrary("unixfile_1_0_0") - thereby
```
jukzi, line 74: if useNatives && favourWindowsNative but not LocalFileNatives.isUsingNatives()
will it load System.loadLibrary("unixfile_1_0_0")?

jukzi: or (unintended) localfile_1_0_0?

feilimb: It will fall-back on the existing implementation in that case, and attempt to load _unixfile_1_0_0.dll_ via the ELSE IF block starting on line 79.

jukzi: doesnt LocalFileNatives.isUsingNatives() alread load localfile_1_0_0?

feilimb: Yes it does. 

If you are getting at the Unix OS workflow - in the case of a Unix based tool configuration (assuming a client does _not_ set the new property _-Declipse.filesystem.favourWindowsNative=true_) there will be no attempt to load the _localfile_1_0_0.dll_ (on Unix based OS) via the new condition on line 73.

For Windows OS workflow if the _-Declipse.filesystem.favourWindowsNative=true_ property is NOT set, the existing order will occur with an attempted load of _unixfile_1_0_0.dll_, followed by a load for _localfile_1_0_0.dll_.  If the _-Declipse.filesystem.favourWindowsNative=true_ property IS set, then there will be _no_ attempted load of _unixfile_1_0_0.dll_, and the first library load will be _localfile_1_0_0.dll_.

jukzi:
I think it's bad to fix only a part of the issue and only with a parameter. if it is a relevant security issue it should work without further parameter  in all scenarios: as far as i understand there are 3 scenarios:

- windows
- unix
- no natives

feilimb:
OK, I had hoped to avoid OS resolution logic as part of fix, but I guess that is the direction of proposed change as per above comment.  I will rewrite an alternative approach and update this PR with a modified approach.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
```diff
@@ -46,19 +47,28 @@ public class LocalFileNativesManager {
 	 * reset the usage of native to the system default
 	 */
 	public static void reset() {
-		setUsingNative(Boolean.parseBoolean(System.getProperty(PROPERTY_USE_NATIVES, String.valueOf(PROPERTY_USE_NATIVE_DEFAULT))));
+		boolean useNatives = Boolean.parseBoolean(System.getProperty(PROPERTY_USE_NATIVES, String.valueOf(PROPERTY_USE_NATIVE_DEFAULT)));
```
jukzi, line 50: please don't touch if it's not necessary

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
```diff
@@ -87,4 +97,4 @@ public static boolean putFileInfo(String fileName, IFileInfo info, int options)
 		return HANDLER.putFileInfo(fileName, info, options);
 	}
 
-}
+}
```
jukzi, line 100: avoid unrelated newlinechange

jukzi:
please squash changes to a single commit.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
```diff
@@ -53,33 +48,27 @@ public class LocalFileNativesManager {
...
-		if (useNatives && favourWindowsNative && LocalFileNatives.isUsingNatives()) {
-			// avoids an invocation of System.loadLibrary("unixfile_1_0_0") - thereby 
-			// avoiding potential DLL hijacking type attack (see: 
-			// https://github.com/eclipse-platform/eclipse.platform/issues/416 for further detail)
-			HANDLER = new LocalFileHandler();
-			nativesAreUsed = true;
-		} else if (useNatives && UnixFileNatives.isUsingNatives()) {
+		final String OS_NAME_WINDOWS_PREFIX = "Windows"; //$NON-NLS-1$
+		final String OS_NAME = System.getProperty("os.name"); //$NON-NLS-1$
+		boolean isWindowsOS = OS_NAME != null && OS_NAME.startsWith(OS_NAME_WINDOWS_PREFIX);
```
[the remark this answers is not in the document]
jukzi: or as in org.eclipse.core.internal.filesystem.local.Convert.isWindows

isWindows = Constants.OS_WIN32.equals(LocalFileSystem.getOS());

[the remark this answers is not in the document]
feilimb: Thanks for the pointer, will use the _org.eclipse.core.internal.filesystem.local.Convert.isWindows_ example.

jukzi:
> Can this be done at a point when Pull Request is finally approved and can be merged to master?

this is a requirement before we approve. merging will happen normally together with approve, or we even skip approve normally.

jukzi:
and you need to sign the legal agreements.

jukzi:
> I have signed the eca so I can't figure out why that pipeline is failing :(

Anonymous (f.breatnach@****.ie)
Could not validate the users ECA sign-off status. Please ensure that the e-mail address of all contributing accounts match a valid Eclipse Foundation account with a valid ECA.

Did you use the signed email address in the commit?

jukzi:
> Short of that the only other thing I can think of is closing this PR, and creating a brand new one with initial commit being from my @gmail address...

try that. retrigger did not help

feilimb:
Closing this PR as the ECA pipeline job is not picking up my corrected email address.  A new PR will be created which has a changeset / diff the same as this one.  I will point to this PR in the new one for reference to all discussion to date.

PR #421, opened 2023-04-24
Fix timing behavior in JobEventTest.testDeadlockRecovery #373

### LINKED ISSUE #373: Random failing org.eclipse.core.tests.runtime.jobs.JobEventTest.testDeadlockRecovery

HeikoKlare:
The test case `org.eclipse.core.tests.runtime.jobs.JobEventTest.testDeadlockRecovery` can randomly fail as it relies on specific operations terminating within specific amounts of time, such as the abortion of a schedule operation due to a deadlock providing stack traces fast enough. Whether this succeeds depends on the performance of the executing infrastructure. 

The change simplifies the test case and assigns proper timeout values: All synchronizing calls between the different threads/jobs now rely on a barrier whose timeout is implemented in a generic way, so that it may only fail if there is really no further progress. The timeout for job listeners (indicating a deadlock) is kept and a significantly higher timeout for the `join` operation is added, as it is only relevant in case of a regression and can thus have a high value to avoid the potential of random failures on slow hardware.

Fixes #373.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobEventTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobEventTest.java
@@ -15,9 +15,6 @@ package org.eclipse.core.tests.runtime.jobs;
 
 import static org.junit.Assert.assertEquals;
 
-import java.util.*;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


HeikoKlare:
Failing tests are unrelated to the test fixed in this PR.

HeikoKlare:
Converted back to draft, since test failed again in last verification build: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4817980287/jobs/8579374539?pr=421

Issue #417, opened 2023-04-20
Random failing IResource.testDelete


HeikoKlare:
The test `org.eclipse.core.tests.resources.IResourceTest.testDelete` fails randomly.
See https://github.com/eclipse-platform/eclipse.platform/actions/runs/4742853133/jobs/8421637788?pr=415#step:7:17743
```
java.lang.AssertionError: invocation 73 should fail, but it doesn't [false,null,F/openProject/1/2/3]
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:81)
	at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:110)
	at org.eclipse.core.tests.resources.TestPerformer.performTestRecursiveLoop(TestPerformer.java:110)
	at org.eclipse.core.tests.resources.TestPerformer.performTest(TestPerformer.java:56)
	at org.eclipse.core.tests.resources.IResourceTest.testDelete(IResourceTest.java:1132)
```

Failure occurred on MacOS.

HeikoKlare:
There is another recent failure of a test using the `TestPerformer`: `org.eclipse.core.tests.resources.LinkedResourceTest.testCopyFolder`.
Maybe there is a relation by using that performer class.

See https://github.com/eclipse-platform/eclipse.platform/actions/runs/4722513909/jobs/8377196657#step:7:18606
```
Caused by: org.eclipse.core.internal.resources.ResourceException: Resource '/ExistingProject/existingFolderInExistingProject/nonExistingFolderInExistingFolder' already exists.
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:309)
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:296)
	at org.eclipse.core.internal.resources.Resource.checkCopyRequirements(Resource.java:248)
	at org.eclipse.core.internal.resources.Resource.assertCopyRequirements(Resource.java:149)
	at org.eclipse.core.internal.resources.Resource.copy(Resource.java:538)
	at org.eclipse.core.tests.resources.LinkedResourceTest$2.invokeMethod(LinkedResourceTest.java:474)
```
Failure occurred on Windows. So if there is a relation, the problems are independent from OS / file system.

fedejeanne:
[...]

PR #424, opened 2023-04-25
Add missing context restore between IResource test runs #417

HeikoKlare:
The `IResource` tests, which use the TestPerformer to execute parameterized delete, copy and move operations, were not implemented properly. The workspace was set up before executing the first parameterization and was then never reset. Thus, all subsequent test iterations worked on the workspace left by the previous iteration. In addition, the out-of-sync resources were synchronized by workspace updates performed at different places in the test code. This includes the background refresh job, which updated the out-of-sync resources at some point in time. In case this happened between the `shouldFail` evaluation of a test iteration and the actual `invokeMethod` execution, the test could fail randomly.

The changes eliminate these shortcomings:
* Whenever one of those tests modifies the workspace, it is restored for the next iteration
* An unexpected background refresh is avoided by explicitly triggering and joining the refresh before each test iteration
* The out-of-sync resources are restored between every two test iterations
* The validations have been adapted, as they have fit to the unexpected workspace states influenced by previous test iterations instead of the proper ones; this particularly includes the proper validation for out-of-sync resources


HeikoKlare:
The touched test cases were erroneous (and prone to random failure) before (see https://github.com/eclipse-platform/eclipse.platform/issues/417#issuecomment-1521772258). Since the situation may only become better and all recent builds of this PR were successful, I'll merge it.

PR #430, opened 2023-04-29
Clean-up OSGi package imports

HannesWell:
- Use properly closed version range
- Remove unused packages
- Remove unnecessary imports of org.osgi.service.component.annotations (they are only necessary at compile-time and Tycho adds them for compilation on the fly).
- Mark optional package-imports that are definitely not optional as such


ruspl-afed:
> Use properly closed version range

What is the benefit of upper bound? It will require effort after update to 2.0.0+ even if used types will remain compatible.

HannesWell:
> > Use properly closed version range
> 
> What is the benefit of upper bound? It will require effort after update to 2.0.0+ even if used types will remain compatible.

As far as I can tell there is no plan for OSGi API version 2 in the foreseeable future. And even if that happens one day I assume that then the types will not remain compatible. Since OSGi is always evangelizes semantic versioning, I expect it will follow that rules and only have a major version bump if there are real breakages.

HannesWell:
So basically this changes just applies OSGi best practices to guard against runtime errors, if one day there is an OSGi API version 2.
Therefore I think this is beneficial. 
Do you still have concerns?

ruspl-afed:
This question was raised in order to understand the policy behind this change and similar changes.
Since similar changes were already merged to other Platform repositories, I'm fine to have this repo aligned with others.

HannesWell:
> This question was raised in order to understand the policy behind this change and similar changes. Since similar changes were already merged to other Platform repositories, I'm fine to have this repo aligned with others.

OK, understand. I know there are some libraries that don't follow the semantic versioning, so one easily gets used to the situation that there is a major version bump without actual breaking changes (e.g. guava to name a big example for at least not broken packages). But as said, I expect OSGi to follow the best practices it evangelizes and to not punish those that are following them.

Bananeweizen:
I find it slightly suspicious to see component.annotations(optional) being removed multiple times. Such optional dependencies have often been used for compile time dependencies in the past. Would they still be useful eventually? (I honestly don't know myself, but this looks very much like jdt.annotations which is also an optional dependency in many projects, to support the null analysis in the compiler)

HannesWell:
> I find it slightly suspicious to see component.annotations(optional) being removed multiple times. Such optional dependencies have often been used for compile time dependencies in the past. Would they still be useful eventually? (I honestly don't know myself, but this looks very much like jdt.annotations which is also an optional dependency in many projects, to support the null analysis in the compiler)

The Annotations in `org.osgi.service.component.annotations` are only processed at build-time for OSGi Declarative Service Components to generate their description XML files:
https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-annotations

At runtime the Service Component Runtime implementation (for example apache Felix SCR) only reads the generated description XML for the files listed as value of the `Service-Component` MANIFEST.MF header. The annotations are not of interest at runtime. In fact they cannot be read at runtime, even if the annotation bundle were present because they have a CLASS retention policy.

The pattern to import the annotation packages as optional was only used to have them available at build-time but to not enforce their presence at runtime. But for some time PDE and Tycho add the service.component.annotation bundle to the compile-classpath in the IDE respectively build, when necessary so that this pattern is not necessary anymore and it is save to remove those imports.

laeubi:
> The annotations are not of interest at runtime

The annotation have `retention=CLASS` so they are effectively not there at runtime at all...
https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/RetentionPolicy.html#CLASS

So even if one wants to read them at runtime it won't be possible.

PR #435, opened 2023-05-03
Improvements/assume in tests

fedejeanne:
Use `assume` to skip tests. Extract methods to check the current platform in tests.


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
```diff
@@ -37,7 +38,7 @@ public class SymlinkTest extends FileSystemTest {
 	 */
 	private static final boolean SYMLINKS_ARE_FIRST_CLASS_FILES_OR_DIRECTORIES = Platform.OS_WIN32
 			.equals(Platform.getOS()) ? true : false;
-	private static String specialCharName = "���� ��� ���� �����"; //$NON-NLS-1$
+	private static String specialCharName = "���� ��� ���� �����"; //$NON-NLS-1$
```
Bananeweizen, line 41: broken encoding

Bananeweizen (changes requested):

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
fedejeanne: You're right, it seems buggy in GitHub. I've checked my local history anyway and it seems I unintentionally changed the code. I just reverted it back to what it was before so the special characters (`äöüß ÄÖÜ àÀâÂ µ²³úá`) should be back in place. 

Thank you!

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/CreateDirectoryTest.java
```diff
@@ -146,13 +147,10 @@ public void testTargetIsFileInLocalFile() {
 	}
 
 	public void testParentDeviceNotExistsInLocalFile() {
-		if (!Platform.getOS().equals(Platform.OS_WIN32)) {
-			return;
-		}
+		Assume.assumeTrue(Platform.getOS().equals(Platform.OS_WIN32));
```
jukzi, line 150: That's a NO-GO as it would permanently add failing test on other platforms instead of "skipping" the testcode.

jukzi (changes requested):

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/CreateDirectoryTest.java
fedejeanne: The tests fail because Tycho can't deal with `Assume` the way it should and it fails the tests instead of skipping them. 

I only noticed it after doing this PR (my bad) and I just found out about this bug (https://github.com/eclipse-tycho/tycho/issues/630).

I'm going to have to drop this PR since skipping tests using `Assume` is not doable at the moment :(

fedejeanne:
Using `Assume` is not a viable solution, there's a Bug in Tycho: https://github.com/eclipse-tycho/tycho/issues/630

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/CreateDirectoryTest.java
jukzi: did you try in the @before?

fedejeanne: hm, interesting, `@Before` will be ignored in this test class, but if I break free from JUnit3 by *not* (indirectly) extending from  `junit.framework.TestCase` then 2 things happen:

1. `@Before` will be called instead of `setup()`
2. `Assume` will skip tests instead of failing them, even when used inside test methods... which is exactly what I wanted in the first place!

I see there are several test classes that indirectly extend from `junit.framework.TestCase`, breaking this chain and moving to JUnit4 is not a trivial task so I will abort this PR and evaluate a better option (maybe migrating some of the test classes in the hierarchy one by one).

Thank you anyway for your comment, at least now I know the problem is solvable by "upgrading" to JUnit4 👍

jukzi: cheers.

vogella:
Please squasch commits or provide individual PR for each commits. Our PR should not contain multiple commits. See https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md

fedejeanne:
@vogella thank you for the hint, I superseded this PR with the new PR #446 and squashed all my commits into 1 :)

Issue #440, opened 2023-05-08
Fix failing tests in FileStoreTest


fedejeanne:
These 3 tests fail:
* `org.eclipse.core.tests.filesystem.FileStoreTest.testCopyAcrossVolumes()`
* `org.eclipse.core.tests.filesystem.FileStoreTest.testCopyFileAcrossVolumes()`
* `org.eclipse.core.tests.filesystem.FileStoreTest.testMoveAcrossVolumes()`

The reason they fail is because they attempt to delete temporary folders which are *not* empty (like `c:\temp` or `d:\temp`) in the `tearDown` method. Since these folders are not empty and they contain _read-only_ files, the tests throw this exception:

```
junit.framework.AssertionFailedError: IResourceTest#clear.99: org.eclipse.core.runtime.CoreException: Problems encountered while deleting files.
	at org.eclipse.core.tests.harness.CoreTest.fail(CoreTest.java:74)
	at org.eclipse.core.tests.resources.ResourceTest.clear(ResourceTest.java:429)
...
```

Looking deeper in the console, something like this shows up:
```
Contains: Could not delete: d:\temp\xyz.
java.nio.file.AccessDeniedException: d:\temp\xyz
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:89)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
...
```

### Preconditions
* There must be (at least) 2 "temp" folders under the root folder of a drive, _e.g._ `c:\temp`, `d:\temp`, ..., `z:\temp`, otherwise the test will be skipped.
* At least one of these folders must contain read-only files

PR #441, opened 2023-05-08
Bugs/file store test

fedejeanne:
Issue: #440 

* Add timestamps to the newly created folders inside already existing `temp` folders in `FileStoreTest`.
* Delete the newly created folders on tear down

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
@@ -22,7 +22,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.stream.Stream;
-
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.IFileStore;
@@ -89,8 +88,10 @@ public class FileStoreTest extends LocalStoreTest {
 
 	private IFileStore getDirFileStore(String path) throws CoreException {
 		IFileStore store = EFS.getFileSystem(EFS.SCHEME_FILE).getStore(new Path(path));
-		store.mkdir(EFS.NONE, null);
-		deleteOnTearDown(store);
+		if (!store.toLocalFile(EFS.NONE, getMonitor()).exists()) {
+			store.mkdir(EFS.NONE, null);
+			deleteOnTearDown(store);
+		}
 		return store;
 	}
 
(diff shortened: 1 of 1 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
```diff
@@ -124,35 +113,41 @@ public void testCopyAcrossVolumes() throws Throwable {
 		IFileStore[] tempDirectories = getFileStoresOnTwoVolumes();
 
 		/* test if we are in the adequate environment */
-		if (tempDirectories == null || tempDirectories.length < 2 || tempDirectories[0] == null
-				|| tempDirectories[1] == null) {
-			return;
-		}
+		Assume.assumeFalse("Can't test without at least 2 temp folders in different volumes", //
```
jukzi, line 116: i don't want a failed test just because i don't have two volumes. just skipping test under this conditions was better

jukzi (changes requested):

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
fedejeanne: I reverted my changes, the test will now be skipped.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
```diff
@@ -90,7 +79,6 @@ public void testBrokenFetchInfo() {
 	private IFileStore getDirFileStore(String path) throws CoreException {
 		IFileStore store = EFS.getFileSystem(EFS.SCHEME_FILE).getStore(new Path(path));
 		store.mkdir(EFS.NONE, null);
-		deleteOnTearDown(store);
```
jukzi, line 93: why is that folder not removed anymore?

jukzi: but for example testCopyFileAcrossVolumes() does only deleteOnTearDown if and only if the test is not skipped!

HeikoKlare: I agree. And even if the test is not skipped, the folder may still be created by the test and then has to be removed. See https://github.com/eclipse-platform/eclipse.platform/pull/441#discussion_r1187251273

fedejeanne: Good point, @jukzi, I missed that one. 
I'm going to take @HeikoKlare's  advice and add it for removal only if it didn't exist before.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
```diff
@@ -17,29 +17,19 @@
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
+import java.util.*;
```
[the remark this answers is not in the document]
fedejeanne: On a closer look, this happened because of a configuration problem. I filed a bug (#442) and opened a new PR (#443)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
```diff
@@ -134,25 +135,32 @@ public void testCopyAcrossVolumes() throws Throwable {
...
+		String subfolderName = "target_" + System.currentTimeMillis();
+
+		IFileStore target = tempSrc.getChild(subfolderName);
 		createDir(target, true);
 		createTree(getTree(target));
 
 		/* c:\temp\target -> d:\temp\target */
-		IFileStore destination = tempDest.getChild("target");
+		IFileStore destination = tempDest.getChild(subfolderName);
+		deleteOnTearDown(destination);
```
jukzi, line 148: why do we still need those deleteOnTearDown here?

HeikoKlare: My is understanding is the following:
If the folder at which the file store root is placed already exists, it will not be deleted on teardown (this is the essential chang ein this PR). Then all child stores have to be deleted on teardown, which is why `deleteOnTearDown` is called for every child.

HeikoKlare:
Consider a file `C:\temp\target` already exists (and is probably even read only), then the test failed as it tried to create the very same file. Adding a timestamp to the file name reduces the risk of such collisions. A name generator that searches for a non-conflicting file name would be even better, but I think this solution is sufficient because of the unlikeliness of a collosions with names containing the current timestamp.

jukzi:
ok, lets try

Issue #434, opened 2023-05-03
Skip tests visibly


fedejeanne:
Several test methods use `return` to skip the test. It's better to use `assume` to skip a method since this makes it clear in the CI environment that a test was not executed.

fedejeanne:
Fixed by #446 
Further improvements (refactoring / migrating to JUnit5) were proposed in #448.

PR #446, opened 2023-05-10
Use "assume" instead of "return" in several tests #434

fedejeanne:
The objective is to let tests be skipped "visibly" i.e. they should be marked as *skipped* instead of *successful* in the test reports. In order to do this, I need to let these tests run with the "JUnit4" runner (the older runner "JUnit38" reports the tests as "failed" when the condition in `Assume.assumeXYZ(...)` is not met).

I also extracted 3 new methods in order to improve the readability of some tests, these are:
* `ResourceTest.isLinux()` and `ResourceTest.isMacOSX()` are similar to the already existing `ResourceTest.isWindows()`
* `CoreTest.assumeCanCreateSymLinks()` uses the already existing `canCreateSymlinks()` and assumes it is `true`

I also removed some unnecessary `return` statements.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
@@ -27,7 +27,12 @@ import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.Platform;
+import org.junit.Assume;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 
+@RunWith(JUnit4.class)
 public class SymlinkTest extends FileSystemTest {
 	/**
 	 * Symbolic links on Windows behave differently compared to Unix-based systems. Symbolic links
@@ -37,7 +42,7 @@ public class SymlinkTest extends FileSystemTest {
 	 */
 	private static final boolean SYMLINKS_ARE_FIRST_CLASS_FILES_OR_DIRECTORIES = Platform.OS_WIN32
 			.equals(Platform.getOS()) ? true : false;
-	private static String specialCharName = "äöüß ÄÖÜ àÀâÂ µ²³úá"; //$NON-NLS-1$
+	private static String specialCharName = "Ã¤Ã¶Ã¼ÃŸ Ã„Ã–Ãœ Ã Ã€Ã¢Ã‚ ÂµÂ²Â³ÃºÃ¡"; //$NON-NLS-1$
 
 	protected IFileStore aDir, aFile; //actual Dir, File
 	protected IFileInfo iDir, iFile, ilDir, ilFile, illDir, illFile;
@@ -105,11 +110,11 @@ public class SymlinkTest extends FileSystemTest {
 		baseStore.delete(EFS.NONE, null);
 	}
 
+	@Test
 	public void testBrokenSymlinkAttributes() {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		long testStartTime = System.currentTimeMillis();
 		makeLinkStructure();
 		//break links by removing actual dir and file
@@ -153,11 +158,11 @@ public class SymlinkTest extends FileSystemTest {
 	}
 
 	// Moving a broken symlink is possible.
+	@Test
 	public void testBrokenSymlinkMove() throws Exception {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		makeLinkStructure();
 		ensureDoesNotExist(aFile);
 		ensureDoesNotExist(aDir);
@@ -199,11 +204,11 @@ public class SymlinkTest extends FileSystemTest {
 	}
 
 	// Removing a broken symlink is possible.
+	@Test
 	public void testBrokenSymlinkRemove() throws Exception {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		makeLinkStructure();
 		ensureDoesNotExist(aFile);
 		ensureDoesNotExist(aDir);
@@ -219,11 +224,11 @@ public class SymlinkTest extends FileSystemTest {
 		assertEquals(infos.length, 0);
 	}
 
+	@Test
 	public void testRecursiveSymlink() throws Exception {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		mkLink(baseStore, "l1", "l2", false);
 		mkLink(baseStore, "l2", "l1", false);
 		IFileStore l1 = baseStore.getChild("l1");
@@ -270,11 +275,11 @@ public class SymlinkTest extends FileSystemTest {
 		assertEquals(infos.length, 1);
 	}
 
+	@Test
 	public void testSymlinkAttributes() {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		makeLinkStructure();
 		assertFalse(iFile.getAttribute(EFS.ATTRIBUTE_SYMLINK));
 		assertFalse(iDir.getAttribute(EFS.ATTRIBUTE_SYMLINK));
@@ -311,11 +316,11 @@ public class SymlinkTest extends FileSystemTest {
 	}
 
 	// Reading from a directory pointed to by a link is possible.
+	@Test
 	public void testSymlinkDirRead() throws Exception {
 		// Only activate this test if testing of symbolic links is possible.
-		if (!canCreateSymLinks()) {
-			return;
-		}
+		assumeCanCreateSymLinks();
+
 		makeLinkStructure();
 		IFileStore childDir = aDir.getChild("subDir");
 		ensureExists(childDir, true);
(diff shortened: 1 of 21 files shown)


vogella:
@HeikoKlare can you review?

HeikoKlare:
Yes, I'll have a look.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -167,6 +167,26 @@ protected static boolean isWindows() {
...
+	/**
+	 * Returns whether the current platform is mac OSX.
+	 *
+	 * @return <code>true</code> if this platform is mac OSX, and <code>false</code>
+	 *         otherwise.
+	 */
+	protected static boolean isMacOSX() {
+		return Platform.getOS().equals(Platform.OS_MACOSX);
+	}
+
```
HeikoKlare, line 189: I'd propose to have these methods in a separate utility class (together with the pre-existing `isWindows` method) to stop fostering inheritance for reusing test utility methods by further messing up test base classes. Then these methods can also be used in other test classes nthat do not inheritfrom `ResourceTest`. Maybe this refactoring could be separated into an independent PR since it is unrelated to the addition of assumes/moving to JUnit 4 anyway.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
```diff
@@ -37,7 +42,7 @@ public class SymlinkTest extends FileSystemTest {
 	 */
 	private static final boolean SYMLINKS_ARE_FIRST_CLASS_FILES_OR_DIRECTORIES = Platform.OS_WIN32
 			.equals(Platform.getOS()) ? true : false;
-	private static String specialCharName = "���� ��� ���� �����"; //$NON-NLS-1$
+	private static String specialCharName = "äöüß ÄÖÜ àÀâÂ µ²³úá"; //$NON-NLS-1$
```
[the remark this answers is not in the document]
fedejeanne: This seems like a bug in GitHub. If you switch to the "split" view (in the "Code" Tab) then you'll see, that the content has not changed. I don't know why Git(Hub) even sees it as a change.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
fedejeanne: Sounds good, I already created an issue for it --> #448

laeubi: Why not add methods to the Platformclass directly? `Platform` already is some kind of utility class, so why we need a utility for a utility that the is used in another utility class?

fedejeanne: @laeubi  that could work too, thank you! I modified the issue #448 , in case you want to follow it :)

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/CoreTest.java
```diff
@@ -55,6 +58,15 @@ public static void debug(String message) {
...
 
+	@Before
+	public void before() throws Exception {
+		setUp();
+	}
+
+	@After
+	public void after() throws Exception {
+		tearDown();
+	}
```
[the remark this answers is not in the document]
fedejeanne: I also thought about this and tried to break the "is-a" relation between the modified classes and `TestCase`. 
I had a number of problems with:
* Several usages of methods like `assertXYZ` need to be changed
* Usages of "base" methods (`assertXYZ` / `failXYZ`) that hat been defined in `CoreTest` need to be changed.
* Several test suites are still defined in the old fashion and add tests by calling `junit.framework.TestSuite.addTest(Test)`

The last one is specially tricky because it means the the test suites need to be completely migrated, which is a major undertake.

I saw my approach as a "quick win". I agree with you, it's not pretty, but is a cheap approach and gets the job done.

Can you live with this technical debt? :)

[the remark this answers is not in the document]
HeikoKlare: Yes, I agree that it will be quite some effort to migrate, in particular because of special test setups like with the session tests.

However, that is independent from where this "cheap approach" is placed. I would still be in favor of having it as local as possible and not placed in the central core test harness plugin.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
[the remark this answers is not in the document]
HeikoKlare: It is actually a change from ISO-8859-1 to UTF-8 encoding. GitHub only displays boths file in the same way as it identifies the different encodings.

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/CoreTest.java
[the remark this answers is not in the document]
fedejeanne: I moved the `before` and `after` methods a bit down in the hierarchy and into `ResourceTest` and `FileSystemTest` (both in the plugin `org.eclipse.core.tests.resources`).

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -200,6 +222,16 @@ public ResourceTest(String name) {
...
+	@Before
+	public void before() throws Exception {
+		setUp();
+	}
+
+	@After
+	public void after() throws Exception {
+		tearDown();
+	}
+
```
HeikoKlare, line 234: I'd propose to make these methods `final` to avoid that subclasses overwrite them with setup/teardown functionality although sticking to JUnit 3 (such that the methods will never be called and added functionality is actually dead code). Please also document the mixture of JUnit 3 and JUnit 4 at these methods to ease comprehension for why the methods are there.

fedejeanne: @HeikoKlare  done

HeikoKlare:
I'd propose to follow the intention of this PR in a slightly different way:
1. Those changes related to identifying the actual OS, as also discussed in #448, should be extracted into a dedicated PR.
2. Those changes related to introducing JUnit 4 assumptions and selectively specifying a JUnit 4 runner in test classes using assumptions should not be made but replaced by a proper JUnit migration (probably directly to JUnit 5) for the following reasons:
    - The proposed changes would introduce a mix of JUnit 3 and JUnit 4 instead of a proper migration, which has to be reverted when completely migrating to a new JUnit version.
    - The introduction of JUnit 4 constructs in Junit 3 tests leads to further refactoring effort when moving to JUnit 5, because the signature of most assert/assume methods changed (parameter order, in particular position of the message).
    - From my point of view, the benefit of replacing the test conditions with JUnit assumptions does not outweigh the drawback of these two points.

fedejeanne:
I agree with 1)
I have to say though, I don't agree with 2) for the following reasons:

> The proposed changes would introduce a mix of JUnit 3 and JUnit 4 instead of a proper migration, **which has to be reverted when completely migrating to a new JUnit version**.

The only changes that need to be reverted when doing a full-blown migration to JUnit5 are:
- Removing all `@RunWith(JUnit4.class)`: it's a simple search-and-replace operation
- Replacing JUnit4 imports with JUnit5 imports: again, search-and-replace

On the other hand, my PR already does some of the work that a full-blown migration from JUnit3 -> JUnit5 would need, that is:
- Adding `@Test` annotations

> The introduction of JUnit 4 constructs in Junit 3 tests leads to further refactoring effort when moving to JUnit 5, because the signature of most assert/assume methods changed (parameter order, in particular position of the message).

My PR does not change any call to existing `assert`-methods, in fact I explicitly avoid such changes because they would only pollute the PR without any added value (all I want is to use `assume`-calls, I don't need to update all JUnit3-calls to do that).

All in all, the purpose of my PR is to use `assume`-calls without the need of a full-blown migration. This PR even has the nice side-effect of actually bringing the code one step closer to a full-blown migration (adding the `@Test` annotations).


I see we kind of monopolized the discussion here, Heiko :-D , but I'm also interested on what the rest has to say about this. Would anyone else care to weigh in ? :) . 

Do you guys think this PR has any chance of being merged or should it be discarded in favor of a full-blown migration?

Thank you in advance for you opinion!

vogella:
> Do you guys think this PR has any chance of being merged or should it be discarded in favor of a full-blown migration?

I'm in favor of merging this one as is (after the freeze period for the next release). Step-wise improvements are more realistic than "lets implement the perfect solution / complete migration".

jukzi:
i like the change as is

fedejeanne:
Thank you @vogella and @jukzi for weighing in :)

Other than the failed build in Linux, this PR looks ready to be merged, right?

Do we need to trigger another build and see if it succeeds in Linux too? If yes, someone else should do since I don't have the necessary rights to do it myself.

HeikoKlare:
Best you rebase on master. Then we have build results with the most recent state and the linux validation is implicitly done again.

jukzi:
> Do we need to trigger another build and see if it succeeds in Linux too? If yes, someone else should do since I don't have the necessary rights to do it myself.

if you stumble upon possibly random failed test please check if there is an issue for that. if yes, quote it and forget about it. if not create one and rebuild to prove its random.

fedejeanne:
> > Do we need to trigger another build and see if it succeeds in Linux too? If yes, someone else should do since I don't have the necessary rights to do it myself.
> 
> if you stumble upon possibly random failed test please check if there is an issue for that. if yes, quote it and forget about it. if not create one and rebuild to prove its random.

@jukzi I will do so, thank you for the tip :) . In this particular case there was no info as to why the build failed, it just said "this build failed" so I assume it must have been some sort of (harware/configuration?) problem with the CI environment.



> Best you rebase on master. Then we have build results with the most recent state and the linux validation is implicitly done again.

@HeikoKlare good point, thanks! I rebased and now all checks are green ✔✔✔ This PR is now ready to be merged :)

HeikoKlare:
Great! Thank you all.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
fedejeanne: See https://github.com/eclipse-platform/eclipse.platform/pull/704

PR #450, opened 2023-05-12
Re-acquire locks via iterative instead of recursive execution #447

### LINKED ISSUE #447: StackOverflowError in LockManager.resumeSuspendedLocks
The call to `LockManager.resumeSuspendedLocks` sometimes results in a `StackOverflowError` because of an endless recursion:
```
Exception in thread "Thread-32" java.lang.StackOverflowError
	at java.base/java.util.Spliterator.getExactSizeIfKnown(Spliterator.java:414)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:526)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.findAny(ReferencePipeline.java:652)
	at org.eclipse.core.internal.jobs.OrderedLock.enqueue(OrderedLock.java:203)
	at org.eclipse.core.internal.jobs.OrderedLock.createSemaphore(OrderedLock.java:146)
	at org.eclipse.core.internal.jobs.OrderedLock.acquire(OrderedLock.java:105)
	at org.eclipse.core.internal.jobs.LockManager$LockState.resume(LockManager.java:58)
	at org.eclipse.core.internal.jobs.LockManager.resumeSuspendedLocks(LockManager.java:316)
	at org.eclipse.core.internal.jobs.OrderedLock.acquire(OrderedLock.java:111)
	at org.eclipse.core.internal.jobs.LockManager$LockState.resume(LockManager.java:58)
	at org.eclipse.core.internal.jobs.LockManager.resumeSuspendedLocks(LockManager.java:316)
	at org.eclipse.core.internal.jobs.OrderedLock.acquire(OrderedLock.java:111)
	at org.eclipse.core.internal.jobs.LockManager$LockState.resume(LockManager.java:58)
	at org.eclipse.core.internal.jobs.LockManager.resumeSuspendedLocks(LockManager.java:316)
	at org.eclipse.core.internal.jobs.OrderedLock.acquire(OrderedLock.java:111)
	at org.eclipse.core.internal.jobs.LockManager$LockState.resume(LockManager.java:58)
	at org.eclipse.core.internal.jobs.LockManager.resumeSuspendedLocks(LockManager.java:316)
        ...
```
This bug seems to exist for quite some time and is the reason for the (Windows) build randomly timeouting after 6 hours. One of the oldest builds failing for this reason is from 17 March 2023: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4445856292/jobs/7805481648#step:5:10990
But the error can be found in every recent Windows build timeouting after 6 hours.

Since the build timed out already before March (but logs do not exist anymore), this bug seems to exist even longer.

HeikoKlare:
When multiple `OrderedLocks` are acquired by different threads, the deadlock recovery mechanism suspending and reacquiring the locks requires an indefinite number of tries until one thread holds all required locks. Since reacquisition is performed by recursive method invocation, the stack can become infinitely large with the change of resulting in a `StackOverflowError`.

These changes replace the recursive lock acqusition by an iterative one, such that still an indefinite number of tries for acquiring a set of locks is required but the chance of resulting in an error is eliminated. The `OrderedLockTest.testComplex`, which was randomly failing on Windows systems due to the recursive implementation and thus disabled in #455, is re-enabled.

The added test case does not deterministically reproduce the erroneous behavior, but since a proper regression test is very hard to define (as specific lock order across a magnitude of retries has to be ensured and coordinated between different threads), it at least executes more sophisticated locking scenarios to ensure proper lock retrieval and deadlock management.

Fixes #447. In particular, with this fix the build timeouts after 6h should hopefully disappear (which were only worked around by #455).

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/LockManager.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/LockManager.java
@@ -49,19 +49,16 @@ public class LockManager {
 		/**
 		 * Re-acquires a suspended lock and reverts to the correct lock depth.
 		 */
-		public void resume() {
-			//spin until the lock is successfully acquired
-			//NOTE: spinning here allows the UI thread to service pending syncExecs
-			//if the UI thread is waiting to acquire a lock.
-			while (true) {
-				try {
-					if (lock.acquire(Long.MAX_VALUE))
-						break;
-				} catch (InterruptedException e) {
-					//ignore and loop
+		public boolean resume() {
+			try {
+				if (lock.acquire(Integer.MAX_VALUE, false)) {
+					lock.setDepth(depth);
+					return true;
 				}
+			} catch (InterruptedException e) {
+				// ignore and retry
 			}
-			lock.setDepth(depth);
+			return false;
 		}
 	}
 
@@ -303,17 +300,26 @@ public class LockManager {
 	 * Resumes all the locks that were suspended while this thread was waiting to acquire another lock.
 	 */
 	void resumeSuspendedLocks(Thread owner) {
-		LockState[] toResume;
-		synchronized (suspendedLocks) {
-			Deque<LockState[]> prevLocks = suspendedLocks.get(owner);
-			if (prevLocks == null)
-				return;
-			toResume = prevLocks.pop();
-			if (prevLocks.isEmpty())
-				suspendedLocks.remove(owner);
+		// spin until the locks are successfully acquired
+		// NOTE: spinning here allows the UI thread to service pending syncExecs
+		// if the UI thread is waiting to acquire a lock.
+		while (true) {
+			LockState[] toResume;
+			synchronized (suspendedLocks) {
+				Deque<LockState[]> prevLocks = suspendedLocks.get(owner);
+				if (prevLocks == null)
+					return;
+				toResume = prevLocks.pop();
+				if (prevLocks.isEmpty())
+					suspendedLocks.remove(owner);
+			}
+			for (LockState element : toResume) {
+				if (!element.resume()) {
+					// Try again if acquiring a lock failed
+					continue;
+				}
+			}
 		}
-		for (LockState element : toResume)
-			element.resume();
 	}
 
 	public void setLockListener(LockListener listener) {
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/OrderedLock.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/OrderedLock.java
@@ -97,6 +97,21 @@ public class OrderedLock implements ILock, ISchedulingRule {
 
 	@Override
 	public boolean acquire(long delay) throws InterruptedException {
+		return acquire(delay, true);
+	}
+
+	/**
+	 * Acquires a lock as defined by {@link #acquire(long)} but only resumes
+	 * suspended locks if {@code resumeSuspendedLocks} is {@code true}.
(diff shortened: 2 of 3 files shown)


runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/LockManager.java
```diff
@@ -303,17 +300,26 @@ void removeLockWaitThread(Thread thread, ISchedulingRule lock) {
...
+				if (prevLocks == null)
+					return;
+				toResume = prevLocks.pop();
+				if (prevLocks.isEmpty())
+					suspendedLocks.remove(owner);
+			}
+			for (LockState element : toResume) {
+				if (!element.resume()) {
+					// Try again if acquiring a lock failed
+					continue;
```
iloveeclipse, line 319: Not sure I understand what this "continue" is supposed to do? Is it does anything - the for loop will just continue. Was it meant to continue the outer while loop may be?

HeikoKlare: You are completely right. The `continue` was supposed to continue the outer loop, such that whenever a resume fails, the loop starts to acquire the locks in the same order again (like the former recursive implementation did). However, it should have been a `break` to achieve that behavior. I've fixed that in d052003.

fedejeanne:
Another occurrence: https://github.com/eclipse-platform/eclipse.platform/actions/runs/4957027490/jobs/8868205300#step:7:6021

iloveeclipse:
The change is not trivial, and since this is not a recent regression, I would propose to postpone with it for 4.29 M1. Objections?

HeikoKlare:
No objections. Makes sense to postpone merge to after 4.28 release 👍 

I propose to have a separate PR that temporarily disables the problematic test case on Windows systems. Then we hopefully already have (more) stable builds for the next weeks while still being aware of regressions (by still executing the test on the other platforms). This PR will then enable the test on Windows again. If you think that's a bad idea, let me know (either here or in the separate PR to come).

iloveeclipse:
> I propose to have a separate PR that temporarily disables the problematic test case on Windows systems.

Sure. Please link to this issue.

HeikoKlare:
I've temporarily disabled the problematic test case on Windows in #455 and changed this PR to re-enable the test.

HeikoKlare:
@iloveeclipse We postponed merging this after 4.28 release. Any objections on merging this now?

iloveeclipse:
> @iloveeclipse We postponed merging this after 4.28 release. Any objections on merging this now?

I've rebased to see if there is something to update after release change.

HeikoKlare:
For documentation: I made a minor change after the review, only affecting the added test case.

Slow CI hardware made the new test case run for quite a long time, particularly on macOS machines, ranging to [60 seconds](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5334722540/jobs/9666892271#step:7:11426) or even [timeouts](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5313675043/jobs/9619885628#step:7:12438).

Thus, I have reduced the number of locks and threads used in that test case to reach acceptable execution times.

HeikoKlare:
Failing test documented in #488.

PR #462, opened 2023-05-18
fixing dependabot.yml format fix: #456

fperezel:


Bananeweizen (approved):

mickaelistria:
@Bananeweizen Feel free to merge those directly. As they're not part of the "payload" and not involved in the official build, it's fine to not delay the merge.

Bananeweizen:
@mickaelistria Only after giving the review I noticed that there is a difference: Quotes around root directory in GH actions part, no quote around root directory for the maven part. I was looking for whether that is relevant, and got interrupted in between.

PR #465, opened 2023-05-26
Bugs/460 test bug297635

### LINKED ISSUE #460: Random failing of TestBug297635
The test `TestBug297635.test2` fails randomly.

One such failures can be seen in https://github.com/eclipse-platform/eclipse.platform/actions/runs/4991112285/jobs/8937471805.

Here's the stack trace:

```
junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:55)
	at junit.framework.Assert.assertNotNull(Assert.java:256)
	at junit.framework.Assert.assertNotNull(Assert.java:248)
	at junit.framework.TestCase.assertNotNull(TestCase.java:391)
	at org.eclipse.core.tests.resources.session.TestBug297635.assertStateTrees(TestBug297635.java:169)
	at org.eclipse.core.tests.resources.session.TestBug297635.test2(TestBug297635.java:131)
...
```

fedejeanne:
This fixes #460 

Some detailed explanation


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -47,9 +55,47 @@ public BundleContext getContext() {
 
 	public void test1() {
```
vogella, line 56: horrible name, maybe have another PR to improve the test method name?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -58,86 +104,73 @@ public void test1() {
...
 			Bundle coreResourcesBundle = Platform.getBundle(ResourcesPlugin.PI_RESOURCES);
 			coreResourcesBundle.stop(Bundle.STOP_TRANSIENT);
-			Bundle b = BundleTestingHelper.installBundle("1", getContext(), ContentTypeTest.TEST_FILES_ROOT + "content/bundle01");
-			BundleTestingHelper.resolveBundles(getContext(), new Bundle[] {b});
+			Bundle b = BundleTestingHelper.installBundle("1", getContext(),
+					ContentTypeTest.TEST_FILES_ROOT + "content/bundle01");
+			BundleTestingHelper.resolveBundles(getContext(), new Bundle[] { b });
 			coreResourcesBundle.start(Bundle.START_TRANSIENT);
 		} catch (MalformedURLException e2) {
-			fail("1.0", e2);
```
vogella, line 95: Maybe have a followup comming which uses better naming for the error message, for example the exception class as String? I know this is used a lot in the test but I never understood the value of using increasing numbers as error messages.

vogella:
@fedejeanne please remove merge commit (see https://www.vogella.com/tutorials/Git/article.html#pullrebase for a general setup).

My comments could be handled in another PR. This one looks fine to me, any other opions @jukzi ?

jukzi:
@vogella This is still draft. Please separate actual fixes from refactoring and give an explanation about the fix for easier review.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -58,86 +104,73 @@ public void test1() {
...
+		// get access to SaveManager#savedStates to verify that tress are being kept
+		// there
 		Map<String, SavedState> savedStates = null;
 		try {
 			Field field = SaveManager.class.getDeclaredField("savedStates");
 			field.setAccessible(true);
 			savedStates = (Map<String, SavedState>) field.get(((Workspace) getWorkspace()).getSaveManager());
 		} catch (IllegalArgumentException e) {
-			fail("2.0", e);
+			fail("5.0", e);
```
laeubi, line 157: Instead of fail, why not simply throw the exception (what would fail he test as well...)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -58,86 +104,73 @@ public void test1() {
...
-
-		// assert the saved state for Bundle01, trees should not be null
-		assertStateTrees(savedStates.get(BUNDLE01_ID), false);
-
+	private void saveUsingSaveManager() {
 		try {
 			((Workspace) getWorkspace()).getSaveManager().save(ISaveContext.SNAPSHOT, true, null, getMonitor());
 		} catch (CoreException e) {
-			fail("4.0", e);
+			fail("7.0", e);
```
laeubi, line 172: same here, simply throw `CoreException`

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
laeubi: The question is more why catch the exception at all and not simply throw it ?!

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -58,86 +104,73 @@ public void test1() {
...
 		try {
 			getWorkspace().save(true, getMonitor());
 		} catch (CoreException e) {
 			fail("3.0", e);
 		}
 	}
 
-	@SuppressWarnings("unchecked")
-	public void test2() {
+	private void reinstallBundle() {
```
laeubi, line 125:
```suggestion
	private void reinstallBundle() throws Exception {
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
fedejeanne: I will, thank you.
I'm also not sure why the numbering but I noticed it's even part of the code templates (if I choose to handle an exception with "surround with try-catch" in a test then the `fail(<number>, e)` is added automatically). Maybe some attempt to quickly identify the line where the exception happened if no stack-traces were present? ... Anyhow, good to know I don't need to continue the tradition :)

iloveeclipse: > Maybe some attempt to quickly identify the line where the exception happened if no stack-traces were present?

Note, the code was written probably at Java 1.2 times, may be ~20 years ago. Whatever didn't worked at that time in Java / Eclipse we will not know anymore, but today such style is simply not needed and wrong.

> it's even part of the code templates

Please create a dedicated PR to change templates to use defaults.

fedejeanne: @iloveeclipse , I updated the templates in #466

HeikoKlare (commented):
This test class is a `WorkspaceSessionTest`. Thus the intention was that between the original test methods `test1` and `test2`, the platform was shut down and restarted.
So from my point of view either
- the two test methods should remain to still have a restart of the platform, or
- the test should not be a `WorkspaceSessionTest` anymore.

I do not see any reasons for restarting the platform in the original bug report (https://bugs.eclipse.org/bugs/show_bug.cgi?id=297635), so probably this stays a proper regression test for that bug when making a simple `ResourceTest` out of it.

fedejeanne:
Good point, I changed it to a `ResourceTest` in c08de7a since I also don't see the point in restarting the platform between tests. I mean I get there might be a memory leak when running a single WS instance for a long time (see [Comment#4](https://bugs.eclipse.org/bugs/show_bug.cgi?id=297635#c4) and [Comment#5](https://bugs.eclipse.org/bugs/show_bug.cgi?id=297635#c5)) but the tests does not seem to address that issue.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -45,26 +53,12 @@ public BundleContext getContext() {
...
 		// register a save participant for the bundle
-		try {
-			getWorkspace().addSaveParticipant(BUNDLE01_ID, TestBug297635.this);
-		} catch (CoreException e) {
-			fail("2.0", e);
-		}
+		addSaveParticipant();
 
 		// create a project with a file
 		IProject project = getWorkspace().getRoot().getProject("Project1");
```
HeikoKlare, line 64: You could think of separating the duplicated logic for creating a file and checking its existance into a method.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
```diff
@@ -45,26 +53,12 @@ public BundleContext getContext() {
 		return Platform.getBundle(PI_RESOURCES_TESTS).getBundleContext();
 	}
 
-	public void test1() {
+	public void testBug() throws Exception {
```
HeikoKlare, line 56: The method becomes much easier to read and comprehend due to this separation of functionality into dedicated methods.
Maybe you could even remove the comments and, if necessary, improve the method names. That could further improve comprehensibility, also w.r.t. the method actually being composed of two chunks that made up the original test methods (install / reinstall).

HeikoKlare (approved):
Changes look good to me. Just two minor comments for possible improvements, which are (from my side) not mandatory to be processed before merging.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
fedejeanne: Done! 
I extracted 2 private methods in 5fd81b6eee542d095e97e18f27ceffe337d070b6:
* `getProject`
* `createFileInProject`

The last one creates a file and makes sure it exists in the WS.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestBug297635.java
fedejeanne: You mean:
1. Rename `testBug` to `testSavedStatesAreDiscardedAfterSnapshot` (or do you have another name in mind?)
2. Remove comment (`/* ... */`) in `reinstallBundle`

If not, could you please give me some example(s)?

HeikoKlare: Sorry for being imprecise: I just meant to remove the comments within the `testBug` method, i.e., the comments just repeating what the method called in the line after the comment is doing, like
```java
// install a bundle
installBundle();
```
The method of the test method can remain as it is.

fedejeanne: I get it now, thanks.

I created/renamed some private methods and removed unnecessary comments in `testBug` in 0347088028c67a0517c11c530f796d077076afbe

Issue #469, opened 2023-05-31
Random failing of IJobManagerTest.testReverseOrder()


fedejeanne:
It happened during [this build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/4988938481/jobs/8932229100#step:7:11196) of the PR #443 

```org.eclipse.core.tests.runtime.jobs.IJobManagerTest.testReverseOrder  Time elapsed: 0.721 s  <<< FAILURE!
junit.framework.AssertionFailedError: 1.1.0 expected:<testReverseOrder(0)(587)> but was:<testReverseOrder(1)(588)>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.TestCase.assertEquals(TestCase.java:238)
	at org.eclipse.core.tests.runtime.jobs.IJobManagerTest.testReverseOrder(IJobManagerTest.java:1778)
```

fedejeanne:
Please observe: this happened after merging #401 so there might be some other issue with the tests in `IJobManagerTest`

PR #471, opened 2023-05-31
Remove test IJobManagerTest.testReverseOrder() #469

fedejeanne:
The test was added in 2003, its intention is not completely clear and it can not be designed in a way that is independent of the hardware. This makes the test fail when running in slow machines and succeed when running in fast machines.

The problem with the test is that the `Job.schedule(long delay)` does not really guarantee the order of execution, it only says that:
>Jobs of equal priority and delay with conflicting schedulingrules are guaranteed to run in the order they are scheduled. No guaranteesare made about the relative execution order of jobs with unrelated or null scheduling rules, or different priorities. 

Which means the test can only try to schedule the jobs with such delays that they will run in reverse order. 

### How to reproduce the error?
Change the delay between the jobs in such a way that 2 consecutive jobs are "too close", e.g.:
```
int[] sleepTimes = new int[] { 600, 400, 1, 0 }; // it was { 600, 400, 200, 5} before
...
for (int i = 0; i < sleepTimes.length; i++) {
	jobs[i].schedule(sleepTimes[i]);
}
```
In this scenario, the 3rd job will be scheduled to run in `1` ms and that will not be enough for the test to even reach the line were the 4th job is scheduled.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
@@ -1750,35 +1750,6 @@ public class IJobManagerTest extends AbstractJobManagerTest {
 		}
 	}
 
-	public void testReverseOrder() {
-		//ensure jobs are run in order from lowest to highest sleep time.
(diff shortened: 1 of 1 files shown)


fedejeanne:
Addresses Bug #469

jukzi:
conflicts with https://github.com/eclipse-platform/eclipse.platform/pull/507

PR #473, opened 2023-06-01
 Fix stale resource link flags after project re-open

### LINKED ISSUE #470: Project description links information is not persisted by LocalMetaArea
In our product, we have a builder that reads linked folders. The builder runs into a NPE, due to stale links information. The stale links information results due to the following sequence:

1. Debug Eclipse with a breakpoint at the start of `Project.reconcileLinksAndGroups()`.
2. Create a Java project.
3. Add a linked source folder to some folder on disk.
4. Close the project.
5. Delete the information about the link from the .project file on disk, e.g. delete the following block (location will be different):
```
    <linkedResources>
	    <link>
		    <name>test</name>
		    <type>2</type>
		    <location>/data/tmp/test</location>
	    </link>
    </linkedResources>
```
6. Open the project in Eclipse.
7. Observe `oldDescription.getLinks()` returns `null` and so nothing is done to refresh the links information.

The information about links in the project description is not persisted in `LocalMetaArea.writePrivateDescription()` and is also not read in `LocalMetaArea.readPrivateDescription()`. We should fix this, to ensure up-to-date links information in the project description after opening a closed project.

trancexpress:
Whenever a project is closed, its resource tree is saved. This includes
linked resources in the project. When the project is re-opened, link
changes in the .project file are not reflected on the projects resource
tree. The old resource tree is read, new information is stored in
ProjectDescription.linkDescriptions, but the old linked resources are
not touched.

This change adjusts Project.open() and Project.close() to set resp.
clear the M_LINK flag of linked resources in the project.

Fixes: https://github.com/eclipse-platform/eclipse.platform/issues/470

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
@@ -29,6 +29,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.IFileStore;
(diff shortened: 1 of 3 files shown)


akurtakov:
What is the status of this one?

iloveeclipse:
> What is the status of this one?

Waits for a review

mickaelistria:
FWIW, I have trouble to properly understand the use-case. @trancexpress would it be possible to add a test that would demonstrate how to trigger the issue?

iloveeclipse (changes requested):
Thanks Simeon, this is a very complex code here and very nice to have a working test.

I've wrote my [understanding of the problem](https://github.com/eclipse-platform/eclipse.platform/issues/470#issuecomment-2077079561) and [proposed solution idea](https://github.com/eclipse-platform/eclipse.platform/issues/470#issuecomment-2077539970).

I've pushed rebased, merged & updated PR here to get it easier for the future work - there were massive changes in the resource tests area on master and so the test was not compilable anymore.

Now it compiles & works again.

I believe the code change itself is not enough, I believe we need a different approach which I've tried to draft in https://github.com/eclipse-platform/eclipse.platform/issues/470#issuecomment-2077539970.

PR #475, opened 2023-06-02
Fix mismatched begin/endOperation

jukzi:
Every prepareOperation() needs a final(!) endOperation() see https://bugs.eclipse.org/bugs/show_bug.cgi?id=124211#c4

In case of Exceptions an endOperation() could have been missing.


mickaelistria:
I don't think the changes in WorkManager are related to the initial intent of this patch. Can you please create a separate PR for those?

mickaelistria:
> It relates: i made the WorkManager methods package private so that i am sure i hit all references - especially to checkOut().

I see it more as a preliminary "technical" change towards elaborating the proper "functional" fix, but not a requirement.
There are risks that someone does use the WorkManager directly despite it's internal, so there are risk that we will have to revert it because of an important enough use-case and incapacity to address it better in a good time frame. So if we can have 2 distinct commits, it would allow to revert -if necessary- one of them without reverting the other thus to keep value in.
1 PR with 2 commits would be OK too.

jukzi:
ok, i will do

jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform/issues/95

PR #478, opened 2023-06-05
Reduce WorkManager method visibility

jukzi:
because checkin/checkout has to be used symmetrically.


jukzi:
see https://github.com/eclipse-platform/eclipse.platform/pull/475#issuecomment-1573795253

PR #479, opened 2023-06-05
Add @FixMethodOrder to ContentDescriptionPerformanceTest #432

### LINKED ISSUE #432: Failing tests under: org.eclipse.core.tests.resources.perf.ContentDescriptionPerformanceTest
The following performance tests are failing in the mentioned class:
* `test2ColdContentDescription()`
* `test3WarmedUpContentDescription()`

The error is the same in both cases:
```
junit.framework.AssertionFailedError: Failed visiting resources: org.eclipse.core.internal.resources.ResourceException: Resource '/bigproject' does not exist.
	at org.eclipse.core.tests.harness.CoreTest.fail(CoreTest.java:74)
	at org.eclipse.core.tests.resources.perf.ContentDescriptionPerformanceTest$1.test(ContentDescriptionPerformanceTest.java:84)
	at org.eclipse.core.tests.harness.PerformanceTestRunner.run(PerformanceTestRunner.java:78)
	at org.eclipse.core.tests.harness.PerformanceTestRunner.run(PerformanceTestRunner.java:52)
	at org.eclipse.core.tests.resources.perf.ContentDescriptionPerformanceTest.doTestContentDescription(ContentDescriptionPerformanceTest.java:87)
...
```

Also, I can't find these tests in the CI runs (see related issue)

fedejeanne:
Fixes #432 

The problem is that `test1CreateWorkspace()` is actually a "setup" and `test4CleanUp()` a "tearDown". The real tests are:
* `test2ColdContentDescription()`
* `test3WarmedUpContentDescription()`

And there is also a dependency between these 2 so they should also run in the given order.

The annotation `@FixMethodOrder(MethodSorters.NAME_ASCENDING)` guarantees the execution order while preserving the original code intact.

I also added `IGNORED_FILES` and ignored _org.eclipse.core.resources.prefs_ since it's not relevant for the test.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/perf/ContentDescriptionPerformanceTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/perf/ContentDescriptionPerformanceTest.java
@@ -13,13 +13,17 @@
  *******************************************************************************/
 package org.eclipse.core.tests.resources.perf;
 
+import java.util.Set;
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.content.IContentDescription;
 import org.eclipse.core.tests.harness.CoreTest;
 import org.eclipse.core.tests.harness.PerformanceTestRunner;
 import org.eclipse.core.tests.resources.ResourceTest;
+import org.junit.FixMethodOrder;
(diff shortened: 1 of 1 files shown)


akurtakov:
Would you please rename to setup/teardown to reduce confusion too?

fedejeanne:
@akurtakov I renamed the methods in 3caa305

akurtakov:
Thanks, merged.

PR #481, opened 2023-06-06
use try-with-resource

jukzi:


akurtakov:
Thanks @jukzi ! These small incremental improvements really help a lot for the project maintenance IMHO.

jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform/issues/488

Issue #490, opened 2023-06-12
WorkManager.isLockAlreadyAcquired() reduced visibility in 4.29


iloveeclipse:
`WorkManager.isLockAlreadyAcquired()` **is not a public API** but it was there **since ever** (2003, see e9154abb3c1eab89e18aa250cd2657c4b5af04f0).

With 28d4794d0b127f3b1c8818a65c08f214384afd69 the method was made package protected, which is technically OK but **unnecessarily** breaks all clients that used non-API code before (Xtext is one example: `org.eclipse.xtext.ui.workspace.WorkspaceLockAccess.isWorkspaceLockedByCurrentThread(IWorkspace)`).

To avoid unnecessarily breaking existing code, this method should be made public again.

This shouldn't be an issue, as we had no issues having this code public for 20 years, should be OK to keep it public for the next 20 years.

I will push a PR.

merks:
When I saw that change come through originally I thought, "Oh, that's must be for my safety and comfort."   But, just like on an airplane where so many things are for our safety and comfort, the comfort part is often a bit questionable at best.  😨

PR #491, opened 2023-06-12
Restore public access on WorkManager.isLockAlreadyAcquired()

iloveeclipse:
WorkManager.isLockAlreadyAcquired() is not a public API but it was there since ever (2003, see e9154ab).

With 28d4794 the method was made package protected, which is technically OK but unnecessarily breaks all clients that used non-API code before (Xtext is one example:
org.eclipse.xtext.ui.workspace.WorkspaceLockAccess.isWorkspaceLockedByCurrentThread(IWorkspace)).

To avoid unnecessarily breaking existing code, this method should be made public again.

This shouldn't be an issue, as we had no issues having this code public for 20 years, should be OK to keep it public for the next 20 years.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/490


iloveeclipse:
Test fail on Mac is unrelated.

PR #498, opened 2023-06-14
Provide static acessors in the ILog interface as an alternative to Platform

laeubi:
Currently one obtains ILog either by injection/di or a static call to Platform.getLog(...)., but is is more semantic to only use ILog interface and also decreases the coupling of the code to a central singleton class.

This is similar to the recent changes in IPath and will also help in decreasing the need to call `Platform.getXXX` methods so we can easier adapt for changes here especially in the context of https://github.com/eclipse-equinox/equinox/pull/253.

A next step would to to replace references, and possibly moving `ILog` to equinox common to reduce the amount of split package classes.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
@@ -14,7 +14,9 @@
  *******************************************************************************/
 package org.eclipse.core.runtime;
 
+import org.eclipse.core.internal.runtime.InternalPlatform;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
 
 /**
  * A log to which status events can be written.  Logs appear on individual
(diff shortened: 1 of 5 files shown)


runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
```diff
@@ -132,4 +134,29 @@ default void error(String message) {
...
 	}
+
+	/**
+	 * Returns the log for the given bundle. If no such log exists, one is created.
+	 *
+	 * @param bundle the bundle whose log is returned
+	 * @return the log for the given bundle
+	 * @since 3.28
+	 */
+	public static ILog getLog(Bundle bundle) {
```
HannesWell, line 145: The Log part seems to be duplicated, since one the usually calls `ILog.getLog(bundle)`.
What about just naming the method `get` or `of` so that one gets more speaking code like `ILog.of(bundle)`.

The same applies for the method below.

HannesWell (commented):
I like the overall idea.
This also reminds me of https://github.com/eclipse-platform/eclipse.platform/pull/33 and I think if we add two new API methods we should add another argument-less convenient method `get()` or `ofContext()` or alike.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
laeubi: Done

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
```diff
@@ -922,7 +922,7 @@ public static long getStateStamp() {
 	 * @since 3.0
 	 */
 	public static ILog getLog(Bundle bundle) {
```
HannesWell, line 924: I think a comment mentioning the new method would be good here and below.
Eventually this method should at least be deprecated.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
```diff
@@ -132,4 +134,45 @@ default void error(String message) {
 	default void error(String message, Throwable throwable) {
 		log(new Status(IStatus.ERROR, getBundle().getSymbolicName(), message, throwable));
 	}
+
```
HannesWell, line 137: Not it that it really matters, but I have the impression that usually such static factories methods are added at the top of the file instead of the bottom.

HannesWell (approved):
Looks good for me. Just a few minor remarks below.

Since the last time it was quite a controversial topic I think it would be good to get some more feedback from others.

vogella:
LGTM

Issue #497, opened 2023-06-14
Get rid of deprecated org.eclipse.core.runtime.Preferences in platform code


laeubi:
org.eclipse.core.runtime.Preferences is deprecated since 2014, but there are still as of today 46 matches.

It would be good to get the reference count down her and seems a good first issue for contributors that want to get in touch with eclipse code or anyone interested in cleaning up things.

As there was a quite large attempt to "move" code to Java 17 it seems to have issues for deprecated references is even worth more.

alex2145:
If I understood correctly, you want to remove some references to the deprecated class `org.eclipse.core.runtime.Preferences` from the runtime module.

This will break all the plugins that use deprecated methods like `Plugin.getPluginPreferences()`.

laeubi:
If we remove the reference how could it break things?

alex2145:
For example, if we remove the method `Plugin.getPluginPreferences()` then all plugins that still use it will not compile anymore.

I don't see why we should replace those references, preferences are now stored according to scopes in the `org.eclipse.core.runtime.preferences.IPreferencesService` part of equinox and plugins should use `InstanceScope.INSTANCE.getNode(yourPluginId)` to set preferences and `DefaultScope.INSTANCE.getNode(yourPluginId)` to set default preferences.

It's my first time looking at this project, I might be misunderstaning.
Can you provide an example reference and what you would replace it with?

laeubi:
> For example, if we remove the method Plugin.getPluginPreferences() then all plugins that still use it will not compile anymore.

The method has to stay, but we should remove any call to this method in platform code.

> I don't see why we should replace those references, preferences are now stored according to scopes in the org.eclipse.core.runtime.preferences.IPreferencesService part of equinox and plugins should use InstanceScope.INSTANCE.getNode(yourPluginId) to set preferences and DefaultScope.INSTANCE.getNode(yourPluginId) to set default preferences.

Because the class is deprecated, that means we do not recommend to use it anymore. So it would be good to replace all usages of that class with the now recommended approach 

https://github.com/eclipse-platform/eclipse.platform/blob/1809821c284452c8ee496dd133f2c4b3a2996a7b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Preferences.java#L90-L93

akurtakov:
[...]

PR #505, opened 2023-06-17
Use equinox preferences APIs in AutoBuildJob class #497

alex2145:
Replace deprecated preferences APIs.
In the implementation of
IPreferenceChangeListener.preferenceChange(PreferenceChangeEvent) do not use the newValue field of the PreferenceChangeEvent object because it is null when autobuild is being turned on (this is caused by the method Preferences.setValue(String, boolean) removing the explicit setting when the value is equal to the default value).

Partially fixes

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
@@ -14,14 +14,26 @@
  *******************************************************************************/
 package org.eclipse.core.internal.events;
 
-import org.eclipse.core.internal.resources.*;
+import org.eclipse.core.internal.resources.ICoreConstants;
+import org.eclipse.core.internal.resources.ResourceException;
+import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
-import org.eclipse.core.resources.*;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.Preferences.PropertyChangeEvent;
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
```diff
@@ -14,22 +14,34 @@
...
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.osgi.framework.Bundle;
 
 /**
  * The job for performing workspace auto-builds, and pre- and post- autobuild
  * notification.  This job is run whenever the workspace changes regardless
  * of whether autobuild is on or off.
  */
-class AutoBuildJob extends Job implements Preferences.IPropertyChangeListener {
+class AutoBuildJob extends Job implements IEclipsePreferences.IPreferenceChangeListener {
```
jukzi, line 44: I highly appreciate cleanup of that deprecated stuff. Thanks!
If you like you could do it even better if AutoBuildJob does not implement anything but creates an private listener instance. That way users of AutoBuildJob could not recognize if the implementation changes like now.

jukzi (approved):

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/AutoBuildJob.java
alex2145: I don't like the idea of having to create another instance just to hide a listener method in an internal class, I prefer the straightforward approach which in my opinion is easier to read.
Users of `AutoBuildJob` should not expect that the class is an instance of `IEclipsePreferences.IPreferenceChangeListener` or invoke `preferenceChange(PreferenceChangeEvent)` directly.

PR #507, opened 2023-06-19
Fix random failing IJobManagerTest.testReverseOrder() #469

HeikoKlare:
The `IJobManagerTest.testReverseOrder()` randomly fails because it contains a race condition. It schedules multiple jobs with decreasing delays and expects them to finish in reverse order. If more time elapses between scheduling two jobs than their difference in delay, this expectation does not hold and the test fails.

This change reduces the test to its minimum: It only validates that two jobs with different delays can overtake, i.e., that a job with "current time + delay" being lower than the same value of a previously scheduled job, the later scheduled one will be executed first. The revised test uses high delays that will not lead to random fails in any considerable environment.

Fixes eclipse-platform/eclipse.platform#469.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
@@ -13,16 +13,39 @@
  *******************************************************************************/
 package org.eclipse.core.tests.runtime.jobs;
 
-import java.util.*;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.stream.Stream;
 import junit.framework.AssertionFailedError;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.jobs.*;
-import org.eclipse.core.tests.harness.*;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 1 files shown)


runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
```diff
@@ -1750,33 +1773,15 @@ protected IStatus run(IProgressMonitor monitor) {
...
-		}
-		Job[] doneOrder = done.toArray(new Job[done.size()]);
-		assertEquals("1.0", jobs.length, doneOrder.length);
-		for (int i = 0; i < doneOrder.length; i++) {
-			assertEquals("1.1." + i, jobs[i], doneOrder[doneOrder.length - 1 - i]);
-		}
+	public void testReverseOrder() throws InterruptedException {
+		// ensure that a job does not wait for one that is scheduled with a high delay
+		Job directlyExecutedJob = new TestJob("Directly executed job", 0, 0);
+		Job delayedJob = new TestJob("Delayed job", 0, 0);
```
jukzi, line 1779: that job should end before tests end to not interfere with other tests. cancel it for example

jukzi (changes requested):

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
HeikoKlare: Yes, thanks! The `cancel` call was missing. Fixed in 890973e.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/IJobManagerTest.java
```diff
@@ -1750,33 +1773,17 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		long delayInSeconds = 20;
+		delayedJob.schedule(delayInSeconds * 1000);
+		directlyExecutedJob.schedule(1);
+		directlyExecutedJob.join(delayInSeconds * 1000 / 2, null);
+		int delayedJobState = delayedJob.getState();
+		delayedJob.cancel();
+		assertThat("delayed job should still be waiting", delayedJobState, is(Job.SLEEPING));
```
jukzi, line 1786: assert before can

jukzi (approved):

Issue #472, opened 2023-06-01
Random failing of Bug_266907.test2ndSession()


fedejeanne:
The test failed during [this PR-Build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5134071947/jobs/9237538412?pr=466#step:7:19642) on May 31, 2023

```
junit.framework.AssertionFailedError: 1.0
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.core.tests.resources.session.Bug_266907.test2ndSession(Bug_266907.java:88)
...
```

fedejeanne:
I think this bug might be related with #460: both test classes extend from `WorkspaceSessionTest` and in both cases it is the 2nd test in the class that fails. I checked the logs in both cases I see this error message right after the execution of the 1st test in the class 

```
!MESSAGE The workspace will exit with unsaved changes in this session.
```

…and this subsequent message right after the 2nd test starts

```
!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.
```

Also, in both cases the 1st test in the class does not really `assert` anything, it just sets the workspace for the 2nd test in the class.

HeikoKlare:
The mentioned tests are `WorkspaceSessionTests`. For each test method, the workspace is reloaded. Thus, these tests do not necessarily assert something in each test method, because some of the methods may only be a setup/cleanup session of the workspace.

Closing the workspace with unsaved changes is not the essential problem here. This test deletes the `.project` file in the first session (so actually there are unsaved changes when closing the workspace), which is recovered while starting the second session. What is going wrong is covered by these entries:
```
!ENTRY org.eclipse.core.resources 2 2 2023-05-31 14:36:14.372
!MESSAGE Save operation warnings.
!SUBENTRY 1 org.eclipse.core.resources 2 234 2023-05-31 14:36:14.373
!MESSAGE The project description file (.project) for 'Project' was missing.  This file contains important information about the project.  A new project description file has been created, but some information about the project may have been lost.
```
For some reason, a save operation is executed after the `.project` file has been deleted. This save operation tries to recover the file, so that the second session can unexpectedly open the project.

[...]

PR #509, opened 2023-06-21
Fix random failing Bug_266907 #472

HeikoKlare:
The regression test for bug 266907 randomly fails. After deleting the `.project` file in the first workspace session of the test, an asynchronously executed save operation non-deterministically restores the `.project` file, which is assumed to be non-existing in the second workspace session.

With this change, the `JobManager` is suspended in the according workspace session to ensure that no asynchronous save operation restoring the `.project` file is executed. This is valid and reasonable since no job executions are required in the test and it prevents an auto-save job from being executed.

A refactoring improving the test class is provided in the subsequent PR #510.

Fixes eclipse-platform/eclipse.platform#472.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/Bug_266907.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/Bug_266907.java
@@ -25,6 +25,7 @@ import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.tests.resources.WorkspaceSessionTest;
 import org.eclipse.core.tests.session.WorkspaceSessionTestSuite;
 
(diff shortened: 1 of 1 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/Bug_266907.java
```diff
@@ -35,6 +36,10 @@ public static Test suite() {
 	}
 
 	public void test1stSession() {
+		// Ensure that no asynchronous save job restores the .project file after
+		// deleting it by suspending the JobManager for this workspace session
+		Job.getJobManager().suspend();
```
fedejeanne, line 41: It might be an overkill for this particular PR, but stopping the `JobManager` may even be desirable most (or even all) subclasses of `WorkspaceSessionTest`, right? 
I mean unless you're specifically testing the `JobManager` itself, it only gets in the way and runs unwanted logic between sessions (a.k.a. between test methods). Or am I seeing it wrong?

HeikoKlare: I agree. That might even be of interest for all `ResourceTests` to improve test isolation and reduce the risk of unintended side effects.
W.r.t. to `WorkspaceSessionTest`: The benefit of these tests in comparison to `ResourceTests` is that a new `JobManager` is started for every session. So there is at least not the chance of leaking a job execution triggered in one test case into the next one.

Issue #464, opened 2023-05-26
unknown saveable // IllegalStateException: Conflicting handlers ContentMergeViewer removeOnHide


jukzi:
Today eclipse came up with errors like these:
```
org.eclipse.ui
Error
Fri May 26 08:30:33 CEST 2023
Conflicting handlers for org.eclipse.compare.switchLeftAndRight: {ActionHandler(org.eclipse.compare.contentmergeviewer.ContentMergeViewer$2@38c4cf29)} vs {ActionHandler(org.eclipse.compare.contentmergeviewer.ContentMergeViewer$2@4fcd84e4)} in: PartImpl (org.eclipse.e4.ui.compatibility.editor)  removeOnHide org.eclipse.compare.CompareEditorContext

java.lang.IllegalStateException: Conflicting handlers: {[ActionHandler(org.eclipse.compare.contentmergeviewer.ContentMergeViewer$2@38c4cf29),  for 'org.eclipse.compare.switchLeftAndRight',  in PartImpl (org.eclipse.e4.ui.compatibility.editor)  removeOnHide org.eclipse.compare.CompareEditorContext, activeWhen=ActivePartExpression(org.eclipse.compare.internal.CompareEditor@9cc33f1), active=true, sourcePriority=1048576, participating=true]} vs {[ActionHandler(org.eclipse.compare.contentmergeviewer.ContentMergeViewer$2@4fcd84e4),  for 'org.eclipse.compare.switchLeftAndRight',  in PartImpl (org.eclipse.e4.ui.compatibility.editor)  removeOnHide org.eclipse.compare.CompareEditorContext, activeWhen=ActivePartExpression(org.eclipse.compare.internal.CompareEditor@9cc33f1), active=true, sourcePriority=1048576, participating=true]} in: PartImpl (org.eclipse.e4.ui.compatibility.editor)  removeOnHide org.eclipse.compare.CompareEditorContext
	at org.eclipse.ui.internal.handlers.LegacyHandlerService$HandlerSelectionFunction.compute(LegacyHandlerService.java:145)
	at org.eclipse.e4.core.internal.contexts.ValueComputation.get(ValueComputation.java:70)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.internalGet(EclipseContext.java:265)
	at org.eclipse.e4.core.internal.contexts.EclipseContext.get(EclipseContext.java:230)
[...]

PR #515, opened 2023-06-23
Revert "Compare: asyncExec propertyChange() to avoid Deadlock #282"

iloveeclipse:
This reverts commit ba2faa704d7a77baefe0d34c65c3a23ad8fbe580 as it causes severe regression with compare editor.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/464


merks:
@jukzi 

This change seems to be quite old and reported fixed a deadlock, and would have been in the release, right?

jukzi:
i committed on M1 two weeks ago.

PR #517, opened 2023-06-26
Allow .project to be linked resource

mickaelistria:


iloveeclipse:
@mickaelistria : please add a test case for such fundamental change. Also please check if all the cases that call the changed code are "happy" with the change (note that it might be called from project.open() call where the underlined project could not yet have known children so calls to project.getFile() would run into trouble).

mickaelistria:
I'm giving up with the idea of trying to run the whole Test Suite with the linked .project, it's very hard to ensure proper synchronization by reacting to various events, and multiple tests are actually designed to not work with linked resources (ie anything that uses the SIMPLE_NATURE). I've started a new Test class with basic tests.
The important matter is mostly to note that this change doesn't cause regression and works OK when changing the ProjectDescription.

Issue #519, opened 2023-06-27
java.lang.NullPointerException: Cannot invoke "org.eclipse.equinox.log.Logger.log(Object, int, String, java.lang.Throwable)"  on shutdown


laeubi:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.equinox.log.Logger.log(Object, int, String, java.lang.Throwable)" because "this.logger" is null
	at org.eclipse.core.internal.runtime.Log.log(Log.java:68)
	at org.eclipse.ui.internal.WorkbenchPlugin.log(WorkbenchPlugin.java:743)
	at org.eclipse.ui.statushandlers.StatusManager.logError(StatusManager.java:313)
	at org.eclipse.ui.statushandlers.StatusManager.handle(StatusManager.java:231)
	at org.eclipse.ui.internal.progress.ProgressManager$1.done(ProgressManager.java:468)
	at org.eclipse.core.internal.jobs.JobListeners.sendEvent(JobListeners.java:153)
	at org.eclipse.core.internal.jobs.JobListeners.sendEventsAsync(JobListeners.java:140)
	at org.eclipse.core.internal.jobs.JobListeners.waitAndSendEvents(JobListeners.java:80)
	at org.eclipse.core.internal.jobs.JobManager.withWriteLock(JobManager.java:524)
	at org.eclipse.core.internal.jobs.JobManager.endJob(JobManager.java:728)
	at org.eclipse.core.internal.jobs.WorkerPool.endJob(WorkerPool.java:117)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:83)
```

PR #520, opened 2023-06-27
Fix NPE in case system is shutting down

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform/issues/519

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/Log.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/Log.java
@@ -15,8 +15,16 @@ package org.eclipse.core.internal.runtime;
 
 import java.util.HashSet;
 import java.util.Set;
-import org.eclipse.core.runtime.*;
-import org.eclipse.equinox.log.*;
+import org.eclipse.core.runtime.ILog;
+import org.eclipse.core.runtime.ILogListener;
+import org.eclipse.core.runtime.ISafeRunnable;
+import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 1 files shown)


runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/Log.java
```diff
@@ -64,6 +72,10 @@ public Bundle getBundle() {
 	 */
 	@Override
 	public void log(final IStatus status) {
+		if (logger == null) {
+			// can happen on system shutdown...
```
iloveeclipse, line 76: Shouldn't we print out an error to System.err in that case?

laeubi: A log could be a warning or info status maybe so not sure if  `System.err` is always a good choice, but if you think it is useful I think  `System.err/out` could be an option at laest.

iloveeclipse: Yes, I do think it is useful, because if something crashes on shutdown and you don't have any trace of it, you can search forever.

tjwatson:
Using a `SeviceTracker` for the primordial service `org.eclipse.equinox.log.ExtendedLogService` which is always registered by the framework is a bit of an overkill and leads to issues once the `ServiceTracker` is closed on shutdown.  One option would be to stash the primordial `org.eclipse.equinox.log.ExtendedLogService` on `InternalPlatform.start` and not use a ServiceTracker and not null out the field with the `org.eclipse.equinox.log.ExtendedLogService` on shutdown either.  Just continue to use it.

PR #522, opened 2023-06-28
Provide static acessors in the interface as an alternative to Platform

laeubi:
Currently one cann add a global loglistenr via the Platform.add/removeLogListener(...)., but is is more semantic to only use ILog interface and also decreases the coupling of the code to a central singleton class.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
@@ -15,6 +15,7 @@
 package org.eclipse.core.runtime;
 
 import org.eclipse.core.internal.runtime.InternalPlatform;
+import org.eclipse.core.internal.runtime.RuntimeLog;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
 
@@ -174,5 +175,41 @@ public interface ILog {
 		} catch (IllegalCallerException e) {
 			return of(ILog.class);
 		}
+	}
+
+	/**
+	 * Adds a log listener that is notified about all log events of the runtime.
+	 * <p>
+	 * Once registered, a listener starts receiving notification as entries are
+	 * added via any <code>ILog.log()</code>. The listener continues to receive
+	 * notifications until it is replaced or removed.
+	 * </p>
+	 *
+	 * @param listener the listener to register
+	 * @see ILog#addLogListener(ILogListener)
+	 * @see #removeLogListener(ILogListener)
+	 * @since 3.31
+	 */
+	public static void addRuntimeLogListener(ILogListener listener) {
+		if (listener == null) {
+			return;
+		}
+		RuntimeLog.addLogListener(listener);
+	}
+
+	/**
+	 * Removes the log listener from receiving notfications about all log events of
+	 * the runtime. If no such listener exists, no action is taken.
+	 *
+	 * @param listener the listener to de-register
+	 * @see ILog#removeLogListener(ILogListener)
+	 * @see #addLogListener(ILogListener)
+	 * @since 3.31
+	 */
+	public static void removeRuntimeLogListener(ILogListener listener) {
+		if (listener == null) {
+			return;
+		}
+		RuntimeLog.removeLogListener(listener);
 	}
 }
--- a/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
-Bundle-Version: 3.30.100.qualifier
+Bundle-Version: 3.31.0.qualifier
 Bundle-SymbolicName: org.eclipse.core.runtime; singleton:=true
 Bundle-Vendor: %providerName
 Bundle-Activator: org.eclipse.core.internal.runtime.PlatformActivator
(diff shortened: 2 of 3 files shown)


HannesWell:
Wouldn't it be better to let the RuntimeLog implement ILog and provide the singleton in the ILog interface `static ILog getRuntimeLog()`?

Then one can add listeners and can Llog messages to it.

laeubi:
> Wouldn't it be better to let the RuntimeLog implement ILog and provide the singleton in the ILog interface `static ILog getRuntimeLog()`?

Better in what regards?

> Then one can add listeners and can Llog messages to it.

I'm not sure if that is actually desired, beside that `RuntimeLog` is an internal class of equinox.commons and can't implement `ILog` anyways so I think it might not be a good choice to expose more as required here, from a users point of view I think the main difference is to get a listener for a specific log or for all log messages.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
```diff
@@ -175,4 +176,34 @@ public static ILog get() {
 			return of(ILog.class);
 		}
 	}
+
+	/**
+	 * Adds the given log listener to the notification list of the platform.
```
HannesWell, line 181:
```suggestion
	 * Adds the given log listener to the notification list of the entire platform.
```

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
```diff
@@ -175,4 +176,34 @@ public static ILog get() {
...
+	 * <p>
+	 * Once registered, a listener starts receiving notification as entries are
+	 * added to plug-in logs via <code>ILog.log()</code>. The listener continues to
```
HannesWell, line 184:
```suggestion
	 * added to any plug-in log via <code>ILog.log()</code>. The listener continues to
```

HannesWell (commented):
> > Wouldn't it be better to let the RuntimeLog implement ILog and provide the singleton in the ILog interface `static ILog getRuntimeLog()`?
> 
> Better in what regards?

I assumed that there would also be a corresponding `logRuntime(IStatus)` method where users could log a status globally. Then a `getRuntimeLog()` would give access to that and would allow to also register listeners in a unified way, all with that method.

But it indeed looks like it is not intended to provide the log as public API to users, so providing those two methods is fine.
I just think that the documentation should be enhanced to emphasize that the listener will receive notifications from all logs of the platform, see the proposals for the add method.

laeubi:
We can for sure enhance the javadoc but keep in mind I just copied this form the Platform code and it is not something new :-)

HannesWell:
> We can for sure enhance the javadoc but keep in mind I just copied this form the Platform code and it is not something new :-)

I know. But I think this is a good opportunity to improve things a little bit. :)
Furthermore since this is now not in the Platform class anymore, I think it is less obvious from the context, that the listener is globally.

HannesWell (approved):
Look good, thanks.
`Runtime` instead of `Platform` is more generic and tied less to the Eclipse Platform.

mickaelistria (changes requested):
Please make commit message more explicit: "the interface" -> "ILog interface"

laeubi:
> Please make commit message more explicit: "the interface" -> "ILog interface"

done

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/ILog.java
```diff
@@ -175,4 +176,40 @@ public static ILog get() {
...
+	 * @param listener the listener to register
+	 * @see ILog#addLogListener(ILogListener)
+	 * @see #removeLogListener(ILogListener)
+	 * @since 3.31
+	 */
+	public static void addRuntimeLogListener(ILogListener listener) {
+		if (listener == null) {
+			return;
+		}
+		RuntimeLog.addLogListener(listener);
```
mickaelistria, line 197: I don't think adding method dereferencing a static helps a lot as it doesn't really make it clearer what happens nor does is make things much simpler. The "RuntimeLog" is a particular instance of the log, internal to Platform, I don't really get why ILog should have particular method to manipulate this particular instance.
Can you please add to this PR examples of usage where this API addition helps noticeably? What about adding a `public static ILog runtimeLog()` accessor if we often need it?

laeubi: The point is to replace` org.eclipse.core.runtime.Platform.addLogListener(ILogListener)` so people are not need to be tied to the "Platform" class an not to invent / provide really new / different things.

mickaelistria: OK, so then having a `public static ILog runtimeLog()` would work, wouldn't it?
My concern with those 2 methods is that they start to mix the responsibility of this interface: it used to be the representation of a Log, then we added public accessors for various Logs (that's a good pattern to provide most probable accessors in the interface, so it's fine too), but this particular methods would then add the responsibility of managing 1 particular Log instance, which seems less consistent.
If we can have users doing `ILog.runtimeLog().addLogListener()` that's probably good and clear enough; without mixing duties in the interface.

laeubi: I'm not sure if we can/should expose the runtime log, the intention seem to be making it only internally accessible but allowing other code to listen for events logged there.

laeubi:
I'm closing this as the "PlatformLog" is not really an `ILog` but something special but using only the interface for log messages, also these methods are mostly used in tests except one place in platform ui where it migth be better to replace it by a direct call.

PR #562, opened 2023-07-04
CharsetManager#setCharsetFor: split function into private subfunctions

Wittmaxi:
This PR refactors a single method of the CharsetManager.

### Purpose
I am preparing to enable a disabled unit test which touches CharsetManager. In order to "get my feet wet", I decided to refactor this method.

### ~Changes in behaviour~
~In case the Name `resourcePath` passed to the method is not a valid resource path, the method currently performed no operation. I introduced an exception that throws `ResourceException` for this case. This might *not* be expected behaviour, I did not find any documentation detailing this. It is possible that other parts rely on `setCharsetFor` to silently fail on invalid resources. I am open for feedback on this!~
I removed all changes in behaviour and made this PR a pure refactoring-PR, preserving behaviour

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
@@ -489,9 +489,8 @@ public class CharsetManager implements IManager {
 	}
 
 	private void setCharsetForHasFailed(IPath resourcePath, BackingStoreException e) throws ResourceException {
-		IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
 		String message = Messages.resources_savingEncoding;
-		throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
+		throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, resourcePath, message, e);
 	}
 
 	@Override
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
@@ -440,43 +440,58 @@ public class CharsetManager implements IManager {
 	}
 
 	public void setCharsetFor(IPath resourcePath, String newCharset) throws CoreException {
-		// for the workspace root we just set a preference in the instance scope
-		if (resourcePath.segmentCount() == 0) {
-			IEclipsePreferences resourcesPreferences = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES);
-			if (newCharset != null)
-				resourcesPreferences.put(ResourcesPlugin.PREF_ENCODING, newCharset);
-			else
-				resourcesPreferences.remove(ResourcesPlugin.PREF_ENCODING);
-			try {
-				resourcesPreferences.flush();
-			} catch (BackingStoreException e) {
-				IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
-				String message = Messages.resources_savingEncoding;
-				throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
-			}
-			return;
+		if (IPath.ROOT.equals(resourcePath)) {
+			setCharsetForRoot(newCharset);
+		} else if (workspace.getRoot().findMember(resourcePath) instanceof Resource resource) {
+			setCharsetForResource(resourcePath, newCharset, resource);
 		}
-		// for all other cases, we set a property in the corresponding project
-		IResource resource = workspace.getRoot().findMember(resourcePath);
-		if (resource != null) {
-			try {
-				// disable the listener so we don't react to changes made by ourselves
-				Preferences encodingSettings = getPreferences(resource.getProject(), true, resource.isDerived(IResource.CHECK_ANCESTORS));
-				if (newCharset == null || newCharset.trim().length() == 0)
-					encodingSettings.remove(getKeyFor(resourcePath));
-				else
-					encodingSettings.put(getKeyFor(resourcePath), newCharset);
-				flushPreferences(encodingSettings, true);
-				if (resource instanceof IProject) {
-					IProject project = (IProject) resource;
-					ValidateProjectEncoding.scheduleProjectValidation(workspace, project);
-				}
-			} catch (BackingStoreException e) {
-				IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
-				String message = Messages.resources_savingEncoding;
-				throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
-			}
+	}
+
+	private void setCharsetForRoot(String newCharset) throws ResourceException {
+		IEclipsePreferences resourcesPreferences = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES);
+		if (newCharset != null) {
+			resourcesPreferences.put(ResourcesPlugin.PREF_ENCODING, newCharset);
+		} else {
+			resourcesPreferences.remove(ResourcesPlugin.PREF_ENCODING);
 		}
+
+		try {
+			resourcesPreferences.flush();
+		} catch (BackingStoreException e) {
+			setCharsetForHasFailed(IPath.ROOT, e);
+		}
+	}
+
+	private void setCharsetForResource(IPath resourcePath, String newCharset, IResource resource)
+			throws ResourceException {
+		try {
+			setResourceEncodingSettings(resourcePath, newCharset, resource);
+			if (resource instanceof Project project) {
+				ValidateProjectEncoding.scheduleProjectValidation(workspace, project);
+			}
+		} catch (BackingStoreException e) {
+			setCharsetForHasFailed(resourcePath, e);
+		}
+	}
+
+	private void setResourceEncodingSettings(IPath resourcePath, String newCharset, IResource resource)
+			throws BackingStoreException {
+		// disable the listener so we don't react to changes made by ourselves
+		Preferences encodingSettings = getPreferences(resource.getProject(), true,
+				resource.isDerived(IResource.CHECK_ANCESTORS));
+
+		if (newCharset == null || newCharset.isBlank()) {
+			encodingSettings.remove(getKeyFor(resourcePath));
+		} else {
+			encodingSettings.put(getKeyFor(resourcePath), newCharset);
+		}
+		flushPreferences(encodingSettings, true);
+	}
+
+	private void setCharsetForHasFailed(IPath resourcePath, BackingStoreException e) throws ResourceException {
+		IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
+		String message = Messages.resources_savingEncoding;
+		throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
 	}
 
 	@Override


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
-		if (resourcePath.segmentCount() == 0) {
-			IEclipsePreferences resourcesPreferences = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES);
-			if (newCharset != null)
-				resourcesPreferences.put(ResourcesPlugin.PREF_ENCODING, newCharset);
-			else
-				resourcesPreferences.remove(ResourcesPlugin.PREF_ENCODING);
-			try {
-				resourcesPreferences.flush();
-			} catch (BackingStoreException e) {
-				IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
```
mickaelistria, line 435: This was actually an ArrayIndexOutOfBoundsException

mickaelistria (commented):
> This might not be expected behaviour, I did not find any documentation detailing this. 

In case of doubt, let's try to avoid changing the behavior unless it is the cause of some existing bug.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
Wittmaxi: "Was" as in "in the old code"? Does the exception still persist?

iloveeclipse:
@Wittmaxi : could you please avoid any merge commits & provide **one** commit with the proposed change? 

Also could you explain what kind of issue you are trying to fix here? If you have an example where the code fails, please provide stack trace, and ideally try to provide a test case for this case.

Wittmaxi:
@iloveeclipse Is the best procedure for me to close this PR and to reopen a new one with all of my commits squashed into a single commit?

This PR has two purposes:
- I want to learn how contributions to eclipse work (first PR! 🥳)
- I want to refactor a method that is touched by a currently [disabled unit test](https://github.com/eclipse-platform/eclipse.platform/blob/77ba2f39aa490ed002fdf72a460482c007198b0c/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/CharsetTest.java#L177) I am working on enabling

mickaelistria:
>  Is the best procedure for me to close this PR and to reopen a new one with all of my commits squashed into a single commit?

The recommended procedure is that you reorganize (squash, rebase...) your commits locally and `push --force` to your `MW_refactor_CharsetManager` branch to update this PR.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
+				resource.isDerived(IResource.CHECK_ANCESTORS));
+		if (isCharsetNameBlank(newCharset))
+			encodingSettings.remove(getKeyFor(resourcePath));
+		else
+			encodingSettings.put(getKeyFor(resourcePath), newCharset);
+		flushPreferences(encodingSettings, true);
+	}
+
+	private boolean isCharsetNameBlank(String newCharset) {
+		return newCharset == null || newCharset.trim().length() == 0;
```
mickaelistria, line 501:
```sugestion
		return newCharset == null || newCharset.isBlank();
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
-				String message = Messages.resources_savingEncoding;
-				throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
+	}
+
+	private void setCharsetForResource(IPath resourcePath, String newCharset, IResource resource)
+			throws ResourceException {
+		try {
+			setResourceEncodingSettings(resourcePath, newCharset, resource);
+			if (resource instanceof IProject) {
+				IProject project = (IProject) resource;
```
mickaelistria:
```suggestion
			if (resource instanceof IProject project) {
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
 
+	private IResource getResourceFromPath(IPath resourcePath) throws ResourceException {
+		IResource resource = workspace.getRoot().findMember(resourcePath);
+		if (resource == null)
+			throw new ResourceException(null);
+		return resource;
+	}
+
+	private boolean isResourceRoot(IPath resourcePath) {
+		return resourcePath.segmentCount() == 0;
```
mickaelistria, line 484:
```suggestion
		return IPath.ROOT.equals(resourcePath);
```
which is explicit enough to inline it directly without intermediary method.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
 			}
+		} catch (BackingStoreException e) {
+			setCharsetForHasFailed(resourcePath, e);
 		}
 	}
 
+	private IResource getResourceFromPath(IPath resourcePath) throws ResourceException {
+		IResource resource = workspace.getRoot().findMember(resourcePath);
+		if (resource == null)
+			throw new ResourceException(null);
```
mickaelistria:
```suggestion
			throw new ResourceException("No resource found for path " + resourcePath);
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
-			try {
-				resourcesPreferences.flush();
-			} catch (BackingStoreException e) {
-				IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
-				String message = Messages.resources_savingEncoding;
-				throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
-			}
-			return;
+	public void setCharsetFor(IPath resourcePath, String newCharset) throws CoreException, ResourceException {
+		if (isResourceRoot(resourcePath))
```
mickaelistria: We usually recommend always adding `{ ... }` as it makes it safer than indentation what is actually included or not.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -421,46 +440,74 @@ public void projectPreferencesChanged(IProject project) {
...
+			if (resource instanceof IProject) {
+				IProject project = (IProject) resource;
+				ValidateProjectEncoding.scheduleProjectValidation(workspace, project);
 			}
+		} catch (BackingStoreException e) {
+			setCharsetForHasFailed(resourcePath, e);
 		}
 	}
 
+	private IResource getResourceFromPath(IPath resourcePath) throws ResourceException {
```
mickaelistria: As this method only has 1 consumer and is trivial to understand (only delegating to `workspace.getRoot().findMember()`, it would IMO be easier to inline and remove it.

mickaelistria (changes requested):
I've added some code suggestions inline.
Splitting setCharsetFor may make sense, extracting reusable bits that are used by both cases too. But in general, many small methods are not preferred over bigger ones as they require many jumps to follow the flow of execution. Methods are welcome sense when they have multiple consumers, when they do a non-trivial or specialized job, if we want to implement unit-testing... but here I don't see a compelling reason to go too small grain; I would even claim that it makes you work of reviving some CharsetTests harder.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
Wittmaxi: Good point! This one will not be part of my final commit anyway, since it introduces new behaviour which we decided on not doing for this commit.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
Wittmaxi: In a newer version of my code, the function is used like this:

```java	
public void setCharsetFor(IPath resourcePath, String newCharset) throws CoreException, ResourceException {
		if (IPath.ROOT.equals(resourcePath)) {
			setCharsetForRoot(resourcePath, newCharset);
		} else if (getResourceFromPath(resourcePath) instanceof Project project) {
			setCharsetForResource(resourcePath, newCharset, project);
		}
}```

I would argue that inlining this behaviour is not readily possible in this `if (xyz instanceof Project project)` without creating a very long line of code. Am I missing something?

Comment (mickaelistria) [reply resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java]: Depends on what you call very long. To me, as long as I can read it on my screen without scrolling and the line is constistent (not doing too much as once), it's fine.

File: resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java (hunk 1 above)
Comment (Wittmaxi) [reply resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java:453]: Is the right procedure for me to open an issue?

**Comment by Wittmaxi** [comment]
Thank you for your reviews, I have addressed them. I am open to more feedback, if any issues still persist. @mickaelistria

**Review by HeikoKlare — APPROVED** [review]
I am in favor of merging the proposed changes. The code becomes much more comprehensible since coherent chunks of functionality are encapsulated into self-describing methods. This conforms to state-of-the-art clean code principles.

I know that it is also a matter of personal preference what is considered "clean code" and to what extent clean code principles may be adopted, but in this case I only see benefits in performing the proposed refactoring.

File: resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java [hunk 8]
```diff
@@ -440,45 +440,60 @@ public void projectPreferencesChanged(IProject project) {
...
-					IProject project = (IProject) resource;
-					ValidateProjectEncoding.scheduleProjectValidation(workspace, project);
-				}
-			} catch (BackingStoreException e) {
-				IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
-				String message = Messages.resources_savingEncoding;
-				throw new ResourceException(IResourceStatus.FAILED_SETTING_CHARSET, project.getFullPath(), message, e);
+	}
+
+	private void setCharsetForRoot(IPath resourcePath, String newCharset) throws ResourceException {
```
mickaelistria, line 450: No need to keep resourcePath then.
```
	private void setCharsetForRoot(String newCharset) throws ResourceException {
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -440,45 +440,60 @@ public void projectPreferencesChanged(IProject project) {
...
+		if (newCharset == null || newCharset.isBlank()) {
+			encodingSettings.remove(getKeyFor(resourcePath));
+		} else {
+			encodingSettings.put(getKeyFor(resourcePath), newCharset);
+		}
+		flushPreferences(encodingSettings, true);
+	}
+
+	private void setCharsetForHasFailed(IPath resourcePath, BackingStoreException e) throws ResourceException {
+		IProject project = workspace.getRoot().getProject(resourcePath.segment(0));
```
mickaelistria, line 492: I resourcePath is root, the `segment(0)` will return `null`;  and `getProject` expects "The given name must be a valid path segment as defined by {@link IPath#isValidSegment(String)}." which is not the case. So this is likely to drive to some error.
I suggest the exception just contain the resourcePath, without trying to resolve it to a project.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -440,45 +440,60 @@ public void projectPreferencesChanged(IProject project) {
...
+			setResourceEncodingSettings(resourcePath, newCharset, resource);
+			if (resource instanceof Project project) {
+				ValidateProjectEncoding.scheduleProjectValidation(workspace, project);
 			}
+		} catch (BackingStoreException e) {
+			setCharsetForHasFailed(resourcePath, e);
 		}
 	}
 
+	private void setResourceEncodingSettings(IPath resourcePath, String newCharset, IResource resource)
```
mickaelistria, line 477: I don't see much value in having these 8 lines extracted out here if the only consumer is a 3 lines method.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
HeikoKlare: Your point is absolutely valid. Since this is pre-existing behavior, I would propose to have a separate PR fixing the behavior and not combine it with the pure refactoring of this PR (see https://github.com/eclipse-platform/eclipse.platform/pull/562#discussion_r1258181011).

Wittmaxi: @mickaelistria The goal of this PR is not to change behavior but to refactor an existing function. I am fine with adding this behavior - I am not sure how having a separate commit is reconcileable with our aim to always have a single commit in each PR. Is such a case an exception to the rules?

HeikoKlare: Having the fix as a separate commit without creating another PR is absolutely fine for me.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
Wittmaxi: The whole refactoring is a [functioncal decomposition](https://en.wikipedia.org/wiki/Functional_decomposition) of setCharsetFor, and so is this function. My goal is for somebody completely new to this unit to be able to read it "newspaper"-style without having to distill the essence of the code.

I am wondering, is there any harm to having this extraction?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
mickaelistria: There is no such requirementof having 1 commit per PR; the goal is to keep things efficient to develop and review. Depending on the change, that can mean different way of doing PRs/commits. Here I believe that an extra commit on this PR is the most efficient, but if you disagree, that's fine, it will be another PR later then.
Note that it's usually more critical to fix bugs than to refactor. Fixing a bug "by the way" is a good way to get refactorings merged in promptly ;)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
mickaelistria: Not much harm, but I just don't buy blindly the argument that more methods is making things clearer and easier to understand. It requires jumps to read and understand things, so it cost cognitive complexity, it adds symbols to the .class (making it slightly bigger)... I know there are dogmas of "no method should be more than X lines", but I've grown too old and grumpy to believe that. Code can be expressive without being scattered.
But it's more for the sake of discussion, not a blocker against merging this PR.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
Wittmaxi: No, from my side there is no need to be dogmatic! I will add a separate commit :)

Wittmaxi:
As discussed [here](https://github.com/eclipse-platform/eclipse.platform/pull/562#discussion_r1258390903), I added a second commit for a bugfix that I tackled during my work on refactoring the function.

The bugfix addresses a possibly faulty exception-message.

mickaelistria (approved):

Wittmaxi:
Thank you for the comment, the commit messages are more descriptive now!

PR #565, opened 2023-07-06
Remove 'Eclipse-LazyStart' headers in favor of 'Bundle-ActivationPolicy'

HannesWell:
... and remove unnecessary 'Bundle-ClassPath: .' entries.

Replace the use of the legacy and Eclipse-specific MANIFEST.MF header 'Eclipse-LazyStart: true' by the OSGi standard compliant header 'Bundle-ActivationPolicy: lazy' (which is already used in most cases.

The default value of 'Bundle-ClassPath' is the dot which is used when the header is not specified. Therefore there is no need to specify it with that value.

### DIFF
--- a/ant/org.eclipse.ant.core/META-INF/MANIFEST.MF
+++ b/ant/org.eclipse.ant.core/META-INF/MANIFEST.MF
@@ -18,5 +18,4 @@ Require-Bundle: org.eclipse.core.variables;bundle-version="[3.1.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.28.0,4.0.0)"
 Bundle-ActivationPolicy: lazy;exclude:="org.eclipse.ant.internal.core.contentDescriber"
 Bundle-RequiredExecutionEnvironment: JavaSE-17
-Bundle-ClassPath: .
 Automatic-Module-Name: org.eclipse.ant.core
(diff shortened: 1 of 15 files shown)


Bananeweizen:
@HannesWell Any chance you have a pointer to some good reading or a rule of thumb for which kind of classes the exclude attribute of the lazy activation policy is useful? I've just noticed that the first time in this PR.

tjwatson:
> @HannesWell Any chance you have a pointer to some good reading or a rule of thumb for which kind of classes the exclude attribute of the lazy activation policy is useful? I've just noticed that the first time in this PR.

To be honest, I think the exclude directive is a case of a bit too much over-engineering.  The idea was that there may be some subset of the bundle that did not require a `BundleContext` and could be used as a "pure" Java library without needing to be activated.  You could use the `exclude` directive for such packages within a bundle to prevent the bundle from being activated when classes are loaded loaded and used from the excluded package. I think the cases where this is useful are very small.

HannesWell:
> @HannesWell Any chance you have a pointer to some good reading or a rule of thumb for which kind of classes the exclude attribute of the lazy activation policy is useful? I've just noticed that the first time in this PR.

I cannot add much, only that I have not used it before and did not even know that this is possible before I encountered it in this repo. Just in case you are interested in the details, it is described in the OSGi chapter [4.4.6.2 Lazy Activation Policy](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.lifecycle.html#i3270445), but it is relatively straight forward.

HannesWell:
Test failures seem unrelated, merging.

Issue #561, opened 2023-07-04
Failure of LinkTest.testAllLinks()


fedejeanne:
The test failed in [this run](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5452492617/jobs/9920076530?pr=560) of the PR #560 

```
java.lang.IllegalArgumentException: Illegal character in scheme name at index 0: %20http://wiki.eclipse.org/Execution_Environment_Descriptions
java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Illegal character in scheme name at index 0: %20http://wiki.eclipse.org/Execution_Environment_Descriptions
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:562)
	at java.base/java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:591)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:689)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:159)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateParallel(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:765)
	at org.eclipse.ua.tests.doc.internal.linkchecker.LinkTest.testAllLinks(LinkTest.java:82)
...
```

fedejeanne:
Failed again in #563

fedejeanne:
Does anyone know how to test this locally?

[...]

PR #566, opened 2023-07-06
Minor simplification of LinkTest

HannesWell:
And investigate why the test is failing.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/561

### DIFF
--- a/ua/org.eclipse.ua.tests.doc/src/org/eclipse/ua/tests/doc/internal/linkchecker/LinkTest.java
+++ b/ua/org.eclipse.ua.tests.doc/src/org/eclipse/ua/tests/doc/internal/linkchecker/LinkTest.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
@@ -101,17 +103,14 @@ public class LinkTest {
 		assertEquals(Collections.emptySet().toString(), linkFailures.toString());
 	}
 
+	private static final Pattern HREF = Pattern.compile("<a href=\"([^\"]+)\"");
+
 	private Set<String> checkLinks(InputStream stream, URI currentDoc, Set<URI> knownPagesURIs) throws IOException {
 		Set<String> res = new HashSet<>();
 		try (BufferedReader in = new BufferedReader(new InputStreamReader(stream))) {
-			String inputLine;
-			while ((inputLine = in.readLine()) != null) {
-				int index = 0;
-				while ((index = inputLine.indexOf("<a href=\"", index)) > 0) {
-					int closeIndex = inputLine.indexOf('"', index + "<a href=\"".length());
-					URI href = URI
-							.create(inputLine.substring(index + "<a href=\"".length(), closeIndex).replace(" ", "%20"));
-					index = closeIndex;
+			for (String inputLine; (inputLine = in.readLine()) != null;) {
+				for (Matcher matcher = HREF.matcher(inputLine); matcher.find();) {
+					URI href = URI.create(matcher.group(1).replace(" ", "%20"));
 					if (href.isAbsolute()) {
 						continue;
 					}


HannesWell:
The offending link is probably on `<help-URL>/help/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_help_index.html`
My guess is that the failure of the LinkTest is caused by a space in some Extension-Point documentation in JDT.
Just created https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/278 to hopeful fix it.

The help-sites are
- https://help.eclipse.org/2023-06/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_jdt_launching_vmInstalls.html&cp%3D3_1_1_20
- https://help.eclipse.org/2023-06/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_jdt_launching_vmInstallTypes.html&cp%3D3_1_1_21

HannesWell:
Even after https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/278 is merged and one I-build has passed the space in beginning is still found.
Stripping the string solves that problem and lets that test pass, but of course it would swallow similar problems in the future.

Does anybody know from which repo the `IWorkbenchHelpSystem` returned by `PlatformUI.getWorkbench().getHelpSystem()` is feed? Do I have to wait for another I-build, for the next milestone or even next release to get the updated version?

Bananeweizen (approved):
I suggest merging your change and not trying to find the original HTML document. AFAIK blanks around URIs are valid in HTML: https://www.w3.org/TR/2014/REC-html5-20141028/infrastructure.html#valid-url-potentially-surrounded-by-spaces

Bananeweizen:
If I get it right, the source containing the blank has already been fixed in https://github.com/eclipse-platform/eclipse.platform/issues/561.

merks:
> If I get it right, the source containing the blank has already been fixed in #561.

Yes that's right.  We did track down the offending files.

ua/org.eclipse.ua.tests.doc/src/org/eclipse/ua/tests/doc/internal/linkchecker/LinkTest.java
```diff
@@ -101,17 +103,14 @@ public void addHits(List<SearchHit> hits, String wordsSearched) {
...
-			while ((inputLine = in.readLine()) != null) {
-				int index = 0;
-				while ((index = inputLine.indexOf("<a href=\"", index)) > 0) {
-					int closeIndex = inputLine.indexOf('"', index + "<a href=\"".length());
-					URI href = URI
-							.create(inputLine.substring(index + "<a href=\"".length(), closeIndex).replace(" ", "%20"));
-					index = closeIndex;
+			for (String inputLine; (inputLine = in.readLine()) != null;) {
+				for (Matcher matcher = HREF.matcher(inputLine); matcher.find();) {
+					URI href = URI.create(matcher.group(1).replace(" ", "%20"));
```
Bananeweizen, line 113: Sorry @HannesWell. I misread your code during first review. We still need to trim the regex result here to deal with URIs starting/ending with blanks.

```suggestion
					URI href = URI.create(matcher.group(1).strip().replace(" ", "%20"));
```

Bananeweizen (changes requested):

HannesWell:
> I suggest merging your change and not trying to find the original HTML document. AFAIK blanks around URIs are valid in HTML: https://www.w3.org/TR/2014/REC-html5-20141028/infrastructure.html#valid-url-potentially-surrounded-by-spaces

Thanks Michael for pointing this out, but unfortunately the help still uses html-4 as mentioned in https://github.com/eclipse-platform/eclipse.platform/issues/561#issuecomment-1627709892. 😞

But I wonder why my changes in the Extension-Point schema are not picked up in an I-build when building the doc?
It looks like the doc is build in the I-builds: 
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/92d943bcd4eeba3187e64fc4f154ab029e234364/eclipse.platform.common/bundles/org.eclipse.jdt.doc.isv/buildDoc.xml#L33

Therefore I assumed the change fix is now included, but I still se it in my latest I-build SDK.

Bananeweizen:
> Thanks Michael for pointing this out, but unfortunately the help still uses html-4 as mentioned in [#561 (comment)](https://github.com/eclipse-platform/eclipse.platform/issues/561#issuecomment-1627709892). 😞

I'm absolutely not an expert on those standards, but looking at the differences between 4 and 5, the href attribute has no relevant changes: https://www.w3.org/TR/html5-diff/#changed-attributes. Therefore URIs starting/ending with blanks should be valid there, too.

Bananeweizen (approved):
Seems my knowledge of the standard is not sufficient enough, so I'm removing my veto.

HannesWell:
> > Thanks Michael for pointing this out, but unfortunately the help still uses html-4 as mentioned in [#561 (comment)](https://github.com/eclipse-platform/eclipse.platform/issues/561#issuecomment-1627709892). 😞
> 
> I'm absolutely not an expert on those standards, but looking at the differences between 4 and 5, the href attribute has no relevant changes: https://www.w3.org/TR/html5-diff/#changed-attributes. Therefore URIs starting/ending with blanks should be valid there, too.

Me neither. But in doubt, I would prefer to be stricter in this case, just to be sure.
And with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1209 the fix in the jdt-doc should be picked up in tonights I-build and the failures should be gone.

HannesWell:
Test is green now.

PR #568, opened 2023-07-07
No need to check freeze period twice

akurtakov:
The shared action already does it so the standalone run is removed.


akurtakov:
No code change and only one action for freeze period run thus merging.

PR #569, opened 2023-07-08
gitignore Ant output folder

Bananeweizen:
In a newly created Oomph setup that output folder contains changes otherwise.


HannesWell (approved):
Yes that one is annoying. Thanks.

PR #571, opened 2023-07-12
Fixed Unit Test for Bug 67606

Wittmaxi:
## The unit-test was faulty for two reasons:
- It checked for the charset of the wrong file
- It ran as an ICoreRunnable in the Workspace context, which meant that
  notifications sent to a CharsetManager could not be seen from within

## Possible problems
- I am not sure if the CharsetManager receives the notification about the file-move in an asynchronous way. If it does, this change introduces a test that might be subject to race conditions. The race conditions would have existed in the test before, if it were enabled.
- I cannot tell precisely why running this Test synchronously makes it work
- I am confident that the [bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=67606) addressed by this test was fixed. However, a tangential bug was not:

## [Bug 3880114](https://bugs.eclipse.org/bugs/show_bug.cgi?id=3880114)

!! these bug reports are NOT duplicates of eachother

When moving files accross two projects with different charsets, we expect the charset of the file to be "sticky", to avoid data corruption. This is currently *not* the case.
I will create another PR introducing a disabled unit test which tests for this behaviour and move on to finding a solution to this faulty behaviour

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
@@ -18,11 +18,30 @@
  *******************************************************************************/
 package org.eclipse.core.internal.resources;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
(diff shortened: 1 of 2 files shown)


Wittmaxi:
Following Discussions with @HeikoKlare, I have decided not to tackle this Bug.

1) My fix for the unit test does not reflect the initial bug anymore
2) The Bug lies deep in the logic of how local settings are broadcast and notified
3) The expected behaviour is not clear enough
4) I could not find any concrete workflow inside of the eclipse IDE which breaks because of this bug

PR #572, opened 2023-07-12
WIP: Test Bug #380114

Wittmaxi:
Unit-Test addressing [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=380114)
This tests what happens if you move a file between projects that have different charset-settings.
We expect the charset to be sticky to avoid corrupting data.

This test needs to be enabled and fixed. This will happen in a subsequent PR, in TDD-style.

## Problems
- I am unsure if it is good to create a new PR for the fix or if it is better to fix the issue in this PR
- I am unsure if changing the behaviour of the file-move accross projects has possible harmful side-effects


jukzi:
Hello Wittmaxi, thanks for contributing. Please enable the test within this commit (squash) to show that it is failing in CI and then add another commit to this PR which is supposed to fix the bug and does not fail the test anymore.

PR #573, opened 2023-07-13
Use equinox preferences APIs in FileSystemResourceManager class #497

alex2145:
Replace deprecated preferences APIs.

Partially fixes


jukzi:
When is the code used? - How could somebody test the change?

alex2145:
Modified code is executed during `FileSystemResourceManager` startup & shutdown and when the value of the preference `refresh.lightweight.enabled` of the `ResourcesPlugin` changes (Window -> Preferences -> General -> Workspace -> Refresh on access).

In my commit there's a Junit test case in case you didn't notice.

jukzi:
works. @alex2145 thanks for contributing

PR #574, opened 2023-07-13
Add command line history to IOConsole/ProcessConsole.

chgenly:
Once enabled, ctrl-up and ctrl-down can be used to move through history.

To enable history use an IOConsolePageParticipant.  When the console is created, Eclipse will create a page book view page, finally the IConolePageParticpant will be created where history can be enabled.

	public class TestsConsolePageParticipant
		implements IConsolePageParticipant {

		@Override
		public void init(IPageBookViewPage page, IConsole console) {
			...
			if (page instanceof IIOConsolePage iocPage) {
				iocPage.setEnableCommandLineHistory(true);

--
IIOConsolePage has been added to expose the setEnableCommandLineHistory method for use by IConsolePageParticipants. IOConsolePage implements IOConsolePage.

IOConsoleViewer has been modified to move the cursor to the end of the line when command line history is enabled.

A unit test for command line history has been added in IOConsoleTests.testHistory()

### DIFF
--- a/debug/org.eclipse.debug.tests/plugin.xml
+++ b/debug/org.eclipse.debug.tests/plugin.xml
@@ -163,4 +163,10 @@
          priority="-1"
          class="org.eclipse.debug.tests.ui.TestVariableValueEditor3"/>
    </extension>
-</plugin>
+   <extension
+         point="org.eclipse.ui.console.consolePageParticipants">
+      <consolePageParticipant
+            class="org.eclipse.debug.tests.TestsConsolePageParticipant"
+            id="org.eclipse.debug.tests.consolePageParticipant">
+      </consolePageParticipant>
+   </extension></plugin>
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/TestsConsolePageParticipant.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/TestsConsolePageParticipant.java
@@ -1,0 +1,39 @@
+package org.eclipse.debug.tests;
+
+import org.eclipse.ui.console.IConsole;
+import org.eclipse.ui.console.IConsolePageParticipant;
+import org.eclipse.ui.console.IIOConsolePage;
+import org.eclipse.ui.console.TextConsole;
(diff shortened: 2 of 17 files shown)


jukzi:
if it is any good, why is it disabled by default? i.e. what is the drawback? Will there be UI support to select an old command from a list?

chgenly:
I think it's good.  It's disabled by default because it changes cursor
behavior.  Which causes unit tests to fail.  I didn't find any
documentation on what the intent of the current cursor behavior was.  So I
opted to not mess with it.

There is no user interface for showing a list and selecting a command from
the history. Just the ability to scroll up and down through history.

It's been a long process, and I haven't heard anything from the reviewer.
So I'd like to get this through before adding more features.

jukzi:
I could review, but i don't see a benefit if the IDE user does not have any benefit from it.

vogella:
You could just change the preference for the tests and enable it for the
IDE by default. Lots of tests do this

Jörg Kubitz ***@***.***> schrieb am Di., 25. Juli 2023, 19:20:

chgenly:
Interesting approach.  I would very much like to enable this feature by
default.  I work on it.

chgenly:
Is there anything I have to do to move this along?

jukzi:
> Is there anything I have to do to move this along?

Fix failing tests and wait for a committer to show interest in your change, if any. recommondation: show the benefit for the community.

chgenly:
Yes, of course, I think I have time to work on the failures tomorrow.  I'm
surprised there were failures.  I guess I didn't run enough of the tests.
I didn't run the whole suite, just all of what I thought were the console
tests.

laeubi:
I think some failures are not because the change but with recent adaption of the target platform, so squash/rebase should help here already.

chgenly:
I've been paying attention to various consoles and have noticed that many command line apps that run in a console support history.  However, none of the apps at my company support history, thus the need for the console to handle it.  Therefore it doesn't make sense to enable history by default. I will remove the changes from my branch which turn on history by default.

chgenly:
Closing:
- Better solution found.  Use a terminal instead of a console.
- Coherent Logix has closed.

Issue #576, opened 2023-07-20
Performance E4 Injector


N1k145:
We have a case where we have around 400k Injected Objects in one of the _injectedObjects_ lists in the E4 Injector.

Injecting those takes about 500 seconds, which is way too long. The main time is lost iterating the list of injected objects and checking if the object is already contained.

merks:
@N1k145 

Thanks for the improvement and for being so receptive to suggestions.

N1k145:
@merks Thank you for the in-depth review and the suggestions.

PR #577, opened 2023-07-20
Performance E4 Injector

N1k145:
fixes: https://github.com/eclipse-platform/eclipse.platform/issues/576

This fix reduces the time to inject 400k from 554s to 47s

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
+++ b/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
@@ -32,6 +32,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -74,8 +75,8 @@ public class InjectorImpl implements IInjector {
 	private final static Short DEFAULT_SHORT = Short.valueOf((short) 0);
 	private final static Byte DEFAULT_BYTE = Byte.valueOf((byte) 0);
 
-	private final Map<PrimaryObjectSupplier, List<WeakReference<?>>> injectedObjects = new WeakHashMap<>();
-	private final Set<WeakReference<Class<?>>> injectedClasses = new HashSet<>();
+	private final Map<PrimaryObjectSupplier, Set<IdentityWeakReference<?>>> injectedObjects = new WeakHashMap<>();
+	private final Set<IdentityWeakReference<Class<?>>> injectedClasses = new LinkedHashSet<>();
 	private final HashMap<Class<?>, Object> singletonCache = new HashMap<>();
 	private final Map<Class<?>, Set<Binding>> bindings = new HashMap<>();
 	private final Map<Class<? extends Annotation>, Map<AnnotatedElement, Boolean>> annotationsPresent = new HashMap<>();
@@ -148,38 +149,28 @@ public class InjectorImpl implements IInjector {
 
 	private void rememberInjectedObject(Object object, PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
-			List<WeakReference<?>> list = injectedObjects.computeIfAbsent(objectSupplier, k -> new ArrayList<>());
-			for (WeakReference<?> ref : list) {
-				if (object == ref.get())
-					return; // we already have it
-			}
-			list.add(new WeakReference<>(object));
+			injectedObjects.computeIfAbsent(objectSupplier,
+					k -> new LinkedHashSet<>()).add(new IdentityWeakReference<>(object));
 		}
 	}
 
 	private boolean forgetInjectedObject(Object object, PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
-			List<WeakReference<?>> list = injectedObjects.get(objectSupplier);
-			if (list != null) {
-				for (Iterator<WeakReference<?>> i = list.iterator(); i.hasNext();) {
-					WeakReference<?> ref = i.next();
-					if (object == ref.get()) {
-						i.remove();
-						return true;
-					}
-				}
+			Set<IdentityWeakReference<?>> set = injectedObjects.get(objectSupplier);
+			if (set != null) {
+				return set.remove(new IdentityWeakReference<>(object));
 			}
 			return false;
 		}
 	}
 
-	private List<WeakReference<?>> forgetSupplier(PrimaryObjectSupplier objectSupplier) {
+	private Set<IdentityWeakReference<?>> forgetSupplier(PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
 			return injectedObjects.remove(objectSupplier);
 		}
 	}
 
-	private List<WeakReference<?>> getSupplierObjects(PrimaryObjectSupplier objectSupplier) {
+	private Set<IdentityWeakReference<?>> getSupplierObjects(PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
 			return injectedObjects.get(objectSupplier);
 		}
@@ -425,7 +416,7 @@ public class InjectorImpl implements IInjector {
 	}
 
 	public void disposed(PrimaryObjectSupplier objectSupplier) {
-		List<WeakReference<?>> references = getSupplierObjects(objectSupplier);
+		Set<IdentityWeakReference<?>> references = getSupplierObjects(objectSupplier);
 		if (references == null)
 			return;
 		Object[] objects = new Object[references.size()];
@@ -668,20 +659,13 @@ public class InjectorImpl implements IInjector {
 
 	private boolean hasInjectedStatic(Class<?> objectsClass) {
 		synchronized (injectedClasses) {
-			for (WeakReference<Class<?>> ref : injectedClasses) {
-				Class<?> injectedClass = ref.get();
-				if (injectedClass == null)
-					continue;
-				if (injectedClass == objectsClass) // use pointer comparison
-					return true;
-			}
-			return false;
+			return injectedClasses.contains(new IdentityWeakReference<>(objectsClass));
 		}
 	}
 
 	private void rememberInjectedStatic(Class<?> objectsClass) {
 		synchronized (injectedClasses) {
-			injectedClasses.add(new WeakReference<>(objectsClass));
+			injectedClasses.add(new IdentityWeakReference<>(objectsClass));
 		}
 	}
 
@@ -1025,5 +1009,35 @@ public class InjectorImpl implements IInjector {
 		boolean isPresent = annotatedElement.isAnnotationPresent(annotation);
 		cache.put(annotatedElement, isPresent);
 		return isPresent;
+	}
+
+	/**
+	 * The IdentityWeakReference extends the {@link WeakReference} with the
+	 * difference that it can be compared to another IdentityWeakReference. The
+	 * compare is done on the Identity of the Referenced Object, this allows us to
+	 * use this element in a Set and have unique objects in it.
+	 */
+	private static class IdentityWeakReference<T> extends WeakReference<T> {
+
+		private final int hashCode;
+
+		IdentityWeakReference(T referent) {
+			super(referent);
+			hashCode = System.identityHashCode(referent);
+		}
+
+		@Override
+		public boolean equals(Object obj) {
+			if (obj instanceof IdentityWeakReference other) {
+				return other.get() == get();
+			}
+			return false;
+		}
+
+		@Override
+		public int hashCode() {
+			return hashCode;
+		}
+
 	}
 }


runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -668,20 +666,14 @@ private void processClass(Object userObject, PrimaryObjectSupplier objectSupplie
...
-			return false;
+			int objectId = System.identityHashCode(objectsClass);
+			return injectedClasses.containsKey(objectId);
 		}
 	}
 
 	private void rememberInjectedStatic(Class<?> objectsClass) {
 		synchronized (injectedClasses) {
-			injectedClasses.add(new WeakReference<>(objectsClass));
+			injectedClasses.put(System.identityHashCode(objectsClass), new WeakReference<>(objectsClass));
```
merks, line 676: Is there an assumption here that there will be no collisions on this hash code?

N1k145: I thought to while implementing, but I dug a bit into it and it has collisions. 
What I found is ObjectUtils.identityToString, but that produces the same output as the default toString()
Which is ClassName@IdentitiyHashCode.

Do you have an idea how we could solve this without collisions and without always iterating the whole list?

N1k145: @merks I looked at the disposal code again. Do you agree that the order of the objects seems to be irrelevant?

My first thought was that they are disposed inverse of creation, but they are disposed in the same order, so the order should be irrelevant?

Then we could use a WeakHashMap where we use the object as key which is held in a weak reference internally. And the WeakReference as value.

merks: Naively I though wherever a List was being used to use instead a HashSet or LinkedHashSet if order is somehow relevant, but looking more closely now, that doesn't work because it's comparing `object == ref.get()`.

I don't see any immediately obvious reason why a weak hash map couldn't be used, but I've never looked closely at this code before...

laeubi: If object equally is relevant an `IdentityHashmap` should be used and as the code is syncronized anyways...

merks: @laeubi The problem is the need for weak references...

laeubi: I must confess that I don't understand the optimization goal here then, maybe it could be explained a bit more in detail. Maybe it would be enough to extend `WeakReference` and override equals/hashcode and then simply use the Set.contains, this hsould have equal performance to a Map.

Beside that System.identityHashCode() might have collisions, but this is not important as long as it is equally distributed, but you then needs to maintain a map of objects as the value and can then iterate over a much smaller set (effectively reimplementing what the first suggestion would archive with a standard set);

N1k145: @laeubi the main optimization goal is the rememberInjectedObject function.
In the original implementation it iterated over the list of all WeakReferences, resolved those and compared the object, and for a list with 400.000 elements that takes a while, basically O(n).

What I am trying is to replace this list with something hash based where a simple and fast contains check can be done.

The complicated thing is the object should be kept in a weak reference, while the contains check has to be done on the actual object.

My initial idea was to keep the hash of the object as a key to the reference, which turned out to be an issue with collusions.

That's why we are discussing which alternative we could use to store a weak reference on the object.

laeubi: As said you have two options:

1. Store `IdentityWeakReference` in a `Set<IdentityWeakReference>` where `IdentityWeakReference` is implemented as below
2. Make  `injectedClasses` a `Map<Long, List<WeakReference>>` to account for hashcode collisions

```
public class IdentityWeakReference extends WeakReference<Object> {

    public IdentityWeakReference(Object referent) {
        super(referent);
    }
    
    @Override
    public boolean equals(Object obj) {
        if (obj instanceof IdentityWeakReference other) {
            return Objects.equal(other.get(), get());
        }
        return false;
    }
    
    @Override
    public int hashCode() {
        return System.identityHashCode(get());
    }

}
```

laeubi: By the way, the current code seem to suffer from leaking `WeakReference` if the are garbage collected they could never be removed so it seems a bit misusing `WeakReference` anyways.

N1k145: Thanks for the example. I don't know where I got the idea from that there is a weakhashset, but with our first example we would build something similar. 

Should I implement this as a nested class or should this be in some reusable location where other data types are provided?

N1k145: @laeubi I updated my PR with our proposal for the `IdentityWeakReference` to keep the order of the elements, just to keep the same behavior as before, I used a LinkedHashSet.

N1k145:
The failing test seams to be unrelated, see #563

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -1026,4 +1019,31 @@ private boolean isAnnotationPresent(AnnotatedElement annotatedElement,
...
+	private static class IdentityWeakReference<T> extends WeakReference<T> {
+
+		IdentityWeakReference(T referent) {
+			super(referent);
+		}
+
+		@Override
+		public boolean equals(Object obj) {
+			if (obj instanceof IdentityWeakReference other) {
+				return Objects.equals(other.get(), get());
```
merks, line 1038: The existing implementation was using == but here you using .equals.  Also, this definition of equals is not consistent with the hashCode implementation which uses the identity hash code.  I think you should simply use == here given this whole class is about "identity".

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -1026,4 +1019,31 @@ private boolean isAnnotationPresent(AnnotatedElement annotatedElement,
...
+			}
+			return false;
+		}
+
+		@Override
+		public int hashCode() {
+			return System.identityHashCode(get());
```
merks, line 1045: The problem here is that if (and of course when) the weak reference is garbage collected, get() will return null and then this class will returning different hashCodes at different times.  That will corrupt the set.   I think it will be necessary to compute the hash code and cache it in the constructor so that it can never change.

merks (changes requested):
It's simpler than before. 😄

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
laeubi: Yes thats probabbly better here.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
laeubi: Good idea.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
N1k145: You are right, I did take a look into Objects equals, but this does both checks, == and equals

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -148,38 +149,36 @@ private void internalInject(Object object, PrimaryObjectSupplier objectSupplier,
...
+				map.add(ref);
 			}
-			list.add(new WeakReference<>(object));
 		}
 	}
 
 	private boolean forgetInjectedObject(Object object, PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
-			List<WeakReference<?>> list = injectedObjects.get(objectSupplier);
+			Set<IdentityWeakReference<?>> list = injectedObjects.get(objectSupplier);
```
merks, line 163: It's not a list anymore so no longer a good name.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -148,38 +149,36 @@ private void internalInject(Object object, PrimaryObjectSupplier objectSupplier,
...
 			if (list != null) {
-				for (Iterator<WeakReference<?>> i = list.iterator(); i.hasNext();) {
-					WeakReference<?> ref = i.next();
-					if (object == ref.get()) {
-						i.remove();
-						return true;
-					}
+				IdentityWeakReference<?> ref = new IdentityWeakReference<>(object);
+				if (list.contains(ref)) {
```
laeubi, line 166: this block can become return `list.remove(ref) != null`

merks, line 166: Here it will be more efficient to do a remove and use the boolean return from that.  Probably just:
```
return list.remove(ref);
```
Though as mentioned above list is a bad name now.  Probably ref can be inlined then too.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -1026,4 +1018,34 @@ private boolean isAnnotationPresent(AnnotatedElement annotatedElement,
...
+
+	/**
+	 * The IdentityWeakReference extends the {@link WeakReference} with the
+	 * difference that it can be compared to another IdentityWeakReference. The
+	 * compare is done on the Identity of the Referenced Object, this allows us to
+	 * use this element in a Set and have unique objects in it.
+	 */
+	private static class IdentityWeakReference<T> extends WeakReference<T> {
+
+		private int hashCode;
```
merks, line 1030: It could be final.

merks (changes requested):
Better still.  Just minor improvements left I think.  👍

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -148,38 +149,32 @@ private void internalInject(Object object, PrimaryObjectSupplier objectSupplier,
...
 	private void rememberInjectedObject(Object object, PrimaryObjectSupplier objectSupplier) {
 		synchronized (injectedObjects) {
-			List<WeakReference<?>> list = injectedObjects.computeIfAbsent(objectSupplier, k -> new ArrayList<>());
-			for (WeakReference<?> ref : list) {
-				if (object == ref.get())
-					return; // we already have it
+			Set<IdentityWeakReference<? extends Object>> set = injectedObjects.computeIfAbsent(objectSupplier,
+					k -> new LinkedHashSet<>());
+			IdentityWeakReference<?> ref = new IdentityWeakReference<>(object);
+			if (!set.contains(ref)) {
```
merks, line 155: Sorry I keep noticing things...  

There is no need to test contains before doing an add; a set won't add a duplicate.  So probably this whole thing can be:
```
injectedObjects.computeIfAbsent(objectSupplier, k -> new LinkedHashSet<>()). add(new IdentityWeakReference<>(object);
```
Then we don't need a name for the "set".

merks (changes requested):
The nice thing is that this is simpler and cleaner than before, and is more efficient.

I think this is my last nitpick.  Sorry...

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
N1k145: I also noticed this, I was thinking about if the check is faster than the adding, I did not really find something about this, so i will change to the one liner

laeubi: In general the one liner is faster.

- **Case 1**: Item is not in the set
- **Case 2**: Item is in the set

With "if/then":

1. This will result in two checks if the item is present (one from the if and one from the add) and one add operation to the internal datastructure
2. This will result in one checks if the item is present (one from the if)

With "oneliner":

1. One check and one add to the internal datastructure
2. one check (internally) and early exit

so for case (1) the oneliner saves one additional check while for (2) it is equally.

merks (approved):
I see no more flaws.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
N1k145: Yes, you are right. I was wondering if the contains check from the set or contains is faster, because those seam to be different implementations. But maybe I thought too much about it.

N1k145:
@merks @laeubi 
Thank you for the reviews and hints in the right direction.

N1k145:
@merks can you merge this PR? I don't have write permissions

merks:
@akurtakov 

I think this change looks like a good improvement and even simplifies the code.  Is there anyone else who ought to review this first?

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/InjectorImpl.java
```diff
@@ -425,7 +416,7 @@ else if (internalRequestor.isOptional())
 	}
 
 	public void disposed(PrimaryObjectSupplier objectSupplier) {
-		List<WeakReference<?>> references = getSupplierObjects(objectSupplier);
+		Set<IdentityWeakReference<?>> references = getSupplierObjects(objectSupplier);
```
szarnekow, line 419: Below we are iterating the set of references outside of the synchronized block. It could be modified concurrently, if I’m understanding this correctly. This was already the case in the old impl. Now I’m wondering why we need the synchronization at all given that it’s flawed.

szarnekow: Another thing: should we remove the refs that have been cleared already? Line 424 could use an iterator and call iterator.remove in the else branch of the nested if

merks: We're iterating over a set/list specific to one supplier though right?  So could another thread really be doing other things with that same supplier at the same time?  I can't say I understand the deeper behavior involved here, so maybe that is possible...

merks: Yes, I think @laeubi already mentioned that there were/are weak reference leaks.  But doesn't the final call to `forgetSupplier(objectSupplier)` just forget the whole list anyway?

szarnekow: It’s certainly a huge improvement in its current shape. Nice work!

The overall design is hard to assess and I’ll leave remarks about the applied synchronization scheme aside.

PR #579, opened 2023-07-21
Disable the cancel button of progress indicator in run configurations accordingly

fedejeanne:
## What it does
Enable/disable the little red button in the progress indicator at the bottom of the "Run configurations..." dialog accordingly. It was always being enabled, which let the user cancel the operation even if the task had been explicitly set to **NOT** cancelable.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/593
Discovered while testing https://github.com/eclipse-pde/eclipse.pde/issues/679

## How to test

* Generate an idle wait by adding this code right at the beginning of `org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.handleLaunchConfigurationSelectionChanged(SelectionChangedEvent)`:

```java
try {
	run(true, false, monitor -> { // NOTICE: the task is NOT cancelable
		monitor.beginTask("Idling for 3 seconds", IProgressMonitor.UNKNOWN); //$NON-NLS-1$
		for (int i = 0; i < 3; i++) {
			Thread.sleep(1000);
		}
	});
} catch (InvocationTargetException | InterruptedException e) {
	e.printStackTrace();
}
```
* Open the _Run configurations..._ dialog (ignore the first progress dialog since that one already works as expected)
* Select something on the left-side panel --> You should see the progress indicator at the bottom and the _cancel_ button should be **deactivated**, like this:
image
* Change the call to `run`, make the task cancelable _i.e._ `run(true, true, monitor -> { ...` --> You should see the progress indicator at the bottom and the _cancel_ button should be **activated**, like this:
image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
@@ -1371,7 +1371,9 @@ public class LaunchConfigurationsDialog extends TitleAreaDialog implements ILaun
 				fLastControl = null;
 			}
 			// Attach the progress monitor part to the cancel button
-			fProgressMonitorPart.attachToCancelComponent(null);
+			if (cancelable) {
+				fProgressMonitorPart.attachToCancelComponent(null);
+			}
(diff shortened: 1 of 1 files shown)


jukzi:
lgtm

HeikoKlare (approved):
Change looks good. Failing test case is due to recent target platform update: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1233

PR #591, opened 2023-07-25
Show progress indicator of "Run configurations..." after delay

fedejeanne:
Wait 1 second before showing the progress indicator in the dialog "Run configurations...". This avoids the flickering that happens when the task is completed in under 1 second.

Contributes to https://github.com/eclipse-pde/eclipse.pde/issues/679

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
@@ -91,6 +91,7 @@ import org.eclipse.swt.widgets.Text;
 import org.eclipse.swt.widgets.ToolBar;
 import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.progress.UIJob;
 import org.eclipse.ui.progress.WorkbenchJob;
 import org.osgi.framework.FrameworkUtil;
 
@@ -1374,8 +1375,12 @@ public class LaunchConfigurationsDialog extends TitleAreaDialog implements ILaun
 			if (cancelable) {
 				fProgressMonitorPart.attachToCancelComponent(null);
 			}
-			fProgressMonitorPart.getParent().setVisible(true);
+
 			fActiveRunningOperations++;
+
+			// only show the progress bar if necessary after a delay
+			UIJob showProgressMonitorPart = createJobToShowProgressMonitorPart();
+			showProgressMonitorPart.schedule(PlatformUI.getWorkbench().getProgressService().getLongOperationTime());
 
 		//do work here collecting enabled states, otherwise to get these states we would need to
 		//perform the validation of the dialog again, which is expensive and would cause flashing of widgets.
@@ -1391,6 +1396,8 @@ public class LaunchConfigurationsDialog extends TitleAreaDialog implements ILaun
 				ModalContext.run(runnable, fork, fProgressMonitorPart, getShell().getDisplay());
 			}
 			finally {
+				showProgressMonitorPart.cancel();
+
 				fActiveRunningOperations--;
 				updateRunnnableControls(true, prev);
 				if (getShell() != null) {
@@ -1407,6 +1414,17 @@ public class LaunchConfigurationsDialog extends TitleAreaDialog implements ILaun
 		else {
 			PlatformUI.getWorkbench().getProgressService().run(fork, cancelable, runnable);
 		}
+	}
+
+	private UIJob createJobToShowProgressMonitorPart() {
+		return UIJob.create("", __ -> { //$NON-NLS-1$
+			// only show the progress monitor if there are still active operations by the
+			// time this job is executed
+			if (fActiveRunningOperations > 0) {
+				fProgressMonitorPart.getParent().setVisible(true);
+			}
+			return Status.OK_STATUS;
+		});
 	}
 
 	/**
--- a/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.ui; singleton:=true
-Bundle-Version: 3.18.100.qualifier
+Bundle-Version: 3.18.200.qualifier
 Bundle-Activator: org.eclipse.debug.internal.ui.DebugUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
```diff
@@ -1374,9 +1375,13 @@ public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable
 			if (cancelable) {
 				fProgressMonitorPart.attachToCancelComponent(null);
 			}
-			fProgressMonitorPart.getParent().setVisible(true);
+
 			fActiveRunningOperations++;
 
+			UIJob showProgressMonitorPart = createJobToShowProgressMonitorPart();
```
HannesWell, line 1381:
```suggestion
					UIJob showProgressMonitorPart = UIJob.create(IInternalDebugCoreConstants.EMPTY_STRING, m -> {
			// only show the progress monitor if there are still active
			// operations by the time this job is executed
			if (fActiveRunningOperations > 0) {
				fProgressMonitorPart.getParent().setVisible(true);
			}
		});
```

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
```diff
@@ -1374,9 +1375,13 @@ public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable
...
+			// only show the progress bar if necessary after 1 second
+			showProgressMonitorPart.schedule(1_000);
```
HannesWell, line 1383: Since potentially not everybody wants a delay, maybe the delay should be configurable in the context?

HannesWell (commented):
Looks reasonable.


If nobody else does in the meantime, I'll try to review this in detail and try it out at the weekend.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
fedejeanne: hm, I see your point but I'd rather have it in a separate method. I think the `run` method is a bit cluttered and a method with a descriptive name helps me "unload my brain" when I read through the code.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
fedejeanne: Do you mean like `DebugUIPlugin.getDefault().getPreferenceStore().getLong("somelong");` ? That would also mean that there needs to be a new preference in the *Preferences* dialog under *Run/Debug > Launching > Launch configurations*, right? 

Or did you mean something else?

fedejeanne: > Since potentially not everybody wants a delay, maybe the delay should be configurable in the context?

I was thinking again about making it configurable and it still doesn't feel right. 

Don't get me wrong, I'm all for customizing the user experience but in this particular case, a low value would mean that the progress bar might flicker and a high value would mean the UI might freeze. Letting the user set this value feels like giving him/her the choice between bad and worse 😅

Maybe you and I can settle for a value and let it hard-coded. Say 500 ms?

merks: There is definitely a tradeoff with configuration preferences that users are unlikely to find and as there are more and more, the user is even less likely to find and understand such things...

HannesWell: Sorry for the late reply.
I meant not configurable for the user but for developers calling this code.

If a dev knows that an operation will definitvly take some time, it would be good to show i immediatly.

Furthermore If we could somehow estimate the remaining runtime(e.g. from a ProgressMonitor, it would be good to consider if it is worth to show it.
E.g. for a 501ms running Task showing Progress after 500ms is pointless.

fedejeanne: > I meant not configurable for the user but for developers calling this code.
Thank you for clarifying it :-) . Do you have any suggestion (code) on how to achieve that?

> Furthermore If we could somehow estimate the remaining runtime(e.g. from a ProgressMonitor, it would be good to consider if it is worth to show it.

Hm, the problem is that the progress monitor does not have any method to see how much work is still to be done so there is no basis to estimate if it's worth showing it.

>  E.g. for a 501ms running Task showing Progress after 500ms is pointless.
Agreed. The timeout is just a simple way of catching all jobs that run in under _X_ ms. I have no way of knowing how long a job is going to take, but In my experience if the job takes > 1s then it will probably take >2 s (i.e. the progress bar will appear for at least a second). Again, this is a very simplistic approach based on my experience. I wanted to provide a solution without burdening the user (or programmer) with the

fedejeanne: > Not really, I would guess that this should somehow be stored in the context? 

Sorry to insist but I don't understand. What _context_ do you mean?

Surely not the `org.eclipse.e4.core.contexts.IEclipseContext`, right?
image

fedejeanne: Ah, you probably meant the `BundleContext` right? I'll look into it on Monday.

HannesWell: > Ah, you probably meant the `BundleContext` right? I'll look into it on Monday.

I think it should be a `org.eclipse.jface.operation.IRunnableContext`, as you are using that in https://github.com/eclipse-pde/eclipse.pde/pull/674/ to show the pop-up that shows the progress indication.
Unfortunately the progress-indicator is not created directly so have to pass trough the threshold value.

HannesWell: Btw. Monday is the last day of development for the upcomming 2023-09 release and I probably won't have the time to review/try this Monday evening (if another Committer does it is of course welcome).

I'll try to review your PR in PDE over the weekend, but cannot promise it.

fedejeanne:
Hello all,
there has been little activity on this one in the past few weeks. Any chance we can resume this PR for 4.30?

mickaelistria:
As build log mentions, a version bump seems necessary:
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.debug.ui: Only qualifier changed for (org.eclipse.debug.ui/3.18.100.v20230911-0736). Expected to have bigger x.y.z than what is available in baseline (3.18.100.v20230802-1257) -> [Help 1]
```

fedejeanne:
> As build log mentions, a version bump seems necessary:
> 
> ```
> [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.debug.ui: Only qualifier changed for (org.eclipse.debug.ui/3.18.100.v20230911-0736). Expected to have bigger x.y.z than what is available in baseline (3.18.100.v20230802-1257) -> [Help 1]
> ```

Thank you for the hint. It's odd, I just bumped the version and I get this error in Eclipse:

image

Any hint on how to get rid of it?

iloveeclipse:
> Any hint on how to get rid of it?

You should set API baseline to 4.29 RC2a. I assume you are still using 4.28 in the IDE.

merks:
FYI, the Oomph setups won't be correct (correct baseline and TP) until Wednesday morning...

merks:
As a workaround, you can do this.  Use Navigate -> Open Setup -> Workspace and copy (select text below) and paste (onto the root Workspace object) these two redirection tasks
```
<?xml version="1.0" encoding="UTF-8"?>
<xmi:XMI xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:setup="http://www.eclipse.org/oomph/setup/1.0">
  <setup:RedirectionTask
      sourceURL="https://download.eclipse.org/eclipse/updates/4.28/R-4.28-202306050440"
      targetURL="https://download.eclipse.org/eclipse/updates/4.29/R-4.29-202309031000"/>
  <setup:RedirectionTask
      sourceURL="https://download.eclipse.org/eclipse/updates/4.29-I-builds"
      targetURL="https://download.eclipse.org/eclipse/updates/4.30-I-builds"/>
</xmi:XMI>
```
Then the baseline task will use the 4.29 release and the target platform will use 4.30 I builds...

image

HannesWell:
> This could be a good compromise since the concept of delayed progress reporting has already been tackled and it seems a delay of 800 ms is the accepted norm. I changed the code to simply use this threshold in a21c1d7.

Sounds good.
While in some cases callers might want to adjust this value, most callers are probably happy with a reasonable default. So I think it is sufficient to add that posibility when there is a demand for it.

I'm still thinking about if it would be good to add means to IProgressMonitor to get the percentage of work done (or at least an Approximation since the value can be outdated quickly)? This would allow to estimate if it is worth to open the progress indicator.
On the other hand it highly depends on the called code if, for example the last 20% really take only 20% of the total time or more  or less. I have seen it often that the indicator got stuck at 80/90% for a long time. And there is even a recommendation to just half the work remaining all the time if the total work is unknown.
So overall I'm not sure if that is worth it.

laeubi:
> I'm still thinking about if it would be good to add means to IProgressMonitor to get the percentage of work done (or at least an Approximation since the value can be outdated quickly)? This would allow to estimate if it is worth to open the progress indicator.

the **implementation** of the monitor should/must handle this so there is no need for an public API at all, so for example ProgressMonitorDialog could have something similar to what Swing uses here:

https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/javax/swing/ProgressMonitor.html#setMillisToDecideToPopup(int)

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
HannesWell: OK, since it is your PR that's your decision. :)
But you can still use the mentioned factory in that method.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
```diff
@@ -1409,6 +1416,20 @@ public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable
 		}
 	}
 
+	private UIJob createJobToShowProgressMonitorPart() {
+		return new UIJob(IInternalDebugCoreConstants.EMPTY_STRING) {
```
HannesWell, line 1420: Besides UIJob.create() I think we can just use an empty string literal here.
IMHO it does not make sense to get that from a constant.

HannesWell:
> > I'm still thinking about if it would be good to add means to IProgressMonitor to get the percentage of work done (or at least an Approximation since the value can be outdated quickly)? This would allow to estimate if it is worth to open the progress indicator.
> 
> the **implementation** of the monitor should/must handle this so there is no need for an public API at all, so for example ProgressMonitorDialog could have something similar to what Swing uses here:
> 
> https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/javax/swing/ProgressMonitor.html#setMillisToDecideToPopup(int)

But the ProgressMonitor is designed differently and does not handle the context, so there must be at least and internal method to get that value.

Any way, since everything additionally seems to be more complex, I think this is fine as a first step, with the two minor points (using the UIJob-factory and the empty string) fixed I would be fine to submit this.

laeubi:
> But the ProgressMonitor is designed differently and does not handle the context, so there must be at least and internal method to get that value.

The value should be passed to the dialog and then it could handle it internally, but users of that Interface must not care nor do we need to store such values in general...

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
fedejeanne: Agreed, done in f50e586b30c24531456f1e19b348228a36044596

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
fedejeanne: Done in 6c443cccd3e7d5912b17855cfcdd8e5535e8663e

HannesWell (approved):
Looks good to me.

Does anybody else plan to review this?

@fedejeanne you need to bump the Micro version segment in in the Manifest order to fix the build failure:

`[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.debug.ui: Only qualifier changed for (org.eclipse.debug.ui/3.18.100.v20230913-0532). Expected to have bigger x.y.z than what is available in baseline (3.18.100.v20230802-1257) -> [Help 1]`

HannesWell:
> > But the ProgressMonitor is designed differently and does not handle the context, so there must be at least and internal method to get that value.
> 
> The value should be passed to the dialog and then it could handle it internally, but users of that Interface must not care nor do we need to store such values in general...

I think we have a missunderstanding here. The amount of work done is interesting do help deside if after the Initial delay it is worth for the remaining time to show the Pop-up.

Making the initial delay configurable is an orthogonal issue.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
```diff
@@ -1374,9 +1375,13 @@ public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable
 			if (cancelable) {
 				fProgressMonitorPart.attachToCancelComponent(null);
 			}
-			fProgressMonitorPart.getParent().setVisible(true);
+
 			fActiveRunningOperations++;
 
+			UIJob showProgressMonitorPart = createJobToShowProgressMonitorPart();
+			// only show the progress bar if necessary after 1 second
```
HeikoKlare, line 1382: This comment should be removed/changed, as the actual delay comes from the progress service (and is actually not 1 second).

HeikoKlare (commented):
Changes look good to me. I only have a minor remark w.r.t. to a code comment.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationsDialog.java
fedejeanne: I changed it

HeikoKlare (approved):

HannesWell (approved):
Thank you @fedejeanne for this PR and your patience.

On the last pass I thought about if `fActiveRunningOperations` should be turned into a `AtomicLong`, but since the code is running in the UI thread in both cases (just later) that's fine.

Submitting.

fedejeanne:
> Thank you @fedejeanne for this PR and your patience.

Thank you for reviewing!

> I'll have a look at the PDE pendant in the next days.

Looking forward to it :-)

laeubi:
> FYI, the Oomph setups won't be correct (correct baseline and TP) until Wednesday morning...

Just to make sure was this meant to be 13.09.2023? The setup still fails for me and I can't use the baseline...

merks:
It doesn't look like this?

image

laeubi:
It seems that Oomph was confused by the pre-req-target project imported as a project in my workspace somehow :-\
At laest now I removed that project and it *seems* to came across the point now...

merks:
Yes, there was a problem caused by the two occurrences and the "wrong" one winning.  That should no longer happen because of the filter in the .project file.

laeubi:
Alright it seems I was now able to get the API error (hurray!):
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1099

PR #595, opened 2023-07-26
Adapt to Lucene version update to 9.7.0 #601

HeikoKlare:
The target platform has been updated with Lucene 9.7.0. This adapts the according index compatibility test case.
**Question:** Should there something else be done than updating the Lucene version in the index compatibility test?

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1233

Failing test can, e.g., be seen in this current build: https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-579/4/testReport/

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/601

### DIFF
--- a/ua/org.eclipse.ua.tests/help/org/eclipse/ua/tests/help/search/PrebuiltIndexCompatibility.java
+++ b/ua/org.eclipse.ua.tests/help/org/eclipse/ua/tests/help/search/PrebuiltIndexCompatibility.java
@@ -47,33 +47,25 @@ import org.osgi.framework.FrameworkUtil;
 /**
  * Verify that older versions of the index can be read by this
  * version of Eclipse.
- *
- * How to maintain this test - if when upgrading to a new version
- * of Lucene one of the IndexReadable tests fails you need to
- * make the following changes:
- * 1. Change the corresponding Compatible() test to expect a result of false
- * 2. Comment out the failing test
- * 3. Change the help system to recognize that version of Lucene as being incompatible
  */
-
 public class PrebuiltIndexCompatibility {
 
 	/**
 	 * Test index built with Lucene 8.4.1
 	 */
 	@Test(expected = IllegalArgumentException.class)
-	public void test8_4_1_IndexReadable() throws Exception {
+	public void test8_4_1_IndexUnreadable() throws Exception {
 		checkReadable("data/help/searchindex/index841");
 	}
 
 	@Test
-	public void test8_4_1LuceneCompatible() {
+	public void test8_4_1LuceneIncompatible() {
 		checkLuceneCompatible("8.4.1", false);
 	}
 
 	@Test
-	public void test9_4_2LuceneCompatible() {
-		checkLuceneCompatible("9.4.2", true);
+	public void testLatestLuceneCompatible() {
+		checkLuceneCompatible(org.apache.lucene.util.Version.LATEST.toString(), true);
 	}
 
 	@Test


merks:
I think the Platform itself stopped building pre-built indexes, so maybe the test itself is just bogus now?

Perhaps @iloveeclipse can say more?   Right now @akurtakov is on vacation, so we shouldn't let him relax...

iloveeclipse:
I do see this test for the first time. It looks to me it makes sense independently if SDK ships indexes or not. But I believe test has to be adopted a bit differently: see 3caa3ab52fb6f8592ca71cff14dc619507dcc14f changes.

HeikoKlare:
I tried to figure out why it might make sense to keep the test. But from my understanding compatibility with indices created by older lucene versions was almost completely removed in https://github.com/eclipse-platform/eclipse.platform/commit/3caa3ab52fb6f8592ca71cff14dc619507dcc14f. Whenever minor or major version increases, the index is considered incompatible:

https://github.com/eclipse-platform/eclipse.platform/blob/f1c95afbe4fe937f7cc085241e97a689258cd86b/ua/org.eclipse.help.base/src/org/eclipse/help/internal/search/SearchIndex.java#L751-L758

So the failing test degrades to a regression test for the `isLuceneCompatible` method of the `SearchIndex` properly comparing the given version with the one used by the search index, which is actually the one of the loaded lucene bundle, i.e., the one provided by the target platform. And since the test case passes a fixed version, it would have to be adapted after each major or minor version change of the lucene bundle in the target platform.

I would propose to either remove the test case or change it to be a regression test for the `isLuceneCompatible` method of the `SearchIndex` in terms of validating that lucene `LATEST` is always considered compatible to the version the search index uses (although such a test might then be better placed in an according test class for the `SearchIndex`). I have updated the PR with a proposal for the latter approach.

merks:
@HeikoKlare 

I agree with your assessment and your changes.  

Not to create more work, but does anything test that an index built with the latest Lucene can then actually be read by that same latest Lucene?  (It's not evident to me quickly looking at the test.)

HeikoKlare:
Thanks for your quick reply, @merks!

The test cases in the touched class `PrebuiltIndexCompatibility` do not check whether an index created with the latest Lucene can be read by that same Lucene. There have only been tests that validated that indices created with older Lucene versions can be read (or cannot be read anymore) by the current Lucene version. That makes sense to me, as the test class seems to be intended for exactly the case of compatibility between Lucene versions.
Just for my understanding: Shouldn't the case that an index created with the latest Lucene version can be read by the same version already be ensured (and tested) by Lucene itself? To me it does not look like you can (easily) persist an index with Lucene that you cannot read with the same Lucene version.

merks:
One would assume that one doesn't have to test that Lucene itself is not buggy.  It seems a fairly reasonable assumption.  And I don't think we should have to add a test for that given there wasn't one before.  Though @iloveeclipse  might want to consider if adding such a test if it is worthwhile.  But probably the downstream products on which he works test this...

I'm fine to merge this as is.

iloveeclipse:
> One would assume that one doesn't have to test that Lucene itself is not buggy

I'm not much in the "help" business. 
@howlger : are you aware about some SDK tests that would check if help search can find something (meaningful)?

HeikoKlare:
Should we merge this one to have the constantly failing test fixed and have more useful CI check results again?

If you think it makes sense to proceed with the discussion on whether some other kind of test would be useful for search functionality, we would move it to a separate issue or a discussion as it won't affect this PR, right?

akurtakov:
Yes, it's better to merge this one to get less failing tests. Further enhancements can be done separately.

howlger:
> > One would assume that one doesn't have to test that Lucene itself is not buggy
> 
> I'm not much in the "help" business. @howlger : are you aware about some SDK tests that would check if help search can find something (meaningful)?

@iloveeclipse Thanks for CCing me. Yes, there are a couple of tests for that here: https://github.com/eclipse-platform/eclipse.platform/tree/master/ua/org.eclipse.ua.tests/help/org/eclipse/ua/tests/help/search

Issue #598, opened 2023-07-26
Notifications about termination of a RuntimeProcess can be lost


HeikoKlare:
When creating a `RuntimeProcess`, it can terminate before it is properly registered for processing notifications, such that is `TERMINATE` event is never propagated.

## Problem Description

When creating a `RuntimeProcess` via `DebugPlugin.newProcess`, it is wrapped into a `ProcessMonitorThread`, which is started directly afterwards. After starting that thread, the process is registered at its `Launch`, which is responsible for registering it for notifications.
https://github.com/eclipse-platform/eclipse.platform/blob/f1c95afbe4fe937f7cc085241e97a689258cd86b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/model/RuntimeProcess.java#L155-L157

The `ProcessMonitorThread` waits for the process to finish and then fires a notification event.

https://github.com/eclipse-platform/eclipse.platform/blob/f1c95afbe4fe937f7cc085241e97a689258cd86b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/model/RuntimeProcess.java#L492-L504

If the process terminates very fast, like the one in `LaunchConfigurationTest.testTerminateTimeStamp`, this can lead to the termination being processed before the process is even registered at its `Launch`. In case of the mentioned test, this leads to sporadic test failures, see #582.

## Solution Proposal

The process should be registered at the `Launch` for notifications before starting the monitor thread that may already handle it termination. I have provided that proposal along with a regression test in #599.

Issue #582, opened 2023-07-24
Random failing LaunchConfigurationTests.testTerminateTimeStamp


HeikoKlare:
`org.eclipse.debug.tests.launching.LaunchConfigurationTests.testTerminateTimeStamp` randomly fails, e.g. https://github.com/eclipse-platform/eclipse.platform/actions/runs/5622957268/job/15236595341

```
org.eclipse.debug.tests.launching.LaunchConfigurationTests.testTerminateTimeStamp -- Time elapsed: 30.05 s <<< FAILURE!
java.lang.AssertionError: Launch did not finished
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.debug.tests.TestUtil.waitWhile(TestUtil.java:138)
	at org.eclipse.debug.tests.AbstractDebugTest.waitWhile(AbstractDebugTest.java:116)
	at org.eclipse.debug.tests.launching.LaunchConfigurationTests.testTerminateTimeStamp(LaunchConfigurationTests.java:1371)
```

The implementation launches a configuration before registering the listener that waits for its termination. This induces a race condition that might be the reason for the failing test.

jukzi:
https://github.com/eclipse-platform/eclipse.platform/actions/runs/5657225255/job/15326634462

```
Launch did not finish
java.lang.AssertionError: Launch did not finish
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.debug.tests.TestUtil.waitWhile(TestUtil.java:138)
	at org.eclipse.debug.tests.AbstractDebugTest.waitWhile(AbstractDebugTest.java:116)
	at org.eclipse.debug.tests.launching.LaunchConfigurationTests.testTerminateTimeStamp(LaunchConfigurationTests.java:1366)

```

HeikoKlare:
Log shows that process has actually terminated but termination event was not processed by listener.

```
Launch termination event did not occur and termination state of launched process is: true
java.lang.AssertionError: Launch termination event did not occur and termination state of launched process is: true
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.debug.tests.TestUtil.waitWhile(TestUtil.java:138)
	at org.eclipse.debug.tests.AbstractDebugTest.waitWhile(AbstractDebugTest.java:116)
	at org.eclipse.debug.tests.launching.LaunchConfigurationTests.testTerminateTimeStamp(LaunchConfigurationTests.java:1367)
```

[...]

PR #599, opened 2023-07-26
Prevent notification losses for termination of RuntimeProcess #598 #582

HeikoKlare:
When creating a `RuntimeProcess`it is wrapped into a `ProcessMonitorThread`, which is started directly afterwards. After starting that thread, the process is registered at its `Launch`, which is responsible for registering it for notifications. If the process terminates too fast, the ProcessMonitorThread already processes the termination event while the process it not registered at its `Launch` for processing that notification yet. Thus, the notification will get lost.

With this change, the process is properly registered before it may terminate and the termination notification may be processed. An according regression test is added. The random failures of test case `LaunchConfigurationTest.testTerminateTimeStamp` should also be fixed by this change.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/598
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/582


debug/org.eclipse.debug.core/core/org/eclipse/debug/core/model/RuntimeProcess.java
```diff
@@ -152,9 +152,9 @@ public RuntimeProcess(ILaunch launch, Process process, String name, Map<String,
 		fThreadNameSuffix = getPidInfo(process, launch);
 
 		fStreamsProxy = createStreamsProxy();
+		launch.addProcess(this);
```
mickaelistria, line 155: Please add a comment about the position being important here, or someone may easily miss it later and refactor in a regressing way.

HeikoKlare: Thanks, @mickaelistria! Makes sense to document that. I have added a comment in e289a48.
In case someone misses the comment, there will still be the added regression test that will fail.

iloveeclipse: It is not nice that the object exposes itself in constructor, but I guess this is nothing we will fix now.
May I suggest to **first** create a monitor (without starting it), just to finish init of the process fields that are supposed to be non-null, before adding the instance to the launch?

HeikoKlare: I fully agree. From my point of view, the constructor does more than a constructor is supposed to do. The process creation might be done by a factory that manages proper instantiation by (1) creating a `RuntimeProcess` with proper information from the `Launch`, (2) registering it at the `Launch`, (3) starting the process (in terms of creating the monitor thread). Then the `RuntimeProcess` would not even have to know about a `Launch`. Actually `RuntimeProcess` is already instantiated in factory methods, but since it's public API, everyone can instantiate it at any place.

Changing the instruction order to finish intialization before adding the instance to the launch totally makes sense. I have adapted that in 1fed561.

PR #600, opened 2023-07-26
Fix warnings in org.eclipse.debug.tests

HeikoKlare:
Contains no functional changes, only renames and other minor refactorings that remove warnings.

### DIFF
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchManagerTests.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchManagerTests.java
@@ -17,7 +17,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import org.eclipse.core.runtime.CoreException;
@@ -49,9 +48,9 @@ public class LaunchManagerTests extends AbstractLaunchTest {
 	 */
 	@Test
 	public void testGenereateConfigName() {
-		String configname = "launch_configuration"; //$NON-NLS-1$
-		String name = getLaunchManager().generateLaunchConfigurationName(configname);
-		assertTrue("the name nust be '" + configname + "'", name.equals(configname)); //$NON-NLS-1$ //$NON-NLS-2$
+		String originalName = "launch_configuration";
+		String generatedName = getLaunchManager().generateLaunchConfigurationName(originalName);
+		assertEquals("unexpected generated configuration name", originalName, generatedName);
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchManagerTests.java
```diff
@@ -83,8 +82,8 @@ public void testGenerateValidName() {
 	public void testGenerateConfigNameReservedName() {
 		if(Platform.OS_WIN32.equals(Platform.getOS())) {
 			String configname = "aux"; //$NON-NLS-1$
-			String name = getLaunchManager().generateUniqueLaunchConfigurationNameFrom(configname);
-			assertEquals("config name should be 'aux'", configname, name); //$NON-NLS-1$
+			String generatedConfigurationName = getLaunchManager().generateUniqueLaunchConfigurationNameFrom(configname);
+			assertEquals("config name should be 'aux'", configname, generatedConfigurationName); //$NON-NLS-1$
```
fedejeanne, line 86:
```suggestion
			assertEquals("config name should be '" + configname + "'", configname, generatedConfigurationName); //$NON-NLS-1$ //$NON-NLS-2$
```

HeikoKlare: I complete removed the unnecessary duplication of the expected names from the error messages in 9635d79.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/viewer/model/VisibleVirtualItemValidator.java
```diff
@@ -46,11 +46,12 @@ public int getLength() {
 	@Override
 	public boolean isItemVisible(VirtualItem item) {
 		int position = 0;
-		while (item.getParent() != null) {
-			position += item.getIndex().intValue();
-			item = item.getParent();
+		VirtualItem currentItem = item;
```
fedejeanne, line 49:
```suggestion
		VirtualItem rootItem = item;
```

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/viewer/model/VisibleVirtualItemValidator.java
```diff
@@ -46,11 +46,12 @@ public int getLength() {
 	@Override
 	public boolean isItemVisible(VirtualItem item) {
 		int position = 0;
```
fedejeanne, line 48:
```suggestion
		int depth= 0;
```

fedejeanne (approved):
lgtm, just some minor optional improvements desired. Thank you for cleaning up :)

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/viewer/model/VisibleVirtualItemValidator.java
HeikoKlare: It's actually about some kind of position and not about depth, as the index within a parent is considered as well.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/viewer/model/VisibleVirtualItemValidator.java
HeikoKlare: It is a variable that reflects the current item during tree traversal. It will have the role of the root item after traversal has completed but calling it `rootItem` during traversal would not be correct in my opinion.

fedejeanne: You're right, I read `currentItem.getParent()` and automatically thought of trees and depth.

fedejeanne (approved):

PR #602, opened 2023-07-27
Remove unnecessary print statements from Tests

CodeLtDave:
This commit enhances the test execution output by removing unnecessary console prints when running the "All runtime tests" configuration. It provides cleaner and more useful feedback during test execution.

Additionally, in the GithubBug_193 test, more descriptive error messages have been added to all assert statements. These messages will provide useful context when a test fails, making it easier to diagnose and fix issues.


HeikoKlare (approved):
Thanks for cleaning up test output and improving error messages!

Issue #525, opened 2023-06-28
Reactivate tests: disabled tests collection


HeikoKlare:
### Current Situation
Several tests have been deactivated over the years for different reasons, such as temporary incompatibilities with a specific platform/OS, bugs in the code, indeterministic behavior of the test etc.

These disablements are done by adding some prefix to `test...` methods in JUnit 3, by removing or commenting out `@Test` in JUnit 4 and 5, or by using `@Ignore` or `@Disabled`. In addition, several tests are only deactivated for specific platforms using early returns or assume statements within the test methods.

### Goal
Since in most cases there is no additional, unique documentation or tagging of these disablements, they are not easy to find. This issue is supposed to be an umbrella to document disabled tests (existing ones and potentially further ones).

The purpose of this issue is to collect the disabled tests to have a central overview and provide the chance to reactivate and, if necessary, fix them. It also serves as a "good first issue" (see https://github.com/eclipse-platform/eclipse.platform.text/pull/212#issuecomment-1594576748) for interested new contributors, which can take any of the collected tests and try to fix them to learn about the Eclipse platform code while, at the same time, already having the chance to make valuable contributions to the project.

### Process
I see the following ideas as good starting points for identifying disabled tests:
- First concentrate on a single project, such as `org.eclipse.core.tests.resources`
- Use different techniques to find disabled tests:
    1. Completely disabled tests:
        - Search for missing or commented out `@Test` annotation (JUnit 4 and 5)
        - Search for tests not starting with `test` (JUnit 3), in particular those starting with `_test`
    2. Tests disabled for specific platforms:
        - Search for accesses to OS constants in `Platform`, such as `OS_WIN32` and validate if they are used to disable tests for specific platforms
        - Search for accesses to methods `isLinux()` etc. in `ResourceTest` subclasses and validate if they are used to disable tests for specific platforms
        - Search for `assumeFalse` or `assumeTrue` statements that make tests only run on specific platforms

#### Examples
- The tests modified in https://github.com/eclipse-platform/eclipse.platform.text/pull/212
- Tests in `org.eclipse.core.tests.resources.regression.Bug_032076`
- The `skipTest` helper in `org.eclipse.core.tests.resources.usecase.SnapshotTest`

### Disabled in Maven
Some tests are disabled in Maven builds, such that integration builds do not reveal whether they fail or not.
[...]

PR #603, opened 2023-07-28
Reactivate tests #525

Michael5601:
Reactivation of working tests that were forgotten after the corresponding bug was fixed. 
Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/525


HeikoKlare (approved):
Thanks for finding and re-enabling tests that were unnecessarily disabled!

PR #605, opened 2023-07-28
Reactivate tests in org.eclipse.core.tests.resources #525

Michael5601:
In this commit, the performance test from the Marker Test class in org.eclipse.core.tests.resources is reactivated because it can be run quickly and successfully. Contributes to #525

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,13 @@
+name: CodeQL call
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: '15 8 * * 1'
+
+jobs:
+  callCodeQLworkflow:
+    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/codeQLworkflow.yml@master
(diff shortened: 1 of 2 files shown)


HeikoKlare:
The enabled method is no actual test method but only some code measuring and printing the time required for processing operations on markers. I gues it was used around 20 years ago when implementing the functionality to get some insights into the implementation's performance. I don't see that it still has any use but may only produce maintenance effort.

I would propose to change this PR (or replace it by a new one) so that the two `_testPerformance*` methods in the `MarkerTest` class are removed. @Michael5601 maybe you want to change it accordingly?

Michael5601:
> The enabled method is no actual test method but only some code measuring and printing the time required for processing operations on markers. I gues it was used around 20 years ago when implementing the functionality to get some insights into the implementation's performance. I don't see that it still has any use but may only produce maintenance effort.
> 
> I would propose to change this PR (or replace it by a new one) so that the two `_testPerformance*` methods in the `MarkerTest` class are removed. @Michael5601 maybe you want to change it accordingly?

I have added a new PR with the requested changes. (See #674)
This PR will be closed now because it has been replaced with #674.

Issue #601, opened 2023-07-27
failing Test PrebuiltIndexCompatibility.test9_4_2LuceneCompatible


jukzi:
on all platforms in i-build since I20230726-0640:

https://download.eclipse.org/eclipse/downloads/drops4/I20230726-0640/testresults/html/org.eclipse.ua.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20230726-0640/testresults/html/org.eclipse.ua.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20230726-0640/testresults/html/org.eclipse.ua.tests_ep429I-unit-win32-java17_win32.win32.x86_64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20230726-0640/testresults/html/org.eclipse.ua.tests_ep429I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html

merks:
There is PR open regarding that issue:

https://github.com/eclipse-platform/eclipse.platform/pull/595

I personally think it's useless to have a test method that tests only if the hard-coded version in that method is equal to the hard-coded version in the target platform.  That's just a maintenance problem, not a test...

PR #607, opened 2023-07-28
AntBuildfileContentDescriber: don't log error - fixes #592

### LINKED ISSUE #592: SaveableCompareEditorInputTest.testModifyAndSaveBothSidesOfCompareEditorXml fails since I20230725-1800
See https://download.eclipse.org/eclipse/downloads/drops4/I20230725-1800/testResults.php

Test `SaveableCompareEditorInputTest | testModifyAndSaveBothSidesOfCompareEditorXml` fails on all platforms since I20230725-1800

This is a regression from https://github.com/eclipse-platform/eclipse.platform/commit/5dc372a0c5002b7f22e5d49eaa1cbf0916455daf

The expected TextMergeViewer is not created for "xml" files, just NullViewer.

```
java.lang.NoSuchFieldException: fLeftat java.base/java.lang.Class.getDeclaredField(Class.java:2610)
at org.eclipse.compare.tests.ReflectionUtils.getField(ReflectionUtils.java:64)
at org.eclipse.team.tests.ui.SaveableCompareEditorInputTest.verifyModifyAndSaveBothSidesOfCompareEditor(SaveableCompareEditorInputTest.java:386)
at org.eclipse.team.tests.ui.SaveableCompareEditorInputTest.testModifyAndSaveBothSidesOfCompareEditorXml(SaveableCompareEditorInputTest.java:439)
```

@jukzi : please check.

jukzi:
https://github.com/eclipse-platform/eclipse.platform/issues/592

### DIFF
--- a/ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
+++ b/ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
@@ -20,7 +20,6 @@ import java.io.Reader;
 
 import javax.xml.parsers.ParserConfigurationException;
 
-import org.eclipse.ant.core.AntCorePlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExecutableExtension;
@@ -59,7 +58,7 @@ public final class AntBuildfileContentDescriber extends XMLContentDescriber impl
 		}
 		catch (SAXException e) {
 			// we may be handed any kind of contents... it is normal we fail to parse
-			AntCorePlugin.log(e);
+			// so we must not log any error here.
 			return INDETERMINATE;
 		}
 		catch (ParserConfigurationException e) {


jukzi:
errors only from https://github.com/eclipse-platform/eclipse.platform/issues/601

ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
```diff
@@ -59,7 +61,9 @@ private int checkCriteria(InputSource contents) throws IOException {
 		}
 		catch (SAXException e) {
 			// we may be handed any kind of contents... it is normal we fail to parse
-			AntCorePlugin.log(e);
+			// must not log an error - or test log listener will fail the test
+			IStatus status = new Status(IStatus.WARNING, AntCorePlugin.PI_ANTCORE, AntCorePlugin.INTERNAL_ERROR, "Error logged from Ant Core: ", e); //$NON-NLS-1$
```
HeikoKlare, line 65: I would suggest to have the message reflect that it is "only" a warning and not an error to have consistent message and severity. You could also add a specific error code for invalid XML like it is already present for malformed URL and the like rather than passing the generic `INTERNAL_ERROR` because this is no real error.

ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
```diff
@@ -59,7 +61,9 @@ private int checkCriteria(InputSource contents) throws IOException {
 		}
 		catch (SAXException e) {
 			// we may be handed any kind of contents... it is normal we fail to parse
-			AntCorePlugin.log(e);
+			// must not log an error - or test log listener will fail the test
```
HeikoKlare, line 64: Could you document the rationale for logging no error but a warning by the fact that the exception thrown here is actually "unexceptional" like mentioned  in the preceeding line ("it is normal we fail to parse") rather than referring to a failing test? The comment currently sounds as if this code is adapted to make some test work but actually the test just revealed that logging an error here may not be reasonable.

ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
HeikoKlare: nitpicky: Log entries (for ANT builds) seem to be provided as NLS strings via properties files that may be localized. Maybe this message should be provided as proper NLS as well?

HeikoKlare (commented):
I agree that no error should be logged when XML parsing fails as passing an invalid XML is "valid" here and logging it as warnings seems to be reasonable. I attach some minor thoughts on the proposed fix.

ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java
jukzi: that's just a copy of org.eclipse.ant.core.AntCorePlugin.log(Throwable) modified to IStatus.WARNING

HeikoKlare (approved):
Completely avoiding a log entry is also fine 👍 

The failing test has been fixed (would disappear after rebase on `master`).

Issue #168, opened 2022-09-13
eclipse is vulnerable to a DLL hijacking attack


vishnusarathashling:
When eclipse is executed it tries to load the following two files:
* unixfile_1_0_0.dll
* git.exe

The problem is that those 2 files are not found in the program’s directory and eclipse.exe tries to find them in the Windows system, then the directories listed in the system path, and finally in the directory where eclipse was launched (current directory).

Loading a DLL or EXE from the directory the program was launched (current directory) is considered a security vulnerability.

It would be better if we could do any of the below:

· Fix the code to not load the listed dll and exe from the current directory.

· Include the missing DLL and exe if they are needed.

vishnusarathashling:
@iloveeclipse @jukzi @akurtakov

akurtakov:
I think you're looking for @niraj-modi who is the windows guy.

vishnusarathashling:
Thanks @akurtakov for delegating me to the right person.

akurtakov:
@jonahgraham One more win32 area looking for help for your "help needed" list.

laeubi:
Just a general remark: If I'm able to place arbitrary executable or dll files on a system then all seems lost anyways...

jonahgraham:
@laeubi I used to be strongly of that opinion too, but the problem is we deliver a dev environment that users are not necessarily experts in using. This error seems similar to the security issue a few years ago of opening an XML (or similar) file that was compromised in the IDE leads your system to being compromised.

I think the common threat model is `git clone <compromised repo>` and then starting eclipse in that repo's directory. VS Code these days runs in a protected mode until you elevate permissions.

As a community Eclipse needs to be in a position that doing something like opening a file or launching eclipse cannot expose user to a vulnerable file, nor more than opening a dodgy word document should (as MS Word also runs in protected mode).

laeubi:
> Can't you get them easily from an update site? bundle-shape: dir and include a dll or exe?

If I can get you to download something malicious from an update-site I can execute arbitrary java code and it is cross-platform ... so why make it more complex :-)

vishnusarathashling:
Are we in a stage to add a fix for this vulnerability? Seems like we missed the chain at some point.

jukzi:
@vishnusarathashling can you provide a stacktrace where those files are loaded?

laeubi:
@jukzi you can take a look here:

[...]

PR #608, opened 2023-07-30
Add sanity checks to prevent DLL hijacking

laeubi:
Currently we only perform a soft check that a given OS is there and rely on failing to load a library, but this allows so called "DLL Hijack" attacs because the JVM search in several places if a library is not found.

This adds some more sanity check even before an attempt is made to load a library with the follwoing steps:

1) a suitable OS must be detected
2) a bundle with a well konw name has to be present in the system 3) the library is accessibly from the bundles classloader

Fix https://github.com/eclipse-platform/eclipse.platform/issues/168

FYI @vishnusarathashling @jonahgraham with this check I would assume the linked issue fixed, as `git.exe` is not searched by Eclipse Platform but by egi/jgit

@tjwatson I'm a bit curious as the bundles do not use any `Bundle-NativeCode` header in the bundle or in the fragment, is this a special extension of Equinox that it searches for `os/<name>/<library>` even if the header is not present? Should we probably migrate this code to using `Bundle-NativeCode` instead?

@jukzi @HannesWell I testes this under linux, could you check with windows that it works well under windows and early exit for the linux/mac checks?
@eclipse-platform/eclipse-platform-committers anyone with mac can verify the fix?

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNatives.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNatives.java
@@ -19,9 +19,16 @@ import java.util.Enumeration;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.IFileSystem;
 import org.eclipse.core.filesystem.provider.FileInfo;
-import org.eclipse.core.internal.filesystem.*;
+import org.eclipse.core.internal.filesystem.FileSystemAccess;
+import org.eclipse.core.internal.filesystem.Messages;
+import org.eclipse.core.internal.filesystem.Policy;
+import org.eclipse.core.internal.filesystem.local.unix.UnixFileNatives;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.osgi.service.environment.Constants;
 import org.eclipse.osgi.util.NLS;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
 
 abstract class LocalFileNatives {
 	private static boolean hasNatives = false;
@@ -33,19 +40,22 @@ abstract class LocalFileNatives {
 	private static final String LIBRARY_NAME = "localfile_1_0_0"; //$NON-NLS-1$
 
 	static {
-		try {
-			System.loadLibrary(LIBRARY_NAME);
-			hasNatives = true;
-			isUnicode = internalIsUnicode();
+		String os = System.getProperty("osgi.os", ""); //$NON-NLS-1$//$NON-NLS-2$
+		if (Constants.OS_WIN32.equals(os) && hasBundle("org.eclipse.core.filesystem.win32.x86_64") && isLibraryPresent()) { //$NON-NLS-1$
 			try {
-				nativeAttributes = nativeAttributes();
+				System.loadLibrary(LIBRARY_NAME);
+				hasNatives = true;
+				isUnicode = internalIsUnicode();
+				try {
+					nativeAttributes = nativeAttributes();
+				} catch (UnsatisfiedLinkError e) {
+					// older native implementations did not support this
+					// call, so we need to handle the error silently
+				}
 			} catch (UnsatisfiedLinkError e) {
-				// older native implementations did not support this
-				// call, so we need to handle the error silently
+				if (isLibraryPresent())
+					logMissingNativeLibrary(e);
 			}
-		} catch (UnsatisfiedLinkError e) {
-			if (isLibraryPresent())
-				logMissingNativeLibrary(e);
 		}
 	}
 
@@ -53,6 +63,21 @@ abstract class LocalFileNatives {
 		String libName = System.mapLibraryName(LIBRARY_NAME);
 		Enumeration<URL> entries = FileSystemAccess.findEntries("/", libName, true); //$NON-NLS-1$
 		return entries != null && entries.hasMoreElements();
+	}
+
+	private static boolean hasBundle(String bsn) {
+		Bundle bundle = FrameworkUtil.getBundle(UnixFileNatives.class);
+		if (bundle != null) {
+			BundleContext bundleContext = bundle.getBundleContext();
+			if (bundleContext != null) {
+				for (Bundle other : bundleContext.getBundles()) {
+					if (other.getSymbolicName().equals(bsn)) {
+						return true;
+					}
+				}
+			}
+		}
+		return false;
 	}
 
 	private static void logMissingNativeLibrary(UnsatisfiedLinkError e) {
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
@@ -18,10 +18,11 @@ import java.nio.file.FileSystems;
 import java.util.Set;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.provider.FileInfo;
-import org.eclipse.core.internal.filesystem.local.nio.*;
+import org.eclipse.core.internal.filesystem.local.nio.DefaultHandler;
+import org.eclipse.core.internal.filesystem.local.nio.DosHandler;
+import org.eclipse.core.internal.filesystem.local.nio.PosixHandler;
 import org.eclipse.core.internal.filesystem.local.unix.UnixFileHandler;
 import org.eclipse.core.internal.filesystem.local.unix.UnixFileNatives;
-import org.eclipse.osgi.service.environment.Constants;
 
 /**
  * <p>Dispatches methods backed by native code to the appropriate platform specific
@@ -57,12 +58,11 @@ public class LocalFileNativesManager {
 	 */
 	public static boolean setUsingNative(boolean useNatives) {
 		boolean nativesAreUsed;
-		boolean isWindowsOS = Constants.OS_WIN32.equals(LocalFileSystem.getOS());
 
-		if (useNatives && !isWindowsOS && UnixFileNatives.isUsingNatives()) {
+		if (useNatives && UnixFileNatives.isUsingNatives()) {
 			HANDLER = new UnixFileHandler();
 			nativesAreUsed = true;
-		} else if (useNatives && isWindowsOS && LocalFileNatives.isUsingNatives()) {
+		} else if (useNatives && LocalFileNatives.isUsingNatives()) {
 			HANDLER = new LocalFileHandler();
 			nativesAreUsed = true;
 		} else {
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/unix/UnixFileNatives.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/unix/UnixFileNatives.java
@@ -18,13 +18,20 @@ import java.io.File;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
+import java.util.regex.Pattern;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.provider.FileInfo;
-import org.eclipse.core.internal.filesystem.*;
+import org.eclipse.core.internal.filesystem.FileSystemAccess;
+import org.eclipse.core.internal.filesystem.Messages;
+import org.eclipse.core.internal.filesystem.Policy;
 import org.eclipse.core.internal.filesystem.local.Convert;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.osgi.service.environment.Constants;
 import org.eclipse.osgi.util.NLS;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
 
 public abstract class UnixFileNatives {
 	private static final String LIBRARY_NAME = "unixfile_1_0_0"; //$NON-NLS-1$
@@ -36,13 +43,19 @@ public abstract class UnixFileNatives {
 	private static final int libattr;
 
 	static {
+		String os = System.getProperty("osgi.os", ""); //$NON-NLS-1$//$NON-NLS-2$
 		boolean _usingNatives = false;
 		int _libattr = 0;
 		try {
-			System.loadLibrary(LIBRARY_NAME);
-			_usingNatives = true;
-			initializeStructStatFieldIDs();
-			_libattr = libattr();
+			//To prevent loading on systems that do not support unixfile first check if this is a supported os and the fragment is present that ships the native...
+			boolean isLinux = Constants.OS_LINUX.equals(os) && hasBundle("org\\.eclipse\\.core\\.filesystem\\.linux\\.(?:aarch64|loongarch64|ppc64le|x86_64)"); //$NON-NLS-1$
+			boolean isMacOs = Constants.OS_MACOSX.equals(os) && hasBundle("org\\.eclipse\\.core\\.filesystem\\.macosx"); //$NON-NLS-1$
+			if ((isLinux || isMacOs) && isLibraryPresent()) {
+				System.loadLibrary(LIBRARY_NAME);
+				_usingNatives = true;
+				initializeStructStatFieldIDs();
+				_libattr = libattr();
+			}
 		} catch (UnsatisfiedLinkError e) {
 			if (isLibraryPresent())
 				logMissingNativeLibrary(e);
@@ -56,6 +69,22 @@ public abstract class UnixFileNatives {
 		String libName = System.mapLibraryName(LIBRARY_NAME);
 		Enumeration<URL> entries = FileSystemAccess.findEntries("/", libName, true); //$NON-NLS-1$
 		return entries != null && entries.hasMoreElements();
+	}
+
+	private static boolean hasBundle(String pattern) {
+		Pattern compile = Pattern.compile(pattern);
+		Bundle bundle = FrameworkUtil.getBundle(UnixFileNatives.class);
+		if (bundle != null) {
+			BundleContext bundleContext = bundle.getBundleContext();
+			if (bundleContext != null) {
+				for (Bundle other : bundleContext.getBundles()) {
+					if (compile.matcher(other.getSymbolicName()).matches()) {
+						return true;
+					}
+				}
+			}
+		}
+		return false;
 	}
 
 	private static void logMissingNativeLibrary(UnsatisfiedLinkError e) {


laeubi:
To test this one should put these breakpoints:

1. UnixFileNatives line 46
2. LocalFileNatives line 43
3. LocalFileNativesManager line 62

depending on the OS you should hit different lines breakpoint and check:

1. it never reaches the point of loading a native lib of an unsupported os (early exit)
2. the native library is still loaded for the matching os

then remove the `org.eclipse.core.filesystem.linux.<your os>` fragment and start again, now you should see that no native lib is ever attempted to be loaded and it falls back to using nio.

jukzi (changes requested):
For windows: `Use fully qualified paths for all calls to LoadLibrary` [1]
I can't see that the patch is doing that.

[1] https://support.microsoft.com/en-us/topic/secure-loading-of-libraries-to-prevent-dll-preloading-attacks-d41303ec-0748-9211-f317-2edc819682e1

laeubi:
@jukzi this is performed by Equinox (under the hood) as the dll is not on a path but inside a jar, so the goal here is that we don't try to load the lib at all if it will not be found (that is what `isLibraryPresent` does).

jukzi:
System.loadLibrary uses multiple paths while System.load() would use absolute path.

laeubi:
> System.loadLibrary uses multiple paths while System.load() would use absolute path.

The Bundle can't use `System.load()` (as there is no absolute path) this is done by `ClasspathManager` of equinox, so we only need to make sure the library will be found by the classpath so the JVM never perform searches anywhere...

HannesWell:
> @jukzi @HannesWell I testes this under linux, could you check with windows that it works well under windows and early exit for the linux/mac checks?

Looks good on Windows. In the static initializer of `UnixFileNatives` isLinux and isMacOs are both false, thus no library load is attempted there and usingNatives is just false. in `LocalFileNatives` the initializer succeeds (although I don't know from where the  `localfile_1_0_0` is loaded and hasNatives is true.

jukzi:
The file is still searched on multiple loacations. Thats wrong:
image
Windows can't load a library from inside a jar. https://stackoverflow.com/questions/4113317/load-library-from-jar

laeubi:
> Windows can't load a library from inside a jar

Windows can't but OSGi can ... please put a breakpoint and check why OSGi is not queried first, if the default search path is used something is wrong and you should see the lib not being loaded at all, so probably there is a bug in Equinox.

jukzi:
OSGi can't when OS can't. it needs to extract file and load it from exactly that single location.

laeubi:
> OSGi can't when OS can't. it needs to extract file and load it from exactly that single location.

If you look at your screenshot it already does exactly that...

merks:
Please forgive me if this is a stupid idea.  Can't org.osgi.framework.Bundle.getResource(String) be used to find the library on the classpath if it exists on the classpath, and then you have the full location?  I.e., like this:

image

laeubi:
This only works if the file is already extracted (e.g. from your workspace) otherwise OSGi will extract is from the jar and use that file, so I don't see any benefit from replicate what OSGi/Equinox is already doing.

merks:
Of course one can use FileLocation.toFileURL too.  I assume, as you suggest, that that's what's done under the covers otherwise.  It seems simpler, i.e. ,fewer changes.  But I don't know if it's actually safer?  It's just a thought...

laeubi:
If it is not safe it should be made more safe by Equinox... so unless someone can show that the dll is issued for loading **outside the fragment of the bundle** I don't see a reason for making things more complex and less portable.

jukzi:
#168 complained that the native libraries are tried to load from multiple location. This patch does not change that. This patch assumes that equinox loads a library from a specific resource location. But equinox does not do that. it tries to find libraries from hooks, several settings and from all org.eclipse.osgi.internal.framework.EquinoxConfiguration.ECLIPSE_LIB_VARIANTS.

merks:
The comment "if it is not safe" suggests that it being safe after this change has not been determined.  I'm not sure if Equinox can make java.lang.System.loadLibrary(String) safe, can it?  

Certainly this type of logic avoids the issue entirely:
```
	static {
		try {
			Bundle bundle = FrameworkUtil.getBundle(LocalFileNatives.class);
			URL resource = bundle.getResource("os/win32/x86_64/localfile_1_0_0.dll"); //$NON-NLS-1$
			System.out.println(resource);
			if (resource != null) {
				Path realPath = Path.of(FileLocator.toFileURL(FileLocator.resolve(resource)).toURI()).toRealPath();
				System.load(realPath.toString());
				System.out.println(realPath);

				// System.loadLibrary(LIBRARY_NAME);
				hasNatives = true;
				isUnicode = internalIsUnicode();
				try {
					nativeAttributes = nativeAttributes();
				} catch (UnsatisfiedLinkError e) {
					// older native implementations did not support this
					// call, so we need to handle the error silently
				}
			}
		} catch (UnsatisfiedLinkError e) {
			if (isLibraryPresent())
				logMissingNativeLibrary(e);
		} catch (IOException | URISyntaxException e1) {
			//$FALL-THROUGH$
		}
	}
```

laeubi:
> complained that the native libraries are tried to load from multiple location

It does not, quote from the original with emphasis by me:

> The problem is that those 2 files are not found in the program’s directory and eclipse.exe tries to **find them in the Windows system**, then the **directories listed in the system path**, and finally **in the directory where eclipse was launched** (current directory).

it does not complain about searching **inside** eclipse, what is desired and what is not a problem at all.

>  I'm not sure if Equinox can make java.lang.System.loadLibrary(String) safe, can it?

@merks please look into how `System.loadLibrary(String)` works... it determines the caller class and the uses the classloader of the caller to search the library and this classlaoder is provided by equinox that the do all the magic "search for the library and possibly extract it from the jar". So how if that is not safe, it should be fixed at the root and not on every location that call is used...

> Certainly this type of logic avoids the issue entirely:

For windows this might seem simpler as **currently** there is only one DLL but please look at the Unix part where we have different OS (linux/macos) and a lot of different architectures, so one effectively has to replicate a lot of the code that Equinox already do to search the "right" library.

merks:
This whole thing seems specific to Windows...

Anyway, I do see a similar problem here:

image

Maybe that branch logic doesn't happen in a real installation but only in a debug launch...

If Equinox ensures that the searching is not done outside the classpath, then that seems okay.  But then I wonder why anything needs to change...

jukzi:
As far as i know Eclipse IDE loads only these two natives: filesystem and swt. Since SWT does not require OSGi it would need another solution then equinox.

laeubi:
> This whole thing seems specific to Windows...

this only depends on because we first check for windows and then for Linux, but if the order is changed somewhere, we add another supported OS/Arch or the fragment is not there at all it will also show up on other os.

> Anyway, I do see a similar problem here:

The "problem" itself is intrinsic to the JVM, the issue **here** is that we try to load a library that will **never** be there on windows  and *after that* try another name (even though I have no clue why one has used a different name for the linux variant) .

So what happens is that on windows there will always be a request to a lib that will never succeed (unless someone swap the order of the ifs).

laeubi:
> Since SWT does not require OSGi it would need another solution then equinox

SWT still uses OSGi, so if one would step into will see that there is more to happen...

jukzi:
uh oh, what are all those llibraries?
image

laeubi:
> uh oh, what are all those llibraries?

These are extracted from your bundles to load them with an absolute path by Equinox... anyways can we try to focus on this particular issue, that is about **we load a library to find out it is available** and not DLL loading in general?!

jukzi:
> Certainly this type of logic avoids the issue entirely:

Even though that solution looks very clean bundle.getResource() does similar obscure file accesses to try around (in debug mode):
image

laeubi:
> Even though that solution looks very clean bundle.getResource() does similar obscure file accesses to try around (in debug mode):

Can you please explain why it should be "obscure"? This is normal behaviour as we are searching thing **inside** the bundle while the problem to solve is that things are searched **outside** the bundle...

jukzi:
obscure: to search for a thing while we know the exact location

laeubi:
> obscure: to search for a thing while we know the exact location

We don't know the location, because this includes classpath scanning and if you know a way to find a thing in the classpath without scanning the classpath then I'm sure many people would be interested in such novel approach...

Also please note that in the case of running from the IDE *more* things are searched than in the real runtime because the classpath also includes special development entries...

merks:
Yes, I agree with @laeubi on his comment/question.  Of course one must search in a bunch of places when looking for something on the classpath, that's completely normal and of course it only searches the classpath, right?

In any case @jukzi your comment suggests that you know the solution.  Perhaps you could sketch it out?

merks:
FYI, I tried this approach in the one place that Oomph loads a dll:
```
        try
        {
          Bundle bundle = Platform.getBundle("org.eclipse.oomph.jreinfo.win32.x86_64"); //$NON-NLS-1$
          URL libraryURL = bundle.getEntry("jreinfo.dll"); //$NON-NLS-1$
          JREInfoPlugin.INSTANCE.log("jreinfo.dll url: " + libraryURL);
          URL fileLibraryURL = FileLocator.toFileURL(libraryURL);
          JREInfoPlugin.INSTANCE.log("jreinfo.dll file URL: " + fileLibraryURL);
          String libraryPath = Path.of(fileLibraryURL.toURI()).toRealPath().toString();
          JREInfoPlugin.INSTANCE.log("jreinfo.dll location: " + libraryPath);
          System.load(libraryPath);
        }
        catch (Exception ex)
        {
          JREInfoPlugin.INSTANCE.log(ex);
        }
```
I.e., I know which bundle (fragment) contains the *.dll and I know the exact path within that jar for the *.dll, so I can use getEntry to get exactly and only that entry from that jar, and then I can force it to be a file URL that I can turn into path that I can load exactly and only that library. 

I tested that it works in the installer via this logging:
```
!ENTRY org.eclipse.oomph.jreinfo 1 0 2023-07-31 14:35:57.804
!MESSAGE jreinfo.dll url: bundleentry://146.fwk912208971/jreinfo.dll
!ENTRY org.eclipse.oomph.jreinfo 1 0 2023-07-31 14:35:57.808
!MESSAGE jreinfo.dll file URL: file:/C:/Users/merks/AppData/Local/Temp/eoi3DD1.tmp/configuration/org.eclipse.osgi/146/0/.cp/jreinfo.dll
!ENTRY org.eclipse.oomph.jreinfo 1 0 2023-07-31 14:35:57.810
!MESSAGE jreinfo.dll location: C:\Users\merks\AppData\Local\Temp\eoi3DD1.tmp\configuration\org.eclipse.osgi\146\0\.cp\jreinfo.dll
```
So in principle it's possible, but it feels a little over kill.  But at least in Oomph, I'd rather err on the side of caution so that I never need to revisit this issue again...

jukzi:
I know the file is currently in the (fragment) bundle "org.eclipse.core.filesystem.win32.x86_64" in the path "/os/win32/x86_64/localfile_1_0_0.dll" and no where else. But i don't know how to get a resource from a specific fragment. Maybe it should not be a fragment.

jukzi:
> jreinfo

that works like it should - single location:

image

merks:
So here is a possible **rough**, brute-force approach/solution that does no searching:
```
	static {
		Bundle bundle = FrameworkUtil.getBundle(LocalFileNatives.class);
		Optional<Bundle> fragment = Arrays.stream(bundle.getBundleContext().getBundles()).filter(it -> "org.eclipse.core.filesystem.win32.x86_64".equals(it.getSymbolicName())).findFirst(); //$NON-NLS-1$
		fragment.ifPresent(it -> {
			try {
				URL entry = it.getEntry("os/win32/x86_64/localfile_1_0_0.dll"); //$NON-NLS-1$
				System.out.println(entry);
				if (entry != null) {
					Path realPath;
					try {
						realPath = Path.of(FileLocator.toFileURL(FileLocator.resolve(entry)).toURI()).toRealPath();
						System.load(realPath.toString());

						hasNatives = true;
						isUnicode = internalIsUnicode();
						try {
							nativeAttributes = nativeAttributes();
						} catch (UnsatisfiedLinkError e) {
							// older native implementations did not support this
							// call, so we need to handle the error silently
						}
					} catch (IOException | URISyntaxException e) {
						//$FALL-THROUGH$
					}
				}
			} catch (UnsatisfiedLinkError e) {
				if (isLibraryPresent())
					logMissingNativeLibrary(e);
			}
		});
	}
```

tjwatson:
> @tjwatson I'm a bit curious as the bundles do not use any Bundle-NativeCode header in the bundle or in the fragment, is this a special extension of Equinox that it searches for os/<name>/<library> even if the header is not present? Should we probably migrate this code to using Bundle-NativeCode instead?

This predates OSGi.  Eclipse platform used to implement its own class loader and had this policy of loading according to environment (os/arch/ws etc.) directories within the plugin JAR.  There are challenges to moving to `Bundle-NativeCode` header (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=201700).  But regardless I don't see that as solving what is being solved here (which I am unclear on).

I don't see what the proposed changes are trying to address.

tjwatson:
> So here is a possible **rough**, brute-force approach/solution that does no searching:

If System.loadLibrary is broken (insecure) then we may as well stop using Java until they "fix" it.

jukzi:
> If System.loadLibrary is broken (insecure) then we may as well stop using Java until they "fix" it.

Java provides System.load() as alternative - which accepts full qualified file name.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#load(java.lang.String)

tjwatson:
> > If System.loadLibrary is broken (insecure) then we may as well stop using Java until they "fix" it.
> 
> Java provides System.load() as alternative - which accepts full qualified file name.
> 
> https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#load(java.lang.String)

But that bypasses anything that the OSGi header `Bundle-NativeCode` would do to find the library in the bundle and requires you to basically repeat all that is done by the framework to setup (extract) the library and place it somewhere to load.  That is more complicated that first observed because the framework also handles when the library needs reloaded by re-extracting to unique locations if the bundle is refreshed.  Is the basic problem that anything that calls System.loadLibrary is broken?

laeubi:
@tjwatson the discussion has sadly diverged from the original topic so here is a summary:

1. We have two "Variants" of a lib, one named "localfile" (windows only) and one named "unixfile" (linux/mac)
2. Now Eclipse first tries to load "unixfile", if that fails it tries to load "localfile" and if that fails fall back to a java only impl

The problem is that if you run on windows, the jvm now always searches for "unixfile.dll" and this is the done as it is found nowhere on different places.

My PR now does the following:

1. Does the OS match
2. Do we have a bundle with a known name
3. Do the classlaoder of our bundle can find an entry with the libname

If all match do the usual stuff because now we are sure the JVM will **not** search for something that can't be found, for the reasons you mentioned I do not really want to use more "special" handling here and instead let OSGi do the underlying ground work, it should just be prevented to load **anything** if we know it can't succeed.

jonahgraham:
Searching is not a problem - it is the where the searching happens that matters. It is totally ok to search through bundles, it is not ok to have unconstrained searches that include CWD or any other location with less privileges than eclipse's location itself.

---

> Now Eclipse first tries to load "unixfile", if that fails it tries to load "localfile" and if that fails fall back to a java only impl

That said, I don't understand why this change is needed. Is the OP correct that unixfile is being searched/loaded on Windows? I don't see that happening and LocalFileNativesManager.setUsingNative(boolean) only uses UnixFileNatives when not on Windows already. 

This change does seem to add some protection against a broken install (where the platform fragment is missing). But it seems it would be nice to do this universally, e.g. stop using loadLibrary everywhere in Eclipse and only use load. 

---

> But that bypasses anything that the OSGi [...]

There is a wider problem with native loading - if you do System.loadLibrary(x) where x is a library that should be in your bundle somewhere, then if it is missing the JVM falls back to searching java.library.path (defaulted to PATH on Windows).

This could be prevented by doing something like this (untested) that still uses all the OSGi stuff, but prevents the fallback searching on PATH:

```java

ClassLoader cl = this.getClassLoader();
String fullPath = cl.findLibrary(name);
if (fullPath == null) {
   // throw an error
}
return System.load(fullPath);
```

If this was provided somewhere standard then all the System.loadLibrary() calls could be replaced with this.

laeubi:
> That said, I don't understand why this change is needed. Is the OP correct that unixfile is being searched/loaded on Windows? I don't see that happening and LocalFileNativesManager.setUsingNative(boolean) only uses UnixFileNatives when not on Windows already.

As described in the PR there was a little check involved now that "fixes" it by checking the OS... this PR just enhances these checks and move them to the static initializer so this can't happen again by accident e.g. if someone refactor the code.

It also guards against the case the no fragment is present at all (e..g you delete the `.win32` or `.linux.xxx` fragments from your install.

jonahgraham:
I don't have a problem with this change - it fixes one of the cases of dlls being loaded badly. I hope that we can resolve the universal problem at some point so we can have a secure installation possible in the future.

tjwatson:
> This could be prevented by doing something like this (untested) that still uses all the OSGi stuff, but prevents the fallback searching on PATH:
> 
> ```java
> ClassLoader cl = this.getClassLoader();
> String fullPath = cl.findLibrary(name);
> if (fullPath == null) {
>    // throw an error
> }
> return System.load(fullPath);
> ```

Perhaps the `org.eclipse.core.runtime.FileLocator` can have a method

```java
  public static String findLibrary(String);
```

And this method would use the caller's class loader to do what you suggest.  But I suggest we force the caller to still call System.load:

```java
    return System.load(FileLocator.findLibrary(name));
```

laeubi:
I just wanted to add that currently `org.eclipse.equinox.common` is not a dependency of the filesystem and it even contains code that says there should be no dependency to e.g. `Platform` even though I don't know why. (see for example org.eclipse.core.internal.filesystem.FileSystemAccess.findEntries(String, String, boolean))

Nerveless I don't see why it should be better to call it directly instead of leaving it ot osgi, because with the `#isLibraryPresent()` call we already make sure the classloader will find the library and therefore not search elsewhere, so what case do we really want to solve here?

laeubi:
> I am unsure OSGi would add that, but I also doubt we want to wait for it.

The code already heavily depends on Equinox then anyways so it won't be a problem to use an equinox specific class until it is there...

laeubi:
And please be aware that there is not **ONE** file even though everyone uses the windows variant because it is super simple as it searches for exactly **ONE** library, please also take a look at `UnixFileNatives` where we need to support:

- macos
- linux
- aarch64
- loongarch64
- ppc64le
- x86_64

fragments...

tjwatson:
> And please be aware that there is not **ONE** file even though everyone uses the windows variant because it is super simple as it searches for exactly **ONE** library, please also take a look at `UnixFileNatives` where we need to support:

Not sure whom should look at this.  If using `ClassLoader.findLibrary` under the covers then it should search just like System.loadLibrary was.  There is a complication that `ClassLoader.findLibrary` is a protected which will force the use of reflection to call on an arbitrary ClassLoader object.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNatives.java
```diff
@@ -55,6 +65,21 @@ private static boolean isLibraryPresent() {
 		return entries != null && entries.hasMoreElements();
 	}
 
+	private static boolean hasBundle(String bsn) {
+		Bundle bundle = FrameworkUtil.getBundle(UnixFileNatives.class);
+		if (bundle != null) {
```
Wittmaxi, line 70: Can we reduce the level of nesting by inverting the conditions?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
```diff
@@ -57,12 +58,11 @@ public static void reset() {
 	 */
 	public static boolean setUsingNative(boolean useNatives) {
 		boolean nativesAreUsed;
-		boolean isWindowsOS = Constants.OS_WIN32.equals(LocalFileSystem.getOS());
 
-		if (useNatives && !isWindowsOS && UnixFileNatives.isUsingNatives()) {
+		if (useNatives && UnixFileNatives.isUsingNatives()) {
```
Wittmaxi, line 62: What about simplifying the nesting?

```java
if (useNatives) {
  HANLDER = new UnixFileHandler();
  nativesAreUsed = LocalFileNatives.isUsingNatives() || UnixFileNaties.isUsingNatives();
} else {
  nativesAreUsed = false;
 ...
}
```

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/unix/UnixFileNatives.java
```diff
@@ -58,6 +71,22 @@ private static boolean isLibraryPresent() {
 		return entries != null && entries.hasMoreElements();
 	}
 
+	private static boolean hasBundle(String pattern) {
+		Pattern compile = Pattern.compile(pattern);
+		Bundle bundle = FrameworkUtil.getBundle(UnixFileNatives.class);
+		if (bundle != null) {
```
Wittmaxi, line 77: See above, can we simplify the nesting by inverting the conditions?

Wittmaxi (changes requested):

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java
laeubi: This will change the behavior, e.g. it creates the handler and is afterwards checking if it actually is sued, that does not make much sense to me, the whole purpose is to prevent the handler from being created and even try to load anything...

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/unix/UnixFileNatives.java
laeubi: I don't see what should be "inverted" here, can you please explain in more detail?

Wittmaxi: naively, I would write 
```java
if (bundle == null)
  return false;
BundleContext bundleContext = bundle.getBundleContext();
if (bundleContext == null)
  return false;
for (Bundle other : bundleContext.getBundles()) {
  if (compile.matcher(other.getSymbolicName()).matches()) {
    return true;
  }
}
return false;
```

PR #610, opened 2023-07-31
Add SECURITY.md

mrybczyn:
Add a SECURITY.md pointing to the main reference file. SECURITY.md contains important information on how to report a vulnerability in the Projects and helps security researchers to find the appropriate communication method.


jukzi:
why is that link needed? when i currently look at https://github.com/eclipse-platform/eclipse.platform/security/policy then it already shows the same info!

PR #615, opened 2023-08-02
Move calls to lock::isConflicting in DeadlockDetector to the end

fedejeanne:
Calls to `ISchedulingRule::isConflicting` can be expensive, especially when they end up calling `ISchedulingRule::contains` and it happens to be a resource container (a workspace, a folder, etc) since they check all children (see those 2 methods in the class
`org.eclipse.core.internal.resources.Resource`).

This PR moves 3 of such calls in the `DeadlockDetector` class all the way to the end and takes advantage of the short-circuit logic in the if-statements, which might end up not calling `isConflicting` at all and saving some computation time.

Contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/983

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/DeadlockDetector.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/DeadlockDetector.java
@@ -17,7 +17,10 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.ArrayList;
 import org.eclipse.core.internal.runtime.RuntimeLog;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
Nice improvement to optimize the evaluation order in the conditions!

Just out of interest: do you have some estimation how high the performance impact is (e.g., with respect to https://github.com/eclipse-platform/eclipse.platform.ui/issues/983)?

fedejeanne:
> Nice improvement to optimize the evaluation order in the conditions!
> 
> Just out of interest: do you have some estimation how high the performance impact is (e.g., with respect to [eclipse-platform/eclipse.platform.ui#983](https://github.com/eclipse-platform/eclipse.platform.ui/issues/983))?

Thank you!

I don't have any real values for it, I just went with my gut. I guess I should have avoided creating high expectation and said _they are unnecessary_ instead of _they are expensive_. Ooops 😅

HeikoKlare:
Even without significant performance improvements, changing the order makes sense to me. The conflict checks seem to be the most expensive of all partial conditions, so it's fine to execute them last. I don't see any order dependencies between or side effects imposed by the partial conditions, so the order should be irrelevant except for performance and comprehensibility.

HeikoKlare (approved):

PR #618, opened 2023-08-03
Reactivate tests #525

Michael5601:
This change reactivates all working disabled tests from project org.eclipse.core.tests.net. The status of the corresponding but is CLOSED WONTFIX but the test runs successfully.  Contributes to #525


Michael5601:
The failing test is random failing. See Issue: https://github.com/eclipse-platform/eclipse.platform/issues/629.

HeikoKlare (changes requested):
The test is not executed in CI builds, as the whole test bundle is deactivated during Maven builds (see https://github.com/eclipse-platform/eclipse.platform/issues/525#issue-1779240019).

The test run fine when started from Eclipse on Windows systems, so the bug has probably been fixed as it appeared on every system (in particular also on Windows), according to the original bug report.
Since the test was disabled because of a that was not actively fixed and since we won't get continuous feedback on whether the test runs fine on all platforms, please re-add a comment with the bug number to the test case explaining that the test once failed due to that bug. In case the test will fail again locally or in integration builds once the bundle can be executed by Tycho, it is easier to find out the reasons when having the original bug at hand.

Michael5601:
> The test is not executed in CI builds, as the whole test bundle is deactivated during Maven builds (see [#525 (comment)](https://github.com/eclipse-platform/eclipse.platform/issues/525#issue-1779240019)).
> 
> The test run fine when started from Eclipse on Windows systems, so the bug has probably been fixed as it appeared on every system (in particular also on Windows), according to the original bug report. Since the test was disabled because of a that was not actively fixed and since we won't get continuous feedback on whether the test runs fine on all platforms, please re-add a comment with the bug number to the test case explaining that the test once failed due to that bug. In case the test will fail again locally or in integration builds once the bundle can be executed by Tycho, it is easier to find out the reasons when having the original bug at hand.

I included the proposed comment for this test.

HeikoKlare (approved):

HeikoKlare:
Revert is provided via #633. It's interesting that the tests are explicitly executed on Jenkins while they are disabled in the Maven build. I was not aware that we execute a different test set during Jenkins integration tests. Do you know how/where those test executions are documented, @iloveeclipse? Are they defined via the `test.xml`?

iloveeclipse:
The (migrated from eclipse.org) https://github.com/eclipse-platform/eclipse.platform.releng/wiki should in theory have an article explaining that, but it doesn't or I'm unable to find this information.

Whoever setup Jenkins / maven build seem not provided explanation how that all is supposed to work together.

SDK build is driven by some shell + maven + ant scripts and **I believe** later are using what is configured in test.xml.

HeikoKlare:
Thanks for the input!

PR #621, opened 2023-08-03
Reactivate tests in org.eclipse.ua.tests #525

Michael5601:
This change reactivates all working disabled tests from project org.eclipse.ua.tests. Bug 210024 has status CLOSED WONTFIX but test still runs successfully. Bug 339274 has status RESOLVED FIXED. Contributes to #525


HeikoKlare (approved):
Thanks for finding the disabled tests and reactivating them.
They succeed locally when started within Eclipse and they are executed successfully in Maven builds. Tests succeed in CI builds on all platforms.

PR #623, opened 2023-08-04
Disable the automatic snapshots for TestBug297635

fedejeanne:
Automatic snapshots might cause `TestBug297635::testBug` to fail if the snapshot is taken right before `assertStateTreesIsNotNull`. The reason for this is that a by taking a snapshot, `org.eclipse.core.internal.resources.SaveManager.forgetSavedTree(String)` will be called and it will delete the state-trees.

This PR disables the automatic snapshots while running tests in TestBug297635 to avoid such random failures.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/460

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
@@ -17,7 +17,10 @@ package org.eclipse.core.internal.resources;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
 import org.eclipse.core.resources.ISaveContext;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 
 /**
@@ -28,6 +31,7 @@ public class DelayedSnapshotJob extends Job {
 	private static final String MSG_SNAPSHOT = Messages.resources_snapshot;
 	private SaveManager saveManager;
 	private Workspace workspace;
+	private volatile boolean suspended;
 
 	public DelayedSnapshotJob(SaveManager manager, Workspace workspace) {
 		super(MSG_SNAPSHOT);
@@ -44,7 +48,7 @@ public class DelayedSnapshotJob extends Job {
 	public IStatus run(IProgressMonitor monitor) {
 		if (monitor.isCanceled())
 			return Status.CANCEL_STATUS;
-		if (!workspace.isOpen()) {
+		if (!workspace.isOpen() || suspended) {
 			return Status.OK_STATUS;
 		}
 		try {
@@ -55,5 +59,23 @@ public class DelayedSnapshotJob extends Job {
 			saveManager.operationCount = 0;
 			saveManager.snapshotRequested = false;
 		}
+	}
+
+	/**
+	 * Suspend automatic snapshots until {@link #resume()} is called.
+	 *
+	 * @see #resume()
+	 */
+	void suspend() {
+		suspended = true;
+	}
+
+	/**
+	 * Resume automatic snapshots. This method does not schedule this Job.
+	 *
+	 * @see #suspend()
+	 */
+	void resume() {
+		suspended = false;
 	}
 }
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
@@ -1464,6 +1464,22 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 	}
 
 	/**
+	 * Tries to determine if the workspace has changes and calls
+	 * {@link #snapshotIfNeeded(boolean)}. If an exception occurs while trying to
+	 * determine whether or not the workspace has changes then the snapshot might be
+	 * taken anyway, the workspace will be simply assumed unchanged.
+	 */
+	private void snapshotIfNeeded() {
+		boolean workspaceHasTreeChanges = false;
+		try {
+			workspaceHasTreeChanges = workspace.hasTreeChanges();
+		} catch (CoreException e) {
+			Policy.log(e);
+		}
+		snapshotIfNeeded(workspaceHasTreeChanges);
+	}
+
+	/**
 	 * Performs a snapshot if one is deemed necessary.
 	 * Encapsulates rules for determining when a snapshot is needed.
 	 * This should be called at the end of every top level operation.
@@ -2204,5 +2220,14 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 		output.writeLong(workspace.nextMarkerId.get());
 		// save the registered sync partners in the synchronizer
 		((Synchronizer) workspace.getSynchronizer()).savePartners(output);
+	}
+
+	public void suspendSnapshotJob() {
+		snapshotJob.suspend();
+	}
+
+	public void resumeSnapshotJob() {
+		snapshotJob.resume();
+		snapshotIfNeeded();
 	}
 }
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
@@ -1578,11 +1578,7 @@ public class Workspace extends PlatformObject implements IWorkspace, ICoreConsta
 				// build() and snapshot() should not fail if they are called.
 				workManager.rebalanceNestedOperations();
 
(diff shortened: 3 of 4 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
```diff
@@ -28,6 +31,7 @@ public class DelayedSnapshotJob extends Job {
 	private static final String MSG_SNAPSHOT = Messages.resources_snapshot;
 	private SaveManager saveManager;
 	private Workspace workspace;
+	private boolean suspended;
```
iloveeclipse, line 34: this must be volatile

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
```diff
@@ -47,6 +51,10 @@ public IStatus run(IProgressMonitor monitor) {
 		if (!workspace.isOpen()) {
 			return Status.OK_STATUS;
 		}
+		if (suspended) {
+			System.out.println("Automatic snapshots are suspended... skipping"); //$NON-NLS-1$
```
iloveeclipse, line 55: and system.out must be removed

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
szarnekow: You might make this an integer and just increment and decrement it when `suspend` or `resume` are being called. This would relieve existing client code from checking the `isSuspended` state before resuming again. `isSuspended` might become obsolete even.

fedejeanne: I made it an `AtomicInteger`, thanks for the tips to you both!

HeikoKlare (changes requested):
I have some general remarks to this proposal:
- The `DelayedSnapshotJob` is completely hidden by the `SaveManager`. The new API exposes that there is some snapshot job running, such that the user has to know about _all_ potential (side) effects of a `resume` and `suspend` operation. Currently, these effects are not documented.
- I am not sure whether the changes lead to consistent and expectable behavior w.r.t. snapshots. For example...
    - Shouldn't the `resume` also at least schedule the job if a snapshot was requested in between?
    - Shouldn't the `SaveManager.shutdown` still write a snapshot? It uses the snapshop job to do so, which was okay by now as the job was completely hidden/managed by the class. By now it can be influenced from outside, which has an unintended side effect on the shutdown behavior.

All in all, I am not sure if a single test is worth exposing internals of the `SaveManager` with all the efforts for documenting and adapting behavior accordingly. At least, I would expect a consistent, documented behavior of the extended `SaveManager`.
Are there further scenarios where disabling the snapshot job would be required, such that these changes are justified by being an enabler for further things?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/DelayedSnapshotJob.java
```diff
@@ -56,4 +61,31 @@ public IStatus run(IProgressMonitor monitor) {
...
+	/**
+	 * Resume automatic snapshots. This method does not schedule this Job.
+	 *
+	 * @see #suspend()
+	 */
+	void resume() {
+		if (suspended.decrementAndGet() < 0) {
+			suspended.set(0);
+		}
+	}
```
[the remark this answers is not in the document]
szarnekow: Agreed. Sorry about that, I didn't consider the semantics of JobManager.suspend / resume. I was more thinking about the usage pattern where a boolean was stored to call / not call resume again. I think that's highly confusing and prone to invalid state in case multiple threads use these methods.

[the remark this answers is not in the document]
fedejeanne: Good point, I haven't thought about the behavior of `suspend` / `resume` in the `JobManager`, only about nesting calls like the ones the `prepare...`/`begin...`/`endOperation` in `Workspace`.

fedejeanne:
>       * Shouldn't the `resume` also at least schedule the job if a snapshot was requested in between?

Good point, the current implementation would have taken the snapshot the next time an operation finished and `snapshotIfNeeded` had been called. I refactored the code and called (a new version of) `snapshotIfNeeded` inside `resumeSnapshotJob` to take a belated snapshot in case one had been requested in between. 

>       * Shouldn't the `SaveManager.shutdown` still write a snapshot? It uses the snapshop job to do so, which was okay by now as the job was completely hidden/managed by the class. By now it can be influenced from outside, which has an unintended side effect on the shutdown behavior.

I wouldn't force that. I think that if it's clear that snapshots are disabled then they will be expected to be disabled all the way, even during shutdown. 

> Are there further scenarios where disabling the snapshot job would be required, such that these changes are justified by being an enabler for further things?

This is the only case I found for now but I haven't looked for other possible scenarios. I assume there aren't many though.

> All in all, I am not sure if a single test is worth exposing internals of the SaveManager with all the efforts for documenting and adapting behavior accordingly.

I see your point: it might be too risky to allow to allow suspending snapshots without ever making sure the original state of the `SaveManager` is restored. What would you say if I change the current implementation and offer a method like this one?:

```java
public void runWithSnapshotsDisabled(Runnable runnable) {
	snapshotJob.suspend();
	try {
		runnable.run();
	} finally {
		snapshotJob.resume();
		snapshotIfNeeded();
	}
}
```
This would still expose the fact that there are snapshots and that they can be deactivated but they could only be deactivated in a controlled environment and the caller wouldn't even have the chance to "forget" activating them afterwards. It would also make it clear in the stack-trace that snapshot have been temporarily deactivated, which would help debugging.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -2205,4 +2205,12 @@ protected void writeWorkspaceFields(DataOutputStream output, IProgressMonitor mo
...
 		((Synchronizer) workspace.getSynchronizer()).savePartners(output);
 	}
+
+	public void suspendSnapshotJob() {
+		snapshotJob.suspend();
+	}
+
+	public void resumeSnapshotJob() {
+		snapshotJob.resume();
+	}
```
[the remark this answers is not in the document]
fedejeanne: These methods might not survive, see [my proposal below](https://github.com/eclipse-platform/eclipse.platform/pull/623#issuecomment-1669427902)

fedejeanne:
I see your point. Do you have any suggestions on how to separate that treat without changing the internal public APIs?

I thought of maybe overloading the constructors of `SaveManager` and `Workspace` with new `protected` constructors to be able to inject my own implementations of `DelayedSnapshotJob` and `SaveManager`, but that has a few downsides:
1. `protected` is IMO another way of saying "is not as public a `public` but you can still change it"
2.  One needs to hack its way into the LC of the whole workspace to be even able to replace the objects (and this could also open another door which can also be easily abused by other methods). 
3. Separating that treat only for a test might be as much an overkill as offering a new internal public API and it would be much more cumbersome, wouldn't you agree?

What do you think about using reflection to replace the `DelayedSnapshotJob` with a NO-OP version of it during tests? I would put the method in a `TestUtil` class, making it only accessible for tests. 

> The scoped disablement has to ensure that is does not fail through conflicting/concurrent calls, i.e., the proposed solution would be too simple as the implicit contract of the method is not fulfilled when performing concurrent calls to it.

Could you please elaborate?

HeikoKlare:
You will probably need to change the API anyway, but since it's `internal` you are basically able to change it. The problem I currently see is that functionality of the snapshot job (maybe even the whole snapshot management) is woven into the `SaveManager`, which makes the originally proposed change difficult because the side effects of suspending the job are unclear. My comment is not really about how the API looks like (may be that `suspend` and `resume` methods for taking snapshots are fine), but about how that API is realized.

>Separating that treat only for a test might be as much an overkill as offering a new internal public API and it would be much more cumbersome, wouldn't you agree?

I don't see why it should be more cumbersome, because you have to check and test every place in the `SaveManager` using the snapshot job anyway, which becomes more easy when separating the functionality into another class implementing the trait.

> I thought of maybe overloading the constructors of SaveManager and Workspace with new protected constructors to be able to inject my own implementations of DelayedSnapshotJob and SaveManager, but that has a few downsides:

> What do you think about using reflection to replace the DelayedSnapshotJob with a NO-OP version of it during tests? I would put the method in a TestUtil class, making it only accessible for tests.

As I mentioned before, in my opinion the need to provide additional API for the test either reveals a broken test or a design flaw and providing some test-specific injection functionality rather looks like a patch for the symptoms than a solution for the root cause to me.

Actually, the original issue is not about providing some API but about a failing test. And looking at the test, it is in a real bad state (validating internal states using reflection). Even if disabling snapshots solves the issue for now, it will fail again when internal implementations change. I think the test could be fixed by rewriting it in a much easier way such that it tests what it is actually supposed to test: https://github.com/eclipse-platform/eclipse.platform/issues/460#issuecomment-1671594343
Maybe than the API extension even becomes obsolete.

fedejeanne:
Superseded by https://github.com/eclipse-platform/eclipse.platform/pull/737

PR #625, opened 2023-08-07
Fix tests in org.eclipse.core.expressions.tests #525

Michael5601:
This commit fixes the tests testLessThanOrEqualToExpression and testGreaterThanOrEqualToExpression of class CountExpression.java in org.eclipse.core.expressions.tests by adding additional functionality similar to the already existing. Contributes to #525.

### DIFF
--- a/runtime/bundles/org.eclipse.core.expressions/src/org/eclipse/core/expressions/CountExpression.java
+++ b/runtime/bundles/org.eclipse.core.expressions/src/org/eclipse/core/expressions/CountExpression.java
@@ -23,25 +23,26 @@ import org.eclipse.core.internal.expressions.Expressions;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IConfigurationElement;
 
-
 /**
  * @since 3.7
  */
 public class CountExpression extends Expression {
 
+	private static final int GREATER_THAN_OR_EQUAL = 9; // [N-
+	private static final int LESS_THAN_OR_EQUAL = 8; // -N]
 	private static final int GREATER_THAN = 7; // (N-
-	private static final int LESS_THAN    = 6; // -N)
-	private static final int ANY_NUMBER   = 5; // *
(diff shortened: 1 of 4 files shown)


Michael5601:
Failing test documented in https://github.com/eclipse-platform/eclipse.platform/issues/628.

runtime/bundles/org.eclipse.core.expressions/src/org/eclipse/core/expressions/CountExpression.java
```diff
@@ -62,68 +63,86 @@ public CountExpression(String size) {
...
 	}
 
 	@Override
 	public EvaluationResult evaluate(IEvaluationContext context) throws CoreException {
-		Object var= context.getDefaultVariable();
+		Object var = context.getDefaultVariable();
 		int size;
 		if (var instanceof Collection) {
-			size= ((Collection<?>)var).size();
+			size = ((Collection<?>) var).size();
```
HeikoKlare, line 118: Please replace the cast with an assignment in the condition, i.e.:
```java
if (var instanceof Collection collection) {
	size = collection.size();
```

HeikoKlare (commented):
Thanks for adding the missing functionality and enabling the according tests!

The changes look sound to me. Tests work as expected. I would only suggest to apply a minor refactoring moving to recent Java features, since the code is touched anyway (see code comment).

runtime/bundles/org.eclipse.core.expressions/src/org/eclipse/core/expressions/CountExpression.java
Michael5601: Thank you for for your reply!

I added the code refactoring in the commit. The tests will run again now and if there is no failing test, the PR can be merged.

HeikoKlare (approved):
With recent improvements, changes are fine for me.
I added the required version bumps for 4.30.

Issue #626, opened 2023-08-07
BuildManager.waitForAutoBuild can cause deadlock


fedejeanne:
When running a `ResourceTest`, if the `JobManager` is suspended by the time the `tearDown` method is invoked then the current thread will wait forever.

PR #627, opened 2023-08-07
Throw exception when build job is waiting and JobManager is suspended

fedejeanne:
If a test suspends the `JobManager` and then waits for the build job to finish then a deadlock occurs. The reason for this is that waiting for a build job while the `JobManager` is not running is an endless wait since `JobManager.waitFor` keeps yielding the execution while the build job is in the `WAITING` state and since the `JobManager` is suspended, the job will never be scheduled.

This PR solves this problem by doing 2 things:
1. It restores the original state of the `JobManager` in the `ResourceTest::tearDown` in case a test suspends it and forgets to resume it. I did this since the original default state of the `JobManager` is `suspended == false` (i.e. "running") which means that restoring the original state would in most cases mean "enable it again" and ergo avoid the deadlock.
2. It throws an `Exception` in case the `JobManager` is suspended and some test wants to wait for a build job that will never end. This is for the (hopefully seldom) case where the fallback described in 1. did not solve the problem. If this happens then probably some other test suspended the `JobManager` and is affecting this test (isolation problem).

I also added 2 tests.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/626

## How to reproduce the error (before this PR)
Create and run this class. It will hang in the `main` thread:

```java
public class HangTest extends ResourceTest {

	@Test
	public void testHang() throws Exception {
		Job.getJobManager().suspend();
	}
}
```

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
@@ -19,22 +19,71 @@
  *******************************************************************************/
 package org.eclipse.core.internal.events;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import org.eclipse.core.internal.resources.*;
 import org.eclipse.core.internal.resources.ComputeProjectOrder.Digraph;
+import org.eclipse.core.internal.resources.ICoreConstants;
+import org.eclipse.core.internal.resources.IManager;
+import org.eclipse.core.internal.resources.Project;
+import org.eclipse.core.internal.resources.ProjectDescription;
+import org.eclipse.core.internal.resources.ResourceStatus;
+import org.eclipse.core.internal.resources.WorkManager;
+import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
 import org.eclipse.core.internal.watson.ElementTree;
-import org.eclipse.core.resources.*;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.jobs.*;
+import org.eclipse.core.resources.IBuildConfiguration;
+import org.eclipse.core.resources.IBuildContext;
+import org.eclipse.core.resources.ICommand;
+import org.eclipse.core.resources.IIncrementalProjectBuilder2;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectDescription;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceDelta;
+import org.eclipse.core.resources.IResourceStatus;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.ISafeRunnable;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.QualifiedName;
+import org.eclipse.core.runtime.SafeRunner;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ILock;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobGroup;
+import org.eclipse.core.runtime.jobs.MultiRule;
 import org.eclipse.osgi.util.NLS;
 import org.osgi.framework.Bundle;
 
 public class BuildManager implements ICoreConstants, IManager, ILifecycleListener {
+
+	public static class JobManagerSuspendedException extends RuntimeException {
+		private static final long serialVersionUID = 330426099787223028L;
+
+		public JobManagerSuspendedException(String message) {
+			super(message);
+		}
+	}
 
 	private static final String BUILDER_INIT = "BuilderInitInfo"; //$NON-NLS-1$
 
@@ -750,12 +799,19 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 	}
 
 	private static void waitFor(Job job) {
+		IJobManager jobManager = Job.getJobManager();
+
 		// Need to loop because jobs can reschedule itself and concurrent running
 		// background jobs change the states too
-		while (!(job.getState() == Job.NONE)) {
+		while (job.getState() != Job.NONE) {
 			// Need to wake up thread to finish as soon as possible:
-			while (!(job.getState() == Job.RUNNING || job.getState() == Job.NONE)) {
-				Job.getJobManager().wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
+			while (job.getState() != Job.RUNNING && job.getState() != Job.NONE) {
+				if (jobManager.isSuspended())
+					throw new JobManagerSuspendedException("The JobManager is suspended, waiting for " //$NON-NLS-1$
+							+ "a job to finish will just block this thread forever. Activate the JobManager again before waiting " //$NON-NLS-1$
+							+ "for a job to finish"); //$NON-NLS-1$
+
+				jobManager.wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
 				Thread.yield();
 				// After wakeup the woken job may go into sleep again when asynchronous
 				// workspace save interrupts autobuild so we need to wait till RUNNING or NONE
@@ -763,7 +819,7 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 			}
 			// Need to wait till job finished:
 			try {
-				Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_BUILD, null);
+				jobManager.join(ResourcesPlugin.FAMILY_AUTO_BUILD, null);
 			} catch (OperationCanceledException | InterruptedException e) {
 				// Ignore
 			}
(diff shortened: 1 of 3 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobTest.java
```diff
@@ -139,11 +153,87 @@ public IProject[] build(int kind, Map<String, String> args, IProgressMonitor mon
...
+	 */
+	private Throwable waitForAutoBuild(long timeoutMillis) throws InterruptedException {
+		try {
+			ForkJoinPool.commonPool()//
+					.submit(() -> getBuildManager().waitForAutoBuild())//
+					.get(timeoutMillis, TimeUnit.MILLISECONDS);
+		} catch (ExecutionException e) {
+			// Since the wait is happening in a Future, the original exception is packed
+			// inside the ExecutionException
+			return e.getCause();
```
HeikoKlare, line 231: I'd propose to have the method rethrow the actual exception to be checked for instead of returning the exception object. Then you do not have to return `null` as a marker and can use a proper `assertThrows` in the calling method.

```suggestion
			throw e.getCause();
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -1127,6 +1131,7 @@ protected void setUp() throws Exception {
 
 	@Override
 	protected void tearDown() throws Exception {
+		 restorePreviousStateOfJobManager();
```
HeikoKlare, line 1134: I would actually expect the `JobManager` to always be running (before a test) and have it running afterwards again. Otherwise a test will have side effects on others. Shouldn't we thus fail the test throwing an exception in case the `JobManager` was not reenabled (like we do after improper cleanup)?

HeikoKlare (commented):
I'd particularly like to have tests fail in case they do not properly clean up (i.e., resume the `JobManager` once they have suspeneded them), so that tests have no side effects on others.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobTest.java
fedejeanne: Good idea, thank you! I just changed it

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
fedejeanne: I changed so the previous state is still restored but I also failed the test. This will not only improve isolation but also let us know which tests are not cleaning up properly.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobTest.java
HeikoKlare: The method should be changed to `void` return type.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
HeikoKlare: I do not see any scenario where the "old" may ever be different from unsuspended. There are only two reasons why that can happen:
1. The JobManager was suspended before the test, i.e., some other test suspended it and did not resume. That is a symptom of improper test isolation and should either bei covered by an assertion at the beginning of the test or not be handled by passing that faulty state to the next test.
2. The test has already suspended the JobManager in its setup logic (as the old state is stored after a test might have disabled the JobManager during its setup instead of storing it before). But hat is exactly what the added functionality is supposed to avoid.

So I would simply resume the `JobManager` here and throw an assertion error if the test has not resumed it by itself.

fedejeanne: >I do not see any scenario where the "old" may ever be different from unsuspended. 

If a previous test which does **not** extend `ResourceTest` stops the `JobManager` then the initial state is "suspended". I didn't want to make any assumptions about which kind of tests ran before or what the "right" state is, I just wanted to make sure every test (that extends `ResourceTest`) cleans up accordingly.

HeikoKlare: That is exactly the first scenario I have described. But if that happens, the initial state for the test is faulty, so it does not make sense to me to restore a faulty state and pass it to the next test.

A test should always assume a "clean state" when starting. And in case it is in doubt that the system is in such a state, it should check and notify about that state rather than dealing with it and passing that state to the next test. So to catch the scenario you described, you would have to assert that the `JobManager` is not suspended when the test starts.

fedejeanne: > That is exactly the first scenario I have described.

I know, I just wanted to emphasize the fact about not making assumptions about which one is the "right" initial state :-) . But I see your point: tests that need to suspend the `JobManager` are the exception; most of the tests run under the assumption that the `JobManager` is **not** suspended. Considering "suspended" as the "faulty state" should be fine.

I changed the code in bcf4c13d8d026c83ba3f2ea62c8715d72c9531f3

HeikoKlare: The documentation is quite clear which state of the `JobManager` is to be considered exceptional ;-)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -1137,6 +1138,26 @@ protected void tearDown() throws Exception {
...
+	}
+
+	private void resumeJobManagerIfNecessary() {
+		if (!Job.getJobManager().isSuspended()) {
+			return;
+		}
+
+		String warningMessage = "The JobManager has been resumed. Please remember to always clean-up yourself after every test "
+				+ "in order to avoid random errors in subsequent tests (i.e. isolate your tests)";
+		TestUtil.log(IStatus.WARNING, getName(), warningMessage);
```
HeikoKlare, line 1157: In case the `JobManager` was suspended and not resumed, a warning will be logged and the test will fail with an assertion error. Is this necessary or wouldn't it be sufficient to throw an assertion error with this message. Then it would be sufficient to call `jobManager.resume()` in the `checkStateOfJobManager()` method.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
```diff
@@ -750,20 +799,27 @@ public void waitForAutoBuildOff() {
...
-			while (!(job.getState() == Job.RUNNING || job.getState() == Job.NONE)) {
-				Job.getJobManager().wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
+			while (job.getState() != Job.RUNNING && job.getState() != Job.NONE) {
+				jobManager.wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
 				Thread.yield();
 				// After wakeup the woken job may go into sleep again when asynchronous
 				// workspace save interrupts autobuild so we need to wait till RUNNING or NONE
 				// (happens after each JUnit class)
+
+				if (jobManager.isSuspended() && job.getState() == Job.WAITING)
```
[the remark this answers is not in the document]
fedejeanne: You're right, it doesn't make sense to wait in case the job is `SLEEPING` either because it would only delay reaching the `WAITING` state which would throw the exception anyway.
I changed it in c7edb1d8cb61bb78fc0348a7eb0b2966a472ef5c

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
```diff
@@ -750,20 +799,27 @@ public void waitForAutoBuildOff() {
...
-			while (!(job.getState() == Job.RUNNING || job.getState() == Job.NONE)) {
-				Job.getJobManager().wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
+			while (job.getState() != Job.RUNNING && job.getState() != Job.NONE) {
+				jobManager.wakeUp(ResourcesPlugin.FAMILY_AUTO_BUILD);
 				Thread.yield();
 				// After wakeup the woken job may go into sleep again when asynchronous
 				// workspace save interrupts autobuild so we need to wait till RUNNING or NONE
 				// (happens after each JUnit class)
+
+				if (jobManager.isSuspended() && job.getState() != Job.RUNNING && job.getState() != Job.NONE)
```
HeikoKlare, line 815: The checks for the state should not be necessary anymore as the loop condition already checked them. In addition: couldn't the condition be checked earlier as waking up jobs and yielding this thread will have not effect in case the job manager is suspended anyway.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -1127,6 +1127,7 @@ protected void setUp() throws Exception {
 
 	@Override
 	protected void tearDown() throws Exception {
+		assertAndRestoreStateOfJobManager();
```
HeikoKlare, line 1130: Sorry that I did not see and comment that one before: asserting the state before doing the rest of the cleanup will leave the system in an inconsistent state after this test in case the job manager was not resumed properly. The assertion should be moved to after every other kind of cleanup.

fedejeanne: Good catch, I hadn't thought about that. I moved the assertion all the way to the end to let the test clean up the rest properly before failing (in 656c74ac8f5aa0509d59604d4d70b03ce492601a)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/ResourceTest.java
```diff
@@ -1135,6 +1136,18 @@ protected void tearDown() throws Exception {
 		cleanup();
 		super.tearDown();
 		FreezeMonitor.done();
+		assertFalse("This test stopped the JobManager, which could have affected other tests.", //
+				wasSuspended);
+	}
+
+	private boolean resumeJobManagerIfNecessary() {
+		// resume the JobManager if necessary
```
HeikoKlare, line 1144: nitpicky: this comment is obsolete

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/AutoBuildJobTest.java
```diff
@@ -139,11 +155,84 @@ public IProject[] build(int kind, Map<String, String> args, IProgressMonitor mon
...
+
+		triggerAutoBuildAndWait();
+		Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_BUILD, null);
+
+		assertEquals("Scheduled calls", 1, scheduled.get());
+		assertEquals("Running calls", 0, running.get());
+
+		assertThrows(JobManagerSuspendedException.class, () -> waitForAutoBuild(2_000));
+
+		// Job.getJobManager().resume();
```
HeikoKlare, line 180: The test currently fails, probably because this line is commented out.

fedejeanne: Done. I also removed the initial call to `resume` from the other test in 3380fb6a3937f2f406c74ced057a2b653e59a41d since the `JobManager` will always be resumed after each test.

HeikoKlare (approved):

PR #630, opened 2023-08-08
Replace Sleep with While Loop in WorkerPoolTest

CodeLtDave:
Summary:
Replaced a 5-second sleep used to wait for worker threads to become idle in the WorkerPoolTest with a while loop. This change includes a timeout of 5 seconds to check when there are fewer than 50 workers, making the test more responsive and reliable.

Changes:
Removed the Thread.sleep(5 * 1000L); line that was used to wait until worker threads become idle. Added a while loop with a 5-second timeout to actively check for the condition of fewer than 50 active worker threads. Updated test annotations.

Benefits:
This change makes the test more robust by actively checking the condition instead of relying on a fixed sleep time. It also aligns with the modern testing approach and ensures the test does not hang indefinitely.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 salesforce.com.
+ * Copyright (c) 2017 salesforce.com and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
```diff
@@ -50,22 +81,38 @@ protected IStatus run(IProgressMonitor monitor) {
...
+			}
 
-		// count worker threads, must be less than WorkerPool.MAX_THREADS
+			if (System.currentTimeMillis() - startTimeInMSec > WORKER_TIMEOUT_IN_MSEC) {
+				Assert.fail("Timeout reached! Too many worker threads active: " + workerThreadCount + ", expected <= "
+						+ MAX_ALLOWED_IDLE_WORKER_THREADS);
+			}
+
+			Thread.yield();
+		}
```
HeikoKlare, line 102: It's a matter of taste, but I'd prefer to make the loop condition explicit in the loop header rather than having a break somewhere in the loop body, i.e. have:
`while(workerThreadCount > MAX_ALLOWED_IDLE_WORKER_THREADS)`
and define the variable before the loop and overwrite it at the end of each iteration with the current value.

HeikoKlare (approved):
Looks like a great performance improvement for the touched test.
Replacing a fixed 5 seconds sleep operation with a periodically checked condition looks good. The execution of the workers will take some more time now, as the main thread uses resources for kind of busy waiting, but since the test now runs for [100 ms](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5797986101/job/15714762998?pr=630#step:7:23004) rather than the more than 5 seconds before, the overhead is absolutely acceptable and the test is not prone to fail because the overhead is likely to exceed the test timeout of 5 seconds.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 salesforce.com.
+ * Copyright (c) 2017 salesforce.com, 2023 Vector Informatik GmbH.
```
jukzi, line 2: the official text is "and others."

jukzi (dismissed):
please avoid merge commits and amend to a single commit per PR

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 salesforce.com.
+ * Copyright (c) 2017 salesforce.com and Others.
```
jukzi, line 2:
```suggestion
 * Copyright (c) 2017 salesforce.com and others.
```

CodeLtDave:
Yes thank you for your tips, I am in the process to fix it.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
CodeLtDave: and thanks you're right

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/WorkerPoolTest.java
```diff
@@ -50,22 +81,33 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		parallelJobStartBarrier.await(BARRIER_TIMEOUT_IN_SEC, TimeUnit.SECONDS);
+
+		long startTimeInMSec = System.currentTimeMillis();
 
-		// this is the ugly part, wait until worker threads become idle
-		Thread.sleep(5 * 1000L);
+		// wait for workerThreadCount to sink below MAX_ALLOWED_IDLE_WORKER_THREADS
+		while (getWorkerThreadCount() <= MAX_ALLOWED_IDLE_WORKER_THREADS) {
+			if (System.currentTimeMillis() - startTimeInMSec > WORKER_TIMEOUT_IN_MSEC) {
+				Assert.fail("Timeout reached! Too many worker threads active: " + getWorkerThreadCount()
```
HeikoKlare, line 91: Please store the `workThreadCount` in a variable that is used for both the loop condition and the assert message. Otherwise it may happen that when checking the loop condition the thread count is still too high but when generating the assert message, it is already below the threshold. Thus the printed message could be quite confusing.

HeikoKlare (changes requested):
In general, the changes are fine for me. There is just one minor issue with the way in which the timeout error message is generated (see comment).

akurtakov:
One hint - rebase your patch on top of master when doing changes. This will make it less likely to have differences between verifications done on the PR and tests run by the I-build.

PR #631, opened 2023-08-10
CharsetDeltaJob: cancel job on shutdown - fixes #355

### LINKED ISSUE #355: CharsetDeltaJob: Workspace was not properly initialized or has already shutdown
I enabled logging output for the ApiTools Application and it sometimes falls with exception:

```
[ERROR] Workspace was not properly initialized or has already shutdown.
java.lang.Exception: Workspace was not properly initialized or has already shutdown.
    at org.eclipse.core.internal.resources.ResourceException.provideStackTrace (ResourceException.java:42)
    at org.eclipse.core.internal.resources.ResourceException.<init> (ResourceException.java:38)
    at org.eclipse.core.internal.resources.Workspace.getWorkManager (Workspace.java:1817)
    at org.eclipse.core.internal.resources.Workspace.endOperation (Workspace.java:1481)
    at org.eclipse.core.internal.resources.CharsetDeltaJob.run (CharsetDeltaJob.java:203)
    at org.eclipse.core.internal.jobs.Worker.run (Worker.java:63)
```

so it seems the `CharsetDeltaJob` should catch and ignore the exception if it can't make sure that it runs on a closing workspace.

jukzi:
fixes "Workspace was not properly initialized or has already shutdown." 
https://github.com/eclipse-platform/eclipse.platform/issues/355

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
@@ -236,6 +236,14 @@ public class CharsetDeltaJob extends Job implements IContentTypeManager.IContent
 	}
 
 	public void shutdown() {
+		try {
+			// try to prevent execution of this job to avoid prevent "already shutdown.":
+			cancel();
+			// if job is already running wait for it to finish:
+			join();
+		} catch (InterruptedException e) {
+			// ignore
+		}
 		IContentTypeManager contentTypeManager = Platform.getContentTypeManager();
 		//if the service is already gone there is nothing to do
 		if (contentTypeManager != null)


jukzi:
fixes locally running org.eclipse.jdt.core.tests.model.JavaSearchBugsTests.testBug178847()

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
```diff
@@ -236,6 +236,12 @@ public void setDisabled(boolean disabled) {
 	}
 
 	public void shutdown() {
+		try {
+			cancel(); // otherwise next job runs into "already shutdown."
```
HeikoKlare, line 240: Makes sense to cancel and join the job during shutdown to stop the job as fast as possible and avoid that the job still runs while the workspace is already shutdown. I would expect this to solve the "already shutdown" issue because _this_ job does not run into "already shutdown" anymore, correct?
So just for understanding: what do you mean by 'otherwise **next** job runs into "already shutdown."'?

jukzi: 'this Job's next execution'

HeikoKlare: thanks for clarification!

HeikoKlare (approved):

PR #632, opened 2023-08-10
Unit-Test: "Copying a file copies it's charset"

Wittmaxi:
One unit test tests the base behaviour of how charsets are moved when files are moved. It stands in opposition to testBug67606 which does not currently run because it is run inside of workspace.run

The second unit test remains disabled. It tests wether "copy" of a file copies it's charset. This is currently not the case and needs to be linked to an issue.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/CharsetTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/CharsetTest.java
@@ -64,6 +64,8 @@ import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.junit.Ignore;
+import org.junit.Test;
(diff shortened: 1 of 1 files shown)


Wittmaxi:
The new disabled unit-tests tests for behaviour described here https://github.com/eclipse-platform/eclipse.platform/issues/634

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/CharsetTest.java
```diff
@@ -193,6 +195,45 @@ public void _testBug67606() throws CoreException {
...
+			final IFile file = project.getFile("file.txt");
+			ensureExistsInWorkspace(file, true);
+			project.setDefaultCharset("FOO", getMonitor());
+			assertEquals("0.9", "FOO", file.getCharset());
+			file.setCharset("BAR", getMonitor());
+			assertEquals("1.0", "BAR", file.getCharset());
+			file.move(project.getFullPath().append("file2.txt"), IResource.NONE, getMonitor());
+			IFile file2 = project.getFile("file2.txt");
+			assertExistsInWorkspace(file2, false);
+			assertEquals("2.0", "BAR", file2.getCharset());
```
HeikoKlare, line 212: Could you please add proper messages for the assertions, such as "file has not the expected default charset" for the first one?

PR #635, opened 2023-08-11
Externalize org.eclipse.tips.ide messages to fix warnings

zelenyhleb:
Externalized org.eclipse.tips.ide strings to make them translatable. Added service class to form html from header-body-footer string set.


ua/org.eclipse.tips.ide/src/org/eclipse/tips/ide/internal/provider/TipHtml.java
```diff
@@ -0,0 +1,39 @@
...
+	private final String footer;
+
+	TipHtml(String header, String body, String footer) {
+		this.header = header;
+		this.body = body;
+		this.footer = footer;
+	}
+
+	TipHtml(String header, String body) {
+		this(header, body, "<br><br><br>"); //$NON-NLS-1$
```
HeikoKlare, line 31: When instantiating this class without a footer, there will now be 7 line breaks at the end of the document (2+2 from the formatting in `get()` and these three). As this was not the case before (documents ended with two or three line breaks), I would like to know whether this is intended, i.e., whether there was so from with too few line breaks in the existing code that is fixed in this refactoring, or whether it is "by accident".

HeikoKlare (commented):
The refactoring looks good. I just have one question regarding a difference in the HTML document formatting w.r.t. the old code (see code comment).

ua/org.eclipse.tips.ide/src/org/eclipse/tips/ide/internal/provider/TipHtml.java
zelenyhleb: Thank you for noting this! My idea was to show three new lines in case if footer is empty (like in `ActionTips.java`). Updated this part a bit.

HeikoKlare: I see. Looks good with recent changes, thanks!

zelenyhleb:
Hi, @HeikoKlare. Is here something more that needs to be done before merging? As I can see, tests failures are unrelated (please correct me if I am wrong)

HeikoKlare:
@zelenyhleb No, there is nothing left to be done here. I will recheck the failing tests and merge on Tuesday (when I am back at work) to be able to react in case unexpected regressions occur. Is that sufficent for you or would you benefit from an earlier merge?

zelenyhleb:
@HeikoKlare Thank you for quick response! Tuesday is ok, I was just afraid if I missed something about this PR.

HeikoKlare (approved):
Test failure is unrelated and documented in #682.
Refactoring looks good, functionality still works as expected.

PR #637, opened 2023-08-12
For ConfigurationSessionTestSuite ensure org.hamcrest.core is added

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1245

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/session/ConfigurationSessionTestSuite.java
+++ b/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/session/ConfigurationSessionTestSuite.java
@@ -44,6 +44,7 @@ import org.junit.Assert;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.Constants;
 import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.Version;
 
 @SuppressWarnings("restriction")
 public class ConfigurationSessionTestSuite extends SessionTestSuite {
@@ -93,6 +94,28 @@ public class ConfigurationSessionTestSuite extends SessionTestSuite {
 		addBundle(org.eclipse.pde.internal.junit.runtime.CoreTestApplication.class); // org.eclipse.pde.junit.runtime
 
 		addBundle(org.hamcrest.CoreMatchers.class); // org.hamcrest.core
+
+		// The org.junit bundle requires an org.hamcrest.core bundle, but as of version
+		// 2.x, the org.hamcrest bundle above provides the actual classes. So we need to
+		// ensure that the actual org.hamcrest.core bundle required by org.junit is
+		// added too.
+		if ("org.hamcrest".equals(FrameworkUtil.getBundle(org.hamcrest.CoreMatchers.class).getSymbolicName())) {
+			Bundle maxHamcrestCoreBundle = null;
+			Version maxHamcrestCoreVersion = null;
+			for (Bundle bundle : FrameworkUtil.getBundle(ConfigurationSessionTestSuite.class).getBundleContext().getBundles()) {
+				if ("org.hamcrest.core".equals(bundle.getSymbolicName())) {
+					Version version = bundle.getVersion();
+					if (maxHamcrestCoreVersion == null || maxHamcrestCoreVersion.compareTo(version) < 0) {
+						maxHamcrestCoreVersion = version;
+						maxHamcrestCoreBundle = bundle;
+					}
+				}
+			}
+			if (maxHamcrestCoreBundle != null) {
+				addBundle(maxHamcrestCoreBundle, null);
+			}
+		}
+
 		addBundle(org.junit.Test.class); // org.junit
 		addBundle(org.junit.jupiter.api.Test.class); // junit-jupiter-api
 		addBundle(org.junit.platform.commons.JUnitException.class); // junit-platform-commons
@@ -127,6 +150,10 @@ public class ConfigurationSessionTestSuite extends SessionTestSuite {
 	public void addBundle(Class<?> classFromBundle, String suffix) {
 		Bundle bundle = FrameworkUtil.getBundle(classFromBundle);
 		Assert.assertNotNull("Class is not from a bundle: " + classFromBundle, bundle);
+		addBundle(bundle, suffix);
+	}
+
+	public void addBundle(Bundle bundle, String suffix) {
 		String url = getBundleReference(bundle, suffix);
 		bundles.add(url);
 	}


HannesWell:
Doesn't this fix https://github.com/eclipse-equinox/equinox/issues/301?

Can you explain why you didn't just added the org.hamcrest.core bundle by its ID as suggested by @laeubi in https://github.com/eclipse-equinox/equinox/issues/301#issuecomment-1675821649?

If hamcrest.core 1.3 is present, hamcrest.core is added twice (which should be a No-Op), and for hamcrest 2.2 `org.hamcrest.core` and `org.hamcrest` are added.

merks:
> Doesn't this fix [eclipse-equinox/equinox#301](https://github.com/eclipse-equinox/equinox/issues/301)?
> 

I tried to force push the change with that link, but apparently I'm too stupid:

image


> Can you explain why you didn't just added the org.hamcrest.core bundle by its ID as suggested by @laeubi in [eclipse-equinox/equinox#301 (comment)](https://github.com/eclipse-equinox/equinox/issues/301#issuecomment-1675821649)?
> 
> If hamcrest.core 1.3 is present, hamcrest.core is added twice (which should be a No-Op), and for hamcrest 2.2 `org.hamcrest.core` and `org.hamcrest` are added.

Because that adds all versions.  I only want to add the highest version.  I also want to be careful that I don't add the same bundle twice, hence only doing this when org.hamcrest is added...

If someone else wants to fix this in a nicer way, I'm more than happy to work on other things!

merks:
@HannesWell  @iloveeclipse 

Shall I merge this so that the osgi tests run again.  We could improve these changes at our leisure after the fact.

iloveeclipse:
Yes, please.

HannesWell:
> I tried to force push the change with that link, but apparently I'm too stupid:

Your commit looks like that, but if you had open a browser to create a PR and want to auto generated text to be updated after a force-push before you created the PR, you probably need to refresh the tab. But I'm sure if that was exactly your issue. :)

> Because that adds all versions. I only want to add the highest version. I also want to be careful that I don't add the same bundle twice, hence only doing this when org.hamcrest is added...
> 
> If someone else wants to fix this in a nicer way, I'm more than happy to work on other things!

> Shall I merge this so that the osgi tests run again. We could improve these changes at our leisure after the fact.

Sure, later is also fine. This should do the job. 👍🏽

merks:
The link was even there when I did a squash and merge!

image

merks:
@iloveeclipse 

Do you need to manually kick off an I-Build to test that this change fixes the problem?

iloveeclipse:
> Do you need to manually kick off an I-Build to test that this change fixes the problem?

Unfortunately there is already a build running, https://ci.eclipse.org/releng/job/Builds/job/I-build-4.29/100/ .

Do you want/need another one, or can you wait 12 hours?

merks:
I would be nice to know if the test run sooner rather than later.  At some point it would be nice to stop chasing the next problem.  😱   But of course I can wait...

iloveeclipse:
OK, here is it: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.29/101/

iloveeclipse:
Hmm.. See https://ci.eclipse.org/releng/job/Builds/job/I-build-4.29/101/console
```
15:10:54  [ERROR] Cannot resolve project dependencies:
15:10:54  [ERROR]   Software being installed: org.eclipse.sdk.ide 4.29.0.qualifier
15:10:54  [ERROR]   Missing requirement: org.eclipse.sdk.feature.group 4.29.0.v20230812-0953 requires 'org.eclipse.equinox.p2.iu; org.eclipse.platform.feature.group [4.29.0.v20230812-0600,4.29.0.v20230812-0600]' but it could not be found
15:10:54  [ERROR]   Cannot satisfy dependency: org.eclipse.sdk.ide 4.29.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.sdk.feature.group 0.0.0
```

???

iloveeclipse:
Is this some temporary network / infra issue? I can see / access https://download.eclipse.org/eclipse/updates/4.29-I-builds/I20230812-0600/
Should I retrigger build again?

merks:
Maybe something needs to be touched?

iloveeclipse:
> org.eclipse.platform.feature.group

Can you see `org.eclipse.platform.feature.group` in the repo https://download.eclipse.org/eclipse/updates/4.29-I-builds/I20230812-0600/ ? If yes, what else should the maven error mean?

merks:
@akurtakov 

Perhaps you have seen such problems before and know if some qualifier update file needs touching, or if this is just a glitch, or what?

laeubi:
> what else should the maven error mean

It is not a maven error, it is a P2 error

> org.eclipse.sdk.feature.group 4.29.0.v20230812-0953 requires 'org.eclipse.equinox.p2.iu; org.eclipse.platform.feature.group [**4.29.0.v20230812-0600,4.29.0.v20230812-0600]**' but it could not be found

The problem is that  `org.eclipse.sdk.feature.group` uses a "build qualified dependency" with an exact version range (most likely because it was replaced by the baseline version) I therfore assume that this feature needs to be touched to pick up changes in `org.eclipse.platform.feature` (what seems to be a never version now).

iloveeclipse:
In the SDK I20230812-0600 I see org.eclipse.platform.feature with exact this version 4.29.0.v20230812-0600
image

So what is missing???

merks:
@iloveeclipse 

There are files like these occasionally need "touching" to force a qualifier update:

image

So @laeubi  is assuming/guessing maybe that's the SDK one.

merks:
Note that this is not about something actually being missing, so no point in looking what's in p2 repositories.  It's as @laeubi suggests...  Somewhere it's using an old dependency when it should be using a newer one because something that should be considered updated is not considered updated and is replaced with something old, that depends on something old and therefore fails.

iloveeclipse:
I know about forceQualifier files, but I don't understand **why**  org.eclipse.platform.feature that **has** version 4.29.0.v20230812-0600 need to be updated if that is already exact the version that was "missed" by p2??

iloveeclipse:
See also https://download.eclipse.org/eclipse/downloads/drops4/I20230812-0600/buildlogs/reporeports/reports/featureNames.html that lists feature versions.

laeubi:
It *should* work to run the aggregator build locally with `-Dcompare-version-with-baselines.skip=false` then Tycho will complain if something has to be bumped (or at least give a warning not sure how the build is exactly configured)

iloveeclipse:
> It _should_ work to run the aggregator build locally

I can do that only on Monday. Should we just trigger new SDK build to see if that was some network issue?

merks:
Why does a feature ever need to be touched?   Generally because one of its included dependencies have changed version but the feature itself isn't considered changed so the old one is substituted (by the baseline comparison) and that one has old include versions that are not satisfied by the IUs in the build/reactor.

merks:
Now I need to go out and will not be available until tomorrow.

Issue #578, opened 2023-07-20
"Attempted to beginRule does not match outer scope"


jukzi:
org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testDependentBuildConfigsSubset()

silently logs
```
!MESSAGE Attempted to beginRule: org.eclipse.core.tests.internal.builders.TimerBuilder$1@49cc8332, does not match outer scope rule: R/
!STACK 0
java.lang.IllegalArgumentException: Attempted to beginRule: org.eclipse.core.tests.internal.builders.TimerBuilder$1@49cc8332, does not match outer scope rule: R/
	at org.eclipse.core.runtime.Assert.isLegal(Assert.java:66)
	at org.eclipse.core.internal.jobs.ThreadJob.illegalPush(ThreadJob.java:138)
	at org.eclipse.core.internal.jobs.ThreadJob.push(ThreadJob.java:402)
	at org.eclipse.core.internal.jobs.ImplicitJobs.begin(ImplicitJobs.java:66)
	at org.eclipse.core.internal.jobs.JobManager.beginRule(JobManager.java:343)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:240)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:303)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:392)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:395)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:506)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:454)
	at org.eclipse.core.internal.events.BuildManager.build(BuildManager.java:536)
	at org.eclipse.core.internal.events.AutoBuildJob.doBuild(AutoBuildJob.java:207)
	at org.eclipse.core.internal.events.AutoBuildJob.run(AutoBuildJob.java:300)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

```

that sounds wrong.

mickaelistria:
Has anything changed in the rules implementation lately? The test are really strongly relying on rules to determine what can be run in parallel and what cannot, and AFAIK they've been stable enough for a long time.

jukzi:
[...]

Issue #201, opened 2022-10-04
random failing ParallelBuildChainTest.testDependentBuildConfigBuildSequentially


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20221003-1800/testresults/html/org.eclipse.core.tests.resources_ep426I-unit-win32-java11_win32.win32.x86_64_11.html
```
junit.framework.AssertionFailedError
at junit.framework.Assert.fail(Assert.java:55)
at junit.framework.Assert.assertTrue(Assert.java:22)
at junit.framework.Assert.assertTrue(Assert.java:31)
at junit.framework.TestCase.assertTrue(TestCase.java:200)
at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testDependentBuildConfigBuildSequentially(ParallelBuildChainTest.java:327)

```

relates to https://github.com/eclipse-platform/eclipse.platform/issues/185 (testDependentProjectsBuildSequentially )?

HeikoKlare:
Closed as failure has not occurred recently.

Issue #185, opened 2022-09-21
random failing ParallelBuildChainTest.testDependentProjectsBuildSequentially


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20220920-2240/testresults/html/org.eclipse.core.tests.resources_ep426I-unit-win32-java11_win32.win32.x86_64_11.html

```
junit.framework.AssertionFailedError
at junit.framework.Assert.fail(Assert.java:55)
at junit.framework.Assert.assertTrue(Assert.java:22)
at junit.framework.Assert.assertTrue(Assert.java:31)
at junit.framework.TestCase.assertTrue(TestCase.java:200)
at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testDependentProjectsBuildSequentially(ParallelBuildChainTest.java:288)
```

relates to https://github.com/eclipse-platform/eclipse.platform/issues/157

jukzi:
also
https://download.eclipse.org/eclipse/downloads/drops4/I20221007-0230/testresults/html/org.eclipse.core.tests.resources_ep426I-unit-win32-java11_win32.win32.x86_64_11.html

PR #641, opened 2023-08-22
Rewrite ParallelBuildChainTest and fix random failures #609

### LINKED ISSUE #609: Random failing ParallelBuildChainTest.testWorkspaceParrallelBuildNoConflict
The test case `org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testWorkspaceParrallelBuildNoConflict` randomly fails:
https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-606/3/testReport/

```
org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testWorkspaceParrallelBuildNoConflict -- Time elapsed: 5.781 s <<< FAILURE!
junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:55)
	at junit.framework.Assert.assertTrue(Assert.java:31)
	at junit.framework.TestCase.assertTrue(TestCase.java:200)
	at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testWorkspaceParrallelBuildNoConflict(ParallelBuildChainTest.java:210)
```

Maybe this is only is matter of slow execution, since according to the asserted states the tested build operation seems to have finished properly but only took too long.

HeikoKlare:
## Issue
The `ParallelBuildChainTest` has random failing test methods. Its implementation relies on specific timing behavior. It spawns builds for different projects or the workspace that take a specific (absolute, hard-coded) amount of time to finish. This has two drawbacks:
1. It induces a race conditions, as whenever execution times exceed these amounts of time (e.g., due to slow infrastructure), the tests will randomly fail.
2. The tests spent much time for unnecessary busy waiting making them require more time / resources than necessary.

## Modification
This change addresses the issues by not relying on builds to finish but on builds to be started. This is sufficient, as the tests are supposed to validate under which circumstances parallel builds are allowed or expected.
To this end, the test now uses either immediately finishing builds or by making them run for a very long time. The long running builds will only be validated regarding whether they have started and will be explicitly aborted when the actual test has finished. This removes the necessity to rely on specific timing behavior. Only one test case validating that a build is not started uses a timeout to wait for, as a blocked build cannot be detected easily.
This change also adds two additional test cases reflecting further combinations of scheduling rules and build configurations.

I know that the PR is a quite large refactoring / rewrite of the test class, but I think the improvements in comprehensibility, performance and hopfully stability are worth the review effort. This test class has been instable for quite some time (see #157, #185, #201, #578, #609).

## Performance
With respect to performance, the change reduced the execution time of the test class on my local computer from ~20 seconds to less than 10 seconds, while the number of test cases increased.
The CI build logs show similar execution times of around 20 seconds for the current test class, so CI builds will hopefully benefit in a similar way.

Fixes #609.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
@@ -13,10 +13,32 @@
  *******************************************************************************/
 package org.eclipse.core.tests.internal.builders;
 
-import java.util.*;
+import static org.eclipse.core.tests.resources.TestUtil.waitForCondition;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.number.OrderingComparison.greaterThanOrEqualTo;
+import static org.hamcrest.number.OrderingComparison.lessThanOrEqualTo;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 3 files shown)


mickaelistria:
Improvements in test are welcome during RC phase. Let's merge it.
thanks1

Issue #648, opened 2023-08-25
Console Scroll Lock is disabled with key "End"


jukzi:
I understand that the key combination ctrl+end (i.e. document end) may make sense to disable scroll lock, but only the key end (line end) in the middle of the console shout not disable scroll lock
![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/51790620/933c223a-90ec-4cb1-b28d-59df5bd4a2b4)

HeikoKlare:
fixed via #649

PR #649, opened 2023-08-28
fix Console Scroll Lock is disabled with key "End" without CTRL #648

jukzi:
only ctrl+end should disable Scroll Lock

https://github.com/eclipse-platform/eclipse.platform/issues/648

### DIFF
--- a/debug/org.eclipse.ui.console/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.ui.console/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ui.console; singleton:=true
-Bundle-Version: 3.13.0.qualifier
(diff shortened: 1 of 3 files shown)


HeikoKlare (commented):
I agree that only CTRL+END should disable scroll lock (and not just pressing END).
The change works as expected.

One observation: Pressing DOWN still disables scroll lock. Personally, I find this quite unintuitive (just like pressing END disabling scroll lock). So I would prefer to apply the CTRL mask check to both END and DOWN/BOTTOM here.

jukzi:
> So I would prefer to apply the CTRL mask check to both END and DOWN/BOTTOM here.

If you reach the bottom of the log with the down key it was intentional to end scroll lock again. Therefore for "down" there is the additional condition checkEndOfDocument().  - However to my experience that condition sometimes yield true to early - when coming near to the end. Would be another topic to fix that.

HeikoKlare:
I see. I mistook the check for `SWT.BOTTOM` to be related to pressing the arrow down button, but it's actually another condition that also checks for end of document, as you point to. So nothing to address in this PR.

Issue #638, opened 2023-08-15
Unknown extension point: 'org.eclipse.core.tests.harness.tests'


jukzi:
With a fresh Oomphed Eclipse IDE i get this error. I have not seen it before.

```
Description	Resource	Path	Location	Type
Unknown extension point: 'org.eclipse.core.tests.harness.tests'	plugin.xml	/org.eclipse.core.tests.resources.saveparticipant	line 5	Plug-in Problem
```
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/307931bb-d322-4d20-bfb4-58ece4d9dc14)

It's used here, but i don't find where it is defined
https://github.com/eclipse-platform/eclipse.platform/blob/master/resources/tests/org.eclipse.core.tests.resources.saveparticipant/plugin.xml

jukzi:
The only way to start those SaveManagerTest i found was to manually start org.eclipse.core.tests.resources.session.SaveParticipantTest.
When i do that test3 fails

```
junit.framework.AssertionFailedError: 6.0
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.TestCase.fail(TestCase.java:223)
	at org.eclipse.core.tests.resources.saveparticipant.SaveManager3Test.testSaveParticipant(SaveManager3Test.java:143)
	at org.eclipse.core.tests.resources.saveparticipant.SaveManagerTest.test3(SaveManagerTest.java:104)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at org.eclipse.jdt.internal.junit.runner.junit3.JUnit3TestReference.run(JUnit3TestReference.java:128)
[...]

PR #650, opened 2023-08-30
test.harness: Delete reference to deleted extension point #638

jukzi:
The SaveManagerTest test can manually started with org.eclipse.core.tests.resources.session.SaveParticipantTest

https://github.com/eclipse-platform/eclipse.platform/issues/638


HeikoKlare (commented):
Removing `plugin.xml` due to non-existent extension point is fine.
The `plugin.xml` has to be removed from the `bin.includes` of the `build.properties` as well.

jukzi:
> The `plugin.xml` has to be removed from the `bin.includes` of the `build.properties` as well.

thanks

HeikoKlare (approved):

Issue #651, opened 2023-08-30
Failing Test: SaveParticipantTest.test3


jukzi:
org.eclipse.core.tests.resources.session.SaveParticipantTest

the test is not used for  almost 20 years. It was once listed as comment in org.eclipse.core.tests.resources.session.AllSessionTests 
from [2d6f69d722d1fb50ad084fc05d3d08c0e3ceac7c](https://github.com/eclipse-platform/eclipse.platform/commit/2d6f69d722d1fb50ad084fc05d3d08c0e3ceac7c#diff-e8f0c12f0f1fdf6b16f0e1c6b5f4c055c1cf672628a4de81838b31e5f5541261R45)

until [31a4b42f024e545077240af22e1349e34425ed07](https://github.com/eclipse-platform/eclipse.platform/commit/31a4b42f024e545077240af22e1349e34425ed07#diff-e8f0c12f0f1fdf6b16f0e1c6b5f4c055c1cf672628a4de81838b31e5f5541261L53)

jukzi:
```
junit.framework.AssertionFailedError: 6.0
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.TestCase.fail(TestCase.java:223)
	at org.eclipse.core.tests.resources.saveparticipant.SaveManager3Test.testSaveParticipant(SaveManager3Test.java:143)
	at org.eclipse.core.tests.resources.saveparticipant.SaveManagerTest.test3(SaveManagerTest.java:104)
```

logging
```
delta is different than expected
Verifying delta for /
Verifying delta for /MyProject
Verifying delta for /MyProject/addedFile
Verifying delta for /Project2
Verifying delta for /Project2/.project
Verifying delta for /Project2/.settings
	Missing expected change
		Actual kind: <ADDED>
		Actual change flags: <0>
Verifying delta for /Project2/.settings/org.eclipse.core.resources.prefs
	Missing expected change
		Actual kind: <ADDED>
		Actual change flags: <0>
Verifying delta for /Project2/addedFile
Verifying delta for /Project2/file110
Verifying delta for /Project2/folder110
Verifying delta for /Project2/folder110/file120
Verifying delta for /Project2/folder111
Verifying delta for /Project2/folder111/file121

```

jukzi:
regression from https://bugs.eclipse.org/bugs/show_bug.cgi?id=479450

HeikoKlare:
fixed via #652

PR #652, opened 2023-08-30
fix SaveParticipantTest.test3 #651

jukzi:
https://github.com/eclipse-platform/eclipse.platform/issues/651

"a new project will always have
.settings/org.eclipse.core.resources.prefs file generated, containing information about project default encoding."

https://bugs.eclipse.org/bugs/show_bug.cgi?id=479450


HeikoKlare (approved):
Test correction is sound and test now runs successfully again.
Should the `SaveParticipantTest` be added to the `AllSessionTest` again?

jukzi:
> Should the `SaveParticipantTest` be added to the `AllSessionTest` again?

we could try that in a separate commit in m1. I don't know why it was never really added. May be it will somehow not wirk un i-build.

jukzi:
fails org.eclipse.ua.tests.doc.internal.linkchecker.TocLinkChecker.testPlatformIsvStatic
https://github.com/eclipse-platform/eclipse.platform/issues/668

Issue #628, opened 2023-08-08
Random failing IJobManagerTest.testOrder


Michael5601:
The test case `org.eclipse.core.tests.runtime.jobs.IJobManagerTest.testOrder` randomly fails:
https://github.com/eclipse-platform/eclipse.platform/pull/625/checks?check_run_id=15677878481
```java
1.1.0 expected:<testOrder(0)(418)> but was:<testOrder(1)(419)>
junit.framework.AssertionFailedError: 1.1.0 expected:<testOrder(0)(418)> but was:<testOrder(1)(419)>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.TestCase.assertEquals(TestCase.java:238)
	at org.eclipse.core.tests.runtime.jobs.IJobManagerTest.testOrder(IJobManagerTest.java:1772)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
```
Locally this test does not fail with my setup. But in PR #625 it failed.

fedejeanne:
This looks a lot like the race condition that `testReverseOrder` suffered from (which was fixed in https://github.com/eclipse-platform/eclipse.platform/commit/060e9361f8f62c0cb1d15f079cfe576e7b0af6a8)

HeikoKlare:
[...]

PR #655, opened 2023-09-04
Fix faulty and random failing IJobManagerTest.testOrder #628

HeikoKlare:
The test case `IJobManagerTest.testOrder` is randomly failing. It is supposed to validate that Jobs are execute in the order in which they are scheduled. To this end, it asserts that the run methods of the Jobs are executed in the order in which the have been scheduled. There is, however, no guarantee for an execution order of the `run` methods but only for the order in which the Jobs are picked by a Worker to be executed.

This change rewrites the test case to only validate that whenever a Job's `run` method is executed the previously scheduled jobs have started running or have already finished as well, i.e., are either in the state RUNNING or NONE.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/628


HeikoKlare:
Failing test is documented #668

Issue #657, opened 2023-09-04
Random failing ParallelBuildChainTest.testIndividualProjectBuilds_ProjectRelaxedRule


HeikoKlare:
The `ParallelBuildChainTest.testIndividualProjectBuilds_ProjectRelaxedRule` is randomly failing: https://github.com/eclipse-platform/eclipse.platform/pull/655/checks?check_run_id=16473097978

```
java.lang.AssertionError: 
too few builds have run in parallel
Expected: a value equal to or greater than <3>
     but: <2> was less than <3>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.assertMinimumNumberOfSimultaneousBuilds(ParallelBuildChainTest.java:425)
	at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.lambda$1(ParallelBuildChainTest.java:140)
	at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.executeIndividualFullProjectBuilds(ParallelBuildChainTest.java:414)
	at org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.testIndividualProjectBuilds_ProjectRelaxedRule(ParallelBuildChainTest.java:138)
```

The tests have recently been rewritten in response to recurring failures and inefficiency (see #609 and #641). The error logs show that one of the rewritten test cases fails because one of the builds that should run in parallel is actually not started in parallel but waits for another build to finish (test execution time is higher than 30 seconds, which is the execution time of a long running builds). I have no clue why the builds are not run in parallel yet. The jobs running the builds are started properly and the scheduling rules allow them to be executed in parallel. Still there is something that randomly blocks one at least one of the build operations. The error cannot be reproduced locally. 

I will at least improve the assertions, as the test should already fail because not all builds have started in time (and log which ones have not been executed). This does not happen because the timeout is higher than the execution time of a single build operation.

iloveeclipse:
> Still there is something that randomly blocks one at least one of the build operations.

May be it's just small number of cores in the target environment, so even if nothing blocks the job, there is no free core to run it?

HeikoKlare:
[...]

PR #658, opened 2023-09-04
Improve error message and fail behavior of ParallelBuildChainTest #657

HeikoKlare:
The `ParallelBuildChainTest` has some flaws in error message design and failure handling, which are addressed by this change:
1. The timeout for builds to start/finish is higher than the execution time of a long running build. Thus in case builds are unexpectedly not run in parallel, the assertion for timely build start/finish is not evaluated as expected.
2. In case an error occurs before starting a build operation, the number of expected builds in the TimerBuilder may be higher than the number of builds that will every be executed. Since the TimerBuilder abortion tries to wait for all expected builds indefinitely, the operation (and thus test tear down) may never terminate.


HeikoKlare:
Failing test documented in #668.

Issue #659, opened 2023-09-05
Race condition in ThreadJob unnecessarily blocks parallel build execution


HeikoKlare:
## Issue
In specific situations, in particular where parallel project builds are running, a `ThreadJob` unnecessary blocks its execution for a long time due to a race condition. In its `waitForRun` method, it checks whether there is a job with a conflicting scheduling rule, which blocks the execution:
https://github.com/eclipse-platform/eclipse.platform/blob/c1e556cdfe60f25f455b623b32c97ede44f1e8ff/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java#L287
The blocking operation uses a monitor `wait` call without a timeout that expects a `notify` call on the `jobStateLock` object of the blocking job: 
https://github.com/eclipse-platform/eclipse.platform/blob/c1e556cdfe60f25f455b623b32c97ede44f1e8ff/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java#L321-L322
Between checking for a blocking job and calling `wait` on the state lock of that job, the blocking job may proceed in its execution. This can also involve the release of the conflicting scheduling rule and the acquisition of a completely different scheduling rule, such that there is actually no conflict between the two jobs anymore. If the blocking job remains in the `RUNNING` state for a long or even indefinite amount of time, the blocked job will wait for an accordinly long time to start (or not start at all), because a notification on the state lock is only performed when the job state changes.

## Example
As a concrete example, building a project acquires the workspace root as its scheduling rule for performing operations _before_ the actual build operation. The actual build operation is then executed with a different, project-specific build rule:
https://github.com/eclipse-platform/eclipse.platform/blob/c1e556cdfe60f25f455b623b32c97ede44f1e8ff/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java#L587-L605
The following execution order leads to an unnecessary block:
1. Job 1 performs the build preparation on the workspace root as its scheduling rule
2. Job 2 starts running and finds Job 1 as a blocking job due to the conflicting scheduling rule
3. Job 1 finished build preparation, releases the scheduling rule and starts the build with its project-specific build rule
4. Job 2 calls `wait` on the `jobStateLock` of Job 2, even though the scheduling rule of the job has changed in between
5. Job 1 runs the build operation and afterwards changes its state from RUNNING
[...]

Issue #488, opened 2023-06-09
random failing ServiceSupplierTestCase.testOptionalReferences


jukzi:
on win
```
java.lang.AssertionError: expected:<1> but was:<2>
	at org.eclipse.e4.core.internal.tests.di.extensions.ServiceSupplierTestCase.testOptionalReferences(ServiceSupplierTestCase.java:267)
```

HeikoKlare:
Example for fail: https://github.com/eclipse-platform/eclipse.platform/actions/runs/5342777773/jobs/9685299025

HeikoKlare:
Test failed again with a different error message in recent build
https://github.com/eclipse-platform/eclipse.platform/pull/595/checks?check_run_id=15421937193

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertNotNull(Assert.java:713)
	at org.junit.Assert.assertNotNull(Assert.java:723)
	at org.eclipse.e4.core.internal.tests.di.extensions.ServiceSupplierTestCase.testOptionalReferences(ServiceSupplierTestCase.java:254)
```

Michael5601:
Failed again in this [run](https://github.com/eclipse-platform/eclipse.platform/pull/605/checks?check_run_id=15704942223) in PR: https://github.com/eclipse-platform/eclipse.platform/pull/605

fedejeanne:
Failed again in [this build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/7030446647/job/19129999755?pr=776) of #776 

```java
org.eclipse.e4.core.internal.tests.di.extensions.ServiceSupplierTestCase.testOptionalReferences -- Time elapsed: 0.070 s <<< FAILURE!
java.lang.AssertionError: expected same:<class org.eclipse.e4.core.internal.tests.di.extensions.DisabledServiceA> was not:<class org.eclipse.e4.core.internal.tests.di.extensions.DisabledServiceB>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotSame(Assert.java:829)
	at org.junit.Assert.assertSame(Assert.java:772)
	at org.junit.Assert.assertSame(Assert.java:783)
	at org.eclipse.e4.core.internal.tests.di.extensions.ServiceSupplierTestCase.testOptionalReferences(ServiceSupplierTestCase.java:270)
```

[...]

PR #660, opened 2023-09-05
Make data setup in ParallelBuildChainTest explicit per test case #659

HeikoKlare:
Define number of projects and number of allowed parallel builds per test case instead of for the whole test class to ensure that test cases can use different values for different purposes. Improve consistency and readability by using explicitly defined and equal values for number of projects and number of parallel builds.

This is a preparatory refactoring for providing a regression test for #659.

Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/659.


HeikoKlare:
Failing test is documented in #488.

PR #661, opened 2023-09-05
Fix race condition in ThreadJob.waitForRun #659

HeikoKlare:
`ThreadJob.waitForRun()` suffers from a race condition. Between checking for a blocking job with a conflicting scheduling rule and starting to wait on the state lock of that blocking job to get notified when the scheduling rule is released, the blocking job may have already changed its state and acquired a different scheduling rule, such that no conflict exists anymore. Since this condition is not reevaluated but waiting on the lock object is started without a timeout, the blocked job unnecessarily waits for the initially blocking job to change its state again, i.e., to finish its execution.

One relevant situation in which the scheduling rule of a job can change is if the job is an internally used `ThreadJob`. They are started whenever a rule is acquired on the `JobManager` and are reused for the same thread. When the same thread acquires multiple rules, one after another, the rule of the `ThreadJob` may effectively change. This, for example, happens during a workspace build operation, which changes the rule for two times. Thus, if the blocking job in `ThreadJob.waitForRun()` is a
`ThreadJob`, it may change its rule, in particular when that job performs a build operation.

With this change, the scheduling rules of blocking and blocked thread are checked for conflicts again before starting to wait on the state lock in the blocked job. This leads to a reevaluation of the block condition in case the scheduling rule of the blocking job has changed in between and avoids the job to unnecessarily block until the blocking job finishes.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/659.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
@@ -142,6 +142,17 @@ public class ParallelBuildChainTest extends AbstractBuilderTest {
 	}
 
 	@Test
+	public void testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule() throws Exception {
+		int numberOfParallelBuilds = 60;
+		var longRunningProjects = createMultipleTestProjects(numberOfParallelBuilds, BuildDurationType.LONG_RUNNING,
+				RuleType.CURRENT_PROJECT_RELAXED);
+		executeIndividualFullProjectBuilds(numberOfParallelBuilds, () -> {
+			assertBuildsToStart(getAllProjects());
+			assertMinimumNumberOfSimultaneousBuilds(longRunningProjects.size());
+		});
+	}
+
+	@Test
 	public void testWorkspaceBuild_NoConflictRule() throws Exception {
 		int numberOfParallelBuilds = 3;
 		setWorkspaceMaxNumberOfConcurrentBuilds(numberOfParallelBuilds);
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
@@ -15,8 +15,14 @@ package org.eclipse.core.internal.jobs;
 
 import java.util.List;
 import org.eclipse.core.internal.runtime.RuntimeLog;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.jobs.*;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.LockListener;
 
 /**
  * Captures the implicit job state for a given thread.
@@ -317,11 +323,19 @@ class ThreadJob extends Job {
 						// The actual exit conditions are listed above at the beginning of
 						// this while loop
 						int state = blockingJob.getState();
-						//ensure we don't wait forever if the blocker is waiting, because it might have yielded to me
-						if (state == Job.RUNNING && canBlock) {
-							blockingJob.jobStateLock.wait();
-						} else if (state != Job.NONE) {
-							blockingJob.jobStateLock.wait(250);
+						// Check that blockingJob has not acquired a different, non-conflicting
+						// scheduling rule since checking for conflicts by JobManager. This can
+						// particularly happen if blockingJob is a ThreadJob that is reused for the
+						// same thread across the acquisition of different rules (see ThreadJob::recycle
+						// and ImplicitJob::newThreadJob) via JobManager::beginRule.
+						if (state != Job.NONE && blockingJob.isConflicting(threadJob)) {
+							// ensure we don't wait forever if the blocker is waiting, because it might have
+							// yielded to me
+							if (state == Job.RUNNING && canBlock) {
+								blockingJob.jobStateLock.wait();
+							} else {
+								blockingJob.jobStateLock.wait(250);
+							}
 						}
 					} catch (InterruptedException e) {
 						// This thread may be interrupted via two common scenarios. 1) If


HeikoKlare:
Any objections on these changes?
Otherwise I would like to merge soon to reduce random test failures (#657) and avoid potential performance losses in parallel job execution (#659).

iloveeclipse:
Would be nice if @jukzi could check this, he recently was involved in Job related changes.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
```diff
@@ -141,6 +141,17 @@ public void testIndividualProjectBuilds_ProjectRelaxedRule() throws Exception {
 		});
 	}
 
+	@Test
+	public void testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule() throws Exception {
+		int numberOfParallelBuilds = 60;
```
jukzi, line 146: i increased it to 600 and got java.lang.AssertionError: not all build jobs have started in time
Expected: (a collection containing <P/longRunningBuildProject0> and a collection containing <P/longRunningBuildProject1> and a collection containing <P/longRunningBuildProject10> and a collection containing <P/longRunningBuildProject100> and a collection containing <P/longRunningBuildProject101> and a collection containing <P/longRunningBuildProject102> and a collection containing <P/longRunningBuildProject103> and a collection containing 
....

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
```diff
@@ -317,11 +323,16 @@ private static ThreadJob waitForRun(final ThreadJob threadJob, IProgressMonitor
...
-							blockingJob.jobStateLock.wait();
-						} else if (state != Job.NONE) {
-							blockingJob.jobStateLock.wait(250);
+						// Check that blockingJob has not acquired a different, non-conflicting
+						// scheduling rule since checking for conflicts by JobManager
+						if (state != Job.NONE && blockingJob.getRule().isConflicting(threadJob.getRule())) {
+							// ensure we don't wait forever if the blocker is waiting, because it might have
+							// yielded to me
+							if (state == Job.RUNNING && canBlock) {
+								blockingJob.jobStateLock.wait();
```
jukzi, line 332: there is still no guarantee that the rule may change between the isConflicting() check and this wait.

jukzi (changes requested):
while it might be an improvement it does not seem to totally fix the issue.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
jukzi: i increased org.eclipse.core.tests.internal.builders.ParallelBuildChainTest.TIMEOUT_IN_MILLIS to not run in timeout and the threads are stuck in
```
"Worker-584: Building longRunningBuildProject85": BLOCKED
    org.eclipse.core.internal.jobs.LockManager.removeLockWaitThread(LockManager.java:287)
    org.eclipse.core.internal.jobs.ThreadJob.waitForRun(ThreadJob.java:350)
    org.eclipse.core.internal.jobs.ThreadJob.joinRun(ThreadJob.java:214)
    org.eclipse.core.internal.jobs.ImplicitJobs.begin(ImplicitJobs.java:100)
    org.eclipse.core.internal.jobs.JobManager.beginRule(JobManager.java:342)
    org.eclipse.core.internal.resources.WorkManager.checkIn(WorkManager.java:124)
    org.eclipse.core.internal.resources.Workspace.prepareOperation(Workspace.java:2403)
    org.eclipse.core.internal.resources.Project$1.run(Project.java:591)
    org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2453)
    org.eclipse.core.internal.resources.Project.internalBuild(Project.java:642)
    org.eclipse.core.internal.resources.Project.build(Project.java:154)
    org.eclipse.core.tests.internal.builders.ParallelBuildChainTest$2.run(ParallelBuildChainTest.java:440)
    org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

```

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
HeikoKlare: I am not sure whether I understand your comment correcly.

You mean that the rule may _not_ have changed since the added `isConflicting()` call in line 328?
Correct me if I'm wrong, but since this code block is synchronized on `blockingJob.jobStateLock`, the rule of the `blockingJob` cannot change between the `isConflicting()` check and the `wait()` call. At least, that is exactly what the the `jobStateLock` is supposed to be used for.
And the rule of `threadJob` cannot change either, as that is the job is currently not executed at all.

In case you refer to the existing conflict check performed in `runNow()`, then in case this `wait()` is reached, the state did actually _not_ change. That is exactly why this `wait()` can be called safely, as a `notify()` will wake up the thread on a state change of the other job.

Maybe you can clarify in which scenario the behavior is incorrect.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
HeikoKlare: Thanks for the validation! I'll try to reproduce and figure out why this still fails.

jukzi: for numberOfParallelBuilds = 300 this PR drastically improves the test. 
sidenote: the wait logic was added with https://bugs.eclipse.org/bugs/show_bug.cgi?id=320329

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
jukzi: The synchronize ensures that bolingJob.getState() can not change. but i don't see a guarantee that the result of "blockingJob.getRule().isConflicting(threadJob.getRule())" can not change during the synchronized block. The ISchedulingRule.isConflicting(ISchedulingRule) can be custom code doing anything.

jukzi: or numberOfParallelBuilds = 300 this PR seems to have the same effect as simply changing .wait() to .wait(250) . May be it would be better to either just never unconditionally wait, or to find the right spot to notify the synchronizer.

HeikoKlare: > The synchronize ensures that bolingJob.getState() can not change. but i don't see a guarantee that the result of "blockingJob.getRule().isConflicting(threadJob.getRule())" can not change during the synchronized block. The ISchedulingRule.isConflicting(ISchedulingRule) can be custom code doing anything.

Okay, I got your point, thanks! However, `ISchedulingRule.isConflicting(ISchedulingRule)` cannoit to anything, as its specification says that "Implementations of this method must be reflexive, symmetric, and consistent". So for the same pair of rules, calling `isConflicting()` multiple times has to yield the same value.

> for numberOfParallelBuilds = 300 this PR seems to have the same effect as simply changing .wait() to .wait(250) . May be it would be better to either just never unconditionally wait, or to find the right spot to notify the synchronizer.

An unconditional wait would also solve the problem and maybe you are right that we should never wait unconditionally. Still I would be in favor to check the conflicting state again before waiting, as the wait is unnecessary in case the conflict got resolved meanwhile.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
HeikoKlare: I finally found out the following: The threads are actually not stuck but only execution only becomes really slow. The reason is that all the jobs start build operations that first acquire the `WorkspaceRoot` rule for notifying about the build. The more jobs do that in parallel, the more complex the dependency graph becomes, which the deadlock detection has to process everytime one of the jobs is not able to acquire the lock. These calculations become very expensive.

You can easily see that progress is still made but hat it becomes very slow when increasing the number of jobs by logging whenever one job is allowed proceed here:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java#L287-L292

On my laptop, 400 jobs can be processed in around 60 seconds, while more jobs will take too long. 300 jobs take around 30 seconds and 200 jobs finish almost instant.

Note that this only happens in case `ThreadJob.waitForRun` is called for a huge number of jobs with conflicting rules at the same time. I am not sure whether this is a likely scenario in realistic use cases.

So in summary, the detected issue is "only" a performance issue, of which I am unsure whether it is worth (or even necessary) to be addressed. From my point of view, it is not related to the change proposed in this PR.

HeikoKlare:
@jukzi Thanks for all the investigation. The issue you found with a large number of jobs seems to be an (already existing) performance issue (see https://github.com/eclipse-platform/eclipse.platform/pull/661#discussion_r1338598774).

While simply changing `wait()` to `wait(250)`, as you propose in https://github.com/eclipse-platform/eclipse.platform/pull/661#discussion_r1338244654, would also address the problem, I still think that re-evaluating the rule conflict is better as it proceeds on the actual state change rather than on a timeout (as the `wait()` call should actually finish due to a notify because of the job state change, which may already have happended once calling `wait()`).

Still it might make sense to replace the unconditional `wait()` call with a call using some timeout to ensure that in case anything yet undetected goes wrong the job will not hang but proceed after some time. However, that may easily hide bugs, so there should probably be some log information about such events. As this is another issue, I would defer it to a different PR.

Do you still have objections to the proposed change?

jukzi:
> While simply changing `wait()` to `wait(250)`, as you propose in [#661 (comment)](https://github.com/eclipse-platform/eclipse.platform/pull/661#discussion_r1338244654), would also address the problem, I still think that re-evaluating the rule conflict is better as it proceeds on the actual state change rather than on a timeout (as the `wait()` call should actually finish due to a notify because of the job state change, which may already have happended once calling `wait()`).

I wonder why i don't see the performance benefit from checking again over wait(250) in the junit test. If there is not measurable impact, then i suggest to make the code less complex over adding additional logic.

HeikoKlare:
You won't see a performance impact in the test for two reasons:
1. When having a small number of jobs, running into the `wait()` after the the rule of the blocking job has changed (i.e., the error case to be fixed by this PR), is an exceptional case. Otherwise the referenced test would have failed more often.
2. When starting many jobs, the impact of a small timeout in a single job does not affect the overall execution time, as the other jobs will proceed in between.

I see the point that using a timeout makes the code simpler while achieving the same result in terms of performance. However, I see three arguments speaking against that solution:
1. The code will only work quite correctly "by accident" (or more correctly: appear to work correctly), as a timeout hides a semantic error (which is waiting for some event that will never occur).
3. When executing the test with 400 jobs, this error occurs around 200 times, i.e., for half of the jobs an unnecessary `wait()` is performed.
2. Adding a timeout for the `wait()` operation is only necessary to ensure that execution proceeds in case of some other potential and yet undetected bug. That's why I would not suggest to set a low timeout value, but some rather high value like 5 or 10 seconds. In case there is no undetected bug in the code, it only makes sense to proceed execution once a notification comes from the blocking thread, as otherwise the thread would continue execution without a chance of being successful as the relevant state has not changed.

That's why I am not in favor of a "simple" solution here. Still, I improved the code by directly checking the jobs for conflicts instead of their rules iun 02fabca.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
jukzi: how can a job releases the scheduling rule while it is running? shouldn' it be constant because it is set only before scheduling?

iloveeclipse: Job can **transfer** rule, not sure if that is what you are asking for.

jukzi: i never did that which method is it?

iloveeclipse: Actually I've mixed up a bit, but both could be interesting
- org.eclipse.core.runtime.jobs.IJobManager.transferRule(ISchedulingRule, Thread)
- org.eclipse.core.runtime.jobs.Job.yieldRule(IProgressMonitor)

jukzi: transferRule does not change jobs rule or state, 
yieldRule doesn't change the rule it changes state RUNNING -> YIELDING which can't happen while jobstate is synchronized, but yes could happen between manager.runNow(threadJob, true) and synchronize. This PR would fix that. Now i see a usecase why it can help

jukzi (approved):

jukzi:
You could add a note to the commit somewhere that it is (for example?) yieldRule() which triggered the deadlock.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
HeikoKlare: It's a bit more tricky: while those places you mention may also lead to the discussed issue, the more relevent situations occur with respect to `ThreadJobs`, so the ones internally used by the `JobManager`. Those `ThreadJobs` are recycled and reused whenever such an internal job for the same thread is required, see:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java#L419

and

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java#L201-L201

So whenever one job acquires multiple rules, one after another, using `JobManager::beginRule`, the same `ThreadJob` will be reused with different rules:
https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java#L340-L343

This is, in particular, the case for build operation, as performed in the failing test. They acquire three rules, the first and last being the `WorkspaceRoot` for notifications about build start/end (which is one leading to conflicts that affect the issue addressed by this PR), and the middle one being the actual build rule, see:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java#L576

HeikoKlare:
> You could add a note to the commit somewhere that it is (for example?) yieldRule() which triggered the deadlock.

Good idea. I have adapted the code comment to reflect why this can happen and added an explanation to the commit message in af5395a. I incorporated the insights concerning the reasons for a change of the scheduling rule in https://github.com/eclipse-platform/eclipse.platform/pull/661#discussion_r1339836397.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
jukzi: Thanks for explanation. Feels like would be better to not recycle ThreadJob. The sideeffects of recycling outweight any potential benefit.
Anyway the fix looks ok.

HeikoKlare: Fully agree. I gave disabling the recycling functionality a short try and did not face any immediate regressions:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java#L225-L228

 In particular, the test we have used in this PR still runs fine and with the same performance when using a high number of jobs. I will create a follow-up issue and PR to investigate whether we can remove the recycling functionality to reduce complexity.

Thanks for the thorough review and validation, @jukzi!

PR #665, opened 2023-09-08
Version bump for 4.30 stream

akurtakov:
Tracked in
eclipse-platform/eclipse.platform.releng.aggregator#1319


akurtakov:
Test failure should be fixed after next I-build as doc bundles needed a touch in order to be regenerated.

Issue #664, opened 2023-09-07
Canceling a progress group doesn't cancel the monitors inside that group


fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform/discussions/654

When canceling a progress group, all monitors in the group should be canceled too.

PR #666, opened 2023-09-08
Don't replace the monitor set by a group when starting a new job

fedejeanne:
* Only set the progress monitor of the Job if it hasn't one already. This makes sure that the progress monitor set via `Job::setProgressGroup` is not replaced with a new one and ergo when the group is canceled, all the monitors in that group are canceled too.

* Add regression test: `JobTest.testSetProgressGroup_progressGroupIsCanceled_monitorInJobIsCanceled()`

* Simplify private method `JobManager::createMonitor(Job)` and remove the annotation `@GuardedBy` in it. 

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/664

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
@@ -23,11 +23,29 @@ import static org.junit.Assert.assertNotEquals;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.concurrent.atomic.*;
-import org.eclipse.core.internal.jobs.*;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.jobs.*;
-import org.eclipse.core.tests.harness.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.concurrent.atomic.AtomicLong;
+import org.eclipse.core.internal.jobs.InternalJob;
+import org.eclipse.core.internal.jobs.JobListeners;
+import org.eclipse.core.internal.jobs.JobManager;
+import org.eclipse.core.internal.jobs.Worker;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.QualifiedName;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobChangeEvent;
+import org.eclipse.core.runtime.jobs.IJobChangeListener;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobChangeAdapter;
+import org.eclipse.core.runtime.jobs.LockListener;
+import org.eclipse.core.tests.harness.FussyProgressMonitor;
+import org.eclipse.core.tests.harness.TestBarrier2;
+import org.eclipse.core.tests.harness.TestJob;
 import org.junit.Assert;
 
 /**
@@ -1600,6 +1618,37 @@ public class JobTest extends AbstractJobTest {
 		waitForState(job, Job.NONE);
 		assertEquals("1.0", IStatus.CANCEL, job.getResult().getSeverity());
 		group.done();
+	}
+
+	public void testSetProgressGroup_propagatesGroupCancellation() throws InterruptedException {
+		Job checkMonitorJob = createJobReactingToCancelRequest();
+
+		final IProgressMonitor progressGroup = Job.getJobManager().createProgressGroup();
+		progressGroup.beginTask("", 2);
+
+		checkMonitorJob.setProgressGroup(progressGroup, 1);
+
+		progressGroup.setCanceled(true);
+
+		checkMonitorJob.schedule();
+		checkMonitorJob.join();
+
+		assertEquals("The job should have been canceled", IStatus.CANCEL, checkMonitorJob.getResult().getSeverity());
+	}
+
+	private Job createJobReactingToCancelRequest() {
+		return new Job("Check if own monitor is canceled") {
+
+			@Override
+			protected IStatus run(IProgressMonitor monitor) {
+
+				if (monitor.isCanceled()) {
+					return Status.CANCEL_STATUS;
+				}
+
+				return Status.OK_STATUS;
+			}
+		};
 	}
 
 	/*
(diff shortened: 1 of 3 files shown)


runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
```diff
@@ -1602,6 +1620,38 @@ protected IStatus run(IProgressMonitor monitor) {
 		group.done();
 	}
 
+	public void testSetProgressGroup_progressGroupIsCanceled_monitorInJobIsCanceled() throws InterruptedException {
```
HeikoKlare, line 1623: I guess the name is meant to follow some pattern like `testMethod_UseCase_ExpectedResult`, but without know that pattern (which would require to have it as default across the project) the test name's intention is hard to understand. Wouldn't something like `testSetProgressGroup_propagatesGroupCancellation` be more comprehensible and equally expressive?

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
```diff
@@ -1602,6 +1620,38 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		Job checkMonitorIsCanceled = createJobThatChecksItsOwnMonitor();
```
HeikoKlare, line 1624: Please reflect in the variable name that it references a job, i.e., `checkMonitorIsCanceledJob` or the like. Otherwise the variable sounds as if it was a method reference.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
```diff
@@ -1602,6 +1620,38 @@ protected IStatus run(IProgressMonitor monitor) {
...
+
+		final IProgressMonitor progressGroup = Job.getJobManager().createProgressGroup();
+		progressGroup.beginTask("", 2);
+
+		checkMonitorIsCanceled.setProgressGroup(progressGroup, 1);
+
+		// let the job do the cancellation
```
HeikoKlare, line 1631: I do not understand that comment as the job does not cancel anything. I don't think that a comment is necessary for comprehension.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
```diff
@@ -1602,6 +1620,38 @@ protected IStatus run(IProgressMonitor monitor) {
...
+		progressGroup.setCanceled(true);
+
+		checkMonitorIsCanceled.schedule();
+		checkMonitorIsCanceled.join();
+
+		assertEquals("The job should have been canceled", IStatus.CANCEL, checkMonitorIsCanceled.getResult().getSeverity());
+	}
+
+	private Job createJobThatChecksItsOwnMonitor() {
```
HeikoKlare, line 1640: nit: this naming sounds as if it was some special case that a job checks its own monitor but actually that's "usualy behavior". Isn't the purpose of the job simply to react to a cancel request, i.e., `createJobReactingToCancelRequest`?

HeikoKlare (commented):
Changes look good to me. I obviously agree with the behavioral change in the `JobManager`, as stated in https://github.com/eclipse-platform/eclipse.platform/discussions/654#discussioncomment-6904513. I have some minor comments regarding code style within the added test.
And I would be in favor of separating the version bump from the functional changes in terms of different commits, but I am not sure if there is a policy for that.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobTest.java
fedejeanne: Changed it to `checkMonitorJob` in a055cfab73062f23630a5873c2eb8c4a7ee1f381

HeikoKlare (approved):

fedejeanne:
> And I would be in favor of separating the version bump from the functional changes in terms of different commits, but I am not sure if there is a policy for that.

I'm not aware of one such policy but I'm 99% sure I have bumped the version in the same PR as the functional changes before. It makes sense to me since bumping the version is mandatory in this case :)

mickaelistria:
There is no strict policy, but the recommendation is to make the change and the bump as 2 distinct commits in the same PR, so the functional part can be cherry-picked to older branches more easily if needed. But we usually won't reject a PR if the bump is included in the same commit as the code change.

fedejeanne:
Thank you both for the hint, I'll take it into account from now on 👍 

I created a PR to mention this in the documentation --> https://github.com/eclipse-platform/.github/pull/151

fedejeanne:
I had to bump another version so I split the commits anyway

HeikoKlare:
Looks like the feature version of `org.eclipse.core.runtime.feature` has to be bumped as well, as the version of the included plugin `org.eclipse.core.jobs` was increased.
I wonder why this was not necessary when merging #625, as it bumped the version of `org.eclipse.core.expressions`, which is also included in that feature.

fedejeanne:
Test feailures documented in https://github.com/eclipse-platform/eclipse.platform/issues/678 and https://github.com/eclipse-platform/eclipse.platform/issues/668 (that last one should be fixed though :-\ )

HeikoKlare:
The commit message of the version bump commit is not correct. It does neither bump the core.tests.runtime bundle version (I guess that was already done by another commit on which this PR was rebased) nor the version of the feature bundle.

I am not sure why the ToC tests (see #668) are failing again, as they succeeded in test runs this morning. There also seems to be some other configuration problem w.r.t. the changed jobs plug-in as the builds fail with validation errors:
```
Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.2:verify (verify) on project org.eclipse.core.jobs: Execute ApiApplication failed: Application returned exit code 10
``` 
 The logs also show this, which I have not seen for the core.jobs plug-in in other builds yet:
```
The following implicit projects are not referenced: 
org.eclipse.platform:org.eclipse.core.jobs:eclipse-plugin:3.15.100-SNAPSHOT
```

fedejeanne:
> The commit message of the version bump commit is not correct. It does neither bump the core.tests.runtime bundle version (I guess that was already done by another commit on which this PR was rebased) nor the version of the feature bundle.

Correct, `org.eclipse.core.tests.runtime` was bumped in 8a3f6d18d3c791880722e68b0ee9fd6f1b996ede and `org.eclipse.core.runtime.feature` in 5531532dad883fb13f1c833983b8a866a0285d11. I reworded the commit.

> I am not sure why the ToC tests (see #668) are failing again, as they succeeded in test runs this morning. There also seems to be some other configuration problem w.r.t. the changed jobs plug-in as the builds fail with validation errors:
> 
> ```
> Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.2:verify (verify) on project org.eclipse.core.jobs: Execute ApiApplication failed: Application returned exit code 10
> ```
> 
> The logs also show this, which I have not seen for the core.jobs plug-in in other builds yet:
> 
> ```
> The following implicit projects are not referenced: 
> org.eclipse.platform:org.eclipse.core.jobs:eclipse-plugin:3.15.100-SNAPSHOT
> ```

No idea why since my changes are rather simple, nothing I haven't done before. Let's see if it's a random failure, I just rebased.

HeikoKlare:
Great, checks now run successfully.
The log message seems to only be related to a "cyclic dependency" as `core.jobs` has been bumped, the platform projects depend on `equinox.registry` which in turn depends on a package from `core.jobs`.

PR #667, opened 2023-09-08
Use jetty-jspc-maven-plugin 10.0.16

akurtakov:


akurtakov:
Syncs jspc compiler with jetty version updated in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1331

PR #672, opened 2023-09-08
Fixes #671 - Move some platform-ui undo utility classes into a core location

### LINKED ISSUE #671: Resource Description classes in o.e.ui.internal.ide.undo used by jdt, better placed in platform
A subset of classes found in org.eclipse.ui.internal.ide.undo is currently used by classes in the jdt.ui bundle, however, these classes would rightly be non-UI operations if they did not have dependencies on org.eclipse.ui. 

Efforts to move this logic down into non-ui bundles, and to assist headless or background process environments make use of as much of eclipse as possible, cannot proceed unless these classes are in a non-ui location that is appropriate for both platform ui's undo logic as well as jdt to access. 

The set of classes appears to be the following:

AbstractResourceDescription.class
ContainerDescription.class
FileDescription.class
FolderDescription.class
IFileContentDescription.class
MarkerDescription.class
ProjectDescription.class

robstryker:


robstryker:
See https://github.com/eclipse-platform/eclipse.platform/issues/671 for description

mickaelistria:
Couldn't those be in an internal but exported with `x-friends` package?

mickaelistria:
I'm a bit worried about the names of all that. For example having a "ProjectDescription" API when there is already `IProjectDescription` as a main object is likely to bring confusion. I'm also not sure about the "descriptors" package which doesn't make it clear what these classes are intended to.
Can you please also show draft PRs for Platform UI and JDT to see how this change would impact consumers?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/descriptions/IFileContentDescription.java
```diff
@@ -0,0 +1,59 @@
...
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.resources.descriptions;
+
+import java.io.InputStream;
+import org.eclipse.core.runtime.CoreException;
+
+/**
+ * IFileContentDescription is a description of a file's content.
```
mickaelistria, line 20: Why not using directly IFile or IFileInfo for that?

mickaelistria:
OK, I start to understand it better that those classes are meant to get a local snapshot of a resource to store in the undo/redo stack and process later. If there are no other consumers than undo/redo, then we can probably just show `undo` in the package name, and instead of "Descriptions" name them "Snapshots".
Couldn't we also move in here that bits that take care of undo/restoring (ie applying the "snapshot" to the workspace) ?

robstryker:
I'm a +1 on changing package name to include an `undo`. I trhink that would add a lot of clarity. 

I'm generally against renaming the classes to 'Snapshots' unless there's a wider consensus that's the correct thing to do. Changing package name is one thing but changing the class name, well, I'm not sure. I'll do it if demanded, but... 

For JDT, the patch to migrate from use of the ui.ide.undo classes to the new core ones would be as simple as changing imports (assuming we don't change the class name, just the package). This is because our primary entry point to creating these objects is by calling `ResourceDescription.fromResource(rootResource);`. 

For eclipse ui, I'd need to investigate more deeply to see how extensively these classes are used. I can see that `org.eclipse.ui.ide.undo,` is an exported package that is public API. Migration would probably involve having existing classes extend those in core.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/AbstractResourceSnapshot.java
```diff
@@ -0,0 +1,140 @@
...
+import org.eclipse.core.resources.ResourceAttributes;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+
+/**
+ * Base implementation of ResourceSnapshot that describes the common attributes
+ * of a resource to be created.
+ *
+ * This class is not intended to be instantiated or used by clients.
```
mickaelistria, line 30: We should maybe remove this line?

mickaelistria (commented):
That looks good overall, just a minor comment inline that we need to address before moving forward.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/AbstractResourceSnapshot.java
robstryker: This line should not be removed. The approved way of accessing these classes is by calling ResourceSnapshot.fromResource(r);

mickaelistria: So couldn't the constructors be only package-visible? And the documentation mention that the right approach to get an instance if `ResourceSnapshot.fromResource()` ?

mickaelistria:
The test failure is also visible on master and doesn't relate to the change (ie actually tests the content of help for last I-Build, not the one that would be generated by this PR).
While I still think we should aim at moving the extension point to make it easier for consumers to use it, I think it's not too bad to have its description in the "legacy" bundle at the moment. So, unless we find a quick and easy solution to the problem (which are yet unclear to me), it would be OK to merge it.

However, I still miss the exact symptom of the problem we're trying to fix. @robstryker do you have a link to some failure somewhere? If it's the current test that we see failing here that we're trying to fix, then the solution can be simply to list the "new" bundle in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/725cacd808359189e38bda85b416ed80d03bda22/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/buildDoc.xml#L54

robstryker:
I'll be refactoring this patch to ensure that org.eclipse.ui.ide can properly consume these in an api-compatible fashion. Please hold off.

mickaelistria:
> Please hold off.

OK, I'm converting to draft in the meantime.

robstryker:
> The test failure is also visible on master and doesn't relate to the change (ie actually tests the content of help for last I-Build, not the one that would be generated by this PR). While I still think we should aim at moving the extension point to make it easier for consumers to use it, I think it's not too bad to have its description in the "legacy" bundle at the moment. So, unless we find a quick and easy solution to the problem (which are yet unclear to me), it would be OK to merge it.
> 
> However, I still miss the exact symptom of the problem we're trying to fix. @robstryker do you have a link to some failure somewhere? If it's the current test that we see failing here that we're trying to fix, then the solution can be simply to list the "new" bundle in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/725cacd808359189e38bda85b416ed80d03bda22/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/buildDoc.xml#L54

I think this comment was meant to be posted to a different issue.  This issue here is not prompted by any test failures.

robstryker:
I have significantly refactored the proposed patch, to add an internal package as the current ui.ide had done. This was an oversight on my part since only 1 of the classes should have been public. 

In order to maintain API compatibility with prior consumers of the ui.ide code, I have chosen to make use of interfaces instead of only abstract classes. This allows the existing UI code to simply wrap the core implementation and implement the same interface. This will  maintain compatability with all of the existing classes and the expected return values. More of that can be discussed on an issue in the UI repository when appropriate.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/IContainerSnapshot.java
```diff
@@ -0,0 +1,59 @@
...
+package org.eclipse.core.resources.undo.snapshot;
+
+import java.net.URI;
+import org.eclipse.core.resources.IResourceFilterDescription;
+
+/**
+ * IContainerSnapshot is a lightweight description that describes a container to
+ * be created.
+ *
+ * This class is not intended to be instantiated by clients.
```
mickaelistria, line 24: This line can be removed as interfaces cannot be instantiated.
However, 1 important thing to settle when introducing this interface API is whether we want to allow other implementations? If we want to restrict the implementations to the ones that are shipped here, we should add a `@noimplements` tag to make PDE warn in case someone wants to implement it.
It's safer to start with as many restrictions as possible and to remove them later than the other way round, as it gives more freedom for further changes.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/IMarkerSnapshot.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.eclipse.core.resources.IMarker;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+
+/**
+ * IMarkerSnapshot is a lightweight snapshot of a marker for the purposes of
+ * undoing.
+ *
+ * This class is not intended to be instantiated by clients.
```
mickaelistria, line 25: Same as above.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/IResourceSnapshot.java
```diff
@@ -0,0 +1,109 @@
...
+
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+
+/**
+ * IResourceSnapshot is a lightweight snapshot that describes the common
+ * attributes of a resource to be created.
+ *
+ * This class is not intended to be extended by clients.
```
mickaelistria, line 26: same as above

mickaelistria (changes requested):

mickaelistria (approved):

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/undo/snapshot/IResourceSnapshot.java
```diff
@@ -0,0 +1,109 @@
...
+	 * Given a resource handle, create an actual resource with the attributes of
+	 * the receiver resource description.
+	 *
+	 * @param resource
+	 *            the resource handle
+	 * @param monitor
+	 *            the progress monitor to be used when creating the resource
+	 * @throws CoreException if creation failed
+	 */
+	public void createExistentResourceFromHandle(IResource resource,
```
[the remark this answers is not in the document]
mickaelistria: I agree it looks like this particular thing can be improved. I'm also curious about why is this method useful when we already have `createResource`. @robstryker in which case would one want to call it instead of `createResource`? What about a new `restoreResource` that would join both?

Issue #668, opened 2023-09-08
Failure of test org.eclipse.ua.tests.doc.internal.linkchecker.TocLinkChecker.testPlatformIsvStatic


fedejeanne:
Failed in [this build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6121624211/job/16615790108?pr=666#step:7:16243) of https://github.com/eclipse-platform/eclipse.platform/pull/666

```java
org.eclipse.ua.tests.doc.internal.linkchecker.TocLinkChecker.testPlatformIsvStatic -- Time elapsed: 0.143 s <<< FAILURE!
java.lang.AssertionError: 
Invalid link in "/org.eclipse.platform.doc.isv/topics_Porting.xml": porting/eclipse_4_29_porting_guide.html
Invalid link in "/org.eclipse.platform.doc.isv/topics_Porting.xml": porting/4.29/faq.html
Invalid link in "/org.eclipse.platform.doc.isv/topics_Porting.xml": porting/4.29/incompatibilities.html
Invalid link in "/org.eclipse.platform.doc.isv/topics_Porting.xml": porting/4.29/recommended.html

	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.ua.tests.doc.internal.linkchecker.TocLinkChecker.doAssert(TocLinkChecker.java:129)
	at org.eclipse.ua.tests.doc.internal.linkchecker.TocLinkChecker.testPlatformIsvStatic(TocLinkChecker.java:84)
```

akurtakov:
This should be fixed (via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/fb48617de266ff5fc19a8921c9ebd2f0e53cf767 ) . Let's see tomorrow's I-build.

HeikoKlare:
As far as I can see, the porting information has only been merged into the `R4_29_maintenance` branch but not into `master`: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1290

Maybe that's the reason why they are missing?

akurtakov:
Isn't https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/c5892bd1d67d98adc0c7c6d387be9d540c2a3ca1 the patch to master?

HeikoKlare:
[...]

PR #674, opened 2023-09-11
Delete tests in org.eclipse.core.tests.resources #525

Michael5601:
In this commit, the deactivated performance tests "_testPerformanceManyResources()"  and "_testPerformanceOneResource()" from MarkerTest.java in org.eclipse.core.tests.resources are deleted because they are not real test methods but rather used for code measuring and printing the time for processing operations on markers. See also https://github.com/eclipse-platform/eclipse.platform/pull/605#issuecomment-1713565424. Contributes to #525


HeikoKlare (approved):

HeikoKlare:
Failing test is documented in #668

PR #675, opened 2023-09-11
Add test to JobGroupTest

fedejeanne:
Add the test method `org.eclipse.core.tests.runtime.jobs.JobGroupTest.testShouldCancel_5()`. This test proves that the cancelation of "sibling" jobs in a `JobGroup` should be done via the **cancelation policy** of the group and not via the monitor.

This PR provides doesn't solve any issue, it merely provides context to this discussion: https://github.com/eclipse-platform/eclipse.platform/discussions/654 (see [comment](https://github.com/eclipse-platform/eclipse.platform/discussions/654#discussioncomment-6967309))

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
@@ -16,14 +16,30 @@ package org.eclipse.core.tests.runtime.jobs;
 
 import static org.junit.Assert.assertNotEquals;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.concurrent.atomic.AtomicLong;
 import junit.framework.AssertionFailedError;
-import org.eclipse.core.runtime.*;
-import org.eclipse.core.runtime.jobs.*;
-import org.eclipse.core.tests.harness.*;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobChangeEvent;
+import org.eclipse.core.runtime.jobs.IJobChangeListener;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobChangeAdapter;
+import org.eclipse.core.runtime.jobs.JobGroup;
+import org.eclipse.core.tests.harness.FussyProgressMonitor;
+import org.eclipse.core.tests.harness.TestBarrier2;
+import org.eclipse.core.tests.harness.TestJob;
 
 /**
  * Tests for {@link JobGroup}.
@@ -1291,6 +1307,74 @@ public class JobGroupTest extends AbstractJobTest {
 		// Verify that the group is canceled in a reasonable time,
 		// i.e only 10 jobs are allowed to run after the shouldCancel method returned true.
 		assertTrue("2.0", numShouldCancelCalled[0] < NUM_JOBS_LIMIT + 10);
+	}
+
+	/**
+	 * Tested scenario: - 2 jobs are started in parallel and the first one returns a
+	 * CANCEL_STATUS while the second one is still running. This lets
+	 * <code>JobGroup.shouldCancel</code> return <code>true</code> which triggers
+	 * the cancelation of any running job.
+	 *
+	 * Expected result: The second job is explicitly canceled
+	 *
+	 */
+	public void testShouldCancel_5() {
+		// the job group allows for 2 threads so both jobs will be started in
+		// parallel
+		JobGroup jobGroup = new JobGroup("JobGroup", 2, 2) {
+			@Override
+			protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFailedJobs,
+					int numberOfCanceledJobs) {
+				// cancel the group as soon as one job in it returns the CANCEL_STATUS
+				return numberOfCanceledJobs > 0;
+			}
+		};
+
+		TestBarrier2 returnsCancelStatusBarrier = new TestBarrier2();
+		TestBarrier2 canceledJobBarrier = new TestBarrier2();
+
+		Job returnsCancelStatusJob = Job.create("The one that cancels", __ -> {
+			// This job running. Wait until the other job is running too.
+			returnsCancelStatusBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
+			canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_RUNNING);
+
+			return Status.CANCEL_STATUS;
+		});
+
+		Job canceledJob = new Job("The one that is explicitly canceled") {
+			@Override
+			protected IStatus run(IProgressMonitor monitor) {
+				// This job running. Wait until the other job is running too.
+				canceledJobBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
+				returnsCancelStatusBarrier.waitForStatus(TestBarrier2.STATUS_RUNNING);
+
+				// give it time: this Job needs to be running at the time the other job finishes
+				canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_DONE);
+				return Status.OK_STATUS;
+			}
+
+			@Override
+			protected void canceling() {
+				canceledJobBarrier.setStatus(TestBarrier2.STATUS_DONE);
+			}
+		};
+
+		returnsCancelStatusJob.setJobGroup(jobGroup);
+		returnsCancelStatusJob.schedule();
+
+		canceledJob.setJobGroup(jobGroup);
+		canceledJob.schedule();
+
+		// Wait here also just in case there is an exception. The main thread is the
+		// right place to let tests fail.
+		canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_DONE);
+
+		waitForCompletion(jobGroup);
+
+		/*
+		 * The fact that we reached the end of the method means that the 2nd job was
+		 * successfully canceled i.e. that the test passed.
+		 */
 	}
 
 	public void testDefaultComputeGroupResult() {


fedejeanne:
I'm a bit baffled about [this error](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6156674859/job/16705833723?pr=675#step:7:22137) (it happens under Linux): 
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.2:verify (verify) on project org.eclipse.help.base: Execute ApiApplication failed: Application returned exit code 10 -> [Help 1]
```

It doesn't say much. Any ideas what could it be or how to fix it?

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,88 @@ public IStatus run(IProgressMonitor monitor) {
...
+			protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFailedJobs,
+					int numberOfCanceledJobs) {
+				// cancel the group as soon as one job in it is canceled
+				return numberOfCanceledJobs > 0;
+			}
+		};
+		for (int i = 0; i < NUM_JOBS; i++) {
+			final int jobNumber = i;
+
+			Job job = new TestJob("TestJob", 1, 1) {
```
HeikoKlare, line 1345: Is a TestJob necessary here if it only performs one tick of one millisecond?

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,88 @@ public IStatus run(IProgressMonitor monitor) {
...
+					if (canceledExplicitly || jobNumber == NUM_JOBS_LIMIT) {
+						// after this cancelation, no more jobs will be scheduled and the running jobs
+						// should be explicitly asked to cancel their execution
+						return Status.CANCEL_STATUS;
+					}
+
+					super.run(monitor);
+					countJobsRan[0]++; // this job ran
+
+					return new Status(IStatus.INFO, "org.eclipse.core.jobs", "Job " + jobNumber);
```
HeikoKlare, line 1361:
```suggestion
					return new Status(IStatus.INFO, JobManager.PI_JOBS, "Job " + jobNumber);
```

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,88 @@ public IStatus run(IProgressMonitor monitor) {
...
+		final int MAX_THREADS = 10;
+
+		// make sure some jobs haven't yet been queued by the time the cancelation takes
+		// place
+		final int NUM_JOBS_LIMIT = MAX_THREADS + 100;
+		final int NUM_JOBS = NUM_JOBS_LIMIT + 1000;
+
+		// counters
+		final int countJobsRan[] = { 0 };
+		final int countJobsCanceled[] = { 0 };
```
HeikoKlare, line 1331: I would prefer to use `AtomicInteger` or the like, as code becomes pretty hard to read when abusing an array with a single element named after the meaning of the first element rather than the meaning of the array.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,88 @@ public IStatus run(IProgressMonitor monitor) {
...
+		}
+
+		// Allow the jobs to proceed to run.
+		barrier.setStatus(TestBarrier2.STATUS_START);
+		waitForCompletion(jobGroup);
+
+		// Some jobs might have already been running by the time the cancelation took
+		// place
+		assertTrue("Too many jobs ran: " + countJobsRan[0], //
+				countJobsRan[0] <= NUM_JOBS_LIMIT + MAX_THREADS);
```
HeikoKlare, line 1384: Please use hamcrest with a proper lessThanOrEqualTo comparison.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
HeikoKlare: In addition, incrementing the first array element from different threads is a race condition.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
HeikoKlare: There is not memory barrier on `countJobsRan`, so this comparison can succeed although it shouldn't.

HeikoKlare (changes requested):
The added test randomly fails due to multiple race conditions. In particular, there is no order of the jobs to be executed, so cancelling the job group once a job with a specific number is executed does not give any guarantee regarding the number of jobs that have run in the end.

To reproduce a failure due to a race condition, add a sleep of several milliseconds into the if clause of the run method. The test will constantly fail.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1313,80 @@ public IStatus run(IProgressMonitor monitor) {
...
+				// give it time: this Job needs to be running at the time the other job finishes
+				awaitLatch(explicitlyCanceled);
+				return Status.OK_STATUS;
+			}
+
+			@Override
+			protected void canceling() {
+				explicitlyCanceled.countDown();
+			}
+		};
```
HeikoKlare, line 1363: I propose the usage of a barrier instead of multiple latches as follows for these reasons:
1. There is no necessity to synchronize the jobs via the main thread, they can simply trigger each other.
2. The barrier already provides the functionality (and more) as defined in `awaitLatch()`.
3. The barrier also prints a stack dump in case the test does not finish. This is important for debugging why one of the jobs did not proceed.

With the proposed change, you can replace all the latch logic in the main thread with a `testBarrier.waitForStatus(TestBarrier2.STATUS_DONE)` that prints a stack trace in case the job waiting for cancellation does not finish.


```suggestion
		Job returnsCancelStatusJob = Job.create("The one that cancels", __ -> {
			testBarrier.setStatus(TestBarrier2.STATUS_START);
			testBarrier.waitForStatus(TestBarrier2.STATUS_RUNNING);
			return Status.CANCEL_STATUS;
		});

		Job canceledJob = new Job("The one that is explicitly canceled") {
			@Override
			protected IStatus run(IProgressMonitor monitor) {
				testBarrier.waitForStatus(TestBarrier2.STATUS_START);
				testBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
				testBarrier.waitForStatus(TestBarrier2.STATUS_DONE);
				return Status.OK_STATUS;
			}

			@Override
			protected void canceling() {
				testBarrier.setStatus(TestBarrier2.STATUS_DONE);
			}
		};
```

HeikoKlare (commented):
I propose the usage of a barrier instead of multiple latches for the reasons I give in the code comment.
In addition, the current implementation does not guarantee that all jobs are finished when the test finishes, as `waitForCompletion` has a timeout and may leave the jobs running. Using a barrier that waits in the main thread for at least the time the jobs themselves wait on that barrier avoids that risk.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
fedejeanne: Thank you for the suggestion. I changed the code but used 2 barriers instead of one. I find it easier to read if every job has its own barrier and they are both orchestrated from the main thread.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1310,76 @@ public IStatus run(IProgressMonitor monitor) {
...
+
+		canceledJob.setJobGroup(jobGroup);
+		canceledJob.schedule();
+
+		// Wait until both jobs are ready to run and let them start running
+		returnsCancelStatusBarrier.waitForStatus(TestBarrier2.STATUS_WAIT_FOR_RUN);
+		canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_WAIT_FOR_RUN);
+
+		returnsCancelStatusBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
+		canceledJobBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
```
HeikoKlare, line 1374: Having two barriers is good to clarify to which job its state blongs. Still, I find the current barrier handling quite "complex". The intended semantics is as follows: Job 1 proceeds only when Job 2 started and vice versa, i.e., each job should set its barrier to "running" once started and wait for the other to be "running" before proceeding. In particular, the barrier is not about having a central mediator (the main thread) but two jobs waiting for specific state changes of the other. So I would still propose to remove the barrier logic from the main thread (execept for the wait for the final execution of the jobs) and leave the coordination up to the jobs.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1310,76 @@ public IStatus run(IProgressMonitor monitor) {
...
+
+		// give it time: the job needs to be canceled
+		canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_DONE);
+
+		assertTrue("The second job should have been explicitly canceled", //
+				jobWasCanceled.get());
```
HeikoKlare, line 1380: This is basically dead code. This assertion can only fail  if the `waitForStatus` call fails and in that case the test already ends after that call. So `jobWasCanceled` is obsolete, as it is already checked through the barrier.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1310,76 @@ public IStatus run(IProgressMonitor monitor) {
...
+		// Wait until both jobs are ready to run and let them start running
+		returnsCancelStatusBarrier.waitForStatus(TestBarrier2.STATUS_WAIT_FOR_RUN);
+		canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_WAIT_FOR_RUN);
+
+		returnsCancelStatusBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
+		canceledJobBarrier.setStatus(TestBarrier2.STATUS_RUNNING);
+
+		// give it time: the job needs to be canceled
+		canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_DONE);
+
```
HeikoKlare, line 1378: You still have to wait for the job group to finish after, as the second job may still be running after the barrier reaches the "done" status.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
fedejeanne: Thank you, I used 2 barriers and removed the orchestration from the main thread, as requested.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
fedejeanne: Good point, thank you!
I added a `waitForCompletion(jobGroup)`

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
fedejeanne: I removed the assertion and added a comment that makes it clear, that reaching the end of the test method means that the test passed.
FWIW, I know it was dead code, I just find it more expressive to have assertions and variables with meaningful names in a test, I think it better expresses the purpose of the test.

HeikoKlare: I see the point. But then there should be a documentation that states why the code is there despite being unnecessary (from a functional perspective). Otherwise, someone maintaining the test will wonder what this (dead) code is supposed to do.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,71 @@ public IStatus run(IProgressMonitor monitor) {
...
+			}
+		};
+
+		returnsCancelStatusJob.setJobGroup(jobGroup);
+		returnsCancelStatusJob.schedule();
+
+		canceledJob.setJobGroup(jobGroup);
+		canceledJob.schedule();
+
+		waitForCompletion(jobGroup);
```
HeikoKlare, line 1368: Now waiting for the barrier state change before waiting job group completion is missing again:
`canceledJobBarrier.waitForStatus(TestBarrier2.STATUS_DONE);`

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1293,6 +1309,71 @@ public IStatus run(IProgressMonitor monitor) {
...
+
+		/*
+		 * The fact that we reached the end of the method means that the 2nd job was
+		 * successfully canceled i.e. that the test passed. This is why there are no
+		 * assertions in this test.
```
HeikoKlare, line 1373: This comment is a bit misleading, as `TestBarrier2::waitForStatus` actually is an assertion (although the name does not express that).

HeikoKlare (changes requested):

HeikoKlare (approved):
The test seems to be sound now. Hope that we did not drop something important during all the iterations 😉 

We can merge after the freeze period.

PR #681, opened 2023-09-13
Reactivate tests in org.eclipse.debug.tests #525

Michael5601:
In this commit the test testCheckReceiver() from the class CheckTests.java in org.eclipse.debug.tests is reactivated and fixed. It is about checking if a checkbox is properly toggled in the ui after changing its state. After fixing the issue the checked box can be clearly seen while debuging the code. Contributes to #525.


HeikoKlare (approved):
The reenabled and adapted test looks good. With #689, the test in now executed in GH actions builds a virtual viewer implementation and succeeds on all platforms. With the actual JFace viewer implementation that is not used during CI builds, the test also succeeds when executed from within an Eclipse IDE at least on Windows.

HeikoKlare:
Thanks for recovering the test!

PR #688, opened 2023-09-15
Fix org.apache.jasper.glassfish dependency in infocenter.product

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1251

### DIFF
--- a/ua/infocenter-web/infocenter-product/infocenter.product
+++ b/ua/infocenter-web/infocenter-product/infocenter.product
@@ -36,7 +36,6 @@ org.osgi.framework.bootdelegation=*
    <plugins>
       <plugin id="com.ibm.icu"/>
       <plugin id="org.apache.felix.scr"/>
-      <plugin id="org.apache.jasper.glassfish"/>
       <plugin id="org.apache.lucene.analysis-common"/>
       <plugin id="org.apache.lucene.analysis-smartcn"/>
       <plugin id="org.apache.lucene.core"/>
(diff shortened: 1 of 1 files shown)


merks:
@iloveeclipse 

Should we just merge this so we can kick off an I-Build?

iloveeclipse:
> @iloveeclipse
> 
> Should we just merge this so we can kick off an I-Build?

Don't we need batch compiler in this product?

iloveeclipse:
In doubt, I would propose to add batch compiler.

merks:
I think it wouldn't hurt and we already realize that we need it (it's package export) in other places so I tend to agree.

Note that we have other "strange dependencies" that I didn't understand their purpose also related to this theme:

https://github.com/eclipse-platform/eclipse.platform/blob/e5f0703cd0427cfda45c9b8c1a02117531b30dfa/ua/org.eclipse.ua.tests/pom.xml#L58-L69

merks:
@iloveeclipse 

FYI, it looks like the continuous-integration/jenkins/pr-head build got past building the infocenter-product:
```
09:03:56.831 [INFO] --- tycho-apitools:4.0.2:verify (verify) @ infocenter-product ---
09:03:56.833 [INFO] 
09:03:56.833 [INFO] ---------< eclipse.platform.ua:org.eclipse.ui.intro.universal >---------
```
Do you still think it's best to include the batch compiler immediately?

merks:
Everything appears to have build, but there continue to be test failures that may well be because of missing things but those things might handled previous via the pom dependencies as mentioned here:

https://github.com/eclipse-platform/eclipse.platform/pull/688#issuecomment-1720934878

@iloveeclipse 

Please advise at how you would like to proceed...  I think/hope that with these changes the I-Build would work, although with test failures...

iloveeclipse:
Compile time doesn't need "jsp compiler" but at runtime product will miss the compiler package for sure.
Ed, please add batch compiler to the product, merge PR & start I Build.

iloveeclipse:
> Note that we have other "strange dependencies" that I didn't understand their purpose also related to this theme:

No idea. I guess an attempt to make similar change like we do now.

merks:
Build running:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.30/15/

merks:
@iloveeclipse 

The build completed.

I assume there are likely to be test failures, though those UA tests are passing for me...

image

Issue #682, opened 2023-09-14
failing linkchecker.ApiDocTest.testTopicsReference


jukzi:
ApiDocTest.testTopicsReference fails i-build I20230913-1800
foer example:
https://download.eclipse.org/eclipse/downloads/drops4/I20230913-1800/testresults/html/org.eclipse.ua.tests.doc_ep430I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

```
expected:<[]> but was:<[ * Unexpected extension points in topics_Reference.xml: org.eclipse.search.textSearchEngine ]>

org.junit.ComparisonFailure: expected:<[]> but was:<[
* Unexpected extension points in topics_Reference.xml:
org.eclipse.search.textSearchEngine
]>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.ua.tests.doc.internal.linkchecker.ApiDocTest.testTopicsReference(ApiDocTest.java:181)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

iloveeclipse:
Regression from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1105.

@mickaelistria : could you please take a look?

iloveeclipse:
Also note [comment](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1349#issuecomment-1719004119) from @akurtakov regarding failing releng test:

> It's in the https://download.eclipse.org/eclipse/downloads/drops4/I20230913-1800/logs.php#javadoc . https://download.eclipse.org/eclipse/downloads/drops4/I20230913-1800/compilelogs/platform.doc.isv.javadoc.txt points that first failure is due to search.[core/ui] split and not being added to javadoc build.

PR #689, opened 2023-09-15
Add VirtualViewerCheckTests for automated execution of JFace CheckTests

HeikoKlare:
Tests using JFace viewers have been replaced by tests on a mock viewer implementation for execution in automated builds. The CheckTests class, however, only has an implementation for a JFace viewer to be executed locally, but no implementation for a mock viewer to be executed in automated builds.

This change adds an according test class and adds it to the test suite for automated builds.

This also serves as a means to enable automated testing of changes proposed in #681.


HeikoKlare:
Test fails are unrelated and documented in #682.

Issue #468, opened 2023-05-30
[Performance] PropertyManager2 iterates the tree for infinite depth for deleting/moving an IFile


lathapatil:
[Bugzilla Issue](https://bugs.eclipse.org/bugs/show_bug.cgi?id=570267) is reproducible in latest builds.

Testcase to reproduce the issue is attached.

Better explanation of the issue at https://www.eclipse.org/forums/index.php/t/1106574/

**Description :**
I have a scenario, where I have around 10K files inside a folder named '_out'. and '_out' in-turn contains other folders and inner files and sub folders.
I'm deleting around 6000 files , which are directly present under the _out folder.

I delete the files by using the following code snippets:
`for(IFile iFile:files){
iFile.delete(true, new NullProgressMonitor());
}`

During the delete of each file, Property of each resource is being deleted, internally in the API Resource.deleteResource(boolean, MultiStatus), using method call to IPropertyManager.deleteResource(IResource). This API in-turn calls PropertyManager2.deleteProperties(IResource, int depth) with INFINITE depth and hence, all the folder and sub-folders inside the location "\workspace\.metadata\.plugins\org.eclipse.core.resources\.projects\projectName\.indexes\8f" are visited and 'Properties.index' files are loaded in each folder and then, checked whether the given IFile has a property in them. Where '8f' folder represents the Bucket for the '_out' folder.

for all the 6000 files, this operation is being repeated and hence, it takes around 15 minutes to delete all the 6000 files.

Ideally, should we just visit the properties.index file under the '8f' and delete the property for the given file and return?

My assumption is that the Properties for any files present directly under the '_out' folder will be saved under the 'properties.index' file in the Bucket corresponding to '_out' folder. (in my scenario 8f folder).

Apologies, if I have misunderstood it. Kindly, clarify.

[TestDeleteFilesPerformanceIssue.zip](https://github.com/eclipse-platform/eclipse.platform/files/11597685/TestDeleteFilesPerformanceIssue.zip)

PR #693, opened 2023-09-18
PropertyManager2 iterates the tree for infinite depth for deleting/moving an IFile #468

lathapatil:
Changing the depth from infinite to zeroth level for IFile resource deletion

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/468


iloveeclipse:
The change makes sense. 
On Linux I see test times around 35 seconds without the patch and 20 seconds with patch.
I've rebased branch to see if there were any errors, I was unable to see jenkins PR state.

Issue #596, opened 2023-07-26
Random failing ProcessConsoleManagerTests.testBug546710_ConsoleCreationRaceCondition


HeikoKlare:
The test case `org.eclipse.debug.tests.console.ProcessConsoleManagerTests.testBug546710_ConsoleCreationRaceCondition` randomly fails, e.g., in this build: https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-579/4/testReport/

```
java.lang.AssertionError: ProcessConsoleManager and LaunchManager got out of sync. expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.debug.tests.console.ProcessConsoleManagerTests.testBug546710_ConsoleCreationRaceCondition(ProcessConsoleManagerTests.java:119)
```

fedejeanne:
Failed again in [this run](https://github.com/eclipse-platform/eclipse.platform/actions/runs/5669508728/job/15363078470) of https://github.com/eclipse-platform/eclipse.platform/pull/591

Michael5601:
Failed again in [this run](https://github.com/eclipse-platform/eclipse.platform/pull/619/checks?check_run_id=15704911192) of https://github.com/eclipse-platform/eclipse.platform/pull/619

HeikoKlare:
Failed again: https://github.com/eclipse-platform/eclipse.platform/runs/19384480853

iloveeclipse:
I haven't seen this failing recently. Should it be closed?

HeikoKlare:
No occurence has been reported/linked for almost two yeras, so let's close this and reopen in case the test fails again.

fedejeanne:
Failed again in #2595 , this time with a slightly different stack trace:

```
[ProcessConsoleManager and LaunchManager got out of sync.
Launches: [SecondMockProcess]
Consoles: ] 
Actual and expected should have same size but actual size is:
  0
while expected size is:
  1
Actual was:
  []
Expected was:
  [org.eclipse.debug.core.Launch@57d441b0]
java.lang.AssertionError: 
[ProcessConsoleManager and LaunchManager got out of sync.
Launches: [SecondMockProcess]
Consoles: ] 
Actual and expected should have same size but actual size is:
  0
while expected size is:
  1
Actual was:
  []
Expected was:
[...]

PR #695, opened 2023-09-18
Use equinox preferences APIs in RefreshManager class #497

lathapatil:
Replacing deprecated Preferences API usage with modern Equinox Preferences API.

Partially fixes

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/refresh/RefreshManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/refresh/RefreshManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2022 IBM Corporation and others.
+ * Copyright (c) 2004, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/refresh/RefreshManager.java
```diff
@@ -12,9 +12,11 @@
  *     IBM - Initial API and implementation
  *     Christoph Läubrich 	- Issue #84 - RefreshManager access ResourcesPlugin.getWorkspace in the init phase
  *     						- Issue #97 - RefreshManager.manageAutoRefresh calls ResourcesPlugin.getWorkspace before the Workspace is fully open
+ *     Latha Patil(ETAS GmbH)	- GitHub Issue #497- Get rid of deprecated org.eclipse.core.runtime.Preferences in platform code
```
HeikoKlare, line 15: Adding the contribution information is optional. If you want to keep it, please also change the year in the first line from 2022 to 2023. 
Nitpicky: I would prefer to format the copyright line consistent with the existing ones, i.e., only containing "Issue ..." instead of "GitHub Issue ...".

HeikoKlare (commented):
Thanks for the adaptation! The change looks good to me.
@lathapatil Could you address the minor comment concerning the copyright information update? And after that, is the PR ready to merge from your side?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/refresh/RefreshManager.java
lathapatil: Thanks for the review . I will adopt the changes and provide for re-review

HeikoKlare:
Looks good, thanks! One final thing: Could you please squash the two commits into a single one, @lathapatil?

lathapatil:
> Looks good, thanks! One final thing: Could you please squash the two commits into a single one, @lathapatil?

Its done !

HeikoKlare (approved):
Failing test is unrelated and already documented in #596.

PR #698, opened 2023-09-19
SearchIndex: avoid access to files outside destDir

jukzi:


ua/org.eclipse.help.base/src/org/eclipse/help/internal/search/SearchIndex.java
```diff
@@ -874,13 +874,18 @@ private void unzipProductIndex() {
 		try (ZipInputStream zis = new ZipInputStream(zipIn)) {
 			ZipEntry zEntry;
 			while ((zEntry = zis.getNextEntry()) != null) {
+				String name = zEntry.getName();
+				File file = new File(destDir, name);
+				if (!file.toPath().normalize().startsWith(destDir.toPath().normalize())) {
+			        throw new RuntimeException("Bad zip entry: "+name); //$NON-NLS-1$
```
HeikoKlare, line 880: Formatting is inconsistent due to whitespaces instead of tabs in this line.

HeikoKlare (commented):
Changes basically look good to me. Just two questions/remarks:
1. Could the thrown exception be more specific, such as the already provided `ZipException`?
2. It would be nice to have regression for the behavior, but the method does not seem to be designed well w.r.t. testability (providing a different index file would be rather difficult).

HeikoKlare:
Failing tests are unrelated and should probably be fixed with #661.

jukzi:
> 1. Could the thrown exception be more specific, such as the already provided `ZipException`?

ZipException is a checked IOException. would require to catch it. As the Exception should not happen ever i don't care the type.

> 2. It would be nice to have regression for the behavior, but the method does not seem to be designed well w.r.t. testability (providing a different index file would be rather difficult).

i do not plan to put effort into a redesign.

ua/org.eclipse.help.base/src/org/eclipse/help/internal/search/SearchIndex.java
jukzi: fixed

HeikoKlare:
> ZipException is a checked IOException. would require to catch it. As the Exception should not happen ever i don't care the type.

True, just wanted to mention that it might be good to throw something more specific than the generic `RuntimeException`.

HeikoKlare (approved):

laeubi:
> > ZipException is a checked IOException. would require to catch it. As the Exception should not happen ever i don't care the type.
> 
> True, just wanted to mention that it might be good to throw something more specific than the generic `RuntimeException`.

I don't think that is really `true` ... reading a zip file will always require to handle IOExceptions anyways so how can an additional check introduce something new be require to catch it?

jukzi:
feel free to further improve it

Issue #701, opened 2023-09-20
Migrate to org.apache.ant 1.10.14


merks:
Orbit is preparing a new version of ant using newer infrastructure:

https://github.com/eclipse-orbit/orbit-simrel/issues/13

There is one minor incompatibility caused by the introduction of org.apache.tools.ant.BuildLogger.getMessageOutputLevel()
```
    /**
     * @return Returns the {@link #setMessageOutputLevel(int) currently set} message output level.
     * The {@code default} implementation of this method returns {@code MSG_INFO}.
     * @since 1.10.13
     */
    default int getMessageOutputLevel() {
        return Project.MSG_INFO;
    }
```
The protected method in org.eclipse.ant.internal.launching.runtime.logger.NullBuildLogger.getMessageOutputLevel() and org.eclipse.ant.tests.core.support.testloggers.TestBuildLogger.getMessageOutputLevel() "override" this with **protected** methods, causing a compile error.

Migration to new versions in general is difficult because the org.apache.ant version is hard coded in the tests.  The target platform is in a different git repository making it a challenge to update both the hard-coded version and the version in the target platform at the same time.  Better would be to compute the version from the available org.apache.ant bundle.

iloveeclipse:
@merks : what is the state of this issue?

merks:
The status is that I need to get SimRel builds working after the migration to here:

https://github.com/eclipse-simrel

The Orbit changes for this Ant version are basically ready, but I need to set up the ci build for it.  As you can see, I've already tested locally that the Ant tests run, so I think there will be no roadblocks...

ETA is likely some time this weekend...

merks:
The target platform updates are in progress here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1387

PR #702, opened 2023-09-20
Change "override" of ant BuildLogger.getMessageOutputLevel() to public

merks:
Modify the ant tests to avoid hard coded versions.

https://github.com/eclipse-platform/eclipse.platform/issues/701

### DIFF
--- a/ant/org.eclipse.ant.tests.ui/Ant Debug Tests/org/eclipse/ant/tests/ui/debug/PropertyTests.java
+++ b/ant/org.eclipse.ant.tests.ui/Ant Debug Tests/org/eclipse/ant/tests/ui/debug/PropertyTests.java
@@ -19,15 +19,23 @@ import org.eclipse.ant.internal.launching.debug.model.AntThread;
 import org.eclipse.ant.internal.launching.debug.model.AntValue;
 import org.eclipse.ant.launching.IAntLaunchConstants;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.debug.core.DebugException;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.debug.core.model.ILineBreakpoint;
 import org.eclipse.debug.core.model.IVariable;
+import org.osgi.framework.Version;
 
 public class PropertyTests extends AbstractAntDebugTest {
 
-	private static final String ANT_VERSION = "Apache Ant(TM) version 1.10.12"; //$NON-NLS-1$
+	private static final String ANT_VERSION;
+
+	static {
+		Version antVersion = Platform.getBundle("org.apache.ant").getVersion(); //$NON-NLS-1$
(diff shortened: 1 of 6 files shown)


ant/org.eclipse.ant.tests.ui/Ant Debug Tests/org/eclipse/ant/tests/ui/debug/PropertyTests.java
```diff
@@ -19,15 +19,23 @@
...
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.debug.core.model.ILineBreakpoint;
 import org.eclipse.debug.core.model.IVariable;
+import org.osgi.framework.Version;
 
 public class PropertyTests extends AbstractAntDebugTest {
 
-	private static final String ANT_VERSION = "Apache Ant(TM) version 1.10.12"; //$NON-NLS-1$
+	private static final String ANT_VERSION;
```
laeubi, line 32: I wonder if a test should actually assert a specific version at all?

merks: I cannot comment on the purpose of such hard coded things because I did not design these tests.   I can only observe that it makes version updates painful and because updates are needed via two separate commits in two separate repositories, it's kind of impossible to do this update.  As such, I simply avoid hard coding the version **and** I avoid questioning the purpose of the tests that need this because I'm personally only interested in upgrading the ant version and ensure that the new ant bundle is properly structure.  Someone else with time can deal with the philosophical questions about the design and purpose of the tests.  😱

akurtakov: IIRC the tests verify messages where the ant version is part of them so it is either parse and ignore the version or hardcode it. This dynamic detection is step in the right direction but if someone wants to improve the ant tests further I'm sure that there is a lot to do.

laeubi: The question for me is just, if the purpose of the test is/was to check that a specific version is present, then you have rendered the purpose useless now and it could be removed. If that was never the purpose it is pointless to check the version as well... so in summary maybe the cases wehre `ANT_VERSION` is used can simply be replaced with a `contains` / `startsWith` or removed all together... but that's just a guess

> because updates are needed via two separate commits in two separate repositories

Maybe these repositories should then be merged?

merks: At the risk of repeating myself, the purpose of this PR is to be able to update Ant versions easily and to ensure that ant 1.10.14 actual works.  Orthogonal to that, anyone, as Alex suggests, is free to redesign and improve the ant tests; I simply don't have the resource for the that nor, quite frankly, significant interest in such an activity.  So if there are further comments down that orthogonal path or down other orthogonal paths, e.g., repo merges, I'll leave those as an exercise for the interested parties.

Issue #448, opened 2023-05-11
[Refactor] Move methods isWindows()/isLinux()/isMacOSX() outside ResourceTest


fedejeanne:
The reason for this issue is mentioned in the PR #446 (see [comment](https://github.com/eclipse-platform/eclipse.platform/pull/446#discussion_r1190047831) and this change should be made only after that PR is merged.

### Option 1: moving to a utility class

These 3 static methods should either be moved to a utility class:
* `org.eclipse.core.tests.resources.ResourceTest.isWindows()`
* `org.eclipse.core.tests.resources.ResourceTest.isLinux()` (still not merged into *master*)
* `org.eclipse.core.tests.resources.ResourceTest.isMacOSX()`  (still not merged into *master*)

### Option 2: move them to the `Platform` class
This would mean: remove these methods and directly call new methods in `Platform`.

## Pre-requisites
PR #446 should be already merged.

fedejeanne:
Thank you @akurtakov, I'll look into that. 

It might be a major change to move to JUnit5 though, so please don't get your hopes too high, he. But I#ll let you know what I find out ;)

akurtakov:
I have made one such migration - it's pretty straightforward though touches many places (making test methods is not mandatory but I did for the sake of testing whether anything will break). See https://github.com/eclipse-platform/eclipse.platform.ui/commit/d179e6ec58ebef3ca376de46cf17dd50aa24e7b5

vogella:
Simple tests migration (import package/ public removal, new asserts) might be possible to migrate with chatgpt. Copy the test into it and ask to migrate to Junit5. One of my clients reported to me that this worked relatively fine for him.

akurtakov:
There is also https://github.com/junit-pioneer/convert-junit4-to-junit5 if anyone is interested .

HeikoKlare:
Option 2 as proposed in https://github.com/eclipse-platform/eclipse.platform/pull/446#discussion_r1190877282 would also be fine. Maybe it would be better to simply have some `Platform.isOS(...)` method to which you can pass the OS constants defined in `Platform` instead of having methods for every platform to ease adaption for further platforms.
However, since new (relevant) platforms do not come up that often it might also be okay to simply have the three methods for the major platforms within the `Platform` class.

fedejeanne:
Quick update (I finally got time to get back to this):
[...]

PR #704, opened 2023-09-20
Add methods to identify the current OS to the class Platform

fedejeanne:
(fulfill the changes requested in https://github.com/eclipse-platform/eclipse.platform/pull/446#discussion_r1190047831)

Move the methods `isLinux()`, `isWindows()` and `isMacOSX()` (renamed to `isMac()`) to the new internal class `org.eclipse.core.runtime.Platform.OS` and add a new method `OS.is(String)` for legacy/flexibility purposes.

Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/448

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/alias/BasicAliasTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/alias/BasicAliasTest.java
@@ -41,6 +41,7 @@ import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.tests.internal.filesystem.wrapper.WrapperFileSystem;
 import org.eclipse.core.tests.resources.ResourceTest;
 import org.junit.Assume;
@@ -263,7 +264,7 @@ public class BasicAliasTest extends ResourceTest {
 	 */
 	@Test
 	public void testBug198571() {
-		Assume.assumeTrue(isWindows());
+		Assume.assumeTrue(OS.isWindows());
 
 		/* look for the adequate environment */
 		String[] devices = findAvailableDevices();
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/MoveTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/MoveTest.java
@@ -27,6 +27,7 @@ import org.eclipse.core.resources.IWorkspaceRunnable;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.runtime.QualifiedName;
 import org.junit.Assume;
 import org.junit.Test;
@@ -49,7 +50,7 @@ public class MoveTest extends LocalStoreTest {
 	 */
 	@Test
 	public void testMoveFileAcrossVolumes() {
-		Assume.assumeTrue(isWindows());
+		Assume.assumeTrue(OS.isWindows());
 
 		/* look for the adequate environment */
 		String[] devices = findAvailableDevices();
@@ -171,7 +172,7 @@ public class MoveTest extends LocalStoreTest {
 	 */
 	@Test
 	public void testMoveFolderAcrossVolumes() {
-		Assume.assumeTrue(isWindows());
+		Assume.assumeTrue(OS.isWindows());
 
 		/* look for the adequate environment */
 		String[] devices = findAvailableDevices();
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/ModelObjectReaderWriterTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/ModelObjectReaderWriterTest.java
@@ -43,13 +43,14 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Platform.OS;
 import org.eclipse.core.tests.resources.ResourceTest;
 import org.xml.sax.InputSource;
 
 public class ModelObjectReaderWriterTest extends ResourceTest {
-	static final IPath LONG_LOCATION = IPath.fromOSString("/eclipse/dev/i0218/eclipse/pffds/fds//fds///fdsfsdfsd///fdsfdsf/fsdfsdfsd/lugi/dsds/fsd//f/ffdsfdsf/fsdfdsfsd/fds//fdsfdsfdsf/fdsfdsfds/fdsfdsfdsf/fdsfdsfdsds/ns/org.eclipse.help.ui_2.1.0/contexts.xml").setDevice(isWindows() ? "D:" : null);
+	static final IPath LONG_LOCATION = IPath.fromOSString("/eclipse/dev/i0218/eclipse/pffds/fds//fds///fdsfsdfsd///fdsfdsf/fsdfsdfsd/lugi/dsds/fsd//f/ffdsfdsf/fsdfdsfsd/fds//fdsfdsfdsf/fdsfdsfds/fdsfdsfdsf/fdsfdsfdsds/ns/org.eclipse.help.ui_2.1.0/contexts.xml").setDevice(OS.isWindows() ? "D:" : null);
 	static final URI LONG_LOCATION_URI = LONG_LOCATION.toFile().toURI();
-	private static final String PATH_STRING = IPath.fromOSString("/abc/def").setDevice(isWindows() ? "D:" : null).toString();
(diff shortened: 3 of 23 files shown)


akurtakov:
Is there intention for these to be used outside of the tests too?
Also there are some other OS_* constants so maybe having single Platform.isOS(String os_name_constant) would be enough and also be more future proof  (e.g. bsd support gets added or next windows no longer be win32 ).

fedejeanne:
> Is there intention for these to be used outside of the tests too? 

I haven't changed any other existing calls of the type `Platform.getOS().equals(Platform.OS_XYZ)` in other productive/tests classes, I wanted to keep the changes to the minimum while getting rid of the utility methods in `ResourceTest`

> Also there are some other OS_* constants so maybe having single Platform.isOS(String os_name_constant) would be enough and also be more future proof (e.g. bsd support gets added or next windows no longer be win32 ).

Good point, I missed those ones. I took your suggestion and replaced the methods with a single method called `isOS`

laeubi:
I think `isWindows()` and alike are still good especially for the case 

> e.g. bsd support gets added or next windows no longer be win32

so one can then handle the case that `isWindows` means `win32` OR `win2050`

fedejeanne:
> I think `isWindows()` and alike are still good especially for the case
> 
> > e.g. bsd support gets added or next windows no longer be win32
> 
> so one can then handle the case that `isWindows` means `win32` OR `win2050`

This could have unintentional effects if an older version of Windows supports different features than a newer version _e.g._ here: https://github.com/eclipse-platform/eclipse.platform/blob/5a3244bad21e311c36c1b4d87f15342217d6a6c9/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java#L43-L44

I'd rather keep the API "low level" in this case and -if necessary- provide another method that checks for the OS-"family" in a different PR. What do you think?

HeikoKlare:
I agree that there should only be one general `isOS()` method instead of methods for every (currently supported) system.
Some further considerations:
1. I do not agree with the name `isOS` as it sounds as if it checks whether the given argument is an OS. Something like `isExecutedOn()` would be easier to comprehend when reading. Another option would be to create a `PlatformOS` class (or an inner `OS` class of `Platform`) encapsulating what is related to the OS on which the platform is executed instead of further extending the already quite large lists of methods provided by `Platform`.
2. What about representing the OS passed to the new method in a typed way (e.g., as an enum literal)? That would improve the API, as you cannot (accidentally) pass faulty strings to the method with unexpected results. And additions like OS families including multiple OSes would then be possible without changing the method (the method would still accept an OS / OS family) but by representing this as relations between OS / OS family.

laeubi:
> From my point of view, the problem with Plattform.getOs().equals(SOME_CONSTANT) is that some explicit string comparison is used for identifying a platform (which is bad from both a maintability/encapculation and a usability perspective).

At least it is the way if one need maximum flexibility

> Instead, Plattform.isOs(SOME_CONSTANT) is capcable of properly encapsulating what is done with the constant to identify the OS

What should be kept in mind is that these constants are already not any raw values, these are already processed (see for example here https://docs.osgi.org/reference/osnames.html for a list of "real" os and their alias) int o a somewhat normalized form, so usually the variants are handled at another place

> And with respect to usability, having isLinux() instead of isOS(OS_LINUX) or the like is rather a matter of laziness than of proper API design, isn't it?

I'm not sure if "API design" really applies to this kind of code as we don't will have any competing implementations nor is this a general purpose data carrier or alike the main advantage for me is that one can use that as a boolean expression in lamda filter style methods like `Platform::isLinux` in contrast to `()->Platform.isOS(Platform.OS_LINUX)` and it is much more obvious / shorter then.

Also I'd like to mention that where this is used, we only use it in a very rough form, e.g. should a test run on windows or linux only, or should i try to use a DLL instead of a so file, it is not really meant as a fine grained feature switch that would more need something like is it an Redhat or Debian or Wayland versusx X11 or Windows XP 32 bit versus Windows 11 ARM and so on.

Sadly especially for SWT there are sometimes some quirks where I use a similar check in user-code, also there I only need a very very rough categorization and really don't mind about any future OSes or ancient OSes that where supported 10 years ago ...

fedejeanne:
So in sum:
* Move the methods to a new inner class `Platform.OS` to avoid bloating `Platform` 
* Rename `isOS` to `isExecutedOn`
* Provide `isExecutedOn(SOME_STRING_CONSTANT)` as well as `isLinux()`, `isWindows()` and `isMacOSX()` for both flexibility and comfort (in lambda expressions)
* No need to provide "comfort" methods for the deprecated OSs (_i.e._ no `isAIX()`,  `isHPUX` etc)
* `isXYZ()` means "is any type of OS of the family _XYZ_" (_e.g._ `isWindows()` means "is _win32_ or _win2050_)

The last point has no implications for this PR since there are no "related" values (constants) for the OSs at the moment, right? 

I left out the part about using `Enum`-Values instead of `String`s as parameter because that would mean that one needs to have the latest version of the `Enum` in the classpath to be able to use the method with newer OSs (or extending/patching the `Enum` locally). I guess one _could_ provide both approaches (`String`s and `Enum`s) but that would complicate the API.

What do you guys think? (👍  / 👎 )

fedejeanne:
@iloveeclipse which point would you like to change/scratch?

iloveeclipse:
In almost every test we have isWindows() isLinux() and is Mac(), so I don't understand the need for extra complicated constants, long method names etc. So basically what https://github.com/eclipse-platform/eclipse.platform/pull/704#issuecomment-1728740800 says.

fedejeanne:
> In almost every test bundle we have some utility class with isWindows() isLinux() and is Mac(), so I don't understand the need for extra complicated constants, long method names etc. So basically what [#704 (comment)](https://github.com/eclipse-platform/eclipse.platform/pull/704#issuecomment-1728740800) says.

In that case I humbly request some clarification from @laeubi about what he meant in [#704 (comment)](https://github.com/eclipse-platform/eclipse.platform/pull/704#issuecomment-1728740800) because I think I interpreted it completely differently: would you please list your proposals/wishes, @laeubi ?

HeikoKlare:
Would adding the following methods address all concerns?
* `Platform.OS.isLinux()`
* `Platform.OS.isWindows()`
* `Platform.OS.isMacOS()`
* `Platform.OS.is(String OS_CONSTANT)` (for checking for other / more specific OSes than with the other methods, as a replacement for `Platform.getOS().equals(OS CONSTANT)`)

iloveeclipse:
> Would adding the following methods address all concerns?

`OS.isMacOS()` -> `OS.isMac()` please

But the rest would be OK.

fedejeanne:
Done in 0af08ad34b992bd0c734a7be4acffc2278527132

fedejeanne:
Test failures documented in https://github.com/eclipse-platform/eclipse.platform/issues/715

fedejeanne:
The [pipeline error](https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-704/7/display/redirect) is unrelated, maybe a connection problem.

```java
[Pipeline] junit
Recording test results
No test report files were found. Configuration error?
Error when executing always post condition:
Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to JNLP4-connect connection from 10.40.62.179/10.40.62.179:46258
		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1784)
		at hudson.remoting.UserRequest$ExceptionResponse.retrieve(UserRequest.java:356)
		at hudson.remoting.Channel.call(Channel.java:1000)
		at hudson.FilePath.act(FilePath.java:1192)
		at hudson.FilePath.act(FilePath.java:1181)
		at hudson.tasks.junit.JUnitParser.parseResult(JUnitParser.java:118)
		at hudson.tasks.junit.JUnitResultArchiver.parse(JUnitResultArchiver.java:157)
		at hudson.tasks.junit.JUnitResultArchiver.parseAndSummarize(JUnitResultArchiver.java:251)
		at hudson.tasks.junit.pipeline.JUnitResultsStepExecution.run(JUnitResultsStepExecution.java:63)
		at hudson.tasks.junit.pipeline.JUnitResultsStepExecution.run(JUnitResultsStepExecution.java:29)
		at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
		at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
		at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
		at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
		at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
		at java.base/java.lang.Thread.run(Thread.java:839)
hudson.AbortException: No test report files were found. Configuration error?
	at hudson.tasks.junit.JUnitParser$ParseResultCallable.invoke(JUnitParser.java:184)
	at hudson.FilePath$FileCallableWrapper.call(FilePath.java:3578)
	at hudson.remoting.UserRequest.perform(UserRequest.java:211)
	at hudson.remoting.UserRequest.perform(UserRequest.java:54)
	at hudson.remoting.Request$2.run(Request.java:377)
	at hudson.remoting.InterceptingExecutorService.lambda$wrap$0(InterceptingExecutorService.java:78)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at hudson.remoting.Engine$1.lambda$newThread$0(Engine.java:125)
	at java.base/java.lang.Thread.run(Thread.java:829)

```

HeikoKlare:
I case no one has objections to the current proposal, I propose to merge after the freeze period.

HeikoKlare:
@laeubi Since you gave so much valuable input to improve the initial proposals, it would be great to have a short notice whether you see your concerns addressed and agree with the current proposal, so that we can merge.

laeubi (approved):
Looks good.

PR #705, opened 2023-09-20
Add generics to new org.eclipse.core.resource.undo.snapshots

mickaelistria:


mickaelistria:
@robstryker Can you please review this? I try to add stronger typing so clients should deal less with cast and potential ClassCastException

robstryker:
Investigating. Thanks for the PR.

robstryker:
I'm fine with a +1 on this. If you think this is better design I'm inclined to agree that updating the code to use more modern features is to our benefit.

mickaelistria:
Thanks. Would it be OK to merge now?

robstryker:
should be fine

PR #706, opened 2023-09-20
Update URL following merge into platform of other repos

jonahgraham:
Amongst possibly other things I think this is the URL that is embedded in MANFIFEST.MF as the location the bundles are stored in git. The proper value for this comes from the parent pom.

### DIFF
--- a/debug/pom.xml
+++ b/debug/pom.xml
@@ -13,19 +13,14 @@
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
-    <groupId>org.eclipse</groupId>
-    <artifactId>eclipse-platform-parent</artifactId>
+    <groupId>org.eclipse.platform</groupId>
+    <artifactId>eclipse.platform</artifactId>
     <version>4.30.0-SNAPSHOT</version>
-    <relativePath>../../eclipse-platform-parent</relativePath>
   </parent>
 
   <groupId>org.eclipse.platform</groupId>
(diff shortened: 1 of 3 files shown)


jonahgraham:
I'm guessing this build will fail as all bundles probably need version bumped?

debug/pom.xml
```diff
@@ -24,7 +24,7 @@
   <packaging>pom</packaging>
 
   <properties>
-    <tycho.scmUrl>scm:git:https://github.com/eclipse-platform/eclipse.platform.debug.git</tycho.scmUrl>
+    <tycho.scmUrl>scm:git:https://github.com/eclipse-platform/eclipse.platform.git</tycho.scmUrl>
```
akurtakov, line 27: It's defined at https://github.com/eclipse-platform/eclipse.platform/blob/master/pom.xml#L27 so I think best is to delete these properties here.

jonahgraham: Thanks - removed and updated PR title and description and contents to make the same update for other repos that were merged into platform

akurtakov:
And it failed because submodules are not using the parent from this repo but define eclipse-platform-parent directly. This should be fixed too.

jonahgraham:
> And it failed because submodules are not using the parent from this repo but define eclipse-platform-parent directly. This should be fixed too.

Thanks for the pointer - I think I made the appropriate update.

akurtakov:
THanks Jonah! These "small" fixes really add up .

Issue #710, opened 2023-09-20
SWTException: Invalid thread access in Memory view


jonahgraham:
This is a regression in Eclipse 4.29 compared to 4.28 - when memory view is refreshed, common after editing a cell.

This is the stacktrace:

```java
!ENTRY org.eclipse.debug.ui 4 120 2023-09-20 16:20:57.444
!MESSAGE Error logged from Debug UI: 
!STACK 0
org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:565)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:481)
	at org.eclipse.swt.widgets.Table.setRedraw(Table.java:3747)
	at org.eclipse.jface.viewers.StructuredViewer.refresh(StructuredViewer.java:1425)
	at org.eclipse.debug.internal.ui.memory.provisional.AbstractAsyncTableRendering.refresh(AbstractAsyncTableRendering.java:2107)
	at org.eclipse.debug.internal.ui.views.memory.renderings.AsyncTableRenderingUpdatePolicy.handleMemoryBlockChanged(AsyncTableRenderingUpdatePolicy.java:145)
	at org.eclipse.debug.internal.ui.views.memory.renderings.AsyncTableRenderingUpdatePolicy.modelChanged(AsyncTableRenderingUpdatePolicy.java:76)
	at org.eclipse.debug.internal.ui.viewers.provisional.AbstractModelProxy$1.run(AbstractModelProxy.java:92)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.debug.internal.ui.viewers.provisional.AbstractModelProxy.fireModelChanged(AbstractModelProxy.java:96)
	at org.eclipse.debug.internal.ui.viewers.update.MemoryBlockProxy$1.handleChange(MemoryBlockProxy.java:54)
	at org.eclipse.debug.internal.ui.viewers.update.EventHandlerModelProxy.dispatchChange(EventHandlerModelProxy.java:237)
	at org.eclipse.debug.internal.ui.viewers.update.EventHandlerModelProxy.handleDebugEvents(EventHandlerModelProxy.java:137)
[...]

PR #712, opened 2023-09-21
Avoid UI access in async viewers which operate on non UI threads

jonahgraham:
The AsynchronousViewer class hierarchy reuses the Viewer interface but adds additional constraints on it, in particular it allows many of its methods to be called from non-UI threads.

See javadoc summary:

https://github.com/eclipse-platform/eclipse.platform/blob/e92b72761b8543358ba9811ea49807f75e98bad5/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/viewers/AsynchronousViewer.java#L57-L72

Commit 3b9c02c4e1f8364bfb05f66fc5d046ab4d2ff8a2 added UI access to method that used to not have any, therefore this patch restores that non-UI implementation locally consistent with other methods in the async viewer that operate on the model.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/710


iloveeclipse:
Thanks Jonah.

PR #718, opened 2023-09-22
StructureDiffViewer.diff: fix Progress #1144

jukzi:
was IllegalStateException toggling whitespace on/off in diff editor

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1144


vogella:
Not sure if that would help you but we also have a tracing option for this:

To see process monitoring bugs:

1.) Create .options file
2.) Add the following:

org.eclipse.equinox.common/debug=true
org.eclipse.equinox.common/progress_monitor=true

3.) Start Eclipse with the -debug flag
4.) Use IDE and check the Error log for the monitoring issue 

See https://github.com/eclipse-pde/eclipse.pde/issues/723

iloveeclipse (approved):

PR #719, opened 2023-09-22
TextConsoleViewer Scrolling

### LINKED ISSUE #669: Unintended / incomprehensible auto-enable/-disable of scroll lock in TextConsoleViewer
## Current Behavior

In a `TextConsoleViewer`, as used by the `Console` view, scroll lock is automatically enabled or disabled under certain conditions. The basic idea seems to be that scroll lock is automatically disabled when navigating to the end of the view. But when pressing down/page-down, when scrolling, or when moving the scroll bar, scroll lock is even disabled if the navigation does not reach the end of the viewer (see also https://github.com/eclipse-platform/eclipse.platform/pull/649#issuecomment-1702374524). More precisely, it is disabled when after navigation the last shown line is less than `X` lines away from the last line of the viewer, where `X` is the line height in pixels. This is obviously unintended (comparing pixels with lines) and is particularly a result of a faulty implementation of `checkEndDocument`. However, it is not completely clear what the expected behavior would be.

More technically:
https://github.com/eclipse-platform/eclipse.platform/blob/f385650ab42fb7275ab164643fe2d1db5040750e/debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsoleViewer.java#L159-L168
The implementation compares line values (`partialBottomIndex`, `lineCount`) with pixels (the scroll bar increment `delta`), which obviously does not give correct results. The method `checkStartDocument` is equally faulty, but simply results in always returning `false`.

## Expected Behavior
It is not completely clear what the currently intended behavior is. But it looks to me as if scroll lock should be disabled in case scrolling reaches the end of the view or if the end is already reached and some down-navigation key is pressed. And it should enable scroll lock whenever some up-navigation is performed if the topmost line has not already been reached.

### Options
I see different options for the expected behavior under certain conditions.

**When to do nothing:**
1. _(Current)_ When the view contains less lines than can be display (i.e. when no scrollbar exists).
2. _(Probably current intention)_ Pressing some up-navigation key while the first line is visible.
3. When pressing some up-navigation key while the cursor is in the first line.

**When to disable scroll lock:**
1. _(Probably current intention)_ As soon as the view is scrolled to the end.
2. _(Probably current intention)_ When pressing some down-navigation key while the last line is shown.
3. When pressing some down-navigation key while the cursor is in the last line.
4. _(Current)_ When moving to the document end via CTRL+END.
5. As soon as the cursor reaches the end of the contents.

**When to enable scroll lock:**
1. _(Probably current intention)_ Whenever an up-navigation is performed and the first line is not visible.
2. _(Current)_ When moving to the line start via HOME.
2. Whenever an up-navigation is performed while the cursor is not in the first line.
3. Whenever any up/down navigation is performed (key press or dragging the scroll bar) and the navigation did not move the cursor to the end of the contents.
3. Whenever any navigation (including left/right) is performed (key press or dragging the scroll bar) and the navigation did not move the cursor to the end of the contents.

### Proposed Option
I propose the following target behavior:
* Nothing is done when pressing some up-navigation key while the cursor is in the first line
* Scroll lock is disabled when:
    * the view is scrolled to the end
    * the cursor is in the last line and some down-navigation is performed
    * CTRL+END is used
* Scroll lock is enabled when:
    * the view is scrolled and is not at the end
    * any vertical navigation is performed other than the ones disabling scroll lock

In particular, I would propose to always enable scroll lock when navigating vertically and not only when pressing some up-navigation key. When I start to navigate within in a console viewer (even using the down button), I usually want to do something at that very position and don't want to have the console keep scrolling.

### Opinions

What opinions do you have with respect to the expected behavior?

sagarrohat:
Improve logic for auto-enable/-disable of scroll lock in TextConsoleViewer #669

Fixes #669


debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsoleViewer.java
```diff
@@ -306,11 +293,13 @@ public void keyPressed(KeyEvent e) {
...
 				return;
 			}
-			if (e.count < 0) { // Mouse dragged down
-				if (checkEndOfDocument()) {
+			if (e.count < 0) { // Mouse scroll down
+				if (isAtEndOfDocument()) {
 					setScrollLock(false);
+				} else {
+					setScrollLock(true);
 				}
```
fedejeanne, line 301:
```suggestion
setScrollLock(!isAtEndOfDocument());			
```

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsoleViewer.java
```diff
@@ -147,51 +144,41 @@ private void scrollToEndOfDocument() {
...
+			return lineCount == lastVisibleLineIndex + 1;
 		}
 		return false;
 	}
 
 	/*
-	 * Check if user preference is enabled for auto scroll lock and the document is empty or the line count is smaller than each
-	 * vertical scroll
+	 * Checks if user preference is enabled for auto scroll lock. and if the view is
+	 * empty or the content is minimal, implying no need for scrolling.
```
fedejeanne, line 168:
```suggestion
	 * Checks if the auto scrolling is not applicable based on: 
	 <ul>
	 <li>Whether or not the appropriate user preference is enabled</li>
	 <li>Whether or not there is enough content to even require scrolling</li>
	 </ul>
```

iloveeclipse:
@sagarrohat : are you still interesting to merge? If yes, please rebase on master and apply all reviewer comments. If not, please close this PR.

PR #723, opened 2023-09-25
Fixes the method setExtension of TipImage.java in org.eclipse.tips.tests #525

Michael5601:
In this commit the method setExtension of TipImage.java is corrected to properly change the extension. Now if setExtension is called, not only the extension is updated but also the base64image because the extension is a part of this image. Mind that for this change the field fBase64Image can't be final because it needs to be updated when the extension changes. setExtension is not used in the constructor anymore to set the fExtension field, rather it is directly set. Also the tests setExtension and setExtension2 are reactivated and updated to test the functionality of setExtension. Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/525.

### DIFF
--- a/ua/org.eclipse.tips.core/src/org/eclipse/tips/core/TipImage.java
+++ b/ua/org.eclipse.tips.core/src/org/eclipse/tips/core/TipImage.java
@@ -41,7 +41,7 @@ public class TipImage {
 	private final URL fURL;
 	private double fAspectRatio = THREE_TO_TWO;
 
-	private final String fBase64Image;
+	private String fBase64Image;
 
 	/**
 	 * Creates a new TipImage with the specified URL which gets read into a base 64
@@ -80,7 +80,6 @@ public class TipImage {
 	 *
 	 * @param base64Image
 	 *            the non-null base64 encoded image according to RFC-2397.
-	 *
 	 * @throws RuntimeException
 	 *             if the string is not valid
 	 * @see TipImage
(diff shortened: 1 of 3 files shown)


ua/org.eclipse.tips.core/src/org/eclipse/tips/core/TipImage.java
```diff
@@ -98,7 +97,6 @@ public TipImage(String base64Image) {
 			int from = base64Image.indexOf('/') + 1;
 			int to = base64Image.indexOf(';');
 			setExtension(base64Image.substring(from, to).trim());
```
HeikoKlare, line 99: Please replace the `setExtension` call with a direct assignment to `fExtension`. First, this avoids the unnecessary replacement in the base 64 representation of the image. Second, calling non-final, non-private methods in a constructor can be dangerous, as a subclass may override them.

HeikoKlare (commented):
Thanks for implementing the missing functionality. When you address the minor comment on the code, could you please also improve the commit message such that it primarily explains the behavior change in `TipImage` and only mentions the fixed and re-enabled tests afterwards?

ua/org.eclipse.tips.core/src/org/eclipse/tips/core/TipImage.java
Michael5601: Thank your for addressing this, I think this change is useful. I included this advice in the commit and also updated the commit message to focus on the behaviour change. Please merge the PR if you think that it is more clear now.

HeikoKlare (commented):
The change looks good. I just wonder what effect this change should have on the bundle version. The change affects public API, which syntactically stays the same but the `setExtension` method was semantically correctly (in terms of Javadoc spec and implementation).
In a strict interpretation, I would say this requires a major version bump, but actually I don't think that makes sense for such a "correction"?

HeikoKlare (approved):
As already mentioned, these changes look good to me. 
I see the slightly changed behavior of the `setExtension` method as a correction/improvement. The previous behavior was kind of broken since the stored file extension and the extension encoded into the image became different, which did not make sense. This can also be seen in the existing tests (which were commented out), as they specified the behavior as it is implemented after this change.

Issue #724, opened 2023-09-28
Remove recycling of ThreadJob in ImplicitJobs


HeikoKlare:
The `JobManager` uses the `ImplicitJobs` implementation to spawn internal jobs for the execution on a specific `ISchedulingRule` within an executed job:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java#L340-L343

This implementation creates instances of `ThreadJob`. These instances are reused, such that the same `ThreadJob` instance is reused for different rules and potentially different jobs. 
_Side note_: Reusing the same `ThreadJob` within the same job for a different rule already led to race conditions (see #659).
This is achieved by a _recycle_ functionality, which caches a `ThreadJob` for reuse once it is not used anymore:

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java#L225-L228

https://github.com/eclipse-platform/eclipse.platform/blob/ca35cee037256b0f6213f84acbf9de28a9650431/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java#L195-L208

This functionality results in the same job occuring with different rules at different places and points in time. This is both unexpected and error prone (as can be see in issue #659).

In addition, the recycling functionality exists since the early days of Eclipse. At that time, this may have been a useful performance improvement, while I don't expect this to still be relevant nowadays.

Referring to https://github.com/eclipse-platform/eclipse.platform/pull/661#discussion_r1339883084, I propose to remove this recycling functionality to reduce code complexity and avoid the unnecessary risk of errors.

iloveeclipse:
> the recycling functionality exists since the early days of Eclipse. At that time, this may have been a useful performance improvement, while I don't expect this to still be relevant nowadays.

I also don't see immediately what could be a good reason for recycling. 
Creating new ThreadJob shouldn't cost much time looking on the code, reusing however involves central JobManager locks.

Just wondering if there is some use case with high frequency creation of ThreadJob objects that can make GC life harder as it should be.

> I propose to remove this recycling functionality to reduce code complexity and avoid the unnecessary risk of errors.

[...]

PR #727, opened 2023-09-28
Remove recycling of ThreadJob in ImplicitJobs #724

HeikoKlare:
The `JobManager` uses the `ImplicitJobs` implementation to spawn internal jobs for the execution on a specific `ISchedulingRule` within an executed job. This implementation creates instances of `ThreadJob`. These instances are reused, such that the same `ThreadJob` instance is reused for different rules and potentially different jobs. This is achieved by a recycle functionality, which caches a `ThreadJob` for reuse once it is not used anymore. This functionality results in the same job occurring with different rules at different places and points in time. This is both unexpected and error prone.

Since this reuse functionality only seems to be a performance improvement whose value does nowadays not outweigh the added complexity and risk of error, this change removes the recycling functionality.

Note that the recyle functionality also fulfilled some kind of dispose functionality, as it assigned `null` values to the fields of `ThreadJob`. I did not find a reason why this should still be necessary when not reusing the `ThreadJobs` since there should be no further references to the object and the GC should clean them up anyway.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/724.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ImplicitJobs.java
@@ -32,11 +32,6 @@ import org.eclipse.core.runtime.jobs.Job;
  */
 class ImplicitJobs {
 
-	/**
-	 * Cached unused instance that can be reused
-	 * @GuardedBy("this")
-	 */
-	private ThreadJob jobCache = null;
 	protected JobManager manager;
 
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@HeikoKlare : I would prefer to merge early in case there is a catch. If you don't plan any changes here, feel free to merge.

HeikoKlare:
Thanks, @iloveeclipse! I don't plan further changes here. I would just like to rebase on master before merging to have another validation on the latest state. I will do so as soon as builds work again and will merge afterwards.

PR #729, opened 2023-10-05
Cleanup of new o.e.resources.undo API

robstryker:
The new org.eclipse.core.resources.undo API has some usability issues. 

The main problem is that it has two functions that take a "T" as a parameter, but the primary method of creating one of these objects is via the `ResourceSnapshotFactory.fromResource(etc)` call. When calling this function, it is very common to pass in an IResource. We won't know in advance what type the return is, whether it be a IResourceSnapshot<IFile> or a IResourceSnapshot<IContainer> or IResourceSnapshot<IProject>, so we cannot store it in an appropriately typed variable.  

This means, when we try to later call `recordStateFromHistory(IFile)` (or IContainer or IProejct, depending on the type of the snapshot type) we are unable to pass in the appropriately typed object. 

In reality, the parameter is duplicative. The IResourceSnapshot already has access to its internal object and we should not need to be passing it in as a parameter. It only adds to confusion. These parameters should be removed.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/undo/snapshot/AbstractResourceSnapshot.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/undo/snapshot/AbstractResourceSnapshot.java
@@ -83,7 +83,7 @@ abstract class AbstractResourceSnapshot<T extends IResource> implements IResourc
 	@Override
 	public T createResource(IProgressMonitor monitor) throws CoreException {
 		T resource = createResourceHandle();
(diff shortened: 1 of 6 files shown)


mickaelistria (approved):
That is really cool. Would it also require to sync change with JDT as well?
I'd be happy to merge it as soon as consumers are ready too (so we don't break builds)

robstryker:
JDT has a patch ready to go here:   https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/849/files

But it fails now obviously.

mickaelistria:
OK, let's go.
Thanks again!

PR #733, opened 2023-10-09
Ensure target exists calling FileSystemResourceManager.write() #103

HeikoKlare:
The `FileSystemResourceManager.write()` method assumes the passed target to exist. Some operations in the `File` class do not properly ensure that. They check the file existence before starting a workspace operation, which allows some other operation, such as a refresh, to remove the file from the workspace between the check and the start of the operation.

With this change, the existence check in the `File` class is performed after starting a workspace operation, such that no other operation is able to remove the file concurrently and make the write operation file.

Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/103.


HeikoKlare:
Closing as change is unnecessary: preparing operation has properly been starting before checking target existence.

PR #735, opened 2023-10-09
Modernized help UI prototype: Fix set scope still set after deletion

howlger:
Fix the issue that a set scope will still be shown as active after it has been deleted.

### DIFF
--- a/ua/org.eclipse.help.webapp/m/index.js
+++ b/ua/org.eclipse.help.webapp/m/index.js
@@ -350,6 +350,12 @@
                             });
                         } else {
                             var deleteButton = createButton(scopesPage, '<%js:scopes_scope_delete_button_label%>', '<%js:scopes_scope_delete_button_description%>', function() {
+
+                                // if the scope to be deleted is set, unset it first
+                                if (searchScope.level == SEARCH_SCOPE_LEVEL_CUSTOM && searchScope.customIndex == scopeNr) {
+                                    setSearchScope([SEARCH_SCOPE_LEVEL_ALL]);
+                                }
+
                                 doScopesOperation('remove&workingSet=' + encodeURIComponent(scopeName));
                             });
                             deleteButton.className = 'b br';


akurtakov:
@howlger Do you intent to make the new one default or at least have a UI preference to be able to switch to it ?

howlger:
> @howlger Do you intent to make the new one default or at least have a UI preference to be able to switch to it ?

[Active help](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/guide/ua_help_content_active.htm) is not yet supported. Should that be done first or should we already add a checkbox to enable the modernized help UI first?

The test failure on Windows does not seem to me to be related to this change.

akurtakov:
> > @howlger Do you intent to make the new one default or at least have a UI preference to be able to switch to it ?
> 
> [Active help](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/guide/ua_help_content_active.htm) is not yet supported. Should that be done first or should we already add a checkbox to enable the modernized help UI first?

Being able to switch to it even without active help would be beneficial.

> 
> The test failure on Windows does not seem to me to be related to this change.

laeubi:
By the way I don't know how active help is implemented but I have created a while back
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/797

This would allow some more active features (not only in help) to be performed when interacting with the "host code" as then one can implement some rest-like api to interact with the embedding code using JS in the browser.

laeubi:
So in the given "Active Help Example" there is currently a link "Open a cheatsheet" if I click on this I get a feedback that I need to run this "local" (whatever this should mean).

With AJAX intercept, the code could b rewritten:

1. The link (or even some textblocks or buttons or whatever Ui elemnt) will be hidden
2. The the JS code will make an AJAX request to for example `activeAction.json?id =org.eclipse.ui.cheatsheets.OpenCheatSheetFromHelpAction` what will be answered by a static json like `{activeActions:false}` what woll result in keep the non-functional UI hidden
3. in "local" mode the call will be intercepted and instead one will get a json like this `{activeActions:true, payload={ ... }}`
4. This will then show the UI and on click perform a `POST activeAction.json` with given payload to the url
5. This can then again be intercepted and perform any action in Java code one desires.



(javascript:liveAction("org.eclipse.ui.cheatsheets", "org.eclipse.ui.cheatsheets.OpenCheatSheetFromHelpAction","org.eclipse.pde.helloworld")).

howlger:
@laeubi Thanks for the link. Currently, the modernized help UI sits on top of the legacy help. Things like displaying search results, which should be better done on server side, are currently done via JavaScript in the web browser. The active help does not yet work because it is bound to the HTML frameset structure of the legacy help. It's probably easy to fix, but it's not done yet.

See the [modernized help UI e.g. here](https://remainsoftware.com/docs/openapi/help/index.jsp) vs. [deactivated with the parameter `legacy`](https://remainsoftware.com/docs/openapi/help/index.jsp?legacy).

> So in the given "Active Help Example" there is currently a link "Open a cheatsheet" if I click on this I get a feedback that I need to run this "local" (whatever this should mean).

The active help works only in Eclipse or in an Eclipse-based application, accessible via _Help > Help Contents_, not in the stand-alone _Information Center_ (e.g. [help.eclipse.org](https://help.eclipse.org)). The idea is to provide links to open a specific dialog or to execute other commands.

ua/org.eclipse.help.webapp/m/index.js
```diff
@@ -344,6 +344,12 @@
                             });
                         } else {
                             var deleteButton = createButton(scopesPage, '<%js:scopes_scope_delete_button_label%>', '<%js:scopes_scope_delete_button_description%>', function() {
+
+                                // if the scope to be deleted is set, unset it first
+                                if (searchScope.l == 4 && searchScope.t == scopeNr) {
```
iloveeclipse, line 349: Please first make the code more clear (e.g. replace magic "4" I see here and above with proper constants) and add some documentation about `searchScope .l, .t, .s` variables meaning.

**After that** please commit the change that fixes the problem.

iloveeclipse (changes requested):
In order to understand & review the change, another change is needed first, see my comment in code.

ua/org.eclipse.help.webapp/m/index.js
howlger: Thanks for reviewing. You mean in a separate pull request, right?

iloveeclipse: As you like. I would just insert "refactor" commit *before* the fix commit and force push two commits on same branch.

howlger: I did it in a separate pull request: https://github.com/eclipse-platform/eclipse.platform/pull/738

I will update this pull request when it get applied.

iloveeclipse: @howlger :  #738 is merged. Could we continue here?

howlger: Thx for merging https://github.com/eclipse-platform/eclipse.platform/pull/738.
Resolved as requested: speaking variable names and constants instead of magic numbers.

iloveeclipse (approved):

iloveeclipse:
Danke Holger, sorry ist bei mir leider untergegangen.

howlger:
> Danke Holger, sorry ist bei mir leider untergegangen.

Thanks for reviewing and merging.

It would also have been up to me to ask. Hopefully I will soon find some time to work on what @akurtakov [says](https://github.com/eclipse-platform/eclipse.platform/pull/735#issuecomment-1752646172).

PR #737, opened 2023-10-09
Fix test for save state cleanup on snapshot save / bug 297635 #460

HeikoKlare:
The existing test `TestBug297635` relies on reflection to test some internal state change of the `SavedState` class that saves temporary states until some save operation. The test was prone to fail because it relied on internal state changes that depend an specific overall system state (e.g., have an unsaved workspace state, so that no concurrent automatic snapshot save is allowed to occur during test execution). It used reflection to access an internal, highly volatile state.

The bug for which the test case serves as a regression test was due to missing cleanup triggered by `SavedState.forgetSavedTree()`. Instead of checking for internal state changes performed by the cleanup, the rewritten test only checks for a call of the according method. To this end, it temporarily inserts a spy on the `SaveManager`. Since a `Workspace` and `SaveManager` are not easy to set up in an isolated way for testing purposes, the test still relies on reflection, but only to inject a spy on the `SaveManager` rather than to validate internal states.

Since the test is not required to be run as a session test anymore, it is moved to the ordinary regression resource tests.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/460.


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
```diff
@@ -0,0 +1,147 @@
...
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		saveParticipantBundle = SaveParticipant.installContainingBundleAndRegister();
+		saveFull();
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		if (saveParticipantBundle != null) {
```
fedejeanne, line 52: Could this condition ever be `false` ? If `saveParticipantBundle == null` then there was an exception in the `setUp` method and the `tearDown` wouldn't run, or am I missing something?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
```diff
@@ -0,0 +1,147 @@
...
+			return bundle;
+		}
+
+		private static BundleContext getContext() {
+			return Platform.getBundle(PI_RESOURCES_TESTS).getBundleContext();
+		}
+
+		private static void registerAtWorkspace() throws CoreException {
+			getWorkspace().addSaveParticipant(SAVE_PARTICIPANT_BUNDLE, new SaveParticipant());
+		}
```
fedejeanne, line 144: These 3 methods don't need to be in this class, you could move them to the outer class.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/AllSessionTests.java
```diff
@@ -25,7 +26,7 @@
 		TestBug12575.class, WorkspaceDescriptionTest.class, TestBug30015.class,
 		TestMasterTableCleanup.class,
 		ProjectPreferenceSessionTest.class, TestBug113943.class, TestCreateLinkedResourceInHiddenProject.class,
-		Bug_266907.class, TestBug297635.class, TestBug323833.class,
+		Bug_266907.class, Bug_297635.class, TestBug323833.class,
```
fedejeanne, line 29: You mentioned in the description that this test does **not** need to run as a session test anymore. That would mean that you can delete this and leave it only in the other suite.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
```diff
@@ -0,0 +1,147 @@
...
+		try {
+			setField(workspace, saveManagerFieldName, spySaveManager);
+			executeOnSpySaveManager.accept(spySaveManager);
+		} finally {
+			setField(workspace, saveManagerFieldName, originalSaveManager);
+		}
+	}
+
+	private static Object getField(Object object, String fieldName)
+			throws NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException {
```
fedejeanne, line 91: NIT: Unless you foresee some exception that can be handled in the future (in a `catch`-block), I would even go as far as throwing only `Exception` in this and in every other method in the class. This would discourage catching the exception and calling `fail` in the future.

Your call though.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/AllSessionTests.java
HeikoKlare: Fixed in 69978f2.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: Fixed in 69978f2.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: Sorry, I do not understand that comment. Why should `tearDown` not be called in case `setUp` throws an exception? This condition is _exactly_ for the case that `setUp` failed due to some exception.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: Yes, they could technically be moved. But why should they? They semantically belong to the `SaveParticipant`, which is why they are placed in that class.

HeikoKlare: > Instantiating, installing and registering is something that is done _on_ a `SaveParticipant` but not _by_ a `SaveParticipant`.

True, that's why it was no instance method but a class method, comparable to having a factory method for a class.
But I think I got your point and agree that the inner class was not well designed, so I restructured it in f9a5b76. Now it makes more sense to me and I hope you agree :-)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: I was not aware of the unexpected JUnit 3 behavior of not executing `tearDown()` when `setUp()` throws an exception. I consider this an actual flaw as it breaks test isolation. I have documented the behavior in #742.

Concerning this PR, we should keep the check during `tearDown()`, as we hopefully fix the behavior for exceptions during `setUp()` by JUnit migration or a JUnit 3 workaround.

laeubi: > I was not aware of the unexpected **JUnit 3** behavior of not executing tearDown() when setUp() throws an exception. I consider this an actual flaw as it breaks test isolation.

I think JUnit3 is really not something to recommend these days, so worry about incomplete / inadequate features seems not help much and better instead move on to JUnit 5 (or 4 at least).

HeikoKlare: > I think JUnit3 is really not something to recommend these days, so worry about incomplete / inadequate features seems not help much and better instead move on to JUnit 5 (or 4 at least).

Totally agree. Junit 3 should never be used to implement new tests (if possible). Here, it's simply legacy, as several tests (including the one adapted in this PR) are still based on JUnit 3 and it seems to be quite some effort to properly migrate them, We had a discussion about that some months ago (I think related to some addition of `assume` statements), but right now I do cannot find it anymore.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
```diff
@@ -0,0 +1,155 @@
...
+	protected void setUp() throws Exception {
+		super.setUp();
+		BundleWithSaveParticipant.install();
+		saveFull();
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		BundleWithSaveParticipant.uninstall();
+		super.tearDown();
```
fedejeanne, line 50: In JUnit5 (if we ever migrate to it), an Exception in the `setup` method could leave the field `bundle` empty and calling `uninstall` would throw an `IllegalStateException` and would keep the test from cleaning up properly (no call to `super.tearDown()` would happen). 

Is this intentional? If not, I would just invert the order of these 2 lines.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
fedejeanne: I agree, thanks for changing it

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
fedejeanne: To be clear: I liked the code better before _i.e._ even though `tearDown` is not being called with the current implementation (because it's a mix between JUnit3 and JUnit4), the check would make sense in JUnit5. I propose you leave the null-check in place so this test will be "future-ready" :-)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: Thanks for the finding! That was not intentional but a "regression" I introduced with the recent change. I fixed that in 19dbe22.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_297635.java
HeikoKlare: Yes, that was an unintended change :-)
I adapted the behavior in 19dbe22.

fedejeanne:
Looks good :-)

HeikoKlare:
Failing test is unrelated and documented in #744.

In addition, checks are only failing because of an API tooling issue, see https://github.com/eclipse-pde/eclipse.pde/issues/782.
PR only changes one file and version bump for containing plug-in has already been made for 4.30 stream.

PR #752, opened 2023-10-20
Defer registering SaveParticipant until IWorkspace service is available

### LINKED ISSUE #748: IllegalStateException: Workspace is already closed or not ready yet
This is a duplicate of #644 (but I don't have karma to reopen that).
The key issue is that `ResourcesPlugin.getWorkspace()` is called during bundle start of the debug.core plugin (https://github.com/eclipse-platform/eclipse.platform/blob/cfb3c5f0f1aa45861b2bf07d13d7a2efca1d82ea/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java#L730)

I see exceptions like these
```
org.osgi.framework.BundleException: Exception in org.eclipse.debug.core.DebugPlugin.start() of bundle org.eclipse.debug.core.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:839)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:767)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1032)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:371)
	at org.eclipse.osgi.container.Module.doStart(Module.java:605)
	at org.eclipse.osgi.container.Module.start(Module.java:468)
	at org.eclipse.osgi.framework.util.SecureAction.start(SecureAction.java:513)
	at org.eclipse.osgi.internal.hooks.EclipseLazyStarter.postFindLocalClass(EclipseLazyStarter.java:117)
	... 100 more
Caused by: java.lang.IllegalStateException: Workspace is already closed or not ready yet. Consider tracking the org.eclipse.core.resources.IWorkspace service (using your favorite technique, e.g. Declarative Services, ServiceTracker, Blueprint, ...) instead of calling the static method here to prevent such issues!
	at org.eclipse.core.resources.ResourcesPlugin.getWorkspace(ResourcesPlugin.java:502)
	at org.eclipse.debug.core.DebugPlugin.start(DebugPlugin.java:730)
	at org.eclipse.osgi.internal.framework.BundleContextImpl$2.run(BundleContextImpl.java:818)
	at org.eclipse.osgi.internal.framework.BundleContextImpl$2.run(BundleContextImpl.java:1)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:571)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:810)
	... 107 more
```

This happens with `org.eclipse.debug.core` (3.21.100.v20230731-1425). 
Due to the fact that this is a race condition (and depends on the bundle loading order) I could not reproduce it with OOTB Eclipse, but once I install https://sling.apache.org/documentation/development/ide-tooling.html from https://nightlies.apache.org/sling/eclipse/

kwin:
This fixes the issue "IllegalStateException: Workspace is already closed or not ready yet" when the bundle is started early

This closes #748

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
@@ -723,7 +723,6 @@ public class DebugPlugin extends Plugin {
 
 			SourceLookupUtils.shutdown();
 			Preferences.savePreferences(DebugPlugin.getUniqueIdentifier());
-			fWorkspaceServiceTracker.getService().removeSaveParticipant(getUniqueIdentifier());
 			fWorkspaceServiceTracker.close();
 		} finally {
 			super.stop(context);
@@ -740,7 +739,7 @@ public class DebugPlugin extends Plugin {
 
 			@Override
 			public IWorkspace addingService(ServiceReference<IWorkspace> reference) {
-				IWorkspace workspace = super.addingService(reference);
+				IWorkspace workspace = context.getService(reference);
 				try {
 					workspace.addSaveParticipant(getUniqueIdentifier(), new ISaveParticipant() {
 						@Override
@@ -767,6 +766,12 @@ public class DebugPlugin extends Plugin {
 					log(e.getStatus());
 				}
 				return workspace;
+			}
+
+			@Override
+			public void removedService(ServiceReference<IWorkspace> reference, IWorkspace service) {
+				service.removeSaveParticipant(getUniqueIdentifier());
+				super.removedService(reference, service);
 			}
 
 		};
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
@@ -739,7 +739,7 @@ public class DebugPlugin extends Plugin {
 
 			@Override
 			public IWorkspace addingService(ServiceReference<IWorkspace> reference) {
-				IWorkspace workspace = context.getService(reference);
+				IWorkspace workspace = super.addingService(reference);
 				try {
 					workspace.addSaveParticipant(getUniqueIdentifier(), new ISaveParticipant() {
 						@Override
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
@@ -33,7 +33,7 @@ import javax.xml.transform.TransformerException;
 
 import org.eclipse.core.resources.ISaveContext;
 import org.eclipse.core.resources.ISaveParticipant;
-import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IAdapterManager;
@@ -76,6 +76,8 @@ import org.eclipse.debug.internal.core.groups.GroupMemberChangeListener;
 import org.eclipse.debug.internal.core.sourcelookup.SourceLookupUtils;
 import org.eclipse.osgi.service.environment.Constants;
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.osgi.util.tracker.ServiceTracker;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.xml.sax.SAXException;
@@ -437,6 +439,11 @@ public class DebugPlugin extends Plugin {
 	private HashMap<String, IConfigurationElement> fProcessFactories = null;
 
 	/**
+	 * Service tracker for the workspace service
+	 */
+	private ServiceTracker<IWorkspace, IWorkspace> fWorkspaceServiceTracker = null;
+
+	/**
 	 * Mode constants for the event notifier
 	 */
 	private static final int NOTIFY_FILTERS = 0;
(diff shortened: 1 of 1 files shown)


debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
```diff
@@ -727,22 +735,43 @@ public void stop(BundleContext context) throws Exception {
...
 		super.start(context);
 		new DebugOptions(context);
-		ResourcesPlugin.getWorkspace().addSaveParticipant(getUniqueIdentifier(),
-				new ISaveParticipant() {
-					@Override
-					public void saving(ISaveContext saveContext) throws CoreException {
-						if (fExpressionManager != null) {
-							fExpressionManager.storeWatchExpressions();
+
+		fWorkspaceServiceTracker = new ServiceTracker<>(context, IWorkspace.class, null) {
```
laeubi, line 739: Instead of extending ServiceTracker you should use a `ServiceTrackerCustomizer`, that can the return a custom object (e.g. a `Runnable` that performs the cleanup in this case seems useful), then this will even work if there are multiple workspaces (even if unlikley), or the workspace plugin shut down before your plugin (what can happen as shutdown might be performed in reverse order), that way you can simply close the tracker to perform the removal

laeubi (changes requested):

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
laeubi: > A ServiceTracker implements already a ServiceTrackerCustomizer

because it is used internally if you don't pass your own in the constructor but is only meant for tracking simple services where no further processing is required.

> and by that allows to only overwrite the methods which are relevant

It does not hurt to leave one method (e.g. update) empty implemented, that is usually much more readable and flexible than extending the tracker.

> What I would rather do is to also do the cleanup in the ServiceTracker (instead of the stop method)

this is what I would recommend here, as otherwhise there is now a way that the stop method return null...

> Returning a Runnable seems overly complex because the ServiceTracker is not supposed to be used with getService() nor does it need to keep track of the bound SaveParticipant.

I won't assume `Runnable` complex at the end it does only matter that cleanup is done when the service about to disappear (or the tracker is closed), not only when the bundle is stopped.

kwin: I now did the cleanup in the servicetrackercustomizer as well. That way it should be called when the tracker is closed or the service disappears. \

> It does not hurt to leave one method (e.g. update) empty implemented, that is usually much more readable and flexible than extending the tracker.

I don't agree with that statement, overriding ServiceTracker also nicely encapsulates dealing with the bundle context and the anonymous inner class should be pretty much straight-forward here.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
```diff
@@ -727,22 +734,49 @@ public void stop(BundleContext context) throws Exception {
...
+
+			@Override
+			public IWorkspace addingService(ServiceReference<IWorkspace> reference) {
+				IWorkspace workspace = super.addingService(reference);
+				try {
+					workspace.addSaveParticipant(getUniqueIdentifier(), new ISaveParticipant() {
```
laeubi, line 744: workspace might be null at this point...

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
```diff
@@ -727,22 +734,49 @@ public void stop(BundleContext context) throws Exception {
...
+						@Override
+						public void prepareToSave(ISaveContext saveContext) throws CoreException {
+						}
+
+						@Override
+						public void doneSaving(ISaveContext saveContext) {
+						}
+					});
+				} catch (CoreException e) {
+					log(e.getStatus());
```
laeubi, line 766: In this case you still are tracking the service and will later try to remove the (non registered) participant, you should free the service here and return null to indicate the service is not tracked.

laeubi (changes requested):

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
kwin: I added a guard now.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/DebugPlugin.java
kwin: Done.

laeubi (approved):
I approve this here as extending ST or using STC is a style question.

What I wonder is if we probably should just push this down to the `ResourcesPlugin` as an abstract class (e.g. `WorkspaceTracker`) with two abstract methods like

- `WorkspaceTracker#onOpen(IWorkspace workspace)`
- `WorkspaceTracker#onClose(IWorkspace workspace)`

so this can easily be reused at other places.

laeubi:
@kwin please verify this on the next integration build (somewhere available around tomorrow)

kwin:
@laeubi Where is the p2 update site of the integration build for platform? I only found the outdated documentation at https://wiki.eclipse.org/Eclipse_Project_Update_Sites... and 2023-12 m2 (i.e. https://download.eclipse.org/technology/epp/packages/2023-12/202310261200/ and https://download.eclipse.org/releases/2023-12/202310271000/) contains an older version of debug.core (3.21.200.v20230920-0608)

iloveeclipse:
You can use update site from last successful IBuild, like from 

https://download.eclipse.org/eclipse/downloads/drops4/I20231028-1800/

this would be then 

https://download.eclipse.org/eclipse/updates/4.30-I-builds/I20231028-1800/

PR #755, opened 2023-10-20
Migrate file system tests to JUnit 4

HeikoKlare:
The `FileSystemTest` and its subclasses are still based on JUnit 3 due to implementing the `CoreTest` class.
In addition the `CoreTest` class as a base class for many tests based on JUnit 3, provides several file system utilities to its subclasses. These utilities are, however, not inherent to the test class but can be provided as independent utilities. This is especially necessary to break up inheritance hierarchies and foster reuse for JUnit 4/5 migration.
 
This change does the following:
* Moves functionalities concerning symlink creation and temp file creation to the `FileSystemHelper`
* Migrates `FileSystemTest` and its subclasses to JUnit 4
* Removes obsolete try/catch blocks by simply making the tests throw the exceptions or asserting the exceptions
* Adds missing try-with-resources blocks for streams


akurtakov (approved):

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/FileSystemHelper.java
```diff
@@ -81,4 +98,92 @@ public static void clear(java.io.File file) {
...
+	 * @throws IllegalStateException in case symlink creation is not supported
+	 * @throws IOException           in case the creation of the symlink failed for
+	 *                               other reasons
+	 */
+	public static void createSymLink(File basedir, String linkName, String linkTarget, boolean isDir)
+			throws IOException {
+		// The following code creates even a link if
+		// Files.createSymbolicLink(new File(basedir, linkName).toPath(), new
+		// File(basedir, linkTarget).toPath());
+		// would throw java.nio.file.FileSystemException "missing rights"
```
HannesWell, line 119: @HeikoKlare can you tell if this is a common scenario (for tests?).
When looking at the code my fist thought was if Files.createSymbolicLink() can be used.

HeikoKlare: @HannesWell Unfortunately, I do not know whether this it is (still) a relevant/common scenario. I have only moved/refactored this code from the `CoreTest` class without challenging the necessity and way it is implemented.

PR #758, opened 2023-10-23
Add JUnit 5 Suite API to ConfigurationSessionTestSuite

HeikoKlare:
The minimal bundle set of the `ConfigurationSessionTestSuite` does not contain the `junit-platform-suite-api` yet. Since it was added to `org.eclipse.test.performance`, which is also contained in that minimal bundle set, dependencies for this test suite cannot properly be resolved anymore. This change adds the missing bundle to the `ConfigurationSessionTestSuite` bundle set.

See issue in https://github.com/eclipse-equinox/equinox/pull/386.


merks:
@HeikoKlare 

Complements on the detective work as well as the speediness of that investigation!  🏅

HeikoKlare:
Thanks, @merks 🙂 Hope this completely fixes the Equinox build issues. In case all checks succeed, I'd propose to immediately merge this to have those builds running properly again.

Issue #709, opened 2023-09-21
[Java 21] "Setting Security Manager is not supported" errors


iloveeclipse:
After starting I20230921-0530 on Java 21 I see following errors reported in the log:
```
eclipse.buildId=4.30.0.I20230921-0530
java.version=21
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data /data/4x_platform_workspace -os linux -ws gtk -arch x86_64

org.eclipse.ant.core
Error
Thu Sep 21 13:33:52 CEST 2023
Setting Security Manager is not supported
```

We should do something on Java 21 to avoid that code being executed or at least mute the error if we can't avoid that.

laeubi:
Platform is currently Java 17 so why is this a bug? What has changed since last discussion here? Any new decisions?
- https://github.com/eclipse-platform/eclipse.platform/discussions/394

iloveeclipse:
> What has changed since last discussion here? 

Java 21 (next LTS) was released

laeubi:
So Platform will upgrade to Java 21 for the 2023-12 release? That would be great :+1:

iloveeclipse:
> So Platform will upgrade to Java 21 for the 2023-12 release? That would be great 👍

No, for sure not, but people / organisations will start Java 21 evaluation & migration and we should make sure that we don't ship software that report errors "by default".

E.g. in Advantest we monitor error log during test executions and report every unexpected warning/error as test fail. So switching our build to Java 21 would lead to test errors in all places where ant could be involved. Of course we can suppress that specific error in tests, but our customers, once we switch product to run on Java 21, will see the errors and report bugs.

That all shouldn't happen if we know where the problem is, so a generic (not client specific) solution is needed.

laeubi:
> No, for sure not, but people / organisations will start Java 21 evaluation & migration and we should make sure that we don't ship software that report errors "by default".

The default Eclipse Packages use Java 17 as a JVM so no errors are reported, this was discussed already here:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=575210

so for me the question is what "bug" in Platform is here to solve?

This is the corresponding bug/issue at ANT: https://bz.apache.org/bugzilla/show_bug.cgi?id=65381
as far as I know @merks is already working on upgrading ant.

[...]

PR #762, opened 2023-10-25
Add original exception to the error log

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/issues/709


laeubi (approved):
Good catch!

iloveeclipse:
The test fails are known side effect of https://github.com/eclipse-equinox/equinox/pull/382.

PR #764, opened 2023-10-25
Fixed some javadoc issues

jukzi:
logged during build:

* InternalAntRunner.java:765: error: reference not found
* Utils: reference not found
* BasicElementLabels.java:125: error: malformed HTML
* TestViewSupportRegistry.java:125: error: unknown tag: ITestViewSupport
* ILaunchObjectProvider.java:69: error: reference not found
* IPDAEventListener.java error: reference not found
* PDAStackCommand.java:22: error: unexpected end tag: </li>
* PreProcessor.java:48: error: unexpected end tag: </p>
* MetadataTreeView.java:160: error: reference not found
* MultiStrategyDumper.java:36: error: invalid use of @return
* AntPage.java:124: error: @param name not found
* ProjectHelper.java:283: error: @param name not found
* NestedProjectCreator.java:118: error: @param name not found
* BuildCommand.java:25: error: tag not supported in HTML5: tt
* MarkerAttributeMap.java:26: error: malformed HTML
* LocalMetaArea.java:419: error: tag not supported in HTML5: tt
* CheatSheetHyperlinkActionFactory.java:63: tag not allowed here: <li>
* CheatSheetManager.java:111: error: @param name not found
* CompositeCheatSheetParser.java:132: warning: @return has already been
specified

...

i did not manage to fix all issues of the build

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupExtension.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupExtension.java
@@ -57,12 +57,12 @@ public class LaunchConfigurationTabGroupExtension {
 	private Map<Set<String>, String> fPerspectives = null;
 
 	/**
-	 * Constructs a launch configuration tab extension based
-	 * on the given configuration element
+	 * Constructs a launch configuration tab extension based on the given
+	 * configuration element
 	 *
-	 * @param element the configuration element defining the
-	 *  attributes of this launch configuration tab extension
-	 * @return a new launch configuration tab extension
+	 * @param element the configuration element defining the attributes of this
+	 *                launch configuration tab extension
+	 * @returns a new launch configuration tab extension
 	 */
 	public LaunchConfigurationTabGroupExtension(IConfigurationElement element) {
 		setConfigurationElement(element);
@@ -79,11 +79,9 @@ public class LaunchConfigurationTabGroupExtension {
 	}
 
 	/**
-	 * Returns the configuration element that defines the attributes
-	 * for this launch configuration tab group extension.
+	 * {@return the configuration element that defines the attributes
+	 * for this launch configuration tab group extension.}
 	 *
-	 * @param configuration element that defines the attributes
-	 *  for this launch configuration tab extension
 	 */
 	protected IConfigurationElement getConfigurationElement() {
 		return fConfig;
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchGroupExtension.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchGroupExtension.java
@@ -46,7 +46,7 @@ public class LaunchGroupExtension implements ILaunchGroup {
 	 *
 	 * @param element the configuration element defining the
 	 *  attributes of this launch group extension
-	 * @return a new launch group extension
+	 * @returns a new launch group extension
 	 */
 	public LaunchGroupExtension(IConfigurationElement element) {
 		setConfigurationElement(element);
@@ -63,11 +63,8 @@ public class LaunchGroupExtension implements ILaunchGroup {
 	}
 
 	/**
-	 * Returns the configuration element that defines the attributes
-	 * for this launch group extension.
-	 *
-	 * @param configuration element that defines the attributes
-	 *  for this launch group extension
+	 * {@return the configuration element that defines the attributes
+	 * for this launch group extension.}
 	 */
 	protected IConfigurationElement getConfigurationElement() {
 		return fConfig;
--- a/ua/org.eclipse.ui.cheatsheets/src/org/eclipse/ui/internal/cheatsheets/registry/CheatSheetRegistryReader.java
+++ b/ua/org.eclipse.ui.cheatsheets/src/org/eclipse/ui/internal/cheatsheets/registry/CheatSheetRegistryReader.java
@@ -222,12 +222,14 @@ public class CheatSheetRegistryReader extends RegistryReader implements IRegistr
 	}
 
 	/**
-	 *	Create and answer a new CheatSheetCollectionElement, configured as a
+	 *	Creates and returns a new CheatSheetCollectionElement, configured as a
 	 *	child of <code>parent</code>
 	 *
 	 *	@return org.eclipse.ui.internal.model.CheatSheetCollectionElement
 	 *	@param parent org.eclipse.ui.internal.model.CheatSheetCollectionElement
-	 *	@param childName java.lang.String
+	 *	@param pluginId java.lang.String
+	 *	@param id java.lang.String
+	 *	@param label java.lang.String
 	 */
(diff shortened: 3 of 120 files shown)


ua/org.eclipse.ui.cheatsheets/src/org/eclipse/ui/internal/cheatsheets/registry/CheatSheetRegistryReader.java
```diff
@@ -222,12 +222,14 @@ protected CheatSheetElement createCheatSheetElement(IConfigurationElement elemen
...
-	 *	Create and answer a new CheatSheetCollectionElement, configured as a
+	 *	Creates and returns a new CheatSheetCollectionElement, configured as a
 	 *	child of <code>parent</code>
 	 *
 	 *	@return org.eclipse.ui.internal.model.CheatSheetCollectionElement
 	 *	@param parent org.eclipse.ui.internal.model.CheatSheetCollectionElement
-	 *	@param childName java.lang.String
+	 *	@param pluginId java.lang.String
+	 *	@param id java.lang.String
+	 *	@param label java.lang.String
```
HeikoKlare, line 232: nit: I see that the documentation was like this before, but maybe it is better to change this into an empty parameter / return value documentation (so that it's obvious that documentation is missing) rather than putting the parameter type into the parameter's documentation?

HeikoKlare (approved):

ua/org.eclipse.ui.cheatsheets/src/org/eclipse/ui/internal/cheatsheets/registry/CheatSheetRegistryReader.java
jukzi: i just want to get the build clean. there are so many unexplained parameters that i doubt anybody will ever document them properly

jukzi:
I need help: why does the build fail "java.lang.OutOfMemoryError: Java heap space" when i fix these errors? 
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3885

iloveeclipse:
> I need help: why does the build fail "java.lang.OutOfMemoryError: Java heap space" when i fix these errors? https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3885

Arrgh, so it wasn't me, see: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1463#issuecomment-1779676895

So this could be really the problem with **parallel** task execution in the platform build, as pointed here: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1463#issuecomment-1779154090

Jenkins instances are **very** limited (4GB RAM, today???), so we should probably disable parallel builds on Jenkins to avoid further frustration.

@laeubi : I see you've enabled this parallel build with "threads 1C" option in efdaeb6b244e26b734ad759151578dc29e496cf6.

Any objections to disable it?

laeubi:
> Any objections to disable it?

Yes, parallel builds reduce the time needed for verification what is a common complaint. As we have 2 CPUs there can only always run two tests at once, so honestly I think if test accumulate > 2 GB of RAM that's way to much and we need to find out whats causing this. Maybe @fredg02 knows how to get a thread-dump or open a helpdesk to find out whats going on.

Nevertheless to verify that one might create a PR to see if it changes something.

iloveeclipse:
> if test accumulate > 2 GB of RAM that's way to much and we need to find out whats causing this

I believe the parallel tasks were creating archives etc, if I understand the maven right, not actual tests? 

Anyway, 2 GB isn't much for an Eclipse process, because we are not talking about "heap used" memory, but about "allocated / not more used / not yet released" + allocated for native code. 

Especially in Java 17 with G1 defaults they have a bug that doesn't release free memory previously allocated by JVM (https://bugs.openjdk.org/browse/JDK-8238686) which we've observed on our RHEL7 too. 

So take ~500mb for SDK "currently used heap", "500 mb" allocated before but not released yet, 500m for native VM runtime overhead and we are already at 1.5 GB for a single task. There isn't much more left for the second one + jenkins process + whatever else running in this VM.

laeubi:
I can not yet reproduce the OOM locally but reducing the maxmem to 2GB for all processes make the `org.eclipse.ua.tests` fail with SIGKILL (with 3GB all runs fine).

HeikoKlare:
@jukzi Can/should we merge this one? Failing checks seem to be unrelated to the changes.
Erroneous Javadoc is recurrently complained by checks in different PRs, so (1) merging this one can help and (2) having this one merged makes it easier to provide dedicated PRs for whatever required Javadoc improvements may be required in other PRs without conflicting with this PR.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchGroupExtension.java
```diff
@@ -46,7 +46,7 @@ public class LaunchGroupExtension implements ILaunchGroup {
 	 *
 	 * @param element the configuration element defining the
 	 *  attributes of this launch group extension
-	 * @return a new launch group extension
+	 * @returns a new launch group extension
```
akurtakov, line 49: This is a typo. There is no returns tag.

jukzi:
ok, lets merge, reducing warning a bit.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupExtension.java
```diff
@@ -57,12 +57,12 @@ public class LaunchConfigurationTabGroupExtension {
...
-	 * on the given configuration element
+	 * Constructs a launch configuration tab extension based on the given
+	 * configuration element
 	 *
-	 * @param element the configuration element defining the
-	 *  attributes of this launch configuration tab extension
-	 * @return a new launch configuration tab extension
+	 * @param element the configuration element defining the attributes of this
+	 *                launch configuration tab extension
+	 * @returns a new launch configuration tab extension
```
szarnekow, line 65: `@returns` should be `@return`

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupExtension.java
```diff
@@ -79,11 +79,9 @@ private void setConfigurationElement(IConfigurationElement element) {
 	}
 
 	/**
-	 * Returns the configuration element that defines the attributes
-	 * for this launch configuration tab group extension.
+	 * {@return the configuration element that defines the attributes
```
szarnekow, line 82: Bogus search and replace?

akurtakov (commented):
Overall satisfied. Just one new typo and ready to submit. 
In general I would appreciate smaller PRs to make it easier to review.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupExtension.java
szarnekow: Nope, it's fine.

HeikoKlare:
Since this is already merged, I've added the correction of the faulty `@returns` tag to #801, which provides some further Javadoc fixes.

Issue #629, opened 2023-08-08
Random failing InputStreamMonitorTests.testInputStreamMonitor


Michael5601:
The test case `org.eclipse.debug.tests.console.InputStreamMonitorTests.testInputStreamMonitor` randomly fails:
```java
Monitor wrote to few bytes. expected:<50> but was:<100>
java.lang.AssertionError: Monitor wrote to few bytes. expected:<50> but was:<100>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.debug.tests.console.InputStreamMonitorTests.testInputStreamMonitor(InputStreamMonitorTests.java:62)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
```
Locally this test does not fail with my setup. But in PR https://github.com/eclipse-platform/eclipse.platform/pull/618 it failed even though my changes don't change anything related.

HeikoKlare:
Should have been fixed by #835.

PR #766, opened 2023-10-25
Select the current tab after displaying them

fedejeanne:
Call `handleTabSelected` right after displaying all tabs in the `LaunchConfigurationTabGroupViewer::displayInstanceTabs` so that the "activated" method of the tab is called.

This is necessary in order to give every tab the chance to compute/show its content when it is first selected by changing between _Run configurations_ in the _Run Configurations Dialog_.

Contributes to https://github.com/eclipse-pde/eclipse.pde/pull/674

## How to test
1. Run *sdk.product*
2.  Open the _Run configurations Dialog_
3. Create 2 run configurations
4. Select one configuration and click on any Tab other than the first tab
5. Switch to the other configuration
 * The same tab should be active in the newly selected _run configuration_
 * The `activated` method of that tab should have been called (the method extends/overrides `org.eclipse.debug.ui.ILaunchConfigurationTab.activated(ILaunchConfigurationWorkingCopy)`)

### _e.g._ testing with _Product Run Configurations (PDE)_
1. Run *sdk.product*
2.  Open the _Run configurations Dialog_
3. Create 2 **Product Run configurations**
4. Select one configuration and click on the Tab **Java Arguments** 
 4.1. Add a breakpoint in `org.eclipse.jdt.debug.ui.launchConfigurations.JavaArgumentsTab.activated(ILaunchConfigurationWorkingCopy)` 
6. Switch to the other configuration
 * The tab **Java Arguments** should be active in the newly selected _run configuration_
 * The execution should stop at the breakpoint

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
@@ -842,8 +842,9 @@ public class LaunchConfigurationTabGroupViewer {
 			fInitializingTabs = false;
 			return;
 		}
+		Class<? extends ILaunchConfigurationTab> lastActiveTabKind = null;
 		if(redrawTabs) {
-			showInstanceTabsFor(type);
+			lastActiveTabKind = showInstanceTabsFor(type);
 		}
 		// show the name area
 		updateVisibleControls(true);
@@ -875,17 +876,25 @@ public class LaunchConfigurationTabGroupViewer {
 		// Turn off initializing flag to update message
 		fInitializingTabs = false;
 
+		// Try to activate the same (type of) tab that was active before.
+		if (!setActiveTab(lastActiveTabKind)) {
+			// The tab with the wanted class wasn't found. Try to activate the first one
+			setActiveTab(0);
+		}
+
 		if (!fViewform.isVisible()) {
 			fViewform.setVisible(true);
 		}
 	}
 
 	/**
-	 * Populate the tabs in the configuration edit area to be appropriate to the current
-	 * launch configuration type.
+	 * Populate the tabs in the configuration edit area to be appropriate to the
+	 * current launch configuration type.
+	 *
 	 * @param configType the type to show tabs for
+	 * @return The class of the last active tab.
 	 */
-	private void showInstanceTabsFor(ILaunchConfigurationType configType) {
+	private Class<? extends ILaunchConfigurationTab> showInstanceTabsFor(ILaunchConfigurationType configType) {
 		// try to keep on same tab
 		Class<? extends ILaunchConfigurationTab> tabKind = null;
 		if (getActiveTab() != null) {
@@ -897,7 +906,7 @@ public class LaunchConfigurationTabGroupViewer {
 			group = createGroup();
 		} catch (CoreException ce) {
 			DebugUIPlugin.errorDialog(getShell(), LaunchConfigurationsMessages.LaunchConfigurationDialog_Error_19, LaunchConfigurationsMessages.LaunchConfigurationDialog_Exception_occurred_creating_launch_configuration_tabs_27,ce); //
-			return;
+			return null;
 		}
 		disposeExistingTabs();
 		fTabGroup = group;
@@ -926,18 +935,8 @@ public class LaunchConfigurationTabGroupViewer {
 				tab.setControl(control.getParent());
 			}
 		}
-		//set the default tab as the first one
-		if (tabs.length > 0) {
-			setActiveTab(tabs[0]);
-		}
-		// select same tab as before, if possible
-		for (ILaunchConfigurationTab t : tabs) {
-			if (t.getClass().equals(tabKind)) {
-				setActiveTab(t);
-				break;
-			}
-		}
 		fDescription = getDescription(configType);
+		return tabKind;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


fedejeanne:
FYI this PR is a **must** in order to be able to merge https://github.com/eclipse-pde/eclipse.pde/pull/674

fedejeanne:
Test failures already documented here: https://github.com/eclipse-platform/eclipse.platform/issues/596, https://github.com/eclipse-platform/eclipse.platform/issues/629, https://github.com/eclipse-platform/eclipse.platform/issues/768.

The new ones were documented here: https://github.com/eclipse-platform/eclipse.platform/issues/770, https://github.com/eclipse-platform/eclipse.platform/issues/771, https://github.com/eclipse-platform/eclipse.platform/issues/772.

None of the test failures seem to be related to the changes introduced in this PR though since no test failed again in the 2nd CI build.

Reviews are welcome!

HeikoKlare (commented):
I see the necessity and goal of the change, but to me it does not look like a proper solution for the issue. Here are my concerns:
- `handleTabSelected()` explicitly avoids a reactivation when the tab index did not change. Is it correct to remove that precondition? I see that if you want to (mis)use the method to reactivate a tab, this condition must be removed, but is that what the method is supposed to do (considering its other usages)? Otherwise it might make more sense to move that reactivation functionality into a separate method that is called in the scenario to be addressed without the precondition of the `handleTabSelected()` method. That might make sense anyway as `handleTabSelected` suggests that it should be called if a tab was selected, but this is not even the case here. It is rather about `reloadTab` or the like.
- The `displayInstanceTabs()` methods does already process all the tabs in its call of `showInstanceTabs()`. It calls `setActiveTab()` for all tabs once and then calls it again for the one that that is to be active afterwards. However, these `setActiveTab` calls finally seem to do nothing else than calling `handleTabSelected`, which does nothing as `fInitializingTabs` is set to `true`, so the method immediately returns without doing anything. So maybe there is something broken in the overall functionality, as it looks to me as if the last lines of code in `showInstanceTabs` are supposed to do exactly what this PR is trying to fix. Please check what that code does. And if it is really not related to what this PR is supposed to achieve, it may at least point to a better solution of the problem.

fedejeanne:
@HeikoKlare I changed it so `handleTabSelected()` will only be called for this specific use case (switching between _Run Configurations_) and **not** in the other 2 use cases so it's safe now, the changes are contained.
It's still necessary to remove the condition `fCurrentTabIndex == fTabFolder.getSelectionIndex()` in order to work, but I tested all 3 use cases where the method `handleTabSelected()` is called and the changes don't cause any undesired behavior.

## The 3 use cases
1. Switching between _Run configurations_
2. Leaving the _JRE_ tab having changed the JRE to a non-modular JRE/Execution environment (_e.g._ Java 1.8)
3. Selecting another launcher for the _Run configuration_ (this is only possible in certain circumstances)

This PR should be ready to be merged now.

fedejeanne:
The failed check is unrelated to the changes, it seems to be a problem with Tycho and the tests in Mac: 

```java
Error:  Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:4.0.4-SNAPSHOT:test (default-test) on 
project org.eclipse.core.tests.resources: Execution default-test of goal 
org.eclipse.tycho:tycho-surefire-plugin:4.0.4-SNAPSHOT:test failed: Exception creating test eclipse runtime: 
zip file is empty -> [Help 1]
```

HeikoKlare:
Please see my previous [comment](https://github.com/eclipse-platform/eclipse.platform/pull/766#pullrequestreview-1701491846). From my understanding, there is already code that is supposed to do what this change is trying to do. Either correct me in case I am wrong or otherwise please fix or replace that code.

fedejeanne:
I re-read your previous comment and made some changes. I pushed the activation of the tabs all the way back to the end of `displayInstanceTabs(...)` (**after** the flag `fInitializingTabs` is switched-off) and I reverted my changes in `handleTabSelected()`. I even got rid of a small bug that activated the first tab unnecessarily.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
```diff
@@ -842,8 +842,9 @@ protected void displayInstanceTabs(boolean redrawTabs) {
 			fInitializingTabs = false;
 			return;
 		}
+		Class<? extends ILaunchConfigurationTab> lastActiveTabKind = null;
```
HeikoKlare, line 845: This should directly be initialized via `getActiveTab().getClass()` and not be retrieved from the `showInstanceTabsFor()` method. If I see correctly, that method must not handle that tab kind at all after removing the restore functionality from that method.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
```diff
@@ -1606,6 +1605,29 @@ protected void errorDialog(CoreException exception) {
 		ErrorDialog.openError(getShell(), null, null, exception.getStatus());
 	}
 
+	/**
+	 * @param tabKind The class of the tab that has to be activated.
+	 * @return <code>true</code> if a tab was activated, <code>false</code>
+	 *         otherwise.
+	 */
```
HeikoKlare, line 1612: Please improve the documentation. It should be clear that the first of the tabs with a matching kind is activated. In case there is no tab of that kind, no activation is performed, and in case there are multiple the first is taken.

HeikoKlare (commented):
I really like the latest solution 👍 In my opinion, it is very comprehensible now. I have only minor comments to the fix itself. And could we add a regression test for it? 

Please also fix the Javadoc of the `refreshTabs0` method when touching it anway.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
fedejeanne: That won't work: `disposeExistingTabs()` is called in between

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
fedejeanne: See https://github.com/eclipse-platform/eclipse.platform/pull/823

fedejeanne:
Thank you for your reviews, @HeikoKlare ! I merged this one so https://github.com/eclipse-pde/eclipse.pde/pull/674 can be also merged.

> And could we add a regression test for it?

Let's put [a pin](https://github.com/vi-eclipse/Eclipse-Platform/issues/24) on that. The new functionality has been tested manually and it's looking fine.

> Please also fix the Javadoc of the refreshTabs0 method when touching it anway.

I didn't touch that one.

HannesWell:
> And could we add a regression test for it?\n\nLet's put a pin on that. The new functionality has been tested manually and it's looking fine.

I want to mention that THW addition of tests ist permitted all the time, so no reason to postpone that from a RelEng/freeze-time point of view. :)

Issue #678, opened 2023-09-12
Random failing BreakpointTests.testDeferredBreakpointsSepVM


fedejeanne:
Failed in [this build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6158974241/job/16712844043#step:7:27530) of https://github.com/eclipse-platform/eclipse.platform/pull/666

```java
org.eclipse.ant.tests.ui.debug.BreakpointTests.testDeferredBreakpointsSepVM -- Time elapsed: 4.353 s <<< FAILURE!
junit.framework.AssertionFailedError: suspended, but not by breakpoint
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.Assert.assertNotNull(Assert.java:256)
	at junit.framework.TestCase.assertNotNull(TestCase.java:399)
	at org.eclipse.ant.tests.ui.debug.BreakpointTests.deferredBreakpoints(BreakpointTests.java:60)
```

fedejeanne:
No reported failures in 1+ years. Closing.

PR #776, opened 2023-10-27
Add explicit casts to avoid implicit narrowing

fedejeanne:
Fix the problem in 2 places by changing the type of the offending variable ("skipped") to int and adding explicit casts.

This is only to see if the CodeQL finding goes away. The code itself does the same since the variable can only contain a value that fits in an int anyway.

Findings:
* https://github.com/eclipse-platform/eclipse.platform/security/code-scanning/50
* https://github.com/eclipse-platform/eclipse.platform/security/code-scanning/51


runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/LazyInputStream.java
```diff
@@ -160,7 +160,7 @@ public long skip(long toSkip) throws IOException {
 		if (toSkip <= 0)
 			return 0;
 		ensureAvailable(toSkip);
-		long skipped = Math.min(toSkip, bufferSize - offset);
+		int skipped = (int) Math.min(toSkip, bufferSize - offset);
```
laeubi, line 163: This seems the wrong approach here, we should use a long where needed!

laeubi (dismissed):
Please don't cast to int to silence a warning, this seems wrong here and one better can use long altogether.

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/LazyInputStream.java
HeikoKlare: I agree that casting to int is not the proper approach here. Either it is not correct that `bufferSize` and/or `offset` are assumed to be in int range, then their types should be changed. Or it is correct that these are both positive int values, then `Math.min` will always return a value within int range (`toSkip` is positive and `bufferSize - offset` is within int range). Then calling `Math.toIntExact` will be safe for the `Math.min` result and could be used instead of an unsafe cast.

fedejeanne: Since the result of `buffersize - offset` must fit in an `int` (they're both `int`s after all), this cast makes sense. The minimum between a positive long and an int fits in an int.

fedejeanne: Sorry, I just saw your comment @HeikoKlare . 

We agree that the current code can't throw an exception but since you mentioned `Math::toIntExact`, I changed the code and made it future-ready (one can never be too sure).

Thanks for the tip.

fedejeanne:
Test failure already documented here: https://github.com/eclipse-platform/eclipse.platform/issues/678

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/LazyInputStream.java
fedejeanne: @laeubi ok?

HeikoKlare: Just to be sure: have you checked that `offset` being an int is actually correct? Since the code is from 2005, it's worth to check. My proposal was only under the assumption that it is actually correct that we deal with int-range values here. But maybe changing to long is even the proper/better solution here.

fedejeanne: > Just to be sure: have you checked that offset being an int is actually correct?

This is imposed by the official Java API, you can see it for example in the 2nd parameter of the method `read(byte[] b, int off, int len)` or in the fact the the `mark` field and the method `mark(int readLimit)` also use `int`. Changing `offset` to a `long` only brings more casts with it so it's not an option.

HeikoKlare: I do not understand how those method parameters should be related to the length of an input stream. The first defines the offset within a given _target_ array to be written to, the other defines when a mark position is to be invalidated. None of that prevents a stream with length `> Integer.MAX_VALUE + 1` to be passed to a `LazyInputStream` and a call of `skip(Integer.MAX_VALUE + 1)`, which will never terminate. Simply try:
```java
new LazyInputStream(new InputStream() {
	@Override
	public int read() throws IOException {
		return 0;
	}
}, 100_000).skip((long) Integer.MAX_VALUE + 1);
```
So I think at least `bufferSize` actually needs to be a `long`, no matter how unlikely it is that such a long stream is really used. Or am I missing something?

fedejeanne:
Rebasing.

Please let me know if you're **not** OK with merging this PR.

HeikoKlare:
I have not reviewed recent changes yet, but would have at least two findings:
1. There are plain casts to int, which should be replaced as mentioned here: https://github.com/eclipse-platform/eclipse.platform/pull/776#discussion_r1374362813
2. From my understanding, the test case will not terminate in case of a regression. This has to be replaced, such that the test actually fails and does not run endlessly.

fedejeanne:
>     1. There are plain casts to int, which should be replaced as mentioned here: [Change several fields in LazyInputStream and LazyReader to long #776 (comment)](https://github.com/eclipse-platform/eclipse.platform/pull/776#discussion_r1374362813)

Those casts are safe since it's a `long % int` operation, no need for the overhead of `Math.toIntExact`
 
>     2. From my understanding, the test case will not terminate in case of a regression. This has to be replaced, such that the test actually fails and does not run endlessly.

I changed the tests: they now test for a regression and also don't run out of heap in the CI build.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/LazyReaderTest.java
```diff
@@ -137,4 +150,35 @@ public void testMarkAndReset() throws IOException {
...
+			@Override
+			public void close() throws IOException {
+				// nothing to close
+			}
+		};
+
+		try (OpenLazyReader objectUnderTest = new OpenLazyReader(infinitelyEmpty, 10)) {
+			// HACK: needed to avoid filling up the RAM with 0's when calling "skip"
+			objectUnderTest.setBufferSize(Integer.MAX_VALUE);
+			objectUnderTest.setOffset(Integer.MAX_VALUE);
```
HeikoKlare, line 177: Please do not add further "for testing purposes only" methods for the class. If a class cannot reach a state of interest, such as a faulty one for a test, by using its API, there is not need to test that case. And in case the state can be achieved via its API, there is no need to add further methods to make testing more convenient.

In this case the used `Reader` mock can, for example, track the number of `read()` calls to identify the situation of interest.

HeikoKlare (commented):
I see two things addressed by this PR:
1. The LazyInputStream implementation is currently only capable of buffering Integer.MAX_VALUE elements. This limit is slightly relaxed.
2. The LazyInputStream implementation contains a bug that can result in a non-terminating execution when passing a value larger than Integer.MAX_VALUE to the skip() method. This is fixed by an exception thrown when exceeding the implementation-dependent capacity of the lazy stream.

Just to consider: This is obviously not a general solution, as there is still a size limit for the buffer. This limit is dynamic in the sense that it depends on the block size. And it is hidden as an implementation detail. So for a user of the class, the limit is not visible, so they will only know what the actual limit is when it is already too late, i.e., when the exception telling them the value is thrown.

What I find even more important is the performance of these changes: They introduce some overhead due to additional checks. Is that (ir)relevant? Maybe it is irrelevant because of the usage profile (these classes are internal, used for a specific purpose, and, as far as I can see, always initialized with a block size of 2^10 = 1024)?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/LazyReaderTest.java
fedejeanne: Not possible: `read` would have to be called so often to get to the interesting edge case that the heap would overflow. The trick is to either hack the `LazyReader` so it doesn't load 2^32 elements into an array. In order to do that all I could come up with was either:
a. Hacking the `bufferSize` and the `offset` (what I did) or 
b. I could declare `blockCapacity` and `loadBlock()` as `protected` and avoid filling up the array in the `OpenLazyReader`. 

Do you think option "b" would be better?

If not, how would you code it?

fedejeanne:
> What I find even more important is the performance of these changes: They introduce some overhead due to additional checks. Is that (ir)relevant? Maybe it is irrelevant because of the usage profile (these classes are internal, used for a specific purpose, and, as far as I can see, always initialized with a block size of 2^10 = 1024)?

I compared old vs new (for both the reader and the stream) by reading a file of 30Mb and the biggest performance hit was 300ms. Most of the time the hit is < 150ms. A small price to pay for the added safety feature.

HeikoKlare:
> I compared old vs new (for both the reader and the stream) by reading a file of 30Mb and the biggest performance hit was 300ms. Most of the time the hit is < 150ms. A small price to pay for the added safety feature.

I have to admit that I cannot derive any conclusions from these values, since I have no idea how this kind of stream is used. This implementation is internal, so we should be able to investigate the consumers. Is the comparison (one call with 30 Mb reading) made for a standard use case of these consumers?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/content/LazyReaderTest.java
HeikoKlare: So this only because the tests run with 2 GB heap space. There was a recent PR increasing that to 4 GB, so maybe it will work now?
eclipse-platform/eclipse.platform.releng.aggregator#1612

fedejeanne:
Since I don't know the "real" use cases either, I created a snippet that creates a temporary file and reads it several times using the old and the new `LazyInputStream`. 

I tested the following file sizes (I checked the sizes of all XML files in my git folder and saw that they go from <1Mb to ~2Mb):
* 1Mb
* 2Mb
* 30Mb
* 300Mb 

You can see the results here: [Runs.zip](https://github.com/eclipse-platform/eclipse.platform/files/13634328/Runs.zip). The performance is pretty much the same so it's not worth worrying about it.

You can use this snippet and create a new class `LazyInputStreamOld` with the content before this PR in order to use it. 
[Snippet.zip](https://github.com/eclipse-platform/eclipse.platform/files/13634344/Snippet.zip).

fedejeanne:
Since the performance looks as good as before, I'm merging. 

Thank you all for your input!

HeikoKlare:
A final remark: I do still not understand the reasons for providing a "test API" instead of running a proper test against the actual API, as mentioned here: https://github.com/eclipse-platform/eclipse.platform/pull/776#discussion_r1417101155
Manipulating internal state of an object to test something is rarely a good idea and will eventually break.

PR #777, opened 2023-10-27
Configure xmx for maven build to be roughly 3.5GB

laeubi:
Lets try this FYI @iloveeclipse


iloveeclipse:
Does it mean, all the tasks spawned by maven will have this limit too? 
Or are they running in same VM anyway, so we just allow maven use almost all RAM on jenkins?

laeubi:
It all depends a bit, usually maven "tasks" are now spawned but running in the same VM, but some mojos (especially tests) allow for a forked execution.

Beside that, "all ram on Jenkins" is actual "all Ram on the Runner", what itself should hopefully only need a small amount, not sure if runners can use, the idea is more to have a dedicated limit, as it runs on the github runners (7 GB) for me it seems "something" has changed that does not allows the JVM to allocate more memory, sadly OOM errors are often not very informative.

iloveeclipse:
Could it be the execution crashed two times at same test?

- https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-777/1/console
- https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-777/2/console

```
!ENTRY org.eclipse.core.tests.resources 1 0 2023-10-27 10:19:02.664
!MESSAGE [testOOME] setUp
/home/jenkins/agent/workspace/eclipse.platform_PR-777@tmp/durable-4879e9b5/script.sh: line 2:   191 Killed                  mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=/home/jenkins/agent/workspace/eclipse.platform_PR-777/.m2/repository -Pbree-libs -Papi-check -Pjavadoc -Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true -Dproject.build.sourceEncoding=UTF-8 -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS -DtrimStackTrace=false
[Pipeline] }
Terminating xvnc.

Fri Oct 27 10:19:09 2023
 ComparingUpdateTracker: 0 pixels in / 0 pixels out
 ComparingUpdateTracker: (1:-nan ratio)
[Pipeline] // xvnc
Post stage
``` 

See also `org.eclipse.core.tests.internal.properties.PropertyManagerTest.testOOME()`

@jukzi : FYI. may be we should revisit the test above considered parallel build & limited Jenkins VM?

laeubi:
> org.eclipse.core.tests.internal.properties.PropertyManagerTest.testOOME()

So we are forcefully generate an OOM and then complain that build fails with OOM .... 😱

laeubi:
> FYI. may be we should revisit the test above considered parallel build & limited Jenkins VM?

Such test (if useful at all) should **run in a forked vm with very tight memory limits** and not as part of a regular test-run... this effectively will make the potentially the whole JVM unstable as it is completely unpredictable when new memory is allocated...

iloveeclipse:
> So we are forcefully generate an OOM 

but we are trying to handle that, and that worked for ~two years...

> and then complain that build fails with OOM ....

in some **other** maven task, at least not in **this** build with Xmx switched to max of Jenkins.

jukzi:
thanks  for the hint Andrey

> So we are forcefully generate an OOM and then complain that build fails with OOM ....

We are forcefully generate an OOM, assuming that Xmx is configured properly. If we run multiple JVMs in parallel Xmx needs to be max divided by degree of parallelism.
We either decrease parallelism, increase mem or decrease Xmx.
We could also disable the test as workaround, but then it will be only a question of time when the issue occurs in other situations.

laeubi:
> but we are trying to handle that, and that worked for ~two years...

There is nothing handled at all, this is only handles that at *this* place but any other code might randomly crash when it unluckily allocates new memory at the wrong time (e.g. any job running in parallel).

> in some other maven task, at least not in this build with Xmx switched to max of Jenkins.

This unnecessarily wastes limited (shared) resources we could better need somewhere else, so please fix that dest or disable it to run only manually, this has now wasted a lot of human and investigation resources as well for the only sake that we do not trust in the JVM cleaning up SoftReferences :-\

> We could also disable the test as workaround, but then it will be only a question of time when the issue occurs in other situations.

It is not as a test should **never** exhaust all **available** resources, thats why I wrote earlier if test require that much memory they are simply broken and no one is about to expect that.

PR #780, opened 2023-10-27
improve PropertyManagerTest.testOOME() memory consumption

jukzi:
allocate more memory each iteration and try to give it back. To avoid OOME on parallel Builds

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
@@ -252,12 +252,19 @@ public class PropertyManagerTest extends LocalStoreTest {
 
 		List<byte[]> wastedMemory = new LinkedList<>();
 		try {
-			int quickAllocationSize = 200_000_000; // 200MB, should be smaller then -Xmx, but big to get OOME quick
+			// 200MB, should be smaller then -Xmx, but big to get OOME quick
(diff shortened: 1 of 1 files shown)


laeubi:
Can we please either use a forked test with **predictable lower memory bounds** or simply disable the test at all?

1. Wasting memory and forcing an OOM can have unpredictable side-effect
2. It is complete unpredictable behavior if the JVM will ever release the memory to the OS even if GC is performed (what also has no grantee to not being a noop) unless the process is exited so other processes are might be affected as well

jukzi:
> Can we please either use a forked test with **predictable lower memory bounds**

can you work on this?

laeubi:
> > Can we please either use a forked test with **predictable lower memory bounds**
> 
> can you work on this?

No but I also didn't introduced a test that exhausts all memory and relies on undefined behavior, so I suggested to simply ignore this (very simple) so it can be run manually if one is in doubt that JVM is not doing what it is supposed to do.

jukzi:
In the current environment the test logs now:
```
Waste memory to force OOME: 536870912/1073741824
Waste memory to force OOME: 536870912
```
i.e it burns half the memory

laeubi:
> i.e it burns half the memory

It still relies on that an OOM will clear out some References what is likely but not specified at all, also it will keep up the high memory barrier for the rest of the test-runtime.

So if you really like to test the ting I would do two things:

1. Assume that the JVM is doing things right
2. Simulate a low memory state by having a "clearReferences()" method that clears the soft references (what is what the JVM would do) and then check the assumptions.

PR #781, opened 2023-10-27
Migrate Runtime tests to JUnit 4

HeikoKlare:
`RuntimeTest` and its sublasses are still based on JUnit 3, as they implement the JUnit 3 `TestCase` and the Eclipse-specific `CoreTest` subclass, respectively. This change migrates those tests in `org.eclipse.core.tests.runtime` to JUnit 4 that are neither job tests, nor session or performance tests.

In total, the change does the following:
- Migrates several tests in `org.eclipse.core.tests.runtime` to JUnit 4
- Replaces all try-catch constructs with either exceptions thrown by the test or `assertThrows`
- Replaces unnecessary custom assert implementations with JUnit 4 asserts or hamcrest matcher calls
- Encapsulates all `AutoClosables` into try-with-resources blocks

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/EclipsePreferencesTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/EclipsePreferencesTest.java
@@ -13,6 +13,17 @@
  *******************************************************************************/
 package org.eclipse.core.tests.internal.preferences;
 
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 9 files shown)


HannesWell:
I didn't went through the changes in detail, but I highly appreciate the modernization of tests. This makes them much simpler and easier to understand. :)

HeikoKlare:
Thanks for the feedback, @HannesWell!
I hope that the modernization finally also reveals one or two bugs in the tests, so that they become a little more stable. But I would also be fine with only improving comprehensibility  :-)

akurtakov:
Thanks @HeikoKlare !

Issue #783, opened 2023-10-27
SymlinkTest fails on Windows after converting to JUnit 4


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20231027-0140/testresults/html/org.eclipse.core.tests.resources_ep430I-unit-win32-java17_win32.win32.x86_64_17.html

Most likely regression from 826efd662897718e6e9fca4663819f57ba9564b5.

@HeikoKlare : please check.

```
Multiple Failures (2 failures) org.junit.TestCouldNotBeSkippedException: Test could not be skipped due to other failures java.lang.NullPointerException: Cannot invoke "org.eclipse.core.filesystem.IFileStore.delete(int, org.eclipse.core.runtime.IProgressMonitor)" because "this.localFileBaseStore" is null

org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
org.junit.TestCouldNotBeSkippedException: Test could not be skipped due to other failures
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.filesystem.IFileStore.delete(int, org.eclipse.core.runtime.IProgressMonitor)" because "this.localFileBaseStore" is null
Suppressed: org.junit.TestCouldNotBeSkippedException: Test could not be skipped due to other failures
Caused by: org.junit.AssumptionViolatedException: Can't create symbolic links in this platform: win32
at org.junit.Assume.assumeTrue(Assume.java:68)
at org.eclipse.core.tests.filesystem.SymlinkTest.setUp(SymlinkTest.java:106)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
Suppressed: java.lang.NullPointerException: Cannot invoke "org.eclipse.core.filesystem.IFileStore.delete(int, org.eclipse.core.runtime.IProgressMonitor)" because "this.localFileBaseStore" is null
at org.eclipse.core.tests.filesystem.FileSystemTest.tearDown(FileSystemTest.java:101)
at org.eclipse.core.tests.filesystem.SymlinkTest.tearDown(SymlinkTest.java:116)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

HeikoKlare:
Yes, it's a regression from #755. I propose to properly restructure the FileSystemTests to avoid the complex and error-prone setup/teardown inheritance coming from the original JUnit 3 implementation, which is not necessary with JUnit 4/5 anymore. I started PR #785 for that.

PR #785, opened 2023-10-27
Simplify and correct FileSystemTests #783

HeikoKlare:
Several test classes extend the `FileSystemTest`. This class only provides some utility functions and always creates a local and an in-memory file store, even if the actual test does not need it.

This change simplifies the test classes by removing the `FileSystemTest` super class. The necessary utilities are moved to a utility class. The file store creation logic is moved to an according JUnit test rule that can be instantiated for a local or in-memory file store and keeps track of setup and teardown.
This also fixes a bug in interwoven, inherited setup and teardown functions (#783)

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/783.


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreCreationRule.java
```diff
@@ -0,0 +1,72 @@
...
+		}
+		fileStore.mkdir(EFS.NONE, null);
+	}
+
+	@Override
+	protected void after() {
+		try {
+			fileStore.delete(EFS.NONE, null);
+		} catch (CoreException e) {
+		}
```
fedejeanne, line 63: Why not log the exception here? It could help debugging

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreCreationRule.java
```diff
@@ -0,0 +1,72 @@
...
+
+	public IFileStore getFileStore() {
+		return fileStore;
+	}
+
+	@Override
+	protected void before() throws Throwable {
+		switch(fileSystemType) {
+		case LOCAL:
+			var fileStoreLocation = FileSystemHelper.getRandomLocation(FileSystemHelper.getTempDir());
```
fedejeanne, line 47: Would it be possible to attach the name of the requestor (the test method and/or class) to the path so that it makes it easier to find the culprit when some resources are not cleaned up properly?
This could even go into the method `getRandomLocation` itself so that even other tests still not using the `FileStoreCreationRule` can benefit from it.

It doesn't need to happen in this PR, just something to keep in mind.

fedejeanne (changes requested):
Looks good, thank you for improving the tests!

I have only 1 small improvement which I would like to have implemented and that is: logging the exception in `FileStoreCreationRule::after`

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreCreationRule.java
HeikoKlare: Good point! I have added logging in dec3328.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreCreationRule.java
HeikoKlare: That's a good idea. I modified the location calculation to append the test name in dec3328.

But you are right that this makes sense for every test, so we should probably enhance the `FileSystemHelper` to accept the test name (it cannot identify it on its own) and add it to generated locations.

fedejeanne (approved):
LGTM.
Thank you for the improvements!

PR #787, opened 2023-10-28
Migrate o.e.core.tools off the o.e.osgi.service.resolver API

HannesWell:


HannesWell:
I wonder if anybody is actually using `o.e.core.tools`? The PDE Spies seem to be a more modern and powerful alternative.
I also checked https://download.eclipse.org/eclipse/updates/ and it looks like it is not even published, although it is currently build.
I'm asking because for the `StateDumper` I was not able to replace that with standard OSGi API because for testing I would need some resolver-state dumps and I don't know how to create them.
And given that this Plugin seems unused anyway, one way to solve the problem would be to simply delete the entire `org.eclipse.core.tools` plugin

There is also the `org.eclipse.core.tools-feature` and the Plugins `org.eclipse.core.tools.resources` and `org.eclipse.core.resources.spysupport` which also seem to be abandoned (but I'm not sure what they are doing exactly).

@akurtakov and @iloveeclipse what do you think? Or do you know a use-case of them? If not and if you are fine with removing them I can provide a follow-up PR to delete them.

merks:
I can literally find no update site that includes them.  Only babel seems to have translations:

image

It looks like these things died very long ago, but no one noticed.  It looks like it's time for a proper burial...

HannesWell:
> I can literally find no update site that includes them. Only babel seems to have translations:

Thanks Ed for further analysis. The babel team will probably also be happy to get rid of some translation.
Btw. do they notice by themself if a Plugin is gone or do they have to be notified?

I'll create a follow-up to propose the removal of the mentioned projects.

merks:
I think it's been a really long time since babel did a new translation.  😱

PR #788, opened 2023-10-28
Delete org.eclipse.core.tools Plugins and Feature

HannesWell:
As suggested in https://github.com/eclipse-platform/eclipse.platform/pull/787#issuecomment-1783768055 this PR proposes to delete the following Plug-ins:
- org.eclipse.core.tools
- org.eclipse.core.tools.resources
- org.eclipse.core.resources.spysupport

and the Feature:
- org.eclipse.core.tools

All of them were not be found in any p2-repository: https://github.com/eclipse-platform/eclipse.platform/pull/787#issuecomment-1783771505

The functionality of `org.eclipse.core.tools` is already provided by the PDE Bundle spies, 

@akurtakov, @iloveeclipse or anybody else: Do you have objections or know any until now undiscovered usages?


iloveeclipse:
> The functionality of `org.eclipse.core.tools` is already provided by the PDE Bundle spies

Are you really sure they provide save functionality?

I do not understand the reason for removal, why it is necessary to delete them?

merks:
@iloveeclipse 

These appear to be completely unused and are not shipped anywhere:

https://github.com/eclipse-platform/eclipse.platform/pull/787#issuecomment-1783771505

I.e., it appears to be dead code...

HannesWell:
> I do not understand the reason for removal, why it is necessary to delete them?

As mentioned in https://github.com/eclipse-platform/eclipse.platform/pull/787#issuecomment-1783768055 there are parts in this Plugin that require the `org.eclipse.osgi.service.resolver` package, which I don't know how to replace, but on the long run I would like to migrate PDE, P2 and every other Plugin in the Platform off so that it can be finally deprecated and maybe one day even removed along with the `org.eclipse.osgi.compatibility.state` fragment.



> > The functionality of `org.eclipse.core.tools` is already provided by the PDE Bundle spies
> 
> Are you really sure they provide save functionality?

I looked it up again and it looks like the PDE Bundle spies actually don't provide dependency information, but I created to add support for that https://github.com/eclipse-pde/eclipse.pde/issues/835 (although I don't plan to work on that immediately).
But as Ed said, the Projects to be removed are completely unused and therefore just waste resources when being build in CI, the workspace or if one wonders what they are for or maybe even works on them (e.g. to migrate it off some outdated API).

iloveeclipse:
> there are parts in this Plugin that require the `org.eclipse.osgi.service.resolver` package, which I don't know how to replace

Please comment this code, and leave the rest in peace. If I remember right, the internal resources tooling offers insights about workspace/markers model that no other tooling provides.

The tooling is not shipped simply because no one except platform developers needs that.

merks:
Could you clarify how we platform developers use it?  Don’t we need to install it to use it?

iloveeclipse:
Just import the project in the workspace and start debugger

merks:
Please assume that I am stupid and don’t have even the first clue what I should see in said started debugger that would be useful to me as a platform developer.  There might be a hidden gem we should better understand.

HannesWell:
> Please assume that I am stupid and don’t have even the first clue what I should see in said started debugger that would be useful to me as a platform developer. There might be a hidden gem we should better understand.

You can simplify create a new Eclipse-Application launch configuraiton and in the Plug-ins Tab select `Launch with: Features selected below` and select e.g. `o.e.platform` and `o.e.core.tools`.
grafik

If you want to have the whole sdk add o.e.sdk instead or only the parts you want from it (but only platform is amazingly fast to start).

Then you can select the prespectives you are interested in:

grafik

The `Plug-in Dependencies` and `Metadata` shows something. But the Runtime Spy does not work anymore since the `org.eclipse.core.tools.runtime.RuntimeSpyPerspective` class is gone (but I cannot tell when).

HannesWell:
> > there are parts in this Plugin that require the `org.eclipse.osgi.service.resolver` package, which I don't know how to replace
> 
> Please comment this code, and leave the rest in peace. If I remember right, the internal resources tooling offers insights about workspace/markers model that no other tooling provides.
> 
> The tooling is not shipped simply because no one except platform developers needs that.

Can you explain me how to create a State-Dump? I didn't find a way to do that.
If I'm able to create such dump, I can try to modernize that along with the reading of dumps.

iloveeclipse (changes requested):
Now I have IDE access, and I've created screenshot with all the unique tools that are proposed to be removed here (those I personally care about):

image

So this PR can't be merged as is, if the tools above will be removed.

iloveeclipse:
> I honestly have no idea what/how/when/who is using this code if at all.

The tools are needed if you debug workspace / resources related issues.

iloveeclipse:
> Can you explain me how to create a State-Dump? I didn't find a way to do that.

Not sure what a state dump is, but from the Metadata Spy view you can select a file that contains resources related binary content, right click and say "Dump selected file". See screenshot I've attached above.

iloveeclipse:
> Just import the project in the workspace and start debugger

I had no IDE access on weekend - I meant import `org.eclipse.core.tools.resources` project & start debugging. The rest see attached screenshot.

iloveeclipse:
> Can you explain me how to create a State-Dump? I didn't find a way to do that.
> If I'm able to create such dump, I can try to modernize that along with the reading of dumps.

See commit 88445068ae9b96a4dea4626344b93638356f070a for explanation. 
Looks like this one can run in a headless way and produce dumps from the platform state for given installation.

If that isn't needed anymore (and probably no one used that for long time) and it is not easily possible to use some other API to do what the StateDumper is supposed to do, I would simply remove StateDumper and it's contribution to `<extension point="org.eclipse.core.tools.metadataDumpers">` extension point.

With that, this PR can be closed, as this seem to be the only reason to remove core/resource tools, and as shown above there is no replacement for the remaining functionality.

HannesWell:
> If that isn't needed anymore (and probably no one used that for long time) and it is not easily possible to use some other API to do what the StateDumper is supposed to do, I would simply remove StateDumper and it's contribution to `<extension point="org.eclipse.core.tools.metadataDumpers">` extension point.

That's fine for me as well, created https://github.com/eclipse-platform/eclipse.platform/pull/796 for that.

Thanks for the screenshots, looking at it in more detail again I also noticed that this plugin even has a Readme, with non trivial content:
https://github.com/eclipse-platform/eclipse.platform/blob/master/runtime/bundles/org.eclipse.core.tools/Readme.html

PR #790, opened 2023-10-28
Fully rely on generated OSGi Declarative Services component-xml

HannesWell:
and remove the ones committed into git.

I verified in the IDE that the generated content is equal or at least equivalent.

### DIFF
--- a/runtime/tests/org.eclipse.e4.core.tests/.gitignore
+++ b/runtime/tests/org.eclipse.e4.core.tests/.gitignore
@@ -1,0 +1,1 @@
+OSGI-INF/*.xml
--- a/runtime/tests/org.eclipse.e4.core.tests/.settings/org.eclipse.pde.ds.annotations.prefs
+++ b/runtime/tests/org.eclipse.e4.core.tests/.settings/org.eclipse.pde.ds.annotations.prefs
@@ -1,0 +1,7 @@
+dsVersion=V1_3
+eclipse.preferences.version=1
+enabled=true
+generateBundleActivationPolicyLazy=true
+path=OSGI-INF
+validationErrorLevel=error
+validationErrorLevel.missingImplicitUnbindMethod=error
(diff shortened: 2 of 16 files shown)


HannesWell:
The tycho-ds-plugin generates the component-xml files into `target/classes/OSGI-INF`, but in a tycho-surefire runtime the project's root is used as starting point if bundle entries are searched.
Although `<project-root>/target/classes` are added to the classpath/classloader too (besides the 'bundle's root location') the classpath/classloader is not considered when searching for bundle entries.

And it looks like the spec says that the SCR only searches for bundle-entries: https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e30931

Consequently a test-plugin does not has its own component-xml available when being executed by surefire.
As a work-around I added another `maven-resources-plugin:copy-resources` mojo execution, to also copy the component xml to the project root.
First I assume I could set the [`path property of the tycho-ds-plugin`](https://tycho.eclipseprojects.io/doc/4.0.3/tycho-ds-plugin/declarative-services-mojo.html#path) to something like `${project.basedir}/OSGI-INF` but unfortunately this is not considered.
@laeubi is there a reason that speaks against using the project root directory as base of the output path instead of `project.build.output`? I assume that this would fix the problem in this case.

laeubi:
> First I assume I could set the [path property of the tycho-ds-plugin](https://tycho.eclipseprojects.io/doc/4.0.3/tycho-ds-plugin/declarative-services-mojo.html#path) to something like ${project.basedir}/OSGI-INF but unfortunately this is not considered.

This is intentional

> is there a reason that speaks against using the project root directory as base of the output path instead of project.build.output

Yes, a mojo is not expsected and should never write to `project.basedir` any build output but to `project.build.output`.
Beside that this would introduce all kind of problems, e.g. if one mixes manual with automatic items or wants to checkin the result (for whatever reason) what is generated in the IDE what will then conflict with jgit plugin detecting changes if both are not 100% match and report it as a dirty working tree.

> I assume that this would fix the problem in this case and would avoid the need for a workaround.

The problem is how the test mojo handle reactor projects but as this is legacy behavior we can't really change this and that's why Tycho no longer recommends to use that packaging type when one wants to use full Tycho feature set.

PR #791, opened 2023-10-29
Add o.e.e4.core.javax.tests as javax-using copy of o.e.e4.core.tests

HannesWell:
and migrate o.e.e4.core.tests to use jakarta-annotations.

All test-cases besides the About-tests are copied from org.eclipse.e4.core.tests
Consequently o.e.e4.core.javax.tests only differ in the fact that the former uses javax-annotations, while the latter uses jakarta-annotations.

When support for javax-annotations is removed from the E4-Injector, o.e.e4.core.javax.tests can be deleted too.


HannesWell:
What is the opinion from the others about this?

It adds quite some lines of code (but since it is a copy of existing Eclipse Code I assume a CQ is not necessary).
Nevertheless I think it is mandatory to ensure a contentious functionality of the E4 Injector with both jakarta and javax annotations, with and without https://github.com/eclipse-platform/eclipse.platform/pull/792.

For https://github.com/eclipse-platform/eclipse.platform/pull/792 the new copied tests already detected a flaw with the handling of provides under javax (which I have to investigate).

mickaelistria:
Yes, I think this critical change and the strong need for backward compatibility really makes the extra code worth it.
And indeed, no CQ is necessary for code that was already approved.

akurtakov:
Please land it in - we have to be sure that both javax and jakarta works for at least 2 more years (starting from the date we announce javax being deprecated).

HannesWell:
> Please land it in

Done. Thank you and Mickael for your partizipation.

Bananeweizen:
@HannesWell where should the jakarta imports come from? I have multiple platform IDEs which all fail to compile because that's not available anywhere, even after triggering Oomph. 
grafik

HannesWell:
> @HannesWell where should the jakarta imports come from? I have multiple platform IDEs which all fail to compile because that's not available anywhere, even after triggering Oomph.

They are in the platform-TP and should be in the I-builds repo.
So everything should work an these annotationens are already required by platform for a few weeks.

I can check my Installations later this day when I have computer access again.

merks:
Note that there are two versions of this bundle and only the 2.x version exports jakarta namespace.

image

Both are in the resolved target platform:

image

Bananeweizen:
Thanks @merks and @HannesWell, I got it working after some more rounds of restart and explicitly triggering Oomph. Not sure what I missed the first time. Sorry for the noise.

merks:
No problem!  I'm glad you ironed it out...

PR #804, opened 2023-11-02
fix some JavaDoc

jukzi:
mostly "@param name not found", "reference not found"

### DIFF
--- a/ant/org.eclipse.ant.launching/src/org/eclipse/ant/internal/launching/AntLaunching.java
+++ b/ant/org.eclipse.ant.launching/src/org/eclipse/ant/internal/launching/AntLaunching.java
@@ -44,7 +44,7 @@ public class AntLaunching extends Plugin {
 	/**
 	 * Size of left-hand column for right-justified task name. Used for Ant Build logging.
 	 *
-	 * @see org.eclipse.ant.internal.ui.antsupport.logger.AntProcessBuildLogger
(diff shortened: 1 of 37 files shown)


Bananeweizen:
Please don't replace inline `{@code foo}` by XML `<code>foo</code>`. They are not the same. The inline tag supports special characters without escaping and is therefore easier to write. Only for multi-line code the XML version should be used, since it preserves line breaks.

jukzi:
> Please don't replace inline `{@code foo}` by XML `<code>foo</code>`. They are not the same. The inline tag supports special characters without escaping and is therefore easier to write. Only for multi-line code the XML version should be used, since it preserves line breaks.

`{@code true}` does not contain any special character. the build does report it as wrong. how to solve?

Bananeweizen:
> `{@code true}` does not contain any special character. the build does report it as wrong. how to solve?

Not sure what build you refer to, can you share a link? `{@code true}` is surely valid Javadoc. The curly brace version only exists since Java 5, so a misconfiguration of the compliance level within the checker could potentially raise an error here (but I think chances for that are low).

jukzi:
https://ci.eclipse.org/platform/job/eclipse.platform/job/master/415/consoleText

jukzi:
my bad, the error complains about AssertionFailedError in the same lines

Bananeweizen (approved):

akurtakov (approved):

PR #806, opened 2023-11-02
Remove some deprecated SubProgressMonitor

jukzi:
new SubProgressMonitor -> SubMonitor.convert


jukzi:
It should be `subMonitor.split()` instead of `SubMonitor.convert()` at some place

PR #808, opened 2023-11-03
Move error logging from ResourceTest to specific requiring test case

HeikoKlare:
The `ResourceTest` class implements error logging that is enabled for every test case in every sub class. But the generated error log is only used within a single test case, so this general logging is unnecessary.

This change moves the logging down to the test case in `RelaxedSchedRuleBuilderTest`, in which it is actually used. It also improves the assertions defined on the error log.


vogella:
LGTM

PR #812, opened 2023-11-03
Propagate exceptions in TestPerformer and simplify affected test classes

HeikoKlare:
The `TestPerformer` for running tests on a data matrix did not allow to throw any exceptions during its execution, thus requiring all exceptions to be handled internally. This unnecessarily complicated exception handling in tests.

This change allows `TestPerformer` implementation to throw exceptions that may then directly be passed out of the test method. The test classes with TestPerformer implementations are cleaned up by removing unnecessary try-catch blocks or by replacing them with assertThrows statements.

This is part of preparatory work for migrating the `ResourceTests` to JUnit 4.


akurtakov:
Needs manual rebase to fix conflicts.

PR #814, opened 2023-11-03
Set a bunch of project encodings to UTF-8

fedejeanne:
It was done using a "Quick Fix" and the value set (UTF-8) is the same value that any other project in the workspace uses. The objective of this commit is to declutter the "Problems" view.

### DIFF
--- a/team/bundles/org.eclipse.compare.core/.settings/org.eclipse.core.resources.prefs
+++ b/team/bundles/org.eclipse.compare.core/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/team/bundles/org.eclipse.compare/.settings/org.eclipse.core.resources.prefs
+++ b/team/bundles/org.eclipse.compare/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/team/bundles/org.eclipse.core.net/.settings/org.eclipse.core.resources.prefs
+++ b/team/bundles/org.eclipse.core.net/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/team/bundles/org.eclipse.jsch.core/.settings/org.eclipse.core.resources.prefs
+++ b/team/bundles/org.eclipse.jsch.core/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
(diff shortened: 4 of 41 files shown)


Bananeweizen:
unfortunately that requires version bumps for all bundles that didn't change yet since last release. I've into the same problem several times with "many projects cleanups".

fedejeanne:
Thank you for the hint, @Bananeweizen! I've bumped (hopefully) all necessary versions.

Out of curiosity, why is it even necessary to bump these versions? I mean why are changes in the encoding of the project considered "API changes"?

iloveeclipse:
> I mean why are changes in the encoding of the project considered "API changes"?

They aren't. Bumping micro version segment is required so the different project content can be consumed by others. The content **is** different if encoding changes.

fedejeanne:
> > I mean why are changes in the encoding of the project considered "API changes"?
> 
> They aren't. Bumping micro version segment is required so the different project content can be consumed by others. The content **is** different if encoding changes.

Thank you for the clarification!

The build failures seemed unrelated to the changes and to the plug-in versions so I triggered the failing jobs again. Let's see how it goes.

fedejeanne:
The build failure still seems unrelated to me.

_[Verify MacOS](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6769378523/job/18396595286?pr=814#step:7:23706)_ fails with:
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.4-SNAPSHOT:verify (verify) on project org.eclipse.ui.intro.universal: Execute ApiApplication failed: InvocationTargetException: Error reading manifest for /Users/runner/.m2/repository/p2/osgi/bundle/com.sun.jna/5.13.0.v20230812-1000/com.sun.jna-5.13.0.v20230812-1000.jar: zip END header not found -> [Help 1]
```

I'd say it's safe to merge, what do you guys think?

iloveeclipse:
Should be unrelated, yes.

HeikoKlare:
> The build failure still seems unrelated to me.
> 
> _[Verify MacOS](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6769378523/job/18396595286?pr=814#step:7:23706)_ fails with:
> 
> ```
> Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.4-SNAPSHOT:verify (verify) on project org.eclipse.ui.intro.universal: Execute ApiApplication failed: InvocationTargetException: Error reading manifest for /Users/runner/.m2/repository/p2/osgi/bundle/com.sun.jna/5.13.0.v20230812-1000/com.sun.jna-5.13.0.v20230812-1000.jar: zip END header not found -> [Help 1]
> ```
> 
> I'd say it's safe to merge, what do you guys think?

I case you are unsure, you can also just re-trigger that single GH Actions job, i.e., only the MacOS build. But in this case it should be save to merge even without.

fedejeanne:
> Should be unrelated, yes.

👍  Merging.

> I case you are unsure, you can also just re-trigger that single GH Actions job, i.e., only the MacOS build

I did, that was the 2nd run and it failed with the exact same error as [the 1st run](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6769378523/job/18395672882?pr=814). Thank you for the hint anyway :-)

PR #820, opened 2023-11-04
Add switch to disable/enable javax-annotation processing for E4-Injector

HannesWell:
The switch can be set via the VM property `eclipse.e4.inject.javax.disabled`.
For now the default value is 'false', which enables processing of 'javax' annotations by default.

Users can specify the VM argument `-Declipse.e4.inject.javax.disabled=true` in order to test an application without processing of 'javax' annotations now,


akurtakov:
I havent' tested the patch but from reading it - it looks good to me.

PR #824, opened 2023-11-06
Fix several Javadoc errors

HeikoKlare:
* Remove invalid @return statements
* Replace or remove invalid @see annotations
* Fix faulty method references
* Fix missing qualification of type references
* Remove obsolete exception specification in test documentations

According Javadoc generation errors can, e.g., be found in the log of https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-812/4

### DIFF
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchConfigurationTests.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchConfigurationTests.java
@@ -1388,7 +1388,7 @@ public class LaunchConfigurationTests extends AbstractLaunchTest implements ILau
 	 * Tests that a launch is properly registered for notifications before a
 	 * process is spawned and may already propagate a termination event.
 	 *
-	 * see <a href="https://github.com/eclipse-platform/eclipse.platform/issues/598">598</a>
+	 * @see <a href="https://github.com/eclipse-platform/eclipse.platform/issues/598">598</a>
 	 */
 	@Test
 	public void testTerminateLaunchListener_Issue598() throws Exception {
@@ -1453,8 +1453,6 @@ public class LaunchConfigurationTests extends AbstractLaunchTest implements ILau
 	/**
 	 * Tests that attributes in a nested map are persisted in alphabetical
 	 * order.
-	 *
-	 * @throws CoreException
 	 */
 	@Test
 	public void testMapAttributePersistence() throws CoreException, IOException {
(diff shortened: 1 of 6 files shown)


debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchConfigurationTests.java
```diff
@@ -1387,8 +1387,8 @@ public void launchesTerminated(ILaunch[] launches) {
 	/**
 	 * Tests that a launch is properly registered for notifications before a
 	 * process is spawned and may already propagate a termination event.
-	 *
-	 * @see https://github.com/eclipse-platform/eclipse.platform/issues/598
+	 * <p>
+	 * Seee https://github.com/eclipse-platform/eclipse.platform/issues/598
```
akurtakov, line 1391: This is pretty wrong (not even counting the typo).
It should become @see <a href="url"> ISSUE</a>

akurtakov:
There are many p tags opened which (if kept) I would rather see closed too for the sake of having proper html.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchConfigurationTests.java
HeikoKlare: If I am not mistaken `@see X` is only valid if `X` is a reference to some Java element (type, method).
Do you mean this should be `@see <a href="ISSUE_LINK"/>`?

HeikoKlare:
Conflicting changes have been merged in #827. I adapted this PR to contain the remaining improvements as well follow-ups to #827.

Bananeweizen:
> There are many p tags opened which (if kept) I would rather see closed too for the sake of having proper html.

HeikoKlare:
Thanks for the clarifications! I also thought that using `p` tags for separation is fine without closing them and I agree that the closing tag is best omitted in case that's valid for the sake of reducing verbosity. In fact, this PR does currently not contain any p tag changes anymore, so it's rather something to consider in further Javadoc changes/improvements.

> Hyperlinking really isn't that well supported. That's why many other projects nowadays use JUnit 5 tags instead, for instance https://junit-pioneer.org/docs/issue/

Using annotations for tagging issues is great. In case we do not want to have sophisticated functionality based on those annotations, we could also simply add a custom `@Issue` annotation that is independent of any JUnit version that just tags the issue id. In case we agree on such an approach, we should address that in a separate issue on which we unify all the different ways of tagging issues that we have right now (in the test method name, in the test class name, in a `@see` tag, in plain-text Javadoc etc.). For this PR, I would propose to stick to the `@see <a href...>` version.

PR #829, opened 2023-11-06
Fix random failing ServiceSupplierTestCase.testOptionalReferences #488

HeikoKlare:
The `ServiceSupplierTestCase.testOptionalReferences` test case randomly fails. The test aims to wait for services to be enabled or disabled, but it only waits until a single value is changed although it asserts that multiple values are changed afterwards. This is a race condition, as the order in which the values are changed or in which the changes are visible is non-deterministic, also due to missing memory barriers.

This change addresses this issue via two means:
1. it makes the fields of the asserted values volatile to ensure that changed values eventually become visible to the UI thread
2. it waits for all values to be changed before asserting their values to allow different orders of value changes

### DIFF
--- a/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/ServiceSupplierTestCase.java
+++ b/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/ServiceSupplierTestCase.java
@@ -73,11 +73,11 @@ public class ServiceSupplierTestCase {
 		@Inject
 		@Optional
 		@Service(filterExpression = "(component=disabled)")
-		TestService disabledService;
+		volatile TestService disabledService;
 
 		@Inject
 		@Service(filterExpression = "(component=disabled)")
-		List<TestService> services;
+		volatile List<TestService> services;
 	}
 
 	private final List<ServiceRegistration<?>> registrations = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/ServiceSupplierTestCase.java
```diff
@@ -73,11 +73,11 @@ public static class TestDisabledBean {
 		@Inject
 		@Optional
 		@Service(filterExpression = "(component=disabled)")
-		TestService disabledService;
+		volatile TestService disabledService;
```
jukzi, line 76: if this Lists are used concurrently its not enough to use volatile, they need to be Concurrent Lists, or synchronized

HeikoKlare: True. The volatile was an attempt to introduce a memory barrier to ensure that changed values become eventually visible, but, of course, that does not apply to content changes of the list.

HeikoKlare: Since the field is injected, one cannot easily influence the type of list, i.e., ensure that it is thread-safe. But gladly OSGi handles that case:

> By far the easiest method is to use field injection of a list of services. If you make this field volatile then DS will inject a new list whenever the set of tracked services changes.
See https://enroute.osgi.org/FAQ/300-declarative-services.html#tracking-multiple-services

So from my understanding, `volatile` should be correct here.

Issue #825, opened 2023-11-06
Random failing ParallelBuildChainTest.testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule


HeikoKlare:
`ParallelBuildChainTest.testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule` randomly fails on Linux I-Builds, e.g., https://download.eclipse.org/eclipse/downloads/drops4/I20231105-1800/testresults/html/org.eclipse.core.tests.resources_ep430I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html

It is random because execution succeeds in some builds: https://download.eclipse.org/eclipse/downloads/drops4/I20231103-0420/testResults.php

The test fails after around 20 seconds, which is the timeout value in `ParallelBuildChainTest`, so maybe the test is only performing too bad. The test case has been introduced with #661.

Exemplary log:
```
[...]

PR #832, opened 2023-11-06
Increase timeout for ParallelBuildChainTest #825

HeikoKlare:
The `ParallelBuildChainTest.testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule` randomly fails. This is potentially due to slow infrastructure, thus this change increases the test timeout value.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/825.


HeikoKlare:
Build failures are unrelated. The PR only increases timeout values according to #825, so it is save to merge.

Issue #834, opened 2023-11-06
Random failing OutputStreamMonitorTests.testUnbufferedOutputStreamMonitor


HeikoKlare:
The test `OutputStreamMonitorTests.testUnbufferedOutputStreamMonitor` is randomly failing, e.g., in https://github.com/eclipse-platform/eclipse.platform/pull/824/checks?check_run_id=18400681970:

```
Monitor read wrong content. expected:<[o]> but was:<[]>
org.junit.ComparisonFailure: Monitor read wrong content. expected:<[o]> but was:<[]>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.eclipse.debug.tests.console.OutputStreamMonitorTests.testUnbufferedOutputStreamMonitor(OutputStreamMonitorTests.java:141)
```

A potential reason is that two listeners (`binaryListener` and `streamListener`) are involved and the assertion validates a condition on the `streamListener` while a previous `waitWhile` loop checks for a condition on the `binaryListener`. So the program may sporadically proceed after one listener has been called, although the other was not.

PR #835, opened 2023-11-06
Fix OutputStreamMonitorTests and InputStreamMonitorTests #834

HeikoKlare:
The `OutputStreamMonitorTests.testUnbufferedOutputStreamMonitor` randomly fails. A potential reason is that the calls of `waitWhile()` wait for one listener to be updated, but an assertion on another listener is validated afterwards. Thus it may happen that one listener was already notified and the program proceeds while the other has not been notified but is then validated. In addition, the `InputStreamMonitorTests` rely on simply sleep-based timing behavior for asynchronous monitors to receive their data.

This change correct the waiting conditions in `OutputStreamMonitorTest`. In addition, it refactors the test class to not overwrite setup logic and to properly assign functionality to listeners instead of the containing test class. In `InputStreamMonitorTests`, it replaces the sleep operations with proper wait operations on conditions.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/834

### DIFF
--- a/debug/org.eclipse.debug.tests/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.tests;singleton:=true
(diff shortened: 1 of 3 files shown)


HeikoKlare:
Merging since affected tests are still randomly failing in current builds, e.g., https://github.com/eclipse-platform/eclipse.platform/pull/852/checks?check_run_id=19126938332

PR #851, opened 2023-11-10
Disable some ecj warnings especially in examples and tests

jukzi:
To reduce warnings in workspace

potentiallyUnclosedCloseable=ignore
discouragedReference=ignore
parameterAssignment=ignore
autoboxing=ignore

Those are no real problem and ignored by the developers for years


laeubi:
> Those are no real problem

I can't really buy that for

> potentiallyUnclosedCloseable=ignore

also 

> ignored by the developers for years

Now developers will even not getting warned when adding **new** problematic code...

Since we are using the Jenkins plugin now for a while, I think it would be a good time to add a barrier that fails the build if new warnings are introduced and lower the barrier as soon as a warning getting fixed...

jukzi:
> I can't really buy that for
> 
> > potentiallyUnclosedCloseable=ignore

All those are only "potentially" and typically not a problem as the called method does close the resource.  And even if i would have overseen a unclosed resource as long as they are only in an test or example they do not harm anybody.

Adding test barriers sounds cool, but the "deprecation" warnings will automatically introduce new warnings without any code changed.

Bananeweizen:
> the "deprecation" warnings will automatically introduce new warnings without any code changed.

No, that's simple to avoid. If we deprecate a method, we have to change all callers in the same repository to not use the deprecated method anymore. If a method gets deprecated in an upgraded dependency, we refactor the code to not call the deprecated method _together_ with the dependency upgrade.
So if deprecation warnings get more, someone didn't follow these practices and either the PR introducing the deprecation tag or the dependency upgrade should have been rejected because of being incomplete.

jukzi:
> No, that's simple to avoid.

"simple"? unfortunately over the years thousands of deperecated calls remained alife and nobody fixes it. In principle i agree that dperecated code should be avoided on deprecation, but it seems impossible to get into a clean state now anymore

Issue #715, opened 2023-09-22
Failure of Bug468PerformanceTest.test


fedejeanne:
Failed in [this build](https://github.com/eclipse-platform/eclipse.platform/actions/runs/6263790505/job/17008968379?pr=704) of https://github.com/eclipse-platform/eclipse.platform/pull/704

```java
org.eclipse.core.tests.internal.properties.Bug468PerformanceTest.test -- Time elapsed: 122.0 s <<< FAILURE!
junit.framework.AssertionFailedError: The expected max time(ms): 60000, actual time(ms): 70916
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.core.tests.internal.properties.Bug468PerformanceTest.test(Bug468PerformanceTest.java:163)
```

jukzi:
i  got `The expected max time(ms): 60000, actual time(ms): 96100`. please verify there is no performance regression. If none just increase timeout

jukzi:
@lathapatil @iloveeclipse  test was added for https://github.com/eclipse-platform/eclipse.platform/issues/468

HeikoKlare:
There already is a PR for increasing the timeout: https://github.com/eclipse-platform/eclipse.platform/pull/852
The time taken by the test differs heavily between test runs, maybe also because of hardware utilization. Right now we have several builds running to get all pending PRs merged, so maybe several jobs are running on the same hardware affecting the times taken by the test.

And there was also some discussion in the PR for increasing the timeout on how to replace that performance test with a functional test, since no performance test is actually required here (and it obviously takes quite some build time, even if timeout is set properly). But as a quick fix I would like to merge the timeout change to have more stable builds and replace the test afterwards.

lathapatil:
> There already is a PR for increasing the timeout: #852

I have increased timeout to 75 secs only and  for Mac OS alone. It might fail again if it is taking beyond 75 seconds in certain situations as mentioned in [comment](https://github.com/eclipse-platform/eclipse.platform/issues/715#issuecomment-1831371650)

HeikoKlare:
Temporary disablement proposed in #891.

One note for the test reimplementation: The test is currently written on JUnit 3, which is pretty outdated. When rewriting the test, please write it for JUnit 4/5 instead.

HeikoKlare:
[...]

PR #852, opened 2023-11-10
Failure of Bug468PerformanceTest.test #715

lathapatil:
Failure of the testcase in Mac OS only.
Fix provided as such performance time limit is increased to 75 seconds from 60 seconds only for Mac OS. The testcase took around 71 seconds for Mac OS in the build
(https://github.com/eclipse-platform/eclipse.platform/actions/runs/6263790505/job/17008968379?pr=704)

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/715


lathapatil:
@fedejeanne could you please review the changes ?

fedejeanne:
@lathapatil I haven't been involved in the fix of the original issue (#468), maybe ask @jukzi / @iloveeclipse ?

FWIW this PR seems reasonable since the performance problem has been fixed by #468 and the test is rather new, so adapting the timeout makes sense. I'm curious though: 
1. Why is the performance so bad in Mac? Is there another underlying issue?
2. Is the performance constant (in all Platforms) or does it vary on each execution of the test?

## One last remark
Performance tests are always hardware-dependent so I would approach this in a different way: if possible, instead of testing the _performance_ (_i.e._ the **total time**) I would test the _performance **improvement**_ _i.e._ you could change the test to try and test either:
a. Total time of "the old way" > total time of "the new way" or (if not possible)
b. Total visits (from `PropertyManager2`) of "the old way" > total visits of "the new way" 

If the previous approaches are not practicable then I would simply test that "the new way" produces the same result as "the old way" _i.e._ that all files are gone. This one wouldn't be a performance test but a "normal" test to show that nothing broke with the fix.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468PerformanceTest.java
```diff
@@ -158,9 +159,18 @@ public void test() throws CoreException {
...
+		long maxTimeMacOS = TimeUnit.MILLISECONDS.convert(75, TimeUnit.SECONDS);
+
 		assertTrue("The expected min time(ms): " + 0 + ", actual time(ms): " + timeTakenForDeletingFiles[0],
 				0 <= timeTakenForDeletingFiles[0]);
-		assertTrue("The expected max time(ms): " + maxTime + ", actual time(ms): " + timeTakenForDeletingFiles[0],
-				timeTakenForDeletingFiles[0] <= maxTime);
+
+		if (OS.isWindows() || OS.isLinux()) {
+			assertTrue("The expected max time(ms): " + maxTime + ", actual time(ms): " + timeTakenForDeletingFiles[0],
+					timeTakenForDeletingFiles[0] <= maxTime);
```
HeikoKlare, line 169: I know this code existed like that, but I would actually prefer a change to a matcher-style assertions when touching the code anyway, as it is much easier to read and maintain, like: `assertThat("expected time exceeded", timeTakenForDeletingFiles[0], lessThanOrEqualTo(maxTime))` (using hamcrest MatcherAssert.assertThat and Matchers.lessThanOrEqualTo).

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468PerformanceTest.java
```diff
@@ -158,9 +159,18 @@ public void test() throws CoreException {
 		}, this.project, IWorkspace.AVOID_UPDATE, new NullProgressMonitor());
 
 		long maxTime = TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES);
+		long maxTimeMacOS = TimeUnit.MILLISECONDS.convert(75, TimeUnit.SECONDS);
```
HeikoKlare, line 162: I would find it easier and more comprehensible to put these kinds of values into a `Duration`-typed variable. Currently, you have to scan all the code for assignments to `maxTime` and `maxTimeMacOS` to know what unit it uses. You can either fix this by putting the unit into the variable name or by using a type that handles the units internally, like `Duration` does. Then you can simply say `Duration maxTimeMacOs = Duration.ofSeconds(75)` and say `maxTimeMacOs.toMillis()` in case you need them in milliseconds.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468PerformanceTest.java
```diff
@@ -158,9 +159,18 @@ public void test() throws CoreException {
...
+
 		assertTrue("The expected min time(ms): " + 0 + ", actual time(ms): " + timeTakenForDeletingFiles[0],
 				0 <= timeTakenForDeletingFiles[0]);
-		assertTrue("The expected max time(ms): " + maxTime + ", actual time(ms): " + timeTakenForDeletingFiles[0],
-				timeTakenForDeletingFiles[0] <= maxTime);
+
+		if (OS.isWindows() || OS.isLinux()) {
```
HeikoKlare, line 167: I would propose to only add a special branch for MacOS and put everything else into the else branch. In the unlikely case that there is some other OS than Windows, Linux and MacOS, nothing will be tested here.

HeikoKlare (commented):
Thanks for working on a fix for that quite often failing test! 

I only have minor comments that are mainly due to the fix conforming to the style of the existing code, while I would propose to apply some simple improvements to the code when extending it anyway.

fedejeanne:
I've been thinking about it over the weekend and I found another reason **not** to add a performance test in this class: since this test is hardware dependent then it's bound to become obsolete over time:
1. If the hardware becomes too fast then "the old way" would probably also pass this test --> The test becomes obsolete
2. If the hardware is too slow then "the new way" would fail this test even though it's still faster than "the old way" --> The `maxTime`s need to be adapted again.
3. If **2.** happens very often then at some point **1.** could happen: hardware upgrade --> the test becomes obsolete

This is more of a general problem with any kind of performance test though, not just this test.

All in all, I would either remove this test or, if possible, rewrite as a benchmark test ("old" vs "new")

lathapatil:
> All in all, I would either remove this test or, if possible, rewrite as a benchmark test ("old" vs "new")

Thanks @fedejeanne for your valuable input . I do agree with your point that asserting against a constant time might not be proper way of testing the fix.

I would rather try to check if I can re write the case to check "old way of total time" > "New way of total time"

HeikoKlare:
If I see correctly, this is supposed to be a regression test for #468. This bug is about unnecessary tree traversal with `delete` calls on every element. Wouldn't it be sufficient to write a regression test for the behavior rather than for its performance? I.e., replace this performance test by one that validates that `delete` is only called on the the root `IFile` in the addressed scenario, but not for elements deeper in the tree?

fedejeanne:
> Wouldn't it be sufficient to write a regression test for the behavior rather than for its performance? I.e., replace this performance test by one that validates that delete is only called on the the root IFile in the addressed scenario, but not for elements deeper in the tree?

That would be in fact my favorite option :) 
That's what I meant in [my first comment](https://github.com/eclipse-platform/eclipse.platform/pull/852#issuecomment-1805772390) when I wrote:
> b. Total visits (from PropertyManager2) of "the old way" > total visits of "the new way"

HeikoKlare:
I see.

> a. Total time of "the old way" > total time of "the new way" or (if not possible)
> b. Total visits (from `PropertyManager2`) of "the old way" > total visits of "the new way"

Just to be sure: I would actually recommend neither of these options. Option "a" is a performance test, which I do not consider appropriate for the case that shall be tested here. Option "b" is still kind of a performance test and would not validate proper behavior but only validate somehow "better" behavior than some baseline (especially in terms of performance).
As mentioned before, I would recommend to convert this into a pure functionality that simply validates whether the new code behaves as expected (depth one traveral in specific situations).

HeikoKlare:
I propose to merge this PR since the test is failing on MacOS very often. Still I would keep the issue open to have this performance test replaced with a functionality test as proposed in the comments. I will document / link to the according ideas posted in this PR in the issue when merging this PR.

lathapatil:
> performance test replaced with a functionality test as proposed in the comments.

Thanks for the input and I will be checking on replacing this with functionality test

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468PerformanceTest.java
jukzi: 75s is not enough. A single test that wastes 90s is not good. please rewrite the test to take <1s. meanwhile it could be disabled.

HeikoKlare: Agreed. Temporary disablement is better than temporary relaxation of the timeout value.
@jukzi Do you want to provide a PR to temporarily disable the test or shall I?

jukzi: please disable - or even remove the test.

HeikoKlare: -> #891
Has to be replaced soon, so I think it's fine to keep it disabled and replace it rather than removing and re-adding it.

lathapatil:
> I propose to merge this PR since the test is failing on MacOS very often. Still I would keep the issue open to have this performance test replaced with a functionality test as proposed in the comments. I will document / link to the according ideas posted in this PR in the issue when merging this PR.

@HeikoKlare The performance test is replaced with functional test. let me know if any further code changes are required .

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468FunctionalTest.java
```diff
@@ -0,0 +1,125 @@
...
+			createFile(tempFolder, TO_BE_DELETED_FILE_NAME_PREFIX + 0);
+
+		}, this.project, IWorkspace.AVOID_UPDATE, new NullProgressMonitor());
+	}
+
+	private void createTestProject() throws CoreException {
+		Bug468FunctionalTest.this.project = ResourcesPlugin.getWorkspace().getRoot()
+				.getProject("TestProject");
+		Bug468FunctionalTest.this.project.create(new NullProgressMonitor());
+		Bug468FunctionalTest.this.project.open(new NullProgressMonitor());
```
HeikoKlare, line 72: Please remove the unnecessary `Bug468FunctionalTest.this.`.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468FunctionalTest.java
```diff
@@ -0,0 +1,125 @@
...
+	 * @throws Exception
+	 *             if any exception happens during the deleting of the project.
+	 */
+	@After
+	public void tearDown() throws Exception {
+		this.project.delete(true, new NullProgressMonitor());
+	}
+
+	Workspace ws;
+	PropertyManager2 manager;
```
HeikoKlare, line 94: Please remove these fields and just declare the variables inside the `test` method.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/Bug468FunctionalTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+//Test case for GitHub Issue 468
+@RunWith(JUnit4.class)
```
HeikoKlare, line 43: This annotation can be removed.

HeikoKlare (commented):
Thank you for updating the test @lathapatil!
From a functional perspective, the test looks good. It serves as a regression test for #468 as intended. 

I have some comments with respect to code style. Please also auto-format the code as it does currently not conform to the formatter settings. In addition, I would propose that you add the test case to the `PropertyManagerTest` class. Then the test case is placed next to other test cases concerning the same functionality. You can then also reuse existing functionality of that test class like a project setup for every test case and the `WorkspaceTestRule` that keeps track of properly cleaning up and makes custom teardown functionality obsolete.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
```diff
@@ -550,4 +555,42 @@ public void testSimpleUpdate() throws CoreException {
...
 	}
 
+	/**
+	 * Whenever a delete operation is called on an IFile it's properties also
+	 * deleted from .index file .This Test case validates for given IFile resource
+	 * Zero depth is calculated to traverse through folders for loading right index
+	 * file and delete its properties . because the required index file is present
+	 * under corresponding bucket of the folder same as the IFile and no need to
+	 * traverse to Infinite depth.
+	 */
```
HeikoKlare, line 565: Please put this comment directly above the test method to have it associated as a method documentation.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
```diff
@@ -550,4 +555,42 @@ public void testSimpleUpdate() throws CoreException {
 		}
 	}
 
+	/**
+	 * Whenever a delete operation is called on an IFile it's properties also
+	 * deleted from .index file .This Test case validates for given IFile resource
```
HeikoKlare, line 560:
```suggestion
	 * deleted from .index file. This Test case validates for given IFile resource
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
```diff
@@ -550,4 +555,42 @@ public void testSimpleUpdate() throws CoreException {
...
+	 * Zero depth is calculated to traverse through folders for loading right index
+	 * file and delete its properties . because the required index file is present
```
HeikoKlare, line 562:
```suggestion
	 * file and delete its properties, because the required index file is present
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
```diff
@@ -550,4 +555,42 @@ public void testSimpleUpdate() throws CoreException {
...
+	 * under corresponding bucket of the folder same as the IFile and no need to
+	 * traverse to Infinite depth.
+	 */
+
+	@Captor
+	ArgumentCaptor<IResource> resourceArgCaptor;
+	@Captor
+	ArgumentCaptor<Integer> depthArgCapture;
```
HeikoKlare, line 570: Is there a reason for having these captors injected into fields of the test class instance? I think it should be sufficient to create them rifht in the test method, then no one may be confused why these captors are injected into fields even though being only used in a single test method:
```java
ArgumentCaptor<IResource> resourceArgCaptor = ArgumentCaptor.forClass(IResource.class);
ArgumentCaptor<Integer> depthArgCapture = ArgumentCaptor.forClass(Integer.class);
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/properties/PropertyManagerTest.java
```diff
@@ -550,4 +555,42 @@ public void testSimpleUpdate() throws CoreException {
...
+
+	@Test
+	public void test() throws CoreException {
```
HeikoKlare, line 573: Please give the test case a more meaningful name, such as `testFileDeleteTraversalDepth`.

HeikoKlare (commented):
Thank you for updating the test case. In general, it looks good now. I only have some minor comments remaining. Most important of them is the test method name, which should be more meaningful than just `test()`.

HeikoKlare (approved):
Test case looks good now, thank you!

Failing tests do not seem to be related to this PR. I will check again soon.

HeikoKlare:
Failing test is independent from this PR and documented in #128.

PR #853, opened 2023-11-10
Initialize cause of some InterruptedExceptions when they represent some cancelation by the user

fedejeanne:
Pack an `OperationCanceledException` inside the `InterruptedException` when this is thrown due to the user canceling an operation. This makes it easier to determine the cause of the `InterruptedException` from outside the methods (i.e. in the catch-clauses)

Same idea as https://github.com/eclipse-platform/eclipse.platform.ui/pull/1259

### DIFF
--- a/runtime/bundles/org.eclipse.core.tools/src/org/eclipse/core/tools/search/FindUnusedMembers.java
+++ b/runtime/bundles/org.eclipse.core.tools/src/org/eclipse/core/tools/search/FindUnusedMembers.java
@@ -17,10 +17,30 @@ package org.eclipse.core.tools.search;
 import java.io.IOException;
 import java.io.Writer;
 import java.lang.reflect.InvocationTargetException;
-import org.eclipse.core.runtime.*;
-import org.eclipse.jdt.core.*;
-import org.eclipse.jdt.core.dom.*;
-import org.eclipse.jdt.core.search.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.SubMonitor;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
+import org.eclipse.jdt.core.IMember;
+import org.eclipse.jdt.core.IMethod;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.Signature;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTParser;
+import org.eclipse.jdt.core.dom.IBinding;
+import org.eclipse.jdt.core.dom.IMethodBinding;
+import org.eclipse.jdt.core.dom.ITypeBinding;
+import org.eclipse.jdt.core.dom.IVariableBinding;
+import org.eclipse.jdt.core.search.IJavaSearchConstants;
+import org.eclipse.jdt.core.search.IJavaSearchScope;
+import org.eclipse.jdt.core.search.SearchEngine;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchParticipant;
+import org.eclipse.jdt.core.search.SearchPattern;
+import org.eclipse.jdt.core.search.SearchRequestor;
 import org.eclipse.jdt.internal.corext.dom.Bindings;
 import org.eclipse.jdt.internal.corext.refactoring.RefactoringScopeFactory;
 import org.eclipse.jface.operation.IRunnableWithProgress;
@@ -202,7 +222,9 @@ public class FindUnusedMembers implements IRunnableWithProgress {
 		try {
 			process(monitor);
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException(e.getMessage());
+			interruptedException.initCause(e);
+			throw interruptedException;
 		} catch (CoreException e) {
 			throw new InvocationTargetException(e);
 		}


fedejeanne:
The check failures don't seem to be related to my changes.

runtime/bundles/org.eclipse.core.tools/src/org/eclipse/core/tools/search/FindUnusedMembers.java
```diff
@@ -202,7 +222,9 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException, Inte
 		try {
 			process(monitor);
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException();
```
HeikoKlare, line 225: Is there a reason why the `InterruptedException` is not initialized with the message of the causing exception here, like it is done at the other places where this kind of exception wrapping is appleid?

HeikoKlare (commented):
I do not really understand why we need all this initializations of new `OperationCanceledExceptions` as the cause for the `InterruptedExceptions`. At the existing places where this kind of code was introduced, an `OperationCanceledException` was caught and an `InterruptedException` was thrown instead. It makes sense to add that exception as the cause because it actually _is_ the cause and contains all the information about the actual exception that happened (with error message, stack trace etc.). So this was about _information loss_.

I also see the point that the places at which the code is introduced here are actually about canceled operations, but I do not understand why adding an `OperationCanceledException` as cause for the thrown `InterruptedException` without any further information is an improvement. `InterruptedException` is a caught exception, so the methods throwing it in case an operation is canceled have to declare it and document it with the information when it is thrown (which is contradictory to what an `InterruptedException` is about). So every existing information is already provided. And actually the change is semantically wrong, because the cause of the `InterruptedException` is not an `OperationCanceledException` but only that an operation is canceled. The `OperationCanceledException` has never been thrown, so it cannot be the cause for any other exception. 

Finally, all these thrown `InterruptedExceptions` are kind of broken API, because the methods throw a (semantically wrong) `InterruptedException` instead of an `OperationCanceledException` A proper fix would be to replace the API, but that's probably not worth the effort concerning all the deprecation and necessary adaptations.

runtime/bundles/org.eclipse.core.tools/src/org/eclipse/core/tools/search/FindUnusedMembers.java
fedejeanne: No reason, it's been added in c116573c4620558fc49a8b18999dfad4bc265c57.
Thank you for noticing!

fedejeanne:
I'm with you: I think this was an attempt to force the user to catch and handle the possible cancelation (which the user itself probably triggered anyway) and the easiest way was to use a checked exception that was broadly available in order to ease the maintenance. I can only presume `InterruptedException` must have been an acceptable choice because:
* It's checked
* It's accessible everywhere
* It rarely happens on its own

This last part being IMO specially controversial since one can not really know how the rest of the code, _e.g._ other libraries, (mis)use this exception.

I think changing this pattern (and replacing `InterruptedException` with a new checked exception) could be the topic of another PR. This PR merely tries to add a bit more information to the existing implementation **without breaking the user's code** for which wrapping the `OperationCanceledException` is a good choice since the user's code may remain unchanged.

HeikoKlare:
> I think changing this pattern (and replacing `InterruptedException` with a new checked exception) could be the topic of another PR.

As mentioned before, I don't think that we can/should replace this. You would need to introduce new API, deprecate the old one, and expect everyone to adapts their API usage to be able to eventually remove the old API. Otherwise, you need a major version bump for breaking the API. I don't think either option is worth the effect and/or effort.

> This PR merely tries to add a bit more information to the existing implementation **without breaking the user's code** for which wrapping the `OperationCanceledException` is a good choice since the user's code may remain unchanged.

Yes, it's good to add the causing exception whenever there is one. So I totally agree with those changes. I only have concerns w.r.t. the changes in which exceptions that have never existed are invented, as they are incorrect.

fedejeanne:
>  I only have concerns w.r.t. the changes in which exceptions that have never existed are invented, as they are incorrect.

The `Exception`s (the actual instances) might have not existed but the conditions for them to exist were given: the user canceled an operation. Before this PR, **any** `InterruptedException` had been interpreted as a cancelation by the user, regardless of the actual reason. This PR adds the reason exactly there were the user **actually wanted to cancel the execution** by instantiating a new `OperationCanceledException`. This is what `ThreadJob::waitForRun` and `SubMonitor.split` do: they check if the monitor has been canceled and they instantiate (and throw) an `OperationCanceledException`, so this is actually correct behavior that was missing in some parts of the code.

HeikoKlare:
> The `Exception`s (the actual instances) might have not existed but the conditions for them to exist were given: the user canceled an operation.

Well, just because something happens that matches to an existing exception type does mean that you have to throw it 🙂 In this case, the exception is only *mis*used to transfer the cancel event from an asynchronous execution to the main thread, nothing else. The information inside the exception is not consumed by anyone.

> Before this PR, **any** `InterruptedException` had been interpreted as a cancelation by the user, regardless of the actual reason.

1. At all the touched code places, the `InterruptedException` is manually thrown as an indicator for a canceled operation for the calling code. There is no other statement that produces an `InterruptedException`.
2. All the code is internal. None of the exceptions is passed to outside a method. They are converted into boolean values before. So how does creating an additional exception, which produces overhead by creating another strack trace, help when this information is not provided to anyone?

fedejeanne:
> 2. All the code is internal. None of the exceptions is passed to outside a method. They are converted into boolean values before. So how does creating an additional exception, which produces overhead by creating another strack trace, help when this information is not provided to anyone?

Now I get you and that's a good point: no need to add information if that information neither reaches the logs nor the user anyway. I've reverted my changes in 3 locations in 2cf0a64bd999359fd318bbe5d5e0378685185ebf

FWIW, I did some digging about `InterruptedException`s and I found [this answer in SO](https://stackoverflow.com/a/26856361/834463). I stand corrected (see [my previous comment](https://github.com/eclipse-platform/eclipse.platform/pull/853#issuecomment-1807699776)): I now think it was a good choice to go for `InterruptedException` since it's been designed exactly for that.
 I think what bothered me the most in the beginning was that our javadocs usually mention a cancelation **triggered by the user** and I always associated the `InterruptedException` with an abnormal termination **triggered by the JVM**. Now that I see that this is not necessarily the case, all is good in the world again :)

I thank you for your detailed input and your patience, without it I wouldn't have learned what I've learned today! 🔝

HeikoKlare (approved):
Looks good. Let's merge after freeze period for 4.31 M1.

PR #854, opened 2023-11-10
Use try-with-resource

jukzi:
and enable explicitlyClosedAutoCloseable=warning


vogella:
Very nice to see these clean-ups, thanks for working on these @jukzi I used to be one of the big promoters of these clean-ups so it is great that you do this these days. Lets do with such mass changes within the regular dev cycle (M1-M3) and not shortly before a release.

NOTE: You may want to do similar things in JDT UI / Debug / Core, so far that this code base did not use the JDT clean-ups which I always found wrong. If JDT does not trust their own clean-ups, why should anyone....

jukzi:
All this fixes are hand crafted. The related jdt clean-up can not find or fix them.

vogella:
> All this fixes are hand crafted. The related jdt clean-up can not find or fix them.

Nice, maybe because we ran the JDT clean-ups already in platform? I'm sure JDT clean-ups will find lots of cases to clean-up in JDT code.

resources/bundles/org.eclipse.core.resources/.settings/org.eclipse.jdt.core.prefs
Bananeweizen, line 1: unrelated changes in these settings

jukzi: true

team/bundles/org.eclipse.team.core/src/org/eclipse/team/internal/core/subscribers/AbstractContentComparator.java
```diff
@@ -52,30 +52,16 @@ public boolean compare(IResource e1, IFileRevision e2, IProgressMonitor monitor)
...
-		try {
-			monitor.beginTask(null, 100);
-			is1 = getContents(e1, Policy.subMonitorFor(monitor, 30));
-			is2 = getContents(e2, Policy.subMonitorFor(monitor, 30));
+		monitor.beginTask(null, 100);
+		try (InputStream is1 = getContents(e1, Policy.subMonitorFor(monitor, 30));
+				InputStream is2 = getContents(e2, Policy.subMonitorFor(monitor, 30))) {
 			return contentsEqual(Policy.subMonitorFor(monitor, 40), is1, is2, shouldIgnoreWhitespace());
+		} catch (IOException e) {
+			return false;
```
HeikoKlare, line 60: Same as above: This is not behavior preserving? Before, with an `IOException` during `close()` the method still returned whatever `contentEquals(...)` delivered, now it returns `false`.

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/patch/ApplyPatchOperation.java
```diff
@@ -229,8 +229,7 @@ public void run() {
 
 	private static IFilePatch[] internalParsePatch(IStorage storage)
 			throws CoreException {
-		BufferedReader reader = Utilities.createReader(storage);
-		try {
+		try (BufferedReader reader = Utilities.createReader(storage)) {
```
HeikoKlare, line 232: Same as above: This is not behavior-preserving. Before, an `IOException` during `close()` did nothing, now it may throw a `CoreException`.

team/examples/org.eclipse.team.examples.filesystem/src/org/eclipse/team/examples/model/ModelObjectDefinitionFile.java
```diff
@@ -86,24 +86,17 @@ public ModelObjectElementFile[] getModelObjectElementFiles() throws CoreExceptio
 	}
 
 	private static String[] readLines(IStorage file) throws CoreException {
-		BufferedReader reader = new BufferedReader(new InputStreamReader(file.getContents()));
-		String line = null;
-		List<String> result = new ArrayList<>();
-		try {
+		try (BufferedReader reader = new BufferedReader(new InputStreamReader(file.getContents()))) {
```
HeikoKlare, line 89: I am not completely sure in this scenario, but I also think this is not behavior preserving: When closing the stream fails, the catch block will throw the `CoreException` and overwrite the `return` statement from the `try` block (unlike before).

HeikoKlare (changes requested):
Please correct me if I am wrong, but I think most of these changes are not behavior preserving. The problem is that in a try-with-resources the `IOException` thrown by `close()` is handled by the same catch block as the exceptions thrown by any operation within the `try` block performed on the stream.

The following example demonstrates the behavior. It applies the refactoring pattern of which different variants are used in this PR to a minimal program:

```java
static class ExceptionOnCloseStream extends InputStream {
	@Override
	public int read() throws IOException {
		return -1;
	}
	
	public void close() throws IOException {
		throw new IOException();
	};
	
};

	closeExplicitly();
	closeAutomatically();
}

public static void closeAutomatically() {
	try(ExceptionOnCloseStream stream = new ExceptionOnCloseStream()) {
		stream.read();
	} catch (IOException e)   {
		System.out.println("exception in implicitly closed stream");
	}
}

public static void closeExplicitly() {
	ExceptionOnCloseStream stream = new ExceptionOnCloseStream();
	try {
		stream.read();
	} catch (IOException e)   {
		System.out.println("exception in explicitly closed stream");
	} finally {
		try {
			stream.close();
		} catch (IOException e) {
			// do nothing
		}
	}
}
```

jukzi:
> Please correct me if I am wrong, but I think most of these changes are not behavior preserving.

That is true. The idea is that an IOException during a close() normally never happens anyway and there is no business-case to treat such an exceptional case different from an IOException while opening the resource.

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/FileSystemComparator.java
```diff
@@ -123,17 +123,15 @@ public Object loadSnapshot(File rootLocation) throws IOException {
 
 	public void saveSnapshot(Object toSave, File rootLocation) throws IOException {
 		File summaryFile = new File(rootLocation, SNAPSHOT_FILE_NAME);
-		PrintWriter out = new PrintWriter(new OutputStreamWriter(new BufferedOutputStream(new FileOutputStream(summaryFile))));
-		Map<?, ?> snapshot = (Map<?, ?>) toSave;
-		try {
+		try (PrintWriter out = new PrintWriter(
+				new OutputStreamWriter(new BufferedOutputStream(new FileOutputStream(summaryFile))))) {
```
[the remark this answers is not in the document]
jukzi: What "StreamEncoder" are you talking about?
All close() are supposed to fail only on a hard jdk Error - like out of memory. In such a case a clean continuation can not be guaranteed anyway. 
see java.lang.AutoCloseable.close() javadoc:
```
Cases where the close operation may fail require careful
     * attention by implementers. It is strongly advised to relinquish
     * the underlying resources and to internally <em>mark</em> the
     * resource as closed, prior to throwing the exception. The {@code
     * close} method is unlikely to be invoked more than once and so
     * this ensures that the resources are released in a timely manner.
     * Furthermore it reduces problems that could arise when the resource
     * wraps, or is wrapped, by another resource.
```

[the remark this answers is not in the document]
HeikoKlare: `OutputStreamWriter` uses a `StreamEncoder` for encoding output into a byte stream. The `OutputStreamWriter` is _not_ an `AutoClosable` and will eventually fail during its `Closable#close()` without its output stream being closed, because the `StreamEncoder` flushes a buffer by trying to write to the file which may fail and leave the underlying `FileOutputStream` open.
Since this is part of the test harness, it could result in a "dangling" file handle that may conflict with further test executions.

HeikoKlare:
> there is no business-case to treat such an exceptional case different from an IOException while opening the resource.

I totally agree. So it's fine to treat both cases equally in new code. I am just a little concerned when changing the existing behavior (i.e., strictly speaking: breaking API), as there might be callers relying on `IOExceptions` during `close()` to be swallowed, no matter how unlikely that is.

HeikoKlare:
But as an addition: you and others are more experienced w.r.t. such kinds of changes in the Eclipse platform, so if you say that this will not be a problem, I am totally fine with the changes. Just wanted to mention that there are these "rather technical" incompatibilities.

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/FileSystemComparator.java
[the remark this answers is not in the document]
jukzi: please doublecheck: the OutputStreamWriter is AutoClosable 
The internally used StreamEncoder will only fail with IOException if the underlying stream throws an IOException during write - in that case a close() is not need anymore.  If you find a counter-proof please report it to JDK.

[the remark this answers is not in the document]
HeikoKlare: You are right, sorry for the confusion.

PR #856, opened 2023-11-11
Deduplicate stored launch configs of nested projects #798

### LINKED ISSUE #798: Stored launch configs of nested projects shown twice
If projects are nested (which is typical for maven reactor and modules), then launch configs stored in the nested project are shown twice in the run drop down. It might be useful to implement equals/hashCode such that the absolute file system paths are compared (or to alternatively filter based on the absolute file system path).

grafik

Bananeweizen:
Fixes #798.

m2e projects can be nested (e.g. modules stored inside a reactor). Launch configs stored the inner nested projects appear twice, because they are not equal with the old implementation checking only name and container. Therefore also check the resolved location if the containers are different.

I've tested this with an m2e IDE containing such nested projects and stored launch configs. The duplicates are gone with the change.


laeubi:
@Bananeweizen master is open do you like to finish this?

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchConfiguration.java
```diff
@@ -651,11 +651,15 @@ public ILaunchConfigurationWorkingCopy getWorkingCopy() throws CoreException {
...
-			return getName().hashCode() + container.hashCode();
+		int result = getName().hashCode();
+		if (container != null) {
+			result += container.hashCode();
+		}
+		IPath location = getLocation();
+		if (location != null) {
+			result += location.hashCode();
 		}
+		return result;
```
HannesWell, line 662: What about just one line?
```
return Objects.hash(getName(), getContainer(), getLocation());
```

HannesWell (commented):
That's a nice enhancement, the issue bothered me as well.
The same could be done for PDE's target files which are also found multiple times, in the preferences and by Oomph (not sure if it is Oomph's fault).

SarikaSinha:
@Bananeweizen 
Can you take this to closure?

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchConfiguration.java
```diff
@@ -369,7 +369,7 @@ public boolean equals(Object object) {
 			LaunchConfiguration config = (LaunchConfiguration) object;
 			if (!config.isWorkingCopy()) {
 				return getName().equals(config.getName()) &&
-					equalOrNull(getContainer(), config.getContainer());
+						(equalOrNull(getContainer(), config.getContainer()) || equalOrNull(getLocation(), config.getLocation()));
```
HannesWell, line 372:
```suggestion
						(Objects.equals(getContainer(), config.getContainer()) || Objects.equals(getLocation(), config.getLocation()));
```

HannesWell (commented):
@Bananeweizen do you have plans to complete this?

HeikoKlare:
@Bananeweizen raising the question again 🙂 do you plan to finish this?

PR #867, opened 2023-11-18
try eclipse tool for finding compiler warnings

Bananeweizen:
See https://github.com/eclipse-platform/.github/issues/144. Need to check the Eclipse CI build afterwards.


Bananeweizen:
did not work

PR #732, opened 2023-10-06
Fix eclipse-platform/eclipse.platform#1173 - bad UI on MacOS merge viewer

zulus:
ToolbarManager invoke `.pack()` in the meantime, and might leave viewer broken (due for example ProgressDialog)

Based on layout code I implemented computeSize() method

Issue #770, opened 2023-10-26
Failure of some tests in ProcessConsoleTests


fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform/pull/766/checks?check_run_id=18039630262

```java
array lengths differed, expected.length=1 actual.length=0; arrays first differed at element [0]; expected:<-84> but was:<end of array>
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:89)
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
	at org.junit.Assert.internalArrayEquals(Assert.java:534)
	at org.junit.Assert.assertArrayEquals(Assert.java:343)
	at org.junit.Assert.assertArrayEquals(Assert.java:354)
	at org.eclipse.debug.tests.console.ProcessConsoleTests.testBinaryOutputToFile(ProcessConsoleTests.java:491)
```

and in https://github.com/eclipse-platform/eclipse.platform/actions/runs/6639804748/job/18080581872?pr=766

```java
org.junit.ComparisonFailure: Wrong content in line 0 expected:<['Native' process started.]> but was:<[]>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.eclipse.debug.tests.console.ProcessConsoleTests.testOutput(ProcessConsoleTests.java:448)
```

jukzi:
Failed also with "Wrong content in line 0 expected" during 
https://github.com/eclipse-platform/eclipse.platform/pull/764/checks?check_run_id=18290526329

PR #869, opened 2023-11-20
Revert "Fix eclipse-platform/eclipse.platform#1173 - bad UI on MacOS merge viewer"

### LINKED ISSUE #868: NPE in ContentMergeViewer
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.ToolBar.getParent()" because the return value of "org.eclipse.jface.action.ToolBarManager.getControl()" is null
	at org.eclipse.compare.contentmergeviewer.ContentMergeViewer.lambda$5(ContentMergeViewer.java:811)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4046)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3662)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:342)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:645)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:342)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:552)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:402)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:255)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:651)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:588)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1459)
```
eclipse.buildId=4.30.0.I20231119-1800
java.version=21

sratz:
Reverts eclipse-platform/eclipse.platform#732

Async-exec caused a NPE, see #868.

Fixes #868.

Also, there are some concerns regarding the actual fix, see https://github.com/eclipse-platform/eclipse.platform/pull/732#discussion_r1383011960

### DIFF
--- a/team/bundles/org.eclipse.compare/build.properties
+++ b/team/bundles/org.eclipse.compare/build.properties
@@ -12,6 +12,7 @@
 #     IBM Corporation - initial API and implementation
 ###############################################################################
 source.. = compare/
+output.. = bin/
 bin.includes = icons/,\
                plugin.xml,\
                .,\
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/ContentMergeViewer.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/ContentMergeViewer.java
@@ -78,7 +78,6 @@ import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
-import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Layout;
 import org.eclipse.swt.widgets.Sash;
 import org.eclipse.swt.widgets.Shell;
@@ -127,15 +126,8 @@ public abstract class ContentMergeViewer extends ContentViewer
 
 	private class ContentMergeViewerLayout extends Layout {
 		@Override
-		public Point computeSize(Composite composite, int wHint, int hHint, boolean force) {
-			if (hHint > SWT.DEFAULT && wHint > SWT.DEFAULT) {
-				return new Point(wHint, hHint);
-			}
-
-			Rectangle r = composite.getClientArea();
-
-
-			return new Point(r.width, r.height);
+		public Point computeSize(Composite c, int w, int h, boolean force) {
+			return new Point(100, 100);
 		}
 
 		@Override
@@ -805,12 +797,8 @@ public abstract class ContentMergeViewer extends ContentViewer
 				ToolBarManager tbm = (ToolBarManager) getToolBarManager(fComposite.getParent());
 				if (tbm != null ) {
 					updateToolItems();
-					Display.getDefault().asyncExec(() -> {
-						// relayout in next tick
-						tbm.update(true);
-						tbm.getControl().getParent().setRedraw(true);
-					});
-
+					tbm.update(true);
+					tbm.getControl().getParent().layout(true);
 				}
 			}
 		}


sratz:
@vogella @iloveeclipse would this revert be fine for RC2? What about the API tooling failures? They are again caused by https://github.com/eclipse-tycho/tycho/issues/3019 but I don't think we should now touch all the other bundles here.

HeikoKlare:
With respect to the API tools failure: The builds for this PR should work once the configuration for `org.eclipse.help.base` is changed, see #862 and #871. It should not be necessary to adapt further configurations for the builds of this PR.

vogella:
Fine for me to merge for rc2

iloveeclipse:
> Fine for me to merge for rc2

No, it needs a workaround for API errors, otherwise it will most likely break the SDK build.

HeikoKlare:
#871 would fix all API errors. Or at least merging #862 should be sufficient for the builds of this PR to work again. The latter would probably be the "save" one to get this merged for the upcoming release.

iloveeclipse:
But that is not acceptable for rc2 as it would touch too many bundles

HeikoKlare:
#862 would only touch a single bundle. If it allows to merge this PR, maybe that's worth it? In particular because the only effect of that PR is that API tools will work for the affected bundle again.

But I only want to emphasize that as an opportunity to still merge this revert PR. I know it's very late in the release cycle and it's totally fine for me if we keep it as it is as changes are too large/risky.

iloveeclipse:
> #862 would only touch a single bundle. If it allows to merge this PR, maybe that's worth it? In particular because the only effect of that PR is that API tools will work for the affected bundle again.

OK, that's good.

> But I only want to emphasize that as an opportunity to still merge this revert PR. I know it's very late in the release cycle and it's totally fine for me if we keep it as it is as changes are too large/risky.

I think we should merge this revert after #862
I will do that now.

iloveeclipse (approved):

HeikoKlare:
Great. The test failures are random failures as documented in #770.

laeubi:
> https://github.com/eclipse-platform/eclipse.platform/pull/862 would only touch a single bundle.

I also wanted to note, that changing the `build.properties` here is not a bundle change as we do not include the file in the bundle and therefore a configuration change that can happen anytime in the release cycle  (@akurtakov might correct me if I'm wrong) as long as the build works (or fix the build) and we do not change things like `bin.includes` (what _have_ an impact on the bundle content).

HeikoKlare:
> I also wanted to note, that changing the `build.properties` here is not a bundle change as we do not include the file in the bundle and therefore a configuration change that can happen anytime in the release cycle

That's also my understanding. But still one note on that: this kind of change to the `build.properties` seems to be treated as a change to the bundle by the API tools, which means that they will require a (micro) version bump. In case the bundle has not been touched in the current release cycle, the change to the `build.properties` will thus require a version bump to make API validation succeed, which then affects the generated artifact.
Maybe I misunderstood something, but when fixing all the `build.properties` for the platform repositories, I came across API tools complaining about missing version increments in case it was not incremented in the current release cycle already, see, e.g., https://github.com/eclipse-platform/eclipse.platform.ui/pull/1321

PR #871, opened 2023-11-21
Correct output and source specifications in build.properties

HeikoKlare:
* Remove source entries in projects without source folders
* Remove duplicate source entries
* Add missing output entries
* Fix project classpaths with unseparated bin folders and adapt .gitignore

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/.gitignore
+++ b/resources/bundles/org.eclipse.core.resources/.gitignore
@@ -1,3 +1,3 @@
 temp.folder
-src_ant_bin
+bin_ant
 ant_tasks
--- a/ua/org.eclipse.help.base/build.properties
+++ b/ua/org.eclipse.help.base/build.properties
@@ -11,10 +11,10 @@
 # Contributors:
 #     IBM Corporation - initial API and implementation
 ###############################################################################
-source..=src/,\
-         src_demo/
-src.includes=schema/,about.html
+source.. = src/,\
+           src_demo/
 output.. = bin/
+src.includes = schema/,about.html
 bin.includes = doc/,\
                plugin.xml,\
                preferences.ini,\
--- a/debug/org.eclipse.unittest.ui/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.unittest.ui/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.unittest.ui
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.unittest.ui;singleton:=true
-Bundle-Version: 1.1.200.qualifier
+Bundle-Version: 1.1.300.qualifier
 Bundle-Activator: org.eclipse.unittest.internal.UnitTestPlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/debug/org.eclipse.unittest.ui/build.properties
+++ b/debug/org.eclipse.unittest.ui/build.properties
@@ -19,6 +19,7 @@ bin.includes = plugin.xml,\
                META-INF/
 
 source.. = src/
+output.. = bin/
 src.includes = about.html,\
                schema/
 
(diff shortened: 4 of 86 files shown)


ua/org.eclipse.help.base/build.properties
```diff
@@ -11,9 +11,10 @@
 # Contributors:
 #     IBM Corporation - initial API and implementation
 ###############################################################################
-source..=src/,\
-         src_demo/
-src.includes=schema/,about.html
+source.. = src/,\
```
fedejeanne, line 14: `o.e.help.base` is already provided by https://github.com/eclipse-platform/eclipse.platform/pull/862

fedejeanne (changes requested):

jukzi:
Thanks, we need such for all repositories.

ua/org.eclipse.help.base/build.properties
```diff
@@ -30,5 +31,6 @@ bin.includes = doc/,\
 jars.compile.order = .,\
                      ant_tasks/helpbase-ant.jar
 source.ant_tasks/helpbase-ant.jar = src_ant/
+output.ant_tasks/helpbase-ant.jar = bin2/
```
jukzi, line 34:
```suggestion
output.ant_tasks/helpbase-ant.jar = bin_ant/
```

resources/bundles/org.eclipse.core.resources/.gitignore
```diff
@@ -1,3 +1,3 @@
 temp.folder
-src_ant_bin
-ant_tasks
\ No newline at end of file
+ant_tasks
+bin_ant
```
jukzi, line 3:
```suggestion
bin_ant
ant_tasks
```

ua/org.eclipse.help.base/build.properties
HeikoKlare: That's true, this one supercedes #862, I forgot to mention that.
What kind of changes do you expect for this PR?

ua/org.eclipse.help.base/build.properties
HeikoKlare: I had that change first, but the `.classpath` is defined for `bin2`, so that would have to be adapted as well, along with the `.gitignore`. And it would result in those resources already existing in `bin2` from previous builds to pop up for being added to the repo (as the .gitignore changes to `bin_ant`). I wanted to avoid that by sticking to the existing output folder name `bin2`, although I would also prefer your proposal of `bin_ant`. Would you prefer that we complete change to `bin_ant` and except that developers might be confused once by resources in `bin2` popping up?

jukzi: ok

resources/bundles/org.eclipse.core.resources/.gitignore
HeikoKlare: Fixed in adc852b

HeikoKlare:
> Thanks, we need such for all repositories

I can go through other repositories whenever I find time to do so. Seems that currently this repository is the most affected one, as builds for PRs in other repos like Platform UI and SWT do not fail, so it should not be as urgent for those repositories as it is for this one, right?

jukzi:
All repositories are failing, just depends which projects are changed. It's best seen on bulk changes like https://github.com/eclipse-platform/eclipse.platform.ui/pull/1287

HeikoKlare:
> All repositories are failing, just depends which projects are changed. It's best seen on bulk changes like [eclipse-platform/eclipse.platform.ui#1287](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1287)

I see. I drafted a PR for Platform UI: https://github.com/eclipse-platform/eclipse.platform.ui/pull/1321

ua/org.eclipse.help.base/build.properties
fedejeanne: No need to supersede, one can build upon it and acknowledge my contribution too... small as it is.

And I thank you for your input in my PR too, specially for [pointing out the missing entry](https://github.com/eclipse-platform/eclipse.platform/pull/862#issuecomment-1820673386) :-)

fedejeanne (approved):

PR #872, opened 2023-11-22
Javadoc: fix unclosed <code>

jukzi:


jukzi:
ignoring unrelated random test failures like https://github.com/eclipse-platform/eclipse.platform/issues/715

PR #873, opened 2023-11-22
PKI Implementation 

JavaJoeS:
PKI integration has been divided into two parts, ui and core for headless systems, as requested.   I believe that I have completed all the requests per the review.   Testing may be accomplished by creating a .pki file in your .m2 directory, or you can go to org.eclipse.core.pki.util and run the TemplateForPKIfile and it will create a pki.template file for you to fill in per your config.
My recommendation is use PKCS12 for simple testing.  PKCS11 requires a vendor implementation such as OpenSC or the like.


JavaJoeS:
Dont forget to remove all of the print statements!

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,35 @@
...
+ org.eclipse.pki.dialog,
+ org.eclipse.pki.exception,
+ org.eclipse.pki.interfaces,
+ org.eclipse.pki.jsse,
+ org.eclipse.pki.pkcs,
+ org.eclipse.pki.pkiselection,
+ org.eclipse.pki.preferences,
+ org.eclipse.pki.ui,
+ org.eclipse.pki.util,
+ org.eclipse.pki.wizard,
```
mickaelistria, line 27: This list must be shrinked to the bare minimal amount of packages that need to be exported. Ideally, nothing needs to be exported.

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,35 @@
...
+ target.classes.org.eclipse.core.pki
```
mickaelistria, line 28: This line is clearly wrong. The `target.classes` is not a package, it's a mistake coming from erroneous project configuration.

mickaelistria (changes requested):
Before we can consider a review
* Please put appropriate license headers to all files
* Please squash your commits into a single one. Make sure the commit message describes efficiently the problem that this feature intends to solve
* Please describe in the PR a way to test it
* Please remove all unused code, commented code and all System.out
* Please use org.eclipse.core.net.pki package name
* Have you considered providing this support directly in existing org.eclipse.core.net/org.eclipse.ui.net bundles?
* If new bundle(s) is necessary, please split those 2 into 1 core and 1 ui bundles; we want security stuff to work even without UI (Eclipse Platform is used without its UI in various cases).

But the most important part is that you didn't yet make clear what problem you're trying to solve that cannot be already solved with proper JVM settings, and how one can verify the problem and the solution. Without it, this code will be forever ignored.

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,35 @@
...
+Bundle-Vendor: %PLUGIN_PROVIDER
+Automatic-Module-Name: org.eclipse.core.pki
+Bundle-ActivationPolicy: lazy
+Bundle-Activator: org.eclipse.pki.auth.AuthenticationPlugin
+Bundle-ClassPath: target/org.eclipse.core.pki-1.0.0.jar,
```
ruspl-afed, line 33: This could be replaced with 
`Bundle-ClassPath: .`

team/bundles/org.eclipse.core.pki/plugin.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ruspl-afed, line 1: missed license header

team/bundles/org.eclipse.core.pki/plugin.xml
```diff
@@ -0,0 +1,19 @@
...
+
+<plugin>
+	<extension point="org.eclipse.ui.preferencePages">
+		<page
+			category="org.eclipse.ui.net.NetPreferences"
+            class="org.eclipse.pki.preferences.PreferencePage"
+            id="org.eclipse.preferences.pkcs"
+            name="Java KeyStore Setup">
```
ruspl-afed, line 9: missed i18n

team/bundles/org.eclipse.core.pki/src/org/eclipse/pki/auth/AuthenticationPlugin.java
```diff
@@ -0,0 +1,1133 @@
...
+import org.eclipse.pki.wizard.PKILoginWizard;
+import org.eclipse.pki.wizard.TrustStoreLoginWizard;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.preferences.ConfigurationScope;
+
+/**
+ * The activator class controls the plug-in life cycle
+ */
+public class AuthenticationPlugin extends AbstractUIPlugin {
```
ruspl-afed, line 76: this one should be split to `core` and `ui` part

team/bundles/org.eclipse.core.pki/src/org/eclipse/pki/auth/AuthenticationPlugin.java
```diff
@@ -0,0 +1,1133 @@
...
+	private static final int DIGITAL_SIGNATURE = 0;
+	private static final int KEY_CERT_SIGN = 5;
+	private static final int CRL_SIGN = 6;
+	
+	private static final TrustManager trustManager = new CdeX509TrustManager();
+	
+    // The shared instance
+    private static AuthenticationPlugin plugin;
+    
+    protected static final String DEFAULT_TRUST_STORE = "cacerts";
```
ruspl-afed, line 99: not clear why one would need `protected` constants for Activator. I hope there is no plan to inherit from this type.

team/bundles/org.eclipse.core.pki/src/org/eclipse/pki/auth/AuthenticationPlugin.java
```diff
@@ -0,0 +1,1133 @@
...
+    
+    /**
+     * Returns the shared instance
+     *
+     * @return the shared instance
+     */
+    public static AuthenticationPlugin getDefault() {
+    	System.out.println("AuthenticationPlugin ---getDefault plugin");
+    	if (plugin == null) {
+    		AuthenticationPlugin auth = new AuthenticationPlugin();
```
ruspl-afed, line 125: Do you plan to support running without Platform for your plug-in? In this case I would consider another functionality entry point instead. 
Moreover, activator should be an internal type that implements hooks for Bundle lifecycle. Ideally activator should not be present at all to avoid startup impact.

ruspl-afed (changes requested):
The majority of files needs license header

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
JavaJoeS: OK, gotta figure out what really is needed if anything

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
JavaJoeS: remove target

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,35 @@
...
+Bundle-ManifestVersion: 2
+Bundle-Name: %PLUGIN_NAME
+Bundle-Localization: plugin
+Bundle-SymbolicName: org.eclipse.core.pki; singleton:=true
+Require-Bundle: org.eclipse.equinox.security;bundle-version="[1.0.0,2.0.0)",
+ org.eclipse.equinox.common;bundle-version="3.12.0",
+ org.eclipse.equinox.preferences;bundle-version="3.2.200",
+ org.eclipse.osgi;bundle-version="3.4.0",
+ org.eclipse.equinox.registry;bundle-version="3.4.0",
+ org.eclipse.ui;bundle-version="3.203.100",
```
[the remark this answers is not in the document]
JavaJoeS: Any help with this greatly appreciated

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
JavaJoeS: Yes...   will do.

team/bundles/org.eclipse.core.pki/plugin.xml
JavaJoeS: Can I grab a header off file already in a bundle, copy and paste to all files needing header????

team/bundles/org.eclipse.core.pki/plugin.xml
JavaJoeS: Where does that go?   on the name portion?

team/bundles/org.eclipse.core.pki/src/org/eclipse/pki/auth/AuthenticationPlugin.java
JavaJoeS: Agree ton of work needed here.

team/bundles/org.eclipse.core.pki/src/org/eclipse/pki/auth/AuthenticationPlugin.java
JavaJoeS: Very good question..  I think the answer should be YES.  Support with or without GUI.  The gotcha on this is
having all the required setup, more so for PKCS11 implementation..

team/bundles/org.eclipse.core.pki/META-INF/MANIFEST.MF
JavaJoeS: When I remove the target/**.jar, then plugin wont work.. Any suggestions?  Im sure its me not being 
so familiar with plugins...

team/bundles/org.eclipse.core.pki/plugin.xml
JavaJoeS: Tried to find another plugin.xml file so I could see where the i18n is supposed to be...  None that I looked at had it??

JavaJoeS:
In a nutshell, today's  internet is tons of URLs which almost all are HTTPS single authentication sites.  They are secured by X509 Certificates that (mostly) have been secured via public Certificate Authorities.  They are single, basic authentication since we all use some browser or other that has the Certificate Authority already added in, or via a Java JDK where they are also embedded.  HOWEVER, more and more companies, private entities, even some gubmints are creating internet sites with URLs that require two way authentication requiring YOU to have your own x509 Certificate in a Key Store signed by Their Certificate Authority.  Eclipse DOES not contain this capability.  The PKI Implementation I am trying to add, will provide this capability.

JavaJoeS:
Where is the best place to write docs to describe usage/testing?  I use my account from Amazon AWS to do lots of coding.
account email is different?  Do I need an additional ECA for that account too?

JavaJoeS:
I believe I have satisfied all the requests/modifications.

PR #878, opened 2023-11-22
InputStreamMonitor: stop on IOException - fixes #877

### LINKED ISSUE #877: debug: java.io.IOException flooding Error log
happend while i terminated a debugee:
```
java.io.IOException: Die Pipe wird gerade geschlossen
	at java.base/java.io.FileOutputStream.writeBytes(Native Method)
	at java.base/java.io.FileOutputStream.write(FileOutputStream.java:367)
	at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:125)
	at java.base/java.io.BufferedOutputStream.implWrite(BufferedOutputStream.java:222)
	at java.base/java.io.BufferedOutputStream.write(BufferedOutputStream.java:200)
	at java.base/java.io.FilterOutputStream.write(FilterOutputStream.java:110)
	at org.eclipse.debug.internal.core.InputStreamMonitor.writeNext(InputStreamMonitor.java:188)
	at org.eclipse.debug.internal.core.InputStreamMonitor.write(InputStreamMonitor.java:169)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
eclipse.buildId=4.30.0.I20231119-1800
java.version=21

image

a bit more silence would be nice.

jukzi:
No chance that OutputStream recovers from IOException

https://github.com/eclipse-platform/eclipse.platform/issues/877

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
@@ -18,8 +18,11 @@ package org.eclipse.debug.internal.core;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
-import java.util.Vector;
+import java.util.Arrays;
+import java.util.Queue;
+import java.util.concurrent.LinkedTransferQueue;
 
+import org.eclipse.core.runtime.Status;
 import org.eclipse.debug.core.DebugPlugin;
 
 /**
@@ -38,12 +41,12 @@ public class InputStreamMonitor {
 	/**
 	 * The queue of output.
 	 */
-	private final Vector<byte[]> fQueue;
+	private final Queue<byte[]> fQueue;
 
 	/**
 	 * The thread which writes to the stream.
 	 */
-	private Thread fThread;
+	private volatile Thread fThread;
 
 	/**
 	 * A lock for ensuring that writes to the queue are contiguous
@@ -53,7 +56,7 @@ public class InputStreamMonitor {
 	/**
 	 * Whether the underlying output stream has been closed
 	 */
-	private boolean fClosed = false;
+	private volatile boolean fClosed = false;
 
 	/**
 	 * The charset of the input stream.
@@ -78,7 +81,7 @@ public class InputStreamMonitor {
 	 */
 	public InputStreamMonitor(OutputStream stream, Charset charset) {
 		fStream = stream;
-		fQueue = new Vector<>();
+		fQueue = new LinkedTransferQueue<>();
 		fLock = new Object();
 		fCharset = charset;
 	}
@@ -104,10 +107,7 @@ public class InputStreamMonitor {
 	 * @param text text to append
 	 */
 	public void write(String text) {
-		synchronized (fLock) {
-			fQueue.add(fCharset == null ? text.getBytes() : text.getBytes(fCharset));
-			fLock.notifyAll();
-		}
+		write(fCharset == null ? text.getBytes() : text.getBytes(fCharset));
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


iloveeclipse (dismissed):
> No chance that OutputStream recovers from IOException

Why do you think so? A process communication might be temporarily interrupted? 
I would not generalize shutdown issue you've experienced with a very generic code here that not only used by JDT.
 
I would rather check in the catch block of the `writeNext` if the `fStream` is closed and don't log an error in that case, plus add a check for the closed stream in the `while (fThread != null && !fClosed)` loop in the `write` method.

jukzi:
> Why do you think so? A process communication might be temporarily interrupted?

https://stackoverflow.com/questions/44958315/in-java-can-a-process-pipe-stream-reader-encountering-an-ioexception-recover
does not provide a proof but nobody said it would ever recover.
I have never seen a broke system pipe recover from error. Even if it would recover after some seconds - the current code would flood your harddisk meanwhile with an incredible rate.
An IOException is not a guarantee that the Stream is closed, so your proposal will not work. The Stream was never closed but always reported the same IOException from native code.

iloveeclipse:
> An IOException is not a guarantee that the Stream is closed, so your proposal will not work. 

So if it is not a guarantee of closed stream, why do you propose to stop writing to it on IOException?

> I have never seen a broke system pipe recover from error.

:-) 

What about my actual proposal to check "stream closed" flag in the while loop? That should not rely on "never seen" arguments and stop endless errors.

jukzi:
> "stream closed" flag

How exactly do you propose to check if stream is closed? i don't see any "stream closed flag" nor would the internal java.io.FileOutputStream.closed be set by the native code.

iloveeclipse:
See my comment above, just add `while (fThread != null && !fClosed)`

jukzi:
> add `while (fThread != null && !fClosed)`

How should that help? fClosed is never set (unless pressing ctrl+z) and writeNext() does not return because queue is not empty

jukzi:
Also the current code is wrong because in case of error fQueue elements are removed and not put back. There is no recovery.

iloveeclipse:
> Also the current code is wrong because in case of error fQueue elements are removed and not put back. There is no recovery.

I don't think it is that important to have a recovery for input here. However I do think it is important not to stop accepting input on an exception.

> > add `while (fThread != null && !fClosed)`
> 
> How should that help? fClosed is never set (unless pressing ctrl+z) and writeNext() does not return because queue is not empty

Interesting, the most obvious expectation that `fClosed` is set on close is not true.
I assume that should do same what I've naively expected from `fClosed`, and should be enough to fix the problem you've observed without changing much in the code behavior:

```
diff --git a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
index 03920a6..4608304 100644
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
@@ -45,3 +45,3 @@
 	 */
-	private Thread fThread;
+	private volatile Thread fThread;
 
@@ -183,3 +183,3 @@
 	protected void writeNext() {
-		while (!fQueue.isEmpty() && !fClosed) {
+		while (!fQueue.isEmpty() && !fClosed && fThread != null) {
 			byte[] data = fQueue.firstElement();
```

jukzi:
> should be enough to fix the problem

Your fix does not solve the problem (tested). The error occurs if the debugee is terminated externally for example using process monitor. Until the observing thread stops the log is flooded. 
Why do you think it's important to accept input if the input can not be forwarded anyway? It's even misleading to accept input if the debugee does not accept input anymore.

You can also test the error with a debugee that closes the InputStream. Run this and paste a multiline content to console - each line will print an error:

```
import java.io.IOException;

public class Main {

		System.out.println("hi");
		System.in.close();
		while (true) {
			// endless
		}
	}
}
```

jukzi:
@iloveeclipse the patch nowsets ` fClosed = true` at the end (also on error) and drops further input to avoid filling a queue that is not emptied anymore.
OK?

iloveeclipse:
Will check tomorrow

jukzi:
> Will check tomorrow

@iloveeclipse  if you want to review before submit please do it

PR #888, opened 2023-11-28
Comparator errors in I20231127-0750

MohananRahul:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1557


akurtakov:
This is the first change in a bundle in this release cycle - you shouldn't touch the bundle but bump it's version to x.y.z+100 as now we get qualifier only warning in https://download.eclipse.org/eclipse/downloads/drops4/I20231128-0510/buildlogs/reporeports/reports/versionChecks.html

PR #896, opened 2023-11-29
Fix random failing ProcessConsoleTests #770

HeikoKlare:
Test test methods testOutput() and testBinaryOutputToFile() in ProcessConsoleTests randomly fail. One possible reason is the short time waiting for a process to finish output and the fixed-time waiting for UI events to be processed. This change replaces both conditions by wait operations that check for the actual conditions to be fulfilled or a timeout to occur. This will also improve performance of these tests.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/770


HeikoKlare:
Since `ProcessConsoleTests` are failing annoyingly often in current builds, I'll give this fix a try.

PR #901, opened 2023-11-30
Fixed unneeded version bump on org.eclipse.debug.ui

iloveeclipse:
This fixes API error saying unnecessary version bump. The bundle was not published yet via IBuild, so we can revert.

Regression from
https://github.com/eclipse-platform/eclipse.platform/pull/800


akurtakov:
I really hope that  https://github.com/eclipse-tycho/tycho/commit/7aec992ac34d3252d4545d6dd69d62fbf62823d8 will make this error catched in the verification build too.

PR #902, opened 2023-11-30
Remove JUnit 3 fail operations from ResourceTest

HeikoKlare:
Replaces fail(String, Throwable) operations in ResourceTest utility methods by rethrowing the exception in order to separate those methods to a utility class and prepare for a migration to JUnit 4/5.

This is part of preparatory work for migrating the `ResourceTests` to JUnit 4.


HeikoKlare:
> IIRC the JUnit template adds these "throws Exception". Maybe send an update to JDT UI for this and ask @akurtakov to review?

What kind of template do you mean? The "throws Exceptions" statements added in this PR are necessary because the utility methods in `ResourceTest` previously passed the exception to JUnit 3 `fail` method, which needs to be replaced by exceptions simply being (re)thrown by the test methods.

PR #906, opened 2023-11-30
Move getWorkspace() method from ResourceTest to utility class #903 

HeikoKlare:
This changes moves the getWorkspace() utility method defined in ResourceTest to the dedicated ResourceTestUtil test class. It prepares for removing the JUnit 3 test inheritances hierarchy to migrate to JUnit 4.

⚠️ Depends on and currently contains the changes of #904 and #905, so has to be merged afterwards.


HeikoKlare:
This became obsolete via #925.

PR #918, opened 2023-12-04
remove unchecked warnings

jukzi:
Depending on the amount of warnings on project the warning has been disabled, Suppressed or solved
To get a clean Problems View on ancient code


mickaelistria:
FWIW, I don't think hiding warnings actually makes the code better. In which cases the "unchecked" warning can't be fixed? I understand it can't for APIs that return raw types. is there something else?

jukzi:
It does not make the code better, but unchecked warnings are useless if the code works. And it does work for years. When the code was written there where no generics. needless to refactor it.
But all those warnings hide warnings that do point to real problems.

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/dialogs/ResourceMappingResourceDisplayArea.java
```diff
@@ -227,7 +226,6 @@ public Object getParent(Object o) {
 		}
 
 		@Override
-		@SuppressWarnings("unchecked")
 		public <T> T getAdapter(Class<T> adapter) {
 			if (adapter == IWorkbenchAdapter.class)
 				return (T) this;
```
laeubi, line 231: unchecked warning can be avoided by 

```
public <T> T getAdapter(Class<T> adapter) {
			if (adapter == IWorkbenchAdapter.class)
				return adapter.cast(this);
```

PR #920, opened 2023-12-04
offer non-deprecated SubProgressMonitor for friends

jukzi:
to get rid of workspace warnings

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/Policy.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/Policy.java
@@ -15,6 +15,7 @@ package org.eclipse.core.internal.utils;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import org.eclipse.core.internal.deprecated.SubProgressMonitor;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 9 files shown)


laeubi (changes requested):
I don't think we should do this for people that are to lazy to upgrade their code, instead if one really needs this as an "unrestricted non deprecated" it could be better copied into the projects private codebase.

jukzi:
> I don't think we should do this for people that are to lazy to upgrade their code, instead if one really needs this as an "unrestricted non deprecated" it could be better copied into the projects private codebase.

Can you please propose a better solution instead of just blocking? We don't have the resources to refactor all deprecated classes. 
ltk.core.refactoring is in platform.ui repository and alone has 30 classes containing SubProgressMonitor

laeubi:
As mentioned earlier, if we don't have resources (yet) just keep the warning there so it could be picked up by someone else later on, but adding *new* (not so) API to hide a previous API deprecation is not really a way forward....

vogella:
ltk.core.refactoring comes from jdt which did not join any effort in code clean-ups during the time in which we deprecated Submonitor. Reintroducing deprecated API only because a project refused to do clean ups in the past is a no go

mickaelistria:
I agree with @laeubi and @vogella , those warnings are valid warnings and shouldn't be just hidden. At some point, ltk refactoring will have to be updated and this is the only good remediation, in the meantime, keeping the warning makes sense.
@jukzi if you're chasing particular warnings, then feel free to do it in your workspace; but I don't think it's best to push a "hide all warnings" policy in the project directly; most of them are relevant and worth being considered one day.

vogella:
I can later see if I can clean up a few, IIRC doing the cleanup in projects which were open to changes like pde was not that much of an effort. You could check the commits, I guess 8 years ago.....

PR #924, opened 2023-12-05
Reduce usage of SubProgressMonitor

vogella:


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/Policy.java
```diff
@@ -179,10 +184,6 @@ public static IProgressMonitor monitorFor(IProgressMonitor monitor) {
 	}
 
 	public static IProgressMonitor subMonitorFor(IProgressMonitor monitor, int ticks) {
-		if (monitor == null)
-			return new NullProgressMonitor();
-		if (monitor instanceof NullProgressMonitor)
-			return monitor;
-		return new SubProgressMonitor(monitor, ticks);
+		return SubMonitor.convert(monitor, ticks);
```
HeikoKlare, line 187: Is this utility method really necessary after the simplification? Couldn't that code just be inlined?

vogella: Yes, SubMonitor.convert can also handle NullProgressMonitor IIRC so the whole thing can be further reduced.

vogella: > Yes, SubMonitor.convert can also handle NullProgressMonitor IIRC so the whole thing can be further reduced.

Ups, I did that already...

vogella: Looking at the "inline" I would prefer not to do this via refactoring, as it looks to me that the code logic can be further simplified, e.g. removing done calls, simplifying the call. I do a few simplifications with another PR, maybe others want to join the effort to simplify the code here.

vogella:
Looks like it is not that simple, as the caller call multiple times the beginTask method....

PR #927, opened 2023-12-05
EXPERIMENT - Removing StringPool

vogella:
StringPool seems not actively being used, this commit removes it as a fast test to see if this breaks something


vogella:
If String de-duplication is still required we can use https://stackoverflow.com/questions/27949213/string-deduplication-feature-of-java-8, IIRC EPP have them already activated.

iloveeclipse:
> StringPool seems not actively being used

And this interesting observation is based on which facts?
Beside this, I assume you run some measurements on different JVM versions from 17 to 21?

vogella:
> > StringPool seems not actively being used
> 
> And this interesting observation is based on which facts? Beside this, I assume you run some measurements on different JVM versions from 17 to 21?

EXPERIMENT....

merks:
I wonder, what will the experiment demonstrate?  Will something verify or refute that memory footprint is negatively impacted?

vogella:
Oh guys, this is marked as draft and EXPERIMENTAL... Lets start to discuss once this becomes a real thing.

merks:
Maybe this is relevant or maybe it's outdated:

https://blog.gceasy.io/2018/12/23/usestringdeduplication/#more-2861

vogella:
@jukzi the internal Eclipse Stringpool is mainly used for Markers, so lots of "normal" duplicated String will be still present even if the marker try to remove duplicates  https://stackoverflow.com/questions/5076099/avoid-duplicate-strings-in-java

vogella:
> @jukzi the internal Eclipse Stringpool is mainly used for Markers, so lots of "normal" duplicated String will be still present even if the marker try to remove duplicates https://stackoverflow.com/questions/5076099/avoid-duplicate-strings-in-java

I think some in your list ("org", "eclipse") are created by https://github.com/search?q=repo%3Aeclipse-equinox%2Fp2+new+String%28&type=code

vogella:
> But I'm not sure we are allowed to discuss anything yet...

Constructive discussion is always welcome.

vogella:
I think it is funny that @jukzi ~ 20 000 ("do not insert") and "insert" strings in his HEAP. Looks like the developer think it is import to have these.

merks:
I'm sure it wasn't intentional to suggest that previous comments were not constructive and hence unwelcome.  😱  👅 

Personally I was simply hoping to understand the goal of the experiment, which I still don't understand.

It's not hard to guess where so many `do not insert` duplicates are coming from:

image

merks:
> > It's not hard to guess where so many `do not insert` duplicates are coming from:
> 
> intersting, then the preference manager could probably intern() them when strings loaded

Yes assuming that does not have horrible performance like it once did in the historical past...

szarnekow:
Since I recently considered the opposite and make the StringPoolJob accessible from plugins, I wonder what the outcome of the experiment was.

vogella:
Closing my experiment, work to reduce existing String Duplication happened via different PR.

szarnekow:
> work to reduce existing String Duplication happened via different PR.

Can you please share a link to that work?

mickaelistria:
> Can you please share a link to that work?

https://github.com/eclipse-equinox/equinox/pull/33

HannesWell:
> Closing my experiment, work to reduce existing String Duplication happened via different PR.

Have they already been created?
In general if String-sharing is removed you probably want to intern the strings before they are set to their final data-structure?

PR #954, opened 2023-12-07
Enable tests

amartya4256:
Enabled some disabled tests which aren't in broken state anymore.
https://github.com/eclipse-platform/eclipse.platform/issues/525


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/CharsetTest.java
```diff
@@ -228,7 +227,6 @@ public void testCharsetMoveOnFileMove() throws CoreException {
 	}
 
 	@Test
-	@Ignore("https://github.com/eclipse-platform/eclipse.platform/issues/634")
```
HeikoKlare, line 231: This test still fails as intended (the according bug has not been fixed yet). Just removing the `@Ignore` statement currently has no effect, as the test is executed with JUnit 3. It is only used for documentation purposes until I have migrated all the `ResourceTest` implementation to JUnit 4. The actual deactivation for JUnit 3 execution is the `_` prefix of  the method name.

jukzi:
@amartya4256 please resolve conflicts

HeikoKlare:
We should probably close this as both tests will fail again when enabled (see my comments). I just did not want to simply close without giving feedback to @amartya4256.

PR #965, opened 2023-12-08
JenkinsFile: eclipse warnings

jukzi:


iloveeclipse:
@jukzi : please check https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1653#issuecomment-1848375531 which is most likely caused by changed compilation settings here.

PR #972, opened 2023-12-09
Use PreferencesService.prependScopeToDefaultDefaultLookupOrder()

HannesWell:
which is introduce with
- https://github.com/eclipse-equinox/equinox/pull/424

so we have to await an I-build in order to have a successful build.


HeikoKlare (approved):
The change looks good to me. Build now runs successfully.
@HannesWell can we merge? Would be nice to have the ECJ deprecation warnings fixed.

HannesWell:
> @HannesWell can we merge? Would be nice to have the ECJ deprecation warnings fixed.

Yes we can, thanks for having a look.

PR #975, opened 2023-12-11
Revert change making jdt compiler use project settings

akurtakov:
Should fix
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1653 .
Further work on the topic could be continued as described in https://github.com/eclipse-platform/eclipse.platform/pull/973#issuecomment-1849538208

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,14 +77,6 @@
             <testClassesDirectory>${tycho.surefire.testClassesDirectory}</testClassesDirectory>
           </configuration>
         </plugin>
-		<plugin>
-			<groupId>org.eclipse.tycho</groupId>
-			<artifactId>tycho-compiler-plugin</artifactId>
-			<configuration>
-				<useProjectSettings>true</useProjectSettings>
-			</configuration>
-		</plugin>
-      </plugins>
     </pluginManagement>
   </build>
 </project>


akurtakov:
Push without waiting as comparator issues shouldn't break the verification build and start I-build to get results ASAP.

jukzi:
Since the comparator issue was analysed anderstood  can this revert be reverted and comparator get a reset?

akurtakov:
@srikanth-sankaran @iloveeclipse  WDYT ? Should we go for preserveLocal and touch all needed bundles?

If that's the path recomended by compiler guys the change should happen in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/3e6518b64e478f6f31375307c185eaf15706d9e6/eclipse-platform-parent/pom.xml#L485 and not be a revert of this one.

iloveeclipse:
Alex, I'm for a buuild that uses (consistent) **project** settings accross all projects in aggregator.
Ideally what project defines in settings should be used in all project builds - IDE, Jenkins, SDK aggregator build.
If that line in parent pom makes it - perfect.
We should check before if all the various Jenkins files and poms don't override that again with something different. And ideally we should check if all the projects **have** compiler settings at all.
Unfortunately I can't commit to drive that, as I'm busy with other tasks.

PR #976, opened 2023-12-11
Fix typo in

akurtakov:
https://github.com/eclipse-platform/eclipse.platform/pull/975


akurtakov:
One more example that trying to rush things almost never leads to good results!

Issue #969, opened 2023-12-08
org.osgi.service.http: Unable to satisfy dependency 


jukzi:
The jenkins console log contains 525 times the same warning https://ci.eclipse.org/platform/job/eclipse.platform/job/master/lastBuild/console

    [WARNING] Problems resolving provisioning plan.: [Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet.http [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http.whiteboard 1.1.1.202109301733 to osgi.contract; (&(osgi.contract=JavaServlet)(version=3.1.0)).]
 

```
Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet [2.1.0,3.0.0).;
Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet.http [2.1.0,3.0.0).;
Unable to satisfy dependency from org.osgi.service.http.whiteboard 1.1.1.202109301733 to osgi.contract; (&(osgi.contract=JavaServlet)(version=3.1.0)).
``` 

Can someone solve that? I also get a similar warning when i start workspace locally
minor: why is it printed so many times?

merks:
It would be so nice to have warning free builds. After all, a warning is a cry to be fixed.

laeubi:
> Can someone solve that?

Se the current ongoing work here
- https://github.com/eclipse-equinox/equinox/pull/403

> why is it printed so many times?

Because it happens on each project when it gets resolved and for each platform and then maybe when tests are resolved as well, this is similar to another warning you you repeatedly in the logs

```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by org.eclipse.ant.internal.core.ant.InternalAntRunner (file:/home/jenkins/agent/workspace/eclipse.platform_master/ant/org.eclipse.ant.tests.ui/target/work/configuration/org.eclipse.osgi/59/0/.cp/lib/antsupportlib.jar)
WARNING: Please consider reporting this to the maintainers of org.eclipse.ant.internal.core.ant.InternalAntRunner
WARNING: System::setSecurityManager will be removed in a future release
```

[...]

PR #991, opened 2023-12-12
Unify auto-save configurations

HeikoKlare:
The auto-save configurations are not streamlined across the projects. This change unifies the configuration in the following way:
* Enable auto-formatting for edited lines
* Activate import organization

The change does explicitly not change/unify any additional auto-save actions.

Added configuration parameters (set to `false`) are only due to further options being added to JDT. In case it is preferred to not have these options added to the configuration files but only the entries for auto-formatting and import organization to be changed, I can adapt the PR accordingly. That would affect:
* sp_cleanup.organize_imports=true
* sp_cleanup.format_source_code=true
* sp_cleanup.format_source_code_changes_only=true


akurtakov:
It might be good idea to have one step further and even enable more save actions by default e.g. simplify_lambda* .

HeikoKlare:
> It might be good idea to have one step further and even enable more save actions by default e.g. simplify_lambda* .

Agreed. I would propose to do that in a follow-up PR (to keep changes separate and as it will affect even more plug-ins) or would you propose to incorporate it into this PR?

HeikoKlare:
Maven warnings in Jenkins build are unrelated, see #969.

PR #994, opened 2023-12-12
ProjectPreferences: intern() qualifier

jukzi:
to safe some memory from duplicate strings

image


vogella:
LGTM, @HannesWell WDYT?

HannesWell (approved):
Looks good to me as well.
Since the qualifier is usually the name of plugin that is controlled by the preference the same names occur often.

PR #998, opened 2023-12-12
Fix warning in plugin.xml due to usage of deprecated extension point

HeikoKlare:


jukzi:
thanks for fixing warnings :-)

PR #1001, opened 2023-12-12
Unify additional auto-save actions across projects

HeikoKlare:
Currently, each project configures its own additional auto-save actions (or none at all). This change unifies the enabled actions across all projects.
This includes the default settings applied by JDT tooling with the following derivations:
* `sp_cleanup.use_blocks=false`: Convert control statement bodies to block; disabled since too many files do not yet follow this formatting guideline
* `sp_cleanup.convert_functional_interfaces=true`: Use lambda where possible and simplify expression
* `sp_cleanup.also_simplify_lambda=true`: Simplify lambda expression and method reference syntax
See also: https://github.com/eclipse-platform/eclipse.platform/pull/991#issuecomment-1851633667

The following options are enabled as default by JDT:
* Change non static accesses to static members using declaring type
* Change indirect accesses to static members to direct accesses (accesses through subtypes)
* Convert to enhanced 'for' loops only if the loop variable is used
* Convert to enhanced 'for' loops
* Primitive type rather then wrapper class
* Remove unused imports
* Add missing `@Override` annotations
* Add missing `@Override` annotations to implementations of interface methods
* Add missing `@Deprecated` annotations
* Remove unnecessary casts
* Remove redundant semicolons
* Remove unnecessary '$NON-NLS$' tags
* Remove trailing white spaces on all lines
* Boolean value rather than comparison
* Single 'if' statement rather than duplicate blocks that fall through
* Use diamond operator

@akurtakov What do you think about this configuration? Any settings that you think should be enabled/disabled?
 
Since the amount of changes is hard to review, here is how I produced the changes:
1. Enabled additional actions in every project
2. Reset the additional actions in every project back to default
3. Changed the settings documented above as derivations from default settings via find-and-replace on the `org.eclipse.jdt.core.prefs` files.


akurtakov:
Looks good to me.

HeikoKlare:
Thanks @akurtakov!
I will merge this now. In case there is some annoying setting in the new configuration, e.g., because it changes too much of existing code, I hope that developers experiencing that will report back. I will also have an eye on upcoming PRs to see if the settings may lead to unexpected changes that are unrelated to the actual contents of a PR.

jukzi:
Causes a lot of unrelated changes in https://github.com/eclipse-platform/eclipse.platform/pull/1113

I suggest to format all files once with these refactorings to not confuse with actual changes.

HeikoKlare:
I did not experience many unrelated changes in a single file so far, which is why I did not consider doing a general "cleanup" and touching several files without a specific need yet (and then unnecessarily losing the overwriting the last change information for the modified lines).
But instead of often having unrelated changes, it of course makes more sense to unify all files once. Since this PR only unified the auto-save configurations but not the cleanup configuration, we should probably unify the "cleanup" configuration with the "auto-save" configuration for all projects and then simply execute a cleanup.

jukzi:
> we should probably unify the "cleanup" configuration with the "auto-save" configuration for all projects and then simply execute a cleanup.

yes please

PR #1002, opened 2023-12-12
IllegalStateException: call to beginTask() missing in the Help view

iloveeclipse:
This patch fixes following warning reported to the error log while navigating between help topics in the Help view after https://github.com/eclipse-platform/eclipse.platform.ui/pull/1077:
```
java.lang.IllegalStateException: call to beginTask() missing
	at org.eclipse.jface.action.StatusLineManager$1.worked(StatusLineManager.java:221)
	at org.eclipse.help.ui.internal.views.BrowserPart$2.changed(BrowserPart.java:148)
	at org.eclipse.swt.browser.WebKit.lambda$15(WebKit.java:2492)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
```
To avoid errors reported by the different monitor instances returned by `IStatusLineManager.getProgressMonitor()`, browser listener has to keep a single local monitor instance.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1398

### DIFF
--- a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java
+++ b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java
@@ -125,23 +125,24 @@ public class BrowserPart extends AbstractFormPart implements IHelpPart {
 			}
 		});
 		browser.addProgressListener(new ProgressListener() {
+			IProgressMonitor monitor;
 
 			@Override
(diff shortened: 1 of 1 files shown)


iloveeclipse:
**Real tests** didn't fail, only some warnings that are treated as errors now.
Ignoring errors, looks like they are unavoidable: https://github.com/eclipse-platform/eclipse.platform/issues/969

PR #1005, opened 2023-12-12
Make the "run/debug as..." menus run asynchronously

fedejeanne:
# What it does
Let the `Run`/`Debug As...` (context) menus load asynchronously so that long-running operations don't block the UI. Show a "placeholder" while computing the contents of the menu.

This PR can only be merged together with:
* https://github.com/eclipse-platform/eclipse.platform.ui/pull/1404

# How to test
* Go to a (test) class
* Right click
* Hover over `Run As...` or `Debug As...`
* Move back and forth between both of them, play with it.
* Run/Debug something (_i.e._ click on the sub-menus)

Everything should work as usual

## A small hack to help testing
Add a random wait to `ContextualLaunchAction.isApplicable(...)`, like this:

```java
private boolean isApplicable(LaunchShortcutExtension ext, IEvaluationContext context) throws CoreException {
	Expression expr = ext.getContextualLaunchEnablementExpression();
	try {
		Thread.sleep((long) Math.random() * 200);
	} catch (InterruptedException e) {
		// ignore
	}
	return ext.evalEnablementExpression(context, expr);
}
```

This should give you enough time to see the placeholder in the menu:

image

After a couple of seconds, you should see the "normal" menu items:

image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIMessages.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIMessages.java
@@ -303,6 +303,8 @@ public class DebugUIMessages extends NLS {
 
 	public static String LaunchShortcutAction_combineLaunchShortcutName;
 
+	public static String ContextualLaunchAction_0;
+
 	static {
 		// load message values from bundle file
 		NLS.initializeMessages(BUNDLE_NAME, DebugUIMessages.class);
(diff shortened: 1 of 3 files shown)


mickaelistria:
Would it be possible to have multiple completable future ( 1 for each `LaunchShortcutExtension`) and to show the already resolved ones and the "Loading" element if some ones are remaining and longer to compute?

fedejeanne:
I'll keep that in mind, it should be possible 👍 

For the moment I'm looking at some other options to achieve the same result without moving functionality away from `ActionContributionItem` (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1404) because I think that will have other implications. Any thoughts on that?

fedejeanne:
@mickaelistria I looked into it but I kept running into corner-cases and other difficulties due to the current state of `org.eclipse.debug.ui.actions.ContextualLaunchAction.fillMenu(Menu)`. Would it be acceptable to have it as initially proposed?

fedejeanne:
Test failure is unrelated: #1026

mickaelistria:
This looks OK and is an improvement over current state.

fedejeanne:
@mickaelistria thank you for reviewing and merging. I will then also merge https://github.com/eclipse-platform/eclipse.platform.ui/pull/1404 since it's a pre-requisite for this PR.

mickaelistria:
OK!

PR #1007, opened 2023-12-12
Jenkinsfile: Disable Quality gate for maven warnings #969

jukzi:
should be revert if #969 is fixed
The Problem is, that almost every build fails because the number of warnings is too flaky and independent of the PRs that they should verify. If they would be at least constant they would not fail the PRs.
image


iloveeclipse (approved):
Thanks.

jukzi:
only warnings, no errors => success
image

laeubi:
> The Problem is, that almost every build fails because the number of warnings is too flaky and independent of the PRs that they should verify. If they would be at least constant they would not fail the PRs.

The warnings are about an possible instability in build, therefore they vary dependent on the build execution order (and that's what the warning is complaining about) so it can never be constant.

PR #1015, opened 2023-12-13
DataTree: simplified Path record

jukzi:
Avoid org.eclipse.core.runtime.Path memory overhead in workspace data
structure
The paths in the workspace are independent of OS. They don't need
device, flags, hash or leading "/" root

+ enabled compiler warning unlikelyEqualsArgumentType, because that was
helpful

see https://github.com/eclipse-equinox/equinox/issues/433 why memory reduction is wanted


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/JPath.java
```diff
@@ -0,0 +1,93 @@
...
+ * Contributors:
+ *     Joerg Kubitz - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.internal.dtree;
+
+import java.util.Arrays;
+import java.util.Objects;
+import org.eclipse.core.runtime.IPath;
+
+public record JPath(String[] segments) {
```
laeubi, line 20: Can you please add javadoc that explains why/what this record is used for.

laeubi:
Looking at the code I get the impression one can simply store the raw `String[] segements` instead of wrapping them in another object (that also has some overhead). Beside that, one might want to intern the segment strings before storage as its likely that the same prefix is be used multiple times.

laeubi:
> see https://github.com/eclipse-equinox/equinox/issues/433 why memory reduction is wanted

Your approach / description does not match the one described in the issue, your solution will still incur from the overhead of arrays creation for single element items, I have posted a suggestion there how to save the need for storing a single array element but has a little runtime overhead (_time for memory_).

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/dtree/JPath.java
jukzi: done

jukzi:
> Looking at the code I get the impression one can simply store the raw `String[] segements` instead of wrapping them in another object (that also has some overhead). 

In principle something like that is possible as future improvement. 
to consider.
1. ElementTree.childIDsCache needs an atomic swap
2. Instead of String[] we could store a compressed Object, but the code will look very silly when we pass Object as method parameters
3. That will be target of value objects https://openjdk.org/jeps/401.

> Beside that, one might want to intern the segment strings before storage as its likely that the same prefix is be used multiple times.

That should be done when the String is created: https://github.com/eclipse-equinox/equinox/pull/432 tree data structure only uses segements that have been created by IPath.

jukzi:
BTW the path is stored segment by segment String in AbstractDataTreeNode.name, only the childIDsCache actually stores/caches complete Paths

jukzi:
> BTW the path is stored segment by segment String in AbstractDataTreeNode.name, only the childIDsCache actually stores/caches complete Paths

Which again seems almost only to be used by org.eclipse.core.internal.resources.Container.getChildren(int) which makes me wonder if either that should be converted to use JPath too or if the whole memory problem could be solved another way. - its only childIDsCache that holds much memory.

laeubi:
> only the childIDsCache actually stores/caches complete Paths

Maybe that should be a soft-cache if it is possible to restore the path on other means? Also you say somethign is stored until next workspace save, maybe we should just store the workspace more often or when memmory gets low?

merks:
It's interesting to see exactly the same problems and exactly the same solutions as with EMF's URI crop up here...

szarnekow:
> It's interesting to see exactly the same problems and exactly the same solutions as with EMF's URI crop up here...

Given that many of the paths will of the same segments as URIs in the same workspace, it might be useful to have the segment pool available here, too. What do you think, Ed?

merks:
I expect there will be "compelling" reasons why there cannot be a dependency on org.eclipse.emf.common.  But yes, reusing SegementSequence.create would likely lead to significant improvements.  It has already solved the problem of interning segments, interning the arrays of strings, and maintaining a weak pool for them, all without "synchronized" locking.  But that last peak folks had  at this resulted in "wow, that's so complicated"...  Implementing an analogous solution from scratch is the likely path forward.

jukzi:
closed in favor of https://github.com/eclipse-platform/eclipse.platform/pull/1019

PR #1018, opened 2023-12-14
Fix faulty Javadoc references in o.e.debug.examples.core

HeikoKlare:
Can be seen in every build, e.g.: https://ci.eclipse.org/platform/job/eclipse.platform/job/master/572/consoleText


akurtakov:
Thank you so much for your work on tests and javadoc! It's so encouraging to find better state in this area in the morning compared to the evening!

PR #1019, opened 2023-12-14
[memory] ElementTree: evict childIDsCache

jukzi:
childIDsCache is only used on the current workspace tree, but was hold on the whole history of the workspace

https://github.com/eclipse-equinox/equinox/issues/433


jukzi:
With this change there is only a single ChildIDsCache (ElementTree.childIDsCache) in memory which is much more effective then deduplicating it's values.

mickaelistria:
Thanks, that's a great improvement!

Issue #1017, opened 2023-12-14
Maven checks randomly fail due to Javadoc errors


HeikoKlare:
The quality gate for Maven leads to random failures of PR checks, see, e.g.: https://github.com/eclipse-platform/eclipse.platform/pull/1014/checks?check_run_id=19609932189

This is the case even though the quality gate has been reduced to only fail on errors: https://github.com/eclipse-platform/eclipse.platform/pull/1007

### Analysis

The logs contain several errors because the optional `javax.*` dependencies cannot be resolved in the Javadoc generation. This is, for example, the case for the `org.eclipse.e4.core.contexts` bundle. These errors will, however, not be reported as Javadoc or Maven errors for some reason. The Maven report for `org.eclipse.e4.core.contexts` does not contain them: https://ci.eclipse.org/platform/job/eclipse.platform/job/master/572/maven-warnings/moduleName.53733659/

Sometimes, however, these errors are reported, see e.g.: https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-1014/2/maven-warnings/Error/source.ce60fa94-2a79-4128-a55b-2f2c00681bde/#3357
Then the Maven quality gate does not pass. From what I have seen in the logs, the error seems to be considered if, and only if, it comes right after one of the resolution warnings:
```
17:33:26.600 [INFO] from resources/tests/org.eclipse.core.tests.resources/pom.xml
17:33:26.600 [INFO] ------------------------[ eclipse-test-plugin ]-------------------------
17:33:26.709 [WARNING] Problems resolving provisioning plan.:
Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet [2.1.0,3.0.0).
Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet.http [2.1.0,3.0.0).
Unable to satisfy dependency from org.osgi.service.http.whiteboard 1.1.1.202109301733 to osgi.contract; (&(osgi.contract=JavaServlet)(version=3.1.0)).
17:33:26.805 [ERROR] MavenReportException: Error while generating Javadoc:
Exit code: 1
/home/jenkins/agent/workspace/eclipse.platform_PR-1014/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/contexts/Active.java:31: error: package javax.inject does not exist
```
[...]

PR #1025, opened 2023-12-15
Fix Cut Off Label on Welcome Page

BeckerWdf:
This is the second attempt as the initial fix needed to be reverted.

Fixes: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1210

### DIFF
--- a/ua/org.eclipse.ui.intro.universal/themes/solstice/html/root.css
+++ b/ua/org.eclipse.ui.intro.universal/themes/solstice/html/root.css
@@ -211,10 +211,9 @@ a .link-extra-div, #page-links a .background-image {
 #action-links a#workbench {
     position : absolute;
     right : 20px;
-    top : 5px;
+    top : 0px;
     width : auto;
     text-align:center;
-    margin-bottom : 0px;
 }
 
 #page-links a#workbench .content-img,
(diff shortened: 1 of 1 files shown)


BeckerWdf:
I don't understand the results of the build.
What's wrong here?
Am I doing something wrong?

iloveeclipse:
> I don't understand the results of the build.

Me either, welcome in the club.

> What's wrong here?

No idea.

> Am I doing something wrong?

No, it looks like "quality gate" that seem to thread maven warnings as test failures. Or it is the "quality gate" that don't understands javadoc errors, see #1017. Or it is something else...

In overall, usually people fail to understand what exactly "quality gate" is supposed to say and usually it turns out to be just noise so everyone learns to ignore all the errors reported, which is exact the opposite of what we want.

There was an attempt to mute it for maven in https://github.com/eclipse-platform/eclipse.platform/pull/1007 (after long discussion on #969), but it seem not work reliably, or it is something else that is now need to be muted.

@jukzi : if you know what current build is complaining about, it would be nice if you could mute that too.

jukzi:
complains about ecj warnings, probably because warning have been just enabled https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/cf74f8dd21c95885cd9e4c96837b814f99e5434b
image
That is supposed to settle for in the future.

jukzi:
ignore that ecj quality gate today

BeckerWdf:
> ignore that ecj quality gate today

So should I merge this change?

Issue #744, opened 2023-10-13
Random failing JobTest.testJoinWithCancelingMonitor


HeikoKlare:
Test test `org.eclipse.core.tests.runtime.jobs.JobTest.testJoinWithCancelingMonitor` is randomly failing, e.g., https://github.com/eclipse-platform/eclipse.platform/pull/737/checks?check_run_id=17676558920.

```
junit.framework.AssertionFailedError: 2.0 expected:<4> but was:<2>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.Assert.assertEquals(Assert.java:234)
	at junit.framework.TestCase.assertEquals(TestCase.java:377)
	at org.eclipse.core.tests.runtime.jobs.JobTest.testJoinWithCancelingMonitor(JobTest.java:1004)
```

See #320. #402 for previous, probably related failures.

HeikoKlare:
Still failing randomly in the same state and with the same output: https://github.com/eclipse-platform/eclipse.platform/pull/1021/checks?check_run_id=19707210567

PR #1030, opened 2023-12-18
Fix random failing JobTest#testJoinWithCancelingMonitor #744

HeikoKlare:
The test case JobTest#testJoinWithCancelingMonitor randomly fails because when validating that the used job (longJob) is still running after some event, it is not even guaranteed that the job has started running at all.
This change ensures that longJob is not only scheduled but also running in case the test requires that. This
affects #testJoinWithCancelingMonitor and #testJoinWithTimeout.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/744


HeikoKlare:
Build failure only due to instable Maven quality gate (too many issues have been "fixed" since reference build).

Issue #742, opened 2023-10-13
Tests do no clean up upon exception in setUp()


HeikoKlare:
## Current Situation

If a `setUp()` method of any JUnit 3 test class, i.e., one extending `TestCase` throws an exception, `tearDown()` will not be called. This is by design, as the [`runBare()` method](https://github.com/junit-team/junit4/blob/main/src/main/java/junit/framework/TestCase.java#L138-L153) executes the `setUp()` method outside of the try block that ensures that `tearDown()` is still called upon exceptions during execution of the actual test method.

Several platform tests are still based on JUnit 3 by extending the `CoreTest` class, which is a sublass of `TestCase` and are thus affected by potentially missing cleanup. This breaks test isolation and since all tests of one plug-in are executed on the same workspace, tests executed subsequently to one that failed during `setUp()` will face an unexpected workspace state, making test results unreliable.

## Reproduction

If you want, you can easily reproduce the behavior:

```java
public class DemoTest extends CoreTest {
	@Override
	protected void setUp() throws Exception {
		throw new IllegalStateException();
	}

	public void test() {
		//
	}

	@Override
	protected void tearDown() throws Exception {
		System.out.println("I was called");
	}
}
```

## Proposal

Until all affected test class are migrated to JUnit 4 / 5 (which ensure that @After methods [are always called, even if @Before methods throw an exception](https://junit.org/junit4/javadoc/4.12/org/junit/After.html)), I propose to adapt the behavior of the `CoreTest` class to call `setUp()` within the try block to improve test isolation by re-implementing the `runBare()` method.

vogella:
+1 to the proposal

Heiko Klare ***@***.***> schrieb am Fr., 13. Okt. 2023, 12:42:

HeikoKlare:
Since not all remaining JUnit 3 tests are subclasses of `CoreTest`, I will first try to migrate as many tests as possible to JUnit 4. I hope that most of the remaining JUnit 3 tests then extend `CoreTest`, such that they are affected by the workaround.

I will provide a PR as soon as I have taken a look at all remaining JUnit 3 tests for the possibility of an easy migration to JUnit 4.

HeikoKlare:
Since now all specializations of `CoreTest` have been migrated to JUnit 4 or otherwise now specialize the general `TestCase` if they are still part of a framework relying on JUnit 3 (such as the session tests), it is not necessary anymore to improve the cleanup in the `CoreTest` class.

[...]

PR #1042, opened 2023-12-18
Deprecate CoreTest class #742

HeikoKlare:
The `CoreTest` class of the Eclipse test harness was used as the base class for many JUnit 3-based tests. Since all its subclasses and other usages have been removed and new tests should not use JUnit 3 anymore, the `CoreTest` class is deprecated for future removal.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/CoreTest.java
+++ b/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/CoreTest.java
@@ -37,7 +37,12 @@ import org.junit.Assume;
 
 /**
  * @since 3.1
+ *
+ * @deprecated Use JUnit 4/5 and their assertions instead, and use utilities
+ *             provided via utility classes such as {@link FileSystemHelper}
+ *             instead of accessing utilities via inheritance.
  */
+@Deprecated(forRemoval = true, since = "2024-03")
 public class CoreTest extends TestCase {
 	/** counter for generating unique random file system locations */
 	protected static int nextLocationCounter = 0;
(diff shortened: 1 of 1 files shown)


vogella:
LGTM

HeikoKlare:
> LGTM

Thanks, @vogella. Do / should we add a notice of (this kind of) deprecation to the N&N?

akurtakov:
These are test classes/helpers so there is no API promise for them. I would say that there is no need for deprecation and they can be removed as soon as not used.
Definetely, this is something that shouldn't be in N&N IMO.

HeikoKlare:
I would be in favor of completely removing the class instead of deprecating it (as it is not used "by us" anymore). But the test harness (including the `CoreTest`) is part of the `org.eclipse.test` feature, which is deployed to update sites and thus universally usable. That's why I considered it to be public API with an according promise. Is there a reason for treating this plug-in/feature differently w.r.t. versioning / deprecation, @akurtakov?

akurtakov:
o.e.test feature is there only for the sake of running tests for each I-build, nothing else. API promises are for content of eclipse-SDK-*-win32-x86_64.zip .

HeikoKlare:
Thanks for that clarification! Then I do not see any reason for deprecating the class and will replace this PR with one directly removing it.
I do not want to stress this topic too much, but I found the info that we make an API promise only for simrel contributions very helpful (and very reasonable). Is that explicitly documented (and I simply missed the information), e.g., in the wiki, or is it rather "implicit knowledge". In the latter case I would be in favor of having it documented (maybe along with information about deprecation policies) at some visible place (e.g. related to information about versioning / API baseline / evolution).

vogella:
I should change the MANIFEST.MF to export its package as x-internal, maybe with some friends to avoid warnings in our test plug-ins, to make sure it is not considered as API

HeikoKlare:
> I should change the MANIFEST.MF to export its package as x-internal, maybe with some friends to avoid warnings in our test plug-ins, to make sure it is not considered as API

We should then do this for all plug-ins that are not deployed in simrel, e.g., `org.eclipse.test`, `org.eclipse.test.performance` (and maybe even non-test-related plug-ins?)

PR #1047, opened 2023-12-19
Remove CoreTest class #742

HeikoKlare:
The `CoreTest` class of the Eclipse test harness was used as the base class for many JUnit 3-based tests. Since all its subclasses and other usages have been removed and new tests should not use JUnit 3 anymore, the `CoreTest` class is removed.

Supercedes the deprecation as proposed in #1042. Removal valid according to https://github.com/eclipse-platform/eclipse.platform/pull/1042#issuecomment-1862218333.
Closes #1042.


vogella:
LGTM

HeikoKlare:
Fail is unrelated, documented in #1057, and also occurs in other PRs such as #1056.

PR #1048, opened 2023-12-19
Add missing dependency and import for Javadoc references

HeikoKlare:
See #1017

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.di.extensions/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.e4.core.di.extensions/META-INF/MANIFEST.MF
@@ -9,5 +9,6 @@ Bundle-RequiredExecutionEnvironment: JavaSE-17
 Export-Package: org.eclipse.e4.core.di.extensions;version="0.16.0"
 Bundle-Localization: fragment
 Import-Package: jakarta.inject;version="[2.0.0,3.0.0)",
- javax.inject;version="[1.0.0,2.0.0)";resolution:=optional
+ javax.inject;version="[1.0.0,2.0.0)";resolution:=optional,
+ org.osgi.framework;version="[1.0.0,2.0.0)"
 Automatic-Module-Name: org.eclipse.e4.core.di.extensions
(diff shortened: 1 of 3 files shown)


runtime/bundles/org.eclipse.e4.core.di.extensions/META-INF/MANIFEST.MF
```diff
@@ -10,4 +10,5 @@ Export-Package: org.eclipse.e4.core.di.extensions;version="0.16.0"
 Bundle-Localization: fragment
 Import-Package: jakarta.inject;version="[2.0.0,3.0.0)",
  javax.inject;version="[1.0.0,2.0.0)";resolution:=optional
+Require-Bundle: org.eclipse.osgi
```
laeubi, line 13: please use import package!

laeubi (changes requested):

runtime/bundles/org.eclipse.e4.core.di.extensions/META-INF/MANIFEST.MF
HeikoKlare: corrected in fbbe34a

laeubi (approved):

runtime/bundles/org.eclipse.e4.core.di.extensions/META-INF/MANIFEST.MF
```diff
@@ -9,5 +9,6 @@ Bundle-RequiredExecutionEnvironment: JavaSE-17
 Export-Package: org.eclipse.e4.core.di.extensions;version="0.16.0"
 Bundle-Localization: fragment
 Import-Package: jakarta.inject;version="[2.0.0,3.0.0)",
- javax.inject;version="[1.0.0,2.0.0)";resolution:=optional
+ javax.inject;version="[1.0.0,2.0.0)";resolution:=optional,
+ org.osgi.framework;version="[1.0.0,2.0.0)"
```
HannesWell, line 13: In order to not loose this dependency on the next dependency clean-up I suggest you replace the fully qualified class-names in `org.eclipse.e4.core.di.extensions.OSGiBundle` by regularly imported classes.
Then removing it would lead to a compiler error.

HeikoKlare: Makes sense, thanks! Fixed with 3681b9b.

HannesWell (approved):
Looks good. Thank you. :)

PR #1052, opened 2023-12-19
Evolving java based apis part3

vogella:


iloveeclipse:
@vogella : I see you have deleted content on old wiki pages before even finish with transition. So now the old content is gone and the new one is broken, as most of the links on https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md are broken.

Can you please restore the old pages so that one can compare before/after state and please do not delete more content before the migration is complete.

vogella:
> @vogella : I see you have deleted content on old wiki pages before even finish with transition. So now the old content is gone and the new one is broken, as most of the links on https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md are broken.
> 
> Can you please restore the old pages so that one can compare before/after state and please do not delete more content before the migration is complete.

Restored https://wiki.eclipse.org/Version_Numbering with additional link to the new page

vogella:
> @vogella : cross references do not work on all new pages. Can you please look into that?

See the commit messages: It does currently not fix the links which are pointing still to the
wiki, once the other documents have been migrated these links will be
adjusted.

PR #1053, opened 2023-12-19
[E4] Remove unnecessary javax.inject.Qualifier meta annotations

HannesWell:
The E4-Injector considers both the `javax.inject.Qualifier` and `jakarta.inject.Qualifier`. Therefore specifying both is not necessary and just specifying the latter is sufficient.

The only I reason I can think of where it would be necessary to keep the annotation would be third-party tools that process the E4 annotations, but I'm not aware of any and they could just implement the same approach as the E4 injector to handle both.

This should also help for https://github.com/eclipse-platform/eclipse.platform/issues/1017.

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.e4.core.contexts/META-INF/MANIFEST.MF
@@ -9,7 +9,6 @@ Bundle-ActivationPolicy: lazy
 Require-Bundle: org.eclipse.e4.core.di
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Import-Package: jakarta.inject;version="[2.0.0,3.0.0)",
- javax.inject;version="[1.0.0,2.0.0)";resolution:=optional,
  org.osgi.framework;version="[1.5.0,2.0.0)",
  org.osgi.service.event;version="[1.3.0,2.0.0)"
 Export-Package: org.eclipse.e4.core.contexts;version="1.7.0",
--- a/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/contexts/Active.java
+++ b/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/contexts/Active.java
@@ -23,12 +23,10 @@ import java.lang.annotation.Target;
  * This annotation can be added to injectable fields ands methods to indicate
  * that the injected value should come from the active context.
  *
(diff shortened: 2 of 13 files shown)


HannesWell:
If we merge this https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1680 is maybe obsolete.
@laeubi or is it still required for the injector bundle itself?

laeubi:
> If we merge this [eclipse-platform/eclipse.platform.releng.aggregator#1680](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1680) is maybe obsolete. @laeubi or is it still required for the injector bundle itself?

I would rather keep this as it is not a runtime requirement and has prevented some javadoc confusions.

HeikoKlare:
I think with this, the dependencies to javax.annotation and javax.inject for javadoc generation should only be necessary for the org.eclipse.e4.core.javax.tests bundle. To keep the parent pom clean, it should be possible to remove the additional dependencies there and either only add the additional dependencies for the maven javadoc plugin in that test bundle or completely disable javadoc generation for test bundles, as already proposed by @laeubi.

laeubi:
I won't spend any time in such "micro optimizing" even though its maybe possible, but javadoc generation happens on different levels in the large and and the small scale so I usually prefer manged configuration as it does not really harm anything here.

HannesWell:
> > If we merge this [eclipse-platform/eclipse.platform.releng.aggregator#1680](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1680) is maybe obsolete. @laeubi or is it still required for the injector bundle itself?
> 
> I would rather keep this as it is not a runtime requirement and has prevented some javadoc confusions.

Acknowledge. Although I agree with Heiko that keeping the poms clean is good in general, in this specific case I also agree with Christoph it is not worth the effort.

HannesWell:
Does anybody want to comment about the change itself or does it look all good to you?

HeikoKlare (approved):
Looks good. Thank you!

HeikoKlare:
@HannesWell Sorry that i have produced a conflict by merging #1048 🙁

HannesWell:
> @HannesWell Sorry that i have produced a conflict by merging #1048 🙁

No problem. I was expecting that when I saw your change 🙂 
I'll resolve it tonight. Or since edits by maintainers are permitted you could also do it after fetching this PR and then push it to the branch of this PR at my fork (you just have to add it as new remote). Just in case you have to not yet been aware of this possibility and want to try this out.

PR #1054, opened 2023-12-19
Remove o.e.osgi.services from infocenter.product

HannesWell:
This hopefully helps with the failing platform build.


HannesWell:
At least locally the platform-build succeeded with this.
I think this is a consequence of https://github.com/eclipse-equinox/equinox/pull/403, but I'm not sure why this failed the build exactly since I didn't get more information (logs etc.) from that build.

Removing o.e.osgi.services should have no drawbacks since transitive dependencies should be installed automatically and the replacement of the former content of o.e.osgi.services is required by the `org.eclipse.equinox.http.servlet` and `org.eclipse.equinox.http.registry` bundle.

PR #1055, opened 2023-12-20
Update the build config to eager and non parallel execution

laeubi:
Parallel execution can speedup the build (as well as lazy resolve) but seem to "confuse everyone except me", so this is to disable both features for the sake of contributor confidence.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,5 +1,2 @@
 -Pbuild-individual-bundles
---threads
-1C
---builder
-smart
+-Dtycho.target.eager=true


HeikoKlare:
Just for understanding: why should parallel build be disabled (i.e., how it does confuse)? Is that argument related to the linear log output mixing up parallel executions?
At least I would be in favor of keeping parallel build in case it does not produce severe problems, as it speeds up the build (at least slightly, since speedup is limited by single test projects taking quite some time for sequential execution). The only relevant issue with parallel build I have seen so far was #1017, but that should now be resolved.

laeubi:
> Is that argument related to the linear log output mixing up parallel executions?

Mostly yes, also it seems to not speed up things much here, this build is 20 minutes while I see similar measures in the history of master branch and the attempt to go for parallel builds was never catched up by anyone except me.
A user can still enable it for a local build.

HeikoKlare:
Okay, let's see if/how the build times change. I find most interest not the overall time for checks (mainly driven by the Mac and Jenkins build) but the one for the GitHub Actions build for Linux, since that is one that usually finishes first and can be used for fast feedback.

laeubi:
Linux build is 13 minutes, maybe because it has more CPU+RAM than our Jenkins nodes, as said I'm all for parallel builds (what also would include having parallel test executions), but as this just produces bad user feedback and waste time with discussions given that the Jenkins node don't have the power to have more benefits from that I'd rather no longer want to take the burden to convince people its a good idea.

Bananeweizen:
@laeubi At work I have reduced the confusion od parallel output a bit by setting extremely many loggers of the different maven plugins to warning level. That way goals still show up in the maven console, but the typical "copied 4 resources" and similar output not of interest are gone (and therefore not mixed). Maybe that would be an alternative here too?

laeubi:
> At work I have reduced the confusion od parallel output a bit by setting extremely many loggers of the different maven plugins

The problem is that I don't see anyone is particularly interested in configuring maven loggers, maybe one can even configure the current project/threadname to be printed like the timestamps.

PR #1058, opened 2023-12-21
Remove obsolete <resolver>p2</resolver> parameter and managed version

laeubi:


HeikoKlare (approved):
👍

Bananeweizen:
@laeubi can you say when that resolver argument became obsolete (Tycho 3, Tycho 4)? I have been removing this from several other projects also in the past, but I'm not exactly clear which version of Tycho qualifies for it. I basically always relied on the POM editor telling me it's invalid.

laeubi:
> can you say when that resolver argument became obsolete (Tycho 3, Tycho 4)

I can only tell that one can't really choose the resolver anyways so it was always `p2` so it must be somewhere before Tycho 1.x or something.

PR #1064, opened 2023-12-21
API removal process migrated from the

vogella:
https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines

https://wiki.eclipse.org/Eclipse/API_Central/API_Removal_Process was originally migrated to https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines, I think it is better stored in the repo as mark-down document


vogella:
@HeikoKlare here we could add the info that test code is not API. I push a change

vogella:
Once this is in I also need to update https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines pointing to here

PR #1069, opened 2024-01-04
Improve array length assertions in resource tests

HeikoKlare:
In order to reduce/avoid usage of ambiguous `assertEquals` (hard to distinguish expected/actual) and prepare for a migration to JUnit 5 with swapped expected/actual parameters in the `assertEquals` signature, this change replaces certain `assertEquals` calls with AssertJ assertions:
* Replace `assertEquals(MESSAGE, EXPECTED, ACTUAL.length)` with
`assertThat(ACTUAL).describedAs(MESSAGE).hasSize(EXPECTED))` and remove obsolete message where possible
* Replace `assertEquals(EXPECTED, ACTUAL.length)` with `assertThat(ACTUAL).hasSize(EXPECTED))`
* Replace `assertEquals(EXPECTED, ACTUAL.length)` with `assertThat(ACTUAL.length, is(EXPECTED))` if `ACTUAL` is a primitive type array
* Replace `hasSize(0)` with `isEmpty()` to improve failure messages
* Combine array size comparison with array contents comparison where possible

### What I Did

Initial changes are the result of two regex replacements:
#### Length assertions with message
Find: `assertEquals\((.*), (.*), (.*)\.length\);`
Insert: `assertThat\($3\)\.as\($1\)\.hasSize\($2)\);`

#### Length assertions without message
Find: `assertEquals\((.*), (.*)\.length\);`
Insert: `assertThat\($2\)\.hasSize\($1)\);`

Further improvements, in particular the combination of size and content assertions, is manual work.

### Benefits
* Improved failure messages:
  * When an empty array is asserted, the non-empty array will be reported in case of a failure
  * When size and content of an array are validated, the test will not fail reporting an improper size but reporting the mismatching elements
* Improved comprehensibility due to matchers giving more semantics to the validations than plain assertEquals
* Eased migration to JUnit 5: `assertEquals` statements would need to adapted during a migration from JUnit 4 to 5 because expected/actual parameters are swapped

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/FilteredResourceTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/FilteredResourceTest.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.core.tests.resources;
 
+import static java.util.function.Predicate.not;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
 import static org.eclipse.core.tests.harness.FileSystemHelper.getRandomLocation;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createInFileSystem;
@@ -20,10 +22,7 @@ import static org.eclipse.core.tests.resources.ResourceTestUtil.createInWorkspac
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createTestMonitor;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createUniqueString;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.removeFromWorkspace;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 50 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/FilteredResourceTest.java
```diff
@@ -429,31 +431,35 @@ public void testCreateFilterOnLinkedFolderAndTarget2() throws Exception {
 		// // Refreshing restores sanity (hides the .cpp files)...
 		// folder.refreshLocal(IResource.DEPTH_INFINITE, getMonitor());
 		members = folder.members();
-		assertEquals("4.3", 1, members.length);
-		assertEquals("4.4", IResource.FILE, members[0].getType());
-		assertEquals("4.5", "foo.h", members[0].getName());
+		assertThat(members).hasSize(1).allSatisfy(member -> {
```
Bananeweizen, line 434: for the specific case of asserting 1 element and then checking it's properties assertThat(collection).singleElement() can sometimes be helpful to shorten several lines to one. E.g. you can get from list to element to property via assertThat(collection).singleElement().extracting(fieldOfTheElement).isEqualTo(foo). Not sure if that is of any help here when you have multiple properties to be extracted. :)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/session/TestSave.java
```diff
@@ -40,13 +42,14 @@ public void test1() throws CoreException {
...
 		IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
-		assertTrue("1.0", root.exists());
+		assertThat(root.exists()).isTrue();
 		IResource[] children = root.members();
-		assertEquals("1.2", 1, children.length);
-		IProject project = (IProject) children[0];
-		assertTrue("1.3", project.exists());
-		assertTrue("1.4", project.isOpen());
-		assertEquals("1.5", PROJECT, project.getName());
+		assertThat(children).singleElement().asInstanceOf(InstanceOfAssertFactories.type(IProject.class))
```
Bananeweizen, line 47: ah, I see you already found singleElement()...

Issue #1075, opened 2024-01-05
Silently failing LaunchConfigurationTabGroupViewerTest#testOnlyDefaultTabInOtherConfigIsActivated


HeikoKlare:
Two test cases added with #860 are broken:
* LaunchConfigurationTabGroupViewerTest#testOnlyDefaultTabInOtherConfigIsActivated()
* LaunchConfigurationTabGroupViewerTest#testOtherTabInOtherConfigIsActivated()

They make false use of the `not` matcher from hamcrest, e.g.:
https://github.com/eclipse-platform/eclipse.platform/blob/ef98ac465fff52d1081a1f66735c749c650ddbff/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/ui/LaunchConfigurationTabGroupViewerTest.java#L133

The test class erroneously imports AssertJ's `assertThat` to which passing a matcher object is valid (other than to Hamcrests `assertThat`), but AssertJ's `assertThat` does nothing if no actual assert method is called afterwards (such as `isTrue()` for a boolean assertion). Thus, the two assertions using `not` within this class do actually not assert anything.

In addition, one of the assertions, namely the one in `LaunchConfigurationTabGroupViewerTest#testOnlyDefaultTabInOtherConfigIsActivated()`, fails if it is corrected. The tabs are initialized although they are assumed to not have been initialized. So there is either a flaw in the test or some bug in the tested code.

A correction could look as follows:
```java
assertThat((((SpyTab) tabs[i]).isInitialized())).describedAs("Tab " + i + " should not have been initialized").isFalse();
```

PR #1076, opened 2024-01-05
Replace Hamcrest assertions with AssertJ assertions

HeikoKlare:
Comprehensive assertions can be defined with different libraries, of which Hamcrest and AssertJ are two of the most popular ones. While Platform tests currently only use Hamcrest at some places, AssertJ has several advantages including:
* Easier to apply and learn due to fluent API rather than matcher providing comprehensive code completion support
* Higher expressiveness by easily chaining multiple assertions on the same object
* Better comprehensibility by chaining instead of nesting calls and avoiding the necessity to start an `assertThat` statement with a custom message if desired

This change migrates the existing usages of Hamcrest to AssertJ (except for one custom Matcher implementation). Doing that it also makes use of String formatting rather than String concatenation with a potential slight performance improvement.

See also the discussion eclipse-platform/.github#177


HeikoKlare:
> This pull request skips 1 test.
> ```
> org.eclipse.debug.tests.ui.LaunchConfigurationTabGroupViewerTest ‑ testOnlyDefaultTabInOtherConfigIsActivated
> ```

This PR corrects a faulty assertion in `LaunchConfigurationTabGroupViewerTest#testOnlyDefaultTabInOtherConfigIsActivated()`, which requires it's temporary disablement as the test fails with the correct assertion, see #1075.

Bananeweizen:
Just one more recommendation: It's perfectly fine to convert existing tests with error message towards AssertJ as(errorMessage) calls. However, that is not the best way. as("something") just changes the description of the unit under test, but the original error message remains.
`assertThat(foo()).as("very long error message").isTrue()` will still produce "Expecting 'very long error message' to be true, but it was false". It just changes the _name_ or _description_. `assertThat(foo).withFailMessage(()->"my custom message").isTrue()` would really replace the complete message. Now you might be wondering why does the "as" method exist in the first place? Take this example:
```
assertThat(familyMember.get(0)).as("father").has...
assertThat(familyMember.get(1)).as("mother").has...
```
A failure in that test will be written "Expecting 'father' to have..."

And let me repeat: Your conversion of the existing tests is fine, no need to change it there. It just allows even more nice error messages in new tests.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/ParallelBuildChainTest.java
```diff
@@ -228,9 +222,10 @@ public void testWorkspaceBuild_ConflictingRule() throws Exception {
 		executeIncrementalWorkspaceBuild(() -> {
 			waitForCondition(() -> TimerBuilder.getStartedProjectBuilds().size() > 1,
 					millisToWaitForUnexpectedParallelBuild);
-			assertThat(
-					"all build jobs have started in time although infinitely running builds with conflicting rules exist",
-					TimerBuilder.getStartedProjectBuilds(), not(containsInAnyOrder(longRunningProjects)));
+			assertThat(longRunningProjects).as(
+					"all build jobs have started in time although infinitely running builds with conflicting rules exist")
+					.anySatisfy(longRunningProject -> assertThat(TimerBuilder.getStartedProjectBuilds())
+							.doesNotContain(longRunningProject));
```
Bananeweizen, line 228:
```suggestion
			assertThat(longRunningProjects)
            assertThat(TimerBuilder.getStartedProjectBuilds())
                            .withFailMessage("all build jobs have started in time although infinitely running builds with conflicting rules exist")
							.doesNotContainAnyOf(longRunningProjects);
```

HeikoKlare: Thank you for your feedback, @Bananeweizen!
I also thought some time about the `as(message)` calls, in particular because some of the existing failure messages do not really make sense as a description for the assertion. Many of them would totally make sense when passed to `withFailureMessage(message)`, but then the failure report will _only_ contain that message and nothing about the expected/actual value. In many cases it can be helpful to have more detailed information about the actual and expected values. However, for pure boolean assertions it totally makes sense to simply use `withFailureMessage()`, which is why I will update the PR according.

Regarding your concrete proposal for one assertion: the proposed assertion is semantically different from the original one. The original one checks that "not all long running project builds have started". The proposed on checks that "none of the long running project builds have started". This assertion looked better with Hamcrest, where you can combine matchers to: `assertThat(TimerBuilder.getStartedProjectBuilds(), not(containsInAnyOrder(longRunningProjects)))`. Unfortunately, I did not find an according assertion in AssertJ, which is why I made this strange construct of assertions realizing "there is a long running project build that has not started". If you know a better way to express that, I'd be happy if you can let me know :-)

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/usecase/IFileTest.java
```diff
@@ -121,45 +116,37 @@ public void testFile() throws CoreException {
...
-		assertThat("no member with name '" + file.getName() + "' exists in folder: " + folder,
-				folder.findMember(file.getName()).exists());
-		assertThat(workspace.getRoot().findMember(file.getFullPath()), is(file));
-		assertThat("no member at path '" + file.getFullPath() + "' exists in workspace",
-				workspace.getRoot().exists(file.getFullPath()));
-		assertThat(file.getLocation(), is(absolutePath));
+		assertThat(file).matches(IFile::exists, "exists")
+				.matches(it -> workspace.getRoot().exists(it.getFullPath()), "is contained in workspace")
+				.matches(it -> folder.findMember(it.getName()).exists(), "is contained in folder: " + folder)
+				.isEqualTo(workspace.getRoot().findMember(file.getFullPath()));
```
HeikoKlare, line 122: I have now also replaced several pure boolean assertions with `matches` assertions that improve the output in case of a failure, in particular because they also output the subject under test. This makes it far easier to provide a helpful error message with minimal custom text.

Bananeweizen: That's an interesting technique that I will remember, thanks for that hint. Regarding earlier discussions: You are completely right about as() and withFailureMessage() sometimes not fitting well, therefore in practice it's sometimes more ugly than the nice theory I painted before. Personally I try to use as() as much as possible, if the assertion itself has helpful text (like isNonEmptyDirectory), but switch to withFailMessage, if the default error message is to generic (like isTrue). Since withFailMessage can take format strings and objects, that's also okay performance wise. However, I'll probably use more of your matches(...) version in the future, as that probably simplifies custom error messages a bit.

Bananeweizen:
👍🏻 LGTM

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/usecase/IFileTest.java
HeikoKlare: As a side note: a method reference used for matching can even be negated easily with `Predication.not()`, which, when statically imported, can be used like this: `assertThat(file).matches(not(IFile::exists), "does not exist")`.

HeikoKlare:
Thanks for your review and all your comments, @Bananeweizen! Feels like this PR was worth the effort as it improves some validations even semantically, and particularly because I I've learned quite some things about validations / AssertJ.

PR #1077, opened 2024-01-07
Properly initialize session tests and test name before logging it

HeikoKlare:
Currently, WorkspaceSessionTests log their name in their setup logic
before actually setting the name, such that "[null] setUp" is always
logged. In addition, two WorkspaceSessionTest implementations overwrite
the general setup logic without a `super` call to the general
initialization. This change corrects the setup statements order and
ensures that all session tests call the general initialization logic.


HeikoKlare:
Before:
image

After:
image

Issue #1084, opened 2024-01-11
Compute launch button tooltip Job logs error because it did not find UI thread to get selection.


raghucssit:
This issue was discovered in workspace logs of our application.

## Problem

LaunchingResourceManager tries to copute tooltip for launch configuration button using a Job which is scheduled by a mouse listener.
This Job needs selection to show correct tooltip for the launch configuration button.
`Selection` is determined by `SelectedResourceManager#getCurrentSelection()` which logs time out error if it did not find UI within 10 secs. 
Sometimes for any reasons UI can be busy for more than 10 secs.

## Log
`!ENTRY org.eclipse.debug.ui 4 4 2023-10-19 21:13:44.682
!MESSAGE To avoid deadlock while executing Display.syncExec() from a non UI thread 'Worker-80: Compute launch button tooltip', operation was cancelled.
!SUBENTRY 1 org.eclipse.debug.ui 4 120 2023-10-19 21:13:44.682
!MESSAGE Worker-80: Compute launch button tooltip thread probably holding a lock and trying to acquire UI lock
!STACK 0
java.lang.IllegalStateException: Call stack for thread Worker-80: Compute launch button tooltip
    at java.management@11.0.10/sun.management.ThreadImpl.dumpThreads0(Native Method)
    at java.management@11.0.10/sun.management.ThreadImpl.getThreadInfo(ThreadImpl.java:485)
    at org.eclipse.debug.internal.ui.stringsubstitution.SelectedResourceManager.reportTimeout(SelectedResourceManager.java:297)
    at org.eclipse.debug.internal.ui.stringsubstitution.SelectedResourceManager.getFromUI(SelectedResourceManager.java:243)
    at org.eclipse.debug.internal.ui.stringsubstitution.SelectedResourceManager.getCurrentSelection(SelectedResourceManager.java:91)
    at org.eclipse.debug.internal.ui.contextlaunching.LaunchingResourceManager.computeLabels(LaunchingResourceManager.java:236)
    at org.eclipse.debug.internal.ui.contextlaunching.LaunchingResourceManager$1$1.run(LaunchingResourceManager.java:145)
    at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)`

PR #1085, opened 2024-01-11
Move selection calculation out of the Job that calculates tool-tip for launch config button.

raghucssit:
SelectedResourceManager#getCurrentSelection() needs UI thread to determine selection. Job sometimes don't get access to UI for more than 10 secs. So it is good to move selection compute out of the job and pass the selection to job.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1084


iloveeclipse (approved):
LGTM

PR #1086, opened 2024-01-11
Provide a session test framework for JUnit 5

HeikoKlare:
The existing implementation of session tests (including SessionTestRunner, SessionTestSuite and WorkspaceSessionTestSuite) relies on JUnit 3 functionality. Because of the changed way in which tests are set up in JUnit 5, they cannot be used to run session tests with JUnit 5. In addition the session test framework is currently overly complex, as it provides some functionality that is not used (anymore).

This commit introduces a new JUnit 5 extension, the SessionTestExtension, to be used for running test cases of a class in
separate Eclipse application sessions with JUnit 5. This extension intercepts test methods calls to start a separate Eclipse session in which the tests are executed. The results are evaluated in the intercepted method execution in way they were evaluated for existing JUnit-3-based session tests. Session that are expected to fail can be annotated with an added @SessionShouldFail annotation.

Functionality is demonstrated in a manually executed sample session test.

<s>Functionality is demonstrated in a manually executed sample session test as well as by application to the session test in the runtime tests project. Since JUnit 5 test suites are not yet supported, the project is adapted to execute tests without suites.</s> 
-> resolved via eclipse-tycho/tycho#3358

<s>In order to get the session test running with Tycho, an adaptation of the test application provided via PDE is necessary, because a faulty class loader is currently used for the test engines during remote execution.</s>
-> resolved via eclipse-pde/eclipse.pde#1047

Contributes to:
- #903

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/RemoteTestExecutor.java
+++ b/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/RemoteTestExecutor.java
@@ -1,0 +1,299 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.core.tests.harness.session;
+
+import static org.eclipse.core.tests.harness.TestHarnessPlugin.PI_HARNESS;
+import static org.eclipse.core.tests.harness.TestHarnessPlugin.log;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.ServerSocket;
+import java.net.Socket;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.AssertionFailedException;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.tests.session.RemoteAssertionFailedError;
+import org.eclipse.core.tests.session.RemoteTestException;
+import org.eclipse.core.tests.session.Setup;
+import org.eclipse.core.tests.session.TestDescriptor;
+
+class RemoteTestExecutor {
+	private static final String REMOTE_EXECUTION_INDICATION_SYSTEM_PROPERY = "org.eclipse.core.tests.session.isRemoteExecution"; //$NON-NLS-1$
+
+	/**
+	 * {@return whether the current test execution is done remotely}
+	 */
+	static boolean isRemoteExecution() {
+		return System.getProperty(REMOTE_EXECUTION_INDICATION_SYSTEM_PROPERY, "").equals(Boolean.toString(true));
+	}
+
+	private final String pluginId;
+	private final String applicationId;
+	private final Setup setup;
+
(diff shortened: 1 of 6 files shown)


HeikoKlare:
The migration of session tests to JUnit 5 requires several dependencies to be added to `org.eclipse.core.tests.harness`, as they have to be provided to `ConfigurationSessionTestSuite` to be able to run with JUnit 5. I am not in favor of having all these dependencies in the test harness bundle, but since the only alternative I see would be a separate bundle for the `ConfigurationSessionTestSuite` and the test harness bundle is only used internally anway, I think that adding these dependencies is acceptable. If anyone objects or has a proposal for improvement, please let me know. Otherwise I would like to merge this soon to be able to proceed the migration away from JUnit 3.

runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/RemoteTestExecutor.java
```diff
@@ -0,0 +1,299 @@
...
+		localSetup.setSystemProperty(REMOTE_EXECUTION_INDICATION_SYSTEM_PROPERY, Boolean.toString(true));
+
+		TestDescriptor descriptor = new TestDescriptor(testClass, testMethod);
+		ResultCollector collector = new ResultCollector(getTestId(testClass, testMethod));
+		localSetup.setEclipseArgument("port", Integer.toString(collector.getPort()));
+		new Thread(collector, "Test result collector").start();
+		IStatus status = launch(localSetup);
+		collector.shutdown();
+		if (!shouldFail) {
+			if (!status.isOK()) {
```
fedejeanne, line 63: This would also throw an exception if the test exited with a warning. Wouldn't it be better to ask `if (status.isError())` to allow for warnings here?

HeikoKlare: Note that the `IStatus` object does not report about test results but about the result of the remote execution process. The warning state is used here to indicate that the process returned a non-zero code. This is how session tests have been implemented (in the `ResultCollector`, which is taken over from the existing implementation). Iit is nothing that is new to the JUnit-5-way of running them.

fedejeanne: You're right, thanks for the clarification: the way it's written is misleading at first but in the code I can see that the exit code comes from:
```java
org.eclipse.core.tests.session.Setup.run()
```
... which in turn, comes from: 
```java
/**
* ...
* @return the process exit value
*/
org.eclipse.core.tests.session.ProcessController.execute()
```
So a "warning" status is actually a non-zero return --> error

HeikoKlare:
PR is _not_ ready to be merged because of faulty JUnit test result aggregation. GitHub reports 102 removed tests, which is because (existing, JUnit-3-based) session tests are not processed correctly when executed with a JUnit 5 runner. The test methods are not assigned to the their test class but to the running test suite, and since many session test methods are simply named `test1`, `test2` etc., they "overwrite" each other in the logs when assigned to the same test suite. This particularly applies to the `AllSessionTests` suite, e.g.:
```
AutomatedResourceTests org.eclipse.core.tests.resources.session.AllSessionTests - test1
```
This subsumes multiple executions of a `test1` method of all session test classes in `AllSessionTests`.

HeikoKlare:
> PR is _not_ ready to be merged because of faulty JUnit test result aggregation. GitHub reports 102 removed tests, which is because (existing, JUnit-3-based) session tests are not processed correctly when executed with a JUnit 5 runner. The test methods are not assigned to the their test class but to the running test suite, and since many session test methods are simply named `test1`, `test2` etc., they "overwrite" each other in the logs when assigned to the same test suite.

Has been fixed via #1334 and #1335.

HeikoKlare:
This PR particularly affects the `ConfigurationSessionTest` class, which is why I checked all users of this class in the Eclipse SDK. I found all other users of that class to be in the Equinox repository. I executed those tests to validate that they all still run successfully.

image

Issue #1087, opened 2024-01-11
Failing org.eclipse.core.tests.internal.preferences.PreferencesServiceTest#testLookupOrder


HeikoKlare:
`org.eclipse.core.tests.internal.preferences.PreferencesServiceTest#testLookupOrder()` is failing, probably because of the `user` scope recently added in Equinox: eclipse-equinox/equinox#446

See for example: https://github.com/eclipse-platform/eclipse.platform/actions/runs/7492737383/job/20398029585

```
1.2: array lengths differed, expected.length=4 actual.length=5; arrays first differed at element [3]; expected:<[default]> but was:<[user]>
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:78)
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:28)
	at org.junit.Assert.internalArrayEquals(Assert.java:534)
	at org.junit.Assert.assertArrayEquals(Assert.java:285)
	at org.eclipse.core.tests.internal.preferences.PreferencesServiceTest.testLookupOrder(PreferencesServiceTest.java:237)
```

PR #1088, opened 2024-01-11
Adapt PreferencesServiceTest#testLookupOrder() to UserScope #1087

HeikoKlare:
Adapts the test case PreferencesServiceTest#testLookupOrder() to reflect the UserScope added to Equinox. Also improves the assertions in the test case.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1087

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
@@ -14,6 +14,7 @@
  *******************************************************************************/
 package org.eclipse.core.tests.internal.preferences;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
@@ -54,6 +55,7 @@ import org.eclipse.core.runtime.preferences.IPreferencesService;
 import org.eclipse.core.runtime.preferences.IScopeContext;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.core.runtime.preferences.PreferenceFilterEntry;
+import org.eclipse.core.runtime.preferences.UserScope;
 import org.eclipse.core.tests.harness.FileSystemHelper;
 import org.eclipse.core.tests.runtime.RuntimeTestsPlugin;
 import org.junit.After;
@@ -218,6 +220,7 @@ public class PreferencesServiceTest {
 		String[] defaultOrder = new String[] {"project", //$NON-NLS-1$
 				InstanceScope.SCOPE, //
 				ConfigurationScope.SCOPE, //
+				UserScope.SCOPE, //
 				DefaultScope.SCOPE};
 		String[] fullOrder = new String[] {"a", "b", "c"};
 		String[] nullKeyOrder = new String[] {"e", "f", "g"};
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@HannesWell This adapts to the `UserScope` added to Equinox in https://github.com/eclipse-equinox/equinox/pull/446. From my understanding, simply adding the `UserScope` in the test case should be correct. Please let me know if anything else should be adapted.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
```diff
@@ -230,70 +233,39 @@ public void testLookupOrder() {
 				() -> service.setDefaultLookupOrder(qualifier, key, new String[] { "a", null, "b" }));
 
 		// nothing set
-		String[] order = service.getDefaultLookupOrder(qualifier, key);
-		assertNull("1.0", order);
-		order = service.getLookupOrder(qualifier, key);
-		assertNotNull("1.1", order);
-		assertArrayEquals("1.2", defaultOrder, order);
+		assertThat(service.getDefaultLookupOrder(qualifier, key)).isNull();
+		assertThat(service.getLookupOrder(qualifier, key)).containsExactly(defaultOrder);
```
HannesWell, line 237: I'm not sure if we have already discused it, but what about the standard JUnit Assertions?
```suggestion
		assertNull(service.getDefaultLookupOrder(qualifier, key));
		assertArrayEquals(defaultOrder, service.getLookupOrder(qualifier, key));
```

HannesWell (approved):
Yes this looks good.
Thank you for the follow-up!

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
HeikoKlare: I have to admit that I really dislike `assertEquals` and `assertArrayEquals` because of their bad comprehensibility.
More objectively, I prefer `assertThat` over `assertEquals`/`assertArrayEquals` for several reasons:
* Easier to understand what is `expected` and what is `actual`. In particular because the order changed between JUnit 4 and JUnit 5, it became even harder to do and understand that correctly. I like the relieve myself as well as readers of the assertions from that burden.
* It is not restricted to comparing lists with lists and arrays with arrays but the type of data strucuture is basically irrelevant. In addition, it provides derivation such as ignoring the order. Using standard JUnit 5 assertions for the "basic" cases and `assertThat` for more complex cases will lead to an unnecessary and incomprehensible mixture (in my opinion).
* It provides better error messages, behaves better when one of the arrays is `null` (which you have to / should check separately with standard JUnit assertions, making the code more verbose) and allows you to chain assertions on a single object.

For that reasons I would like to migrate equals assertions to AssertJ (see also #177) for a JUnit 5 migration, as they have to be touched anyway (swapped actual/expected) and then benefit from everything mentioned above.

HeikoKlare:
I'll merge this one now to have the failing test fixed. But feel free to still comment on the usage of `assertEquals` vs, `assertThat`, @HannesWell.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
HannesWell: Sorry for the late reply.
I tend to agree that the assertj assertions are easier to comprehend, but on the other hand I prefer more standard-like solutions over 'custom' ones. And in this case I would consider the default Junit assertions more standard, with the usual advantages of more standard solutions.

But it would probably be good to discuss this in a wider audience and get general consensus to avoid edit-wars.


> It is not restricted to comparing lists with lists and arrays with arrays but the type of data strucuture is basically irrelevant. In addition, it provides derivation such as ignoring the order. Using standard JUnit 5 assertions for the \"basic\" cases and assertThat for more complex cases will lead to an unnecessary and incomprehensible mixture (in my opinion).

I would not consider mixing so bad, it would also make it clear that it's not a simple assertion.

HeikoKlare: > but on the other hand I prefer more standard-like solutions over 'custom' ones

I agree with you and in simple cases like `assertNull(X)`, the standard assertion is probably better (or at least equally good) than `assertThat(X).isNull()`. In particular, in these cases the error messages are comparably expressive. Even though I would say that it does not really matter which of the two you use if you use AssertJ (or some other 3rd party library) in that project/file anyway.

> I would not consider mixing so bad, it would also make it clear that it's not a simple assertion.

Also agreed. I withdraw my comment regard mixing "standard" and "custom" assertions.

> But it would probably be good to discuss this in a wider audience and get general consensus to avoid edit-wars.

I already started a discussion on which custom assertion library we should use, to ensure that we do not start to mix them up: https://github.com/eclipse-platform/.github/discussions/177
Probably you are right that we should also have a quick agreement on when to use custom assertions at all. I am actually not concerned of edit wars, because as long as an assertion change improves something (like comphrensibility or error output), I consider it valuable. With respect to my change this discussion relates to I would say that it was 50% valuable (`assertThat` instead of `assertArrayEquals`) and 50% not (`assertThat` instead of `assertNull`).

A final point I want to make again is that I consider all the standard equality assertions really bad for two reasons:
1. They have pretty bad method signatures (taking two arguments of the same type), in which you easily mix up expected and actual values. Unfortunately, Java lacks a concept of extension methods like C# or Xtend, in which you can define something like `isEqualTo(actual, expected)` and use is in a fluent way as `actual.IsEqualTo(expected)`.
2. The error messages of standard assertions are not very helpful. Consider this example (assertions to be used as alternatives):
```java
int [] actual = new int[] { 1, 2, 3};
int [] expected = new int[] { 1, 2, 3, 4};
assertArrayEquals(actual, expected);
assertThat(actual).containsExactly(expected);
```
Standard assertions give you:
```
array lengths differ, expected: <3> but was: <4>
```
AssertJ gives you: 
```
Expecting actual:
  [1, 2, 3]
to contain exactly (and in same order):
  [1, 2, 3, 4]
but could not find the following elements:
  [4]
```

PR #1059, opened 2023-12-21
Revert "Remove CoreTest class #742"

iloveeclipse:
This reverts commit f408f25e21eba85f6ecf728d1ade26dd8d4fbd38.

The code is still used in Equinox tests.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1684

PR #1090, opened 2024-01-12
Remove CoreTest class #742

HeikoKlare:
The `CoreTest` class of the Eclipse test harness was used as the base class for many JUnit 3-based tests. Since all its subclasses and other usages have been removed and new tests should not use JUnit 3 anymore, the `CoreTest` class is removed.

This reapplied the removal from #1047, which was reverted in #1059 due to remaining users of `CoreTest` in Equinox. The usages in Equinox have finally been cleaned up with eclipse-equinox/equinox#458

I have also checked the following for not having remaining usages:
* Platform, Platform UI, SWT
* Releng Aggregator, Releng Buildtools
* JDT Core/Debug/UI
* PDE


Bananeweizen:
+1 for anything removing JUnit 3. :)

PR #1092, opened 2024-01-14
Remove reference to obsolete swt.binaries repository

HannesWell:
See https://github.com/eclipse-platform/eclipse.platform.swt.binaries/issues/2


HannesWell:
Verification builds don't touch this, therefore there is no need to await them.

PR #1094, opened 2024-01-17
Simplify ExpressionTests

akurtakov:
* Use List.of to simplify test data creation
* Use assertThrows to simplify testing for exceptions
* Use assertFalse instead of assertTrue(!condition)


akurtakov:
One test failure in https://download.eclipse.org/eclipse/downloads/drops4/I20240116-1800/testresults/html/org.eclipse.core.expressions.tests_ep431I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html made me look into the test and had to first make it easier for eyes to grasp before looking in the test itself.

PR #1111, opened 2024-01-22
FileStoreUtil extend NPE information

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1886

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/FileStoreUtil.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/FileStoreUtil.java
@@ -46,13 +46,7 @@ public final class FileStoreUtil {
 	private static int compareNormalisedUri(URI uri1, URI uri2) {
 		int c;
 		// avoid to use IPath here due to high ephemeral memory allocation (Bug 570896)
-		if ((c = compareStringOrNull(uri1.getAuthority(), uri2.getAuthority())) != 0)
-			return c;
-		if ((c = compareStringOrNull(uri1.getScheme(), uri2.getScheme())) != 0)
-			return c;
-		if ((c = comparePathSegments(uri1.getPath(), uri2.getPath())) != 0)
-			return c;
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/FileStoreUtil.java
```diff
@@ -135,9 +135,15 @@ private static int compareStringOrNull(String string1, String string2) {
...
-		if (compare != 0)
-			return compare;
+		try {
+			String scheme1 = fileStore1.getFileSystem().getScheme();
+			String scheme2 = fileStore2.getFileSystem().getScheme();
+			int compare = compareStringOrNull(scheme1, scheme2);
+			if (compare != 0)
+				return compare;
+		} catch (NullPointerException e) {
+			throw new RuntimeException("fileStore1" + fileStore1.getClass() + ':' + fileStore1 + " fileStore2:" + fileStore2.getClass() + ':' + fileStore2, e); //$NON-NLS-1$//$NON-NLS-2$
```
szarnekow, line 145: I acknowledge that the stack trace showed that `getFileSystem` returned null, so it is probably ok. But this line here will create another NPE if fileStore1 or fileStore2 are null.

Note that the message could use some spaces, too.

What do you think about retaining the original exception type:
```
NullPointerException enhanced = new NullPointerException(msg);
enhanced.initCause(e);
throw enhanced;
```

szarnekow: Something like this:

```java
String msg = String.format("%s (fileStore1: %s; fileStore2: %s)", e.getMessage, fileStore1== null? null : fileStore1.getClass().getName(), fileStore2== null? null : fileStore2.getClass().getName());
NullPointerException enhanced = new NullPointerException(msg);
enhanced.initCause(e);
throw enhanced;
```

jukzi: does not matter, it snot checked exception anyway. the goal is only to get the wrong implementation class to then fix the underlying error

PR #1113, opened 2024-01-22
use try-with-resource where hinted

jukzi:


jukzi:
There are a lot of unrelated changes automatic done on save due to 5c071d399a438b14110b40d5332a3150b00d83df 
-  @HeikoKlare is that intended?

HeikoKlare:
The unification of auto-save options was, let's say, rather "expected" or "accepted" than "intended" to produce some unrelated changes when modifying files. See #1001

PR #1121, opened 2024-01-22
fix Potential resource leak warnings

jukzi:
replace FileUtil.safeClose with inlined try-with-resource

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
@@ -17,7 +17,6 @@
 package org.eclipse.core.internal.utils;
 
 import java.io.ByteArrayInputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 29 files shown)


jukzi:
image

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
```diff
@@ -339,8 +310,8 @@ public static IPath toPath(URI uri) {
 	public static final void transferStreams(InputStream source, OutputStream destination, String path,
 			IProgressMonitor monitor) throws CoreException {
 		SubMonitor subMonitor = SubMonitor.convert(monitor);
-		try {
-			try {
+		try (source) {
```
szarnekow, line 313: No need for any actual change but FYI: `try (source; destination) {..}` would have reduced the nesting level by one.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/RefreshLocalJavaFileBuilder.java
```diff
@@ -36,18 +40,17 @@ protected IProject[] build(int kind, Map<String, String> args, IProgressMonitor
 		IFile file = project.getFile("A.java");
 		IPath localLocation = project.getLocation().append(file.getName());
 		java.io.File localFile = localLocation.toFile();
-		FileOutputStream out = null;
-		try {
+		try (FileOutputStream out = new FileOutputStream(localFile)) {
 			if (localFile.exists()) {
```
szarnekow, line 44: At this point we have an open output stream on `localFile` and we delete the file before we start writing to it. 
Previously it used to be the case that the stream was made after deleting the file.
If I try this locally, the file doesn't exist after `out` was closed:

```java
		File tempDir = Files.createTempDir();
		File localFile = new File(tempDir, "a.txt");
		localFile.createNewFile();
		try (FileOutputStream out = new FileOutputStream(localFile)) {
			if (localFile.exists()) {
				localFile.delete();
			}
			out.write("public class A {}".getBytes());
		}
		try (FileInputStream in = new FileInputStream(localFile)) {
			in.readAllBytes();
		}
```

jukzi: thanks -> https://github.com/eclipse-platform/eclipse.platform/pull/1129

PR #1129, opened 2024-01-23
RefreshLocalJavaFileBuilder: use java.nio

jukzi:


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/RefreshLocalJavaFileBuilder.java
```diff
@@ -40,15 +40,9 @@ protected IProject[] build(int kind, Map<String, String> args, IProgressMonitor
...
-			if (localFile.exists()) {
-				localFile.delete();
-			}
-
-			out.write("public class A {}".getBytes());
-		} catch (IOException streamCloseIgnored) {
-			// ignore;
-		} catch (Exception e) {
+		try {
+			Files.writeString(localFile.toPath(), "public class A {}");
```
szarnekow, line 44: Nit: This will write always with UTF8 - the old solution wrote with the platform default locale. Both of these are bonkers, I assume.

We might want to make the charset explicit by using

```suggestion
			Files.writeString(localFile.toPath(), "public class A {}", Charset.forName(file.getCharset());
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/RefreshLocalJavaFileBuilder.java
```diff
@@ -14,8 +14,8 @@
  *******************************************************************************/
 package org.eclipse.core.tests.internal.builders;
 
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.file.Files;
```
szarnekow, line 18:
```suggestion
import java.nio.file.Files;
import java.nio.charset.Charset;
```

szarnekow (approved):

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/RefreshLocalJavaFileBuilder.java
jukzi: The text written is ASCII only, it does not matter.

Issue #1114, opened 2024-01-22
NPE during "Refreshing Workspace" - IFileStore.getFileSystem()" is null - often when switching between git branches 


chrisrueger:
This is a copy of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1886 because I opened it in the wrong repo (with my other account). 

There has already been a PR by @jukzi  related to this:  https://github.com/eclipse-platform/eclipse.platform/pull/1111 

Original issue description: 

I recently noticed the following exception quite often when switching between git branches (using external Github Desktop git client... not Eclipse itself)
Probably because the underlying filesystem content has changed. But maybe the NPE can be handled. 

I cannot remember seeing this exception much earlier.. so from a gut feeling I would say I noticed it in 2023-12. 

## Exception

An internal error occurred during: "Refreshing workspace".

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.filesystem.IFileSystem.getScheme()" because the return value of "org.eclipse.core.filesystem.IFileStore.getFileSystem()" is null
	at org.eclipse.core.internal.filesystem.FileStoreUtil.compareFileStore(FileStoreUtil.java:138)
	at org.eclipse.core.filesystem.IFileStore.compareTo(IFileStore.java:583)
	at java.base/java.util.TreeMap.getEntryUsingComparator(TreeMap.java:374)
	at java.base/java.util.TreeMap.getEntry(TreeMap.java:344)
	at java.base/java.util.TreeMap.get(TreeMap.java:279)
	at org.eclipse.core.internal.resources.AliasManager$LocationMap.matchingResourcesDo(AliasManager.java:203)
	at org.eclipse.core.internal.resources.AliasManager.internalComputeAliases(AliasManager.java:545)
	at org.eclipse.core.internal.resources.AliasManager.computeAliases(AliasManager.java:430)
	at org.eclipse.core.internal.localstore.RefreshLocalAliasVisitor.deleteResource(RefreshLocalAliasVisitor.java:53)
	at org.eclipse.core.internal.localstore.RefreshLocalVisitor.synchronizeExistence(RefreshLocalVisitor.java:195)
	at org.eclipse.core.internal.localstore.RefreshLocalVisitor.visit(RefreshLocalVisitor.java:302)
	at org.eclipse.core.internal.localstore.UnifiedTree.accept(UnifiedTree.java:119)
[...]

PR #1137, opened 2024-01-24
Make NullFileSystem a singleton

iloveeclipse:
/dev/null shouldn't exist multiple times :-)

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1114

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
@@ -15,9 +15,18 @@ package org.eclipse.core.internal.filesystem;
 
 import java.net.URI;
 import java.util.HashMap;
-import org.eclipse.core.filesystem.*;
+import org.eclipse.core.filesystem.EFS;
+import org.eclipse.core.filesystem.IFileStore;
+import org.eclipse.core.filesystem.IFileSystem;
 import org.eclipse.core.filesystem.provider.FileSystem;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionDelta;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.IRegistryChangeEvent;
+import org.eclipse.core.runtime.IRegistryChangeListener;
+import org.eclipse.core.runtime.RegistryFactory;
 import org.eclipse.osgi.util.NLS;
 
 /**
@@ -72,9 +81,15 @@ public class InternalFileSystemCore implements IRegistryChangeListener {
 			IConfigurationElement element = (IConfigurationElement) result;
 			FileSystem fs = (FileSystem) element.createExecutableExtension("run"); //$NON-NLS-1$
 			fs.initialize(scheme);
-			//store the file system instance so we don't have to keep recreating it
-			registry.put(scheme, fs);
-			return fs;
+			synchronized (this) {
+				result = registry.get(scheme);
+				if (result instanceof IFileSystem) {
+					return (IFileSystem) result;
+				}
+				//store the file system instance so we don't have to keep recreating it
+				registry.put(scheme, fs);
+				return fs;
+			}
 		} catch (CoreException e) {
 			//remove this invalid file system from the registry
 			registry.remove(scheme);
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
@@ -29,7 +29,7 @@ public class NullFileSystem extends FileSystem {
 	/**
 	 * The singleton instance of this file system.
 	 */
-	private static IFileSystem instance;
+	private static final IFileSystem INSTANCE = EFS.getNullFileSystem();
 
 	/**
 	 * Returns the instance of this file system
@@ -37,7 +37,7 @@ public class NullFileSystem extends FileSystem {
 	 * @return The instance of this file system.
 	 */
 	public static IFileSystem getInstance() {
-		return instance;
+		return INSTANCE;
 	}
 
 	/**
@@ -45,7 +45,6 @@ public class NullFileSystem extends FileSystem {
 	 */
 	public NullFileSystem() {
 		super();
-		instance = this;
 	}
 
 	@Override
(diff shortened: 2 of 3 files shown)


laeubi (approved):
Looks good, maybe one want to even implement hashCode/equals so that all `NullFileSystem` instances are equal (I think that was probably the whole intend of the previous instantiation in constructor) because now there will be two instances, one from the extension point and one from the INSTANCE.

iloveeclipse:
> maybe one want to even implement hashCode/equals

Hmm. Let see first what tests are saying, I will push second version with hashcode/equals after that.

laeubi:
@iloveeclipse I now created a version that removes the singelton instance altogether:

- https://github.com/eclipse-platform/eclipse.platform/pull/1138

iloveeclipse:
OK, there is a test that fails now:
```
java.lang.NullPointerException: Cannot invoke "String.compareTo(String)" because the return value of "org.eclipse.core.filesystem.IFileSystem.getScheme()" is null
	at org.eclipse.core.tests.filesystem.FileStoreTest.testSortOrder(FileStoreTest.java:684)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```



> I now created a version that removes the singelton instance altogether:

Good idea. See my second change that simply uses `EFS.getNullFileSystem()`.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
```diff
@@ -26,26 +26,20 @@
...
-	private static IFileSystem instance;
-
 	/**
 	 * Returns the instance of this file system
 	 *
 	 * @return The instance of this file system.
 	 */
 	public static IFileSystem getInstance() {
-		return instance;
+		return EFS.getNullFileSystem();
```
laeubi, line 35: Nice trick!

jukzi:
image
mind to fix LocalFileSystem.instance too?

iloveeclipse:
> mind to fix LocalFileSystem.instance too?

Yes. Will push in a moment.

laeubi:
I wanted to note that the "issue" only arises if the API is used incorrectly and using EFS.getXXX will add another roundtrip, I therefore think we should get rid of the instance all all togehter see my other PR

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
jukzi: i don't like that this implementation unnecessarily looks up in a hashmap introducing unneeded computational complexity for an often used function.

jukzi (changes requested):
better just use
private static final IFileSystem instance = EFS.getLocalFileSystem();
private static final IFileSystem instance = EFS.getNullFileSystem();

jukzi (approved):

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
```diff
@@ -72,9 +81,15 @@ public IFileSystem getFileSystem(String scheme) throws CoreException {
 			IConfigurationElement element = (IConfigurationElement) result;
 			FileSystem fs = (FileSystem) element.createExecutableExtension("run"); //$NON-NLS-1$
 			fs.initialize(scheme);
-			//store the file system instance so we don't have to keep recreating it
-			registry.put(scheme, fs);
-			return fs;
+			synchronized (this) {
+				result = registry.get(scheme);
```
jukzi, line 85: nit: could use computeIfAbsent instead of get/put, but not important

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
```diff
@@ -72,9 +81,15 @@ public IFileSystem getFileSystem(String scheme) throws CoreException {
 			IConfigurationElement element = (IConfigurationElement) result;
 			FileSystem fs = (FileSystem) element.createExecutableExtension("run"); //$NON-NLS-1$
 			fs.initialize(scheme);
-			//store the file system instance so we don't have to keep recreating it
-			registry.put(scheme, fs);
-			return fs;
+			synchronized (this) {
```
laeubi, line 84: You should move the part from the config element below the part that checks if there is already an instance

iloveeclipse: > You should move the part from the config element below the part that checks if there is already an instance

No, we don't know if that code takes any locks (which is allowed) and so we might run into deadlock running that code with another lock aquired.

iloveeclipse:
Let's merge for the next IBuild to see if there is anything in SDK that doesn't like this code.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
laeubi: But now you create (and init) a FS on each call but probably never use it...

iloveeclipse: > But now you create (and init) a FS on each call but probably never use it...

Not on each call - only for these threads that might init this in parallel. Once initialized, the value from map is used. Extension points allow multiple creation of contributions, and changing existing implementation to use a lock to avoid that is again breaking client code.

laeubi: the returned map is unsyncronized, so it is not safe to modify it while other threads are already reading it (what could happen here). So then the first read must be syncronized as well, or a concurrent map must be used with computeIfAbsent if you don't like to lock ...

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
iloveeclipse: The bad thing is that map **can** contain value but if it is not a FileSystem, we still have to put here.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/InternalFileSystemCore.java
iloveeclipse: Correct, I tend to forget that, see https://github.com/eclipse-platform/eclipse.platform/pull/1144

PR #1138, opened 2024-01-24
Remove static singleton for NullFileSystem / LocalFileSystem

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform/issues/1114

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileStore.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileStore.java
@@ -13,13 +13,22 @@
  *******************************************************************************/
 package org.eclipse.core.internal.filesystem;
 
-import java.io.*;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import org.eclipse.core.filesystem.*;
+import org.eclipse.core.filesystem.EFS;
+import org.eclipse.core.filesystem.IFileInfo;
+import org.eclipse.core.filesystem.IFileStore;
+import org.eclipse.core.filesystem.IFileSystem;
 import org.eclipse.core.filesystem.provider.FileInfo;
 import org.eclipse.core.filesystem.provider.FileStore;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 
 /**
  * A null file store represents a file whose location is unknown,
@@ -29,13 +38,26 @@ import org.eclipse.core.runtime.*;
  */
 public class NullFileStore extends FileStore {
 	private final IPath path;
+	protected final IFileSystem fileSystem;
 
 	/**
 	 * Creates a null file store
 	 * @param path The path of the file in this store
 	 */
-	public NullFileStore(IPath path) {
+	public NullFileStore(IPath path, IFileSystem fileSystem) {
 		Assert.isNotNull(path);
+		Assert.isLegal(fileSystem instanceof NullFileSystem, "fileSystem must extend NullFileSystem"); //$NON-NLS-1$
+		this.fileSystem = fileSystem;
+		this.path = path;
+	}
+
+	/**
+	 * For test only! Extenders need to extend {@link #getFileSystem()} as well to return their desired instance of the file system!
+	 * @param path The path of the file in this store
+	 */
+	@Deprecated(forRemoval = true, since = "will be deleted 2024-06")
+	public NullFileStore(IPath path) {
+		this.fileSystem = EFS.getNullFileSystem();
 		this.path = path;
 	}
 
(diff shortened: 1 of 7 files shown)


resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileStore.java
```diff
@@ -29,13 +38,25 @@
...
 public class NullFileStore extends FileStore {
 	private final IPath path;
+	protected final IFileSystem fileSystem;
 
 	/**
 	 * Creates a null file store
 	 * @param path The path of the file in this store
 	 */
-	public NullFileStore(IPath path) {
+	public NullFileStore(IPath path, IFileSystem fileSystem) {
```
iloveeclipse, line 47: Why adding an argument / potentially breaking clients that might have used it for years (and yes, this is not API)?

However, I've pushed now second commit that uses `EFS.getNullFileSystem()` https://github.com/eclipse-platform/eclipse.platform/pull/1137.

iloveeclipse (changes requested):

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileStore.java
laeubi: > Why adding an argument / potentially breaking clients that might have used it for years (and yes, this is not API)?

Because it is not API and it is obviously misusing that, I think it is acceptable this for a extender but not for a public constructor. Do you know of any client that can be affected?

laeubi: I now added a "compatibility" constructor ...

laeubi: > But from my experience, clients may use everything that is public and is there for nearly 20 years, so if possible I would always avoid breaking them for no obvious reason.

That's why we mark it internal so clients are aware they can break anytime. And as we can see here this even introduced a bug in BndTools, as the misuses the `NullFileStore` for a custom implementation that is completely unrelated, anyways with the fallback everything should now be fine.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
```diff
@@ -27,34 +27,29 @@
...
+	public NullFileSystem() {
+		super();
+	}
 
 	/**
 	 * Returns the instance of this file system
 	 *
 	 * @return The instance of this file system.
 	 */
+	@Deprecated(forRemoval = true)
```
akurtakov, line 41: When you deprecate something with forRemova=true please add since with proper value e.g. "2024-03" so it's known when it can be removed.

laeubi: @akurtakov it is internal api so it can be removed anytime....

jukzi (changes requested):
i'd prefer a solution that does not deprecate a useful API. the API was not bad but the implementation

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
akurtakov: I missed that. I'm stepping back here :)

laeubi:
> i'd prefer a solution that does not deprecate a useful API. the API was not bad but the implementation

This is **NOT** API ... the API is `ECF.getLocalFileSystem()`  / `ECF.getNullFileSystem()` ...

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileSystem.java
laeubi: It was only @iloveeclipse was concerned that someone somewhere is using our internal API even though as it obviously never has worked reliable...

jukzi:
Everything can be considered api https://www.hyrumslaw.com/ :-)

Anyway If its not public API just remove or fix it, instead of deprecation

laeubi:
> Everything can be considered api

Yes its java API ... but not "Platform Exposed Public API" and it was never called except from internal implementation or tests thats why no one seen this "bug" unless BndTools started to misuses the class and ran into NPE ...

> Anyway If its not public API just remove or fix it, instead of deprecation

I would just remove this but @iloveeclipse was concerned so this seems a good compromise.

iloveeclipse:
> Anyway If its not public API just remove or fix it, instead of deprecation

In ideal world or acting as code owner of an internal closed source code base I would fully agree with that.
But we are not in the ideal world and not in a closed source application where we control everything.

The code in question is there since 2005 / 55144e527494a643000b319a24d6b5f072ed4203.

As we see from the bug report #1114, clients **are** using it because they can, and my personal opinion is that there is no need to break them if we can avoid it.

vogella:
In general internal API be deleted / changed at any time but as concern were raised, why not add the since tag "will be deleted 2024-06" so that users have one release to react.

jukzi:
there is a much easier solution https://github.com/eclipse-platform/eclipse.platform/pull/1137#pullrequestreview-1840905879

laeubi:
> In general internal API be deleted / changed at any time but as concern were raised, why not add the since tag "will be deleted 2024-06" so that users have one release to react.

Done ...



> there is a much easier solution [#1137 (review)](https://github.com/eclipse-platform/eclipse.platform/pull/1137#pullrequestreview-1840905879)

THis does not solve the problem that people are using the internal API in a wrong way ;-)

laeubi:
It seems obsolete now...

PR #1140, opened 2024-01-24
Provisional api guidelines clarification

vogella:


vogella:
@akurtakov the document still speaks about "Before" and "After" the API freeze. Do we still have a official API freeze or has this been removed? I think it has been removed and would update the document if you agree.

akurtakov:
There is still API freeze - it's the Wed before RC1 as can be seen at https://eclipse.dev/eclipse/development/plans/freeze_plan_4_30.php (for the last release).

PR #1147, opened 2024-01-24
Add Ant test.xml for org.eclipse.tips.tests

HeikoKlare:
This adds a test.xml for Ant-based execution of tests in org.eclipse.tips.tests. It is a follow-up to the addition of these tests to the Maven build in #1119. 

This is supposed to enable the execution of the tips tests in I-Builds. I have checked the wiki, namely
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Automated-Testing#automated-unit-and-integration-testing
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Eclipse-Test-Framework

but I am still not sure whether anything else has to be changed (and whether the specification is completely correct) to enable the tests. The wiki refers to some "overall, main test.xml" ([here](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Automated-Testing#appendix)), but the link is broken.

### DIFF
--- a/ua/org.eclipse.tips.tests/build.properties
+++ b/ua/org.eclipse.tips.tests/build.properties
@@ -15,6 +15,7 @@ source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                icons/,\
+               test.xml,\
                .
 src.includes = about.html
 pom.model.property.testClass = org.eclipse.tips.tests.AllTipsTests
--- a/ua/org.eclipse.tips.tests/test.xml
+++ b/ua/org.eclipse.tips.tests/test.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0"?>
+<project name="Tips Tests" default="run" basedir=".">
+	<!-- The property ${eclipse-home} should be passed into this script -->
+	<property name="eclipse-home" value="${basedir}/../.."/>
+	
+	<property name="plugin-name" value="org.eclipse.tips.tests"/>
+	<property name="library-file" value="${eclipse-home}/plugins/org.eclipse.test/library.xml"/>
+	<property name="resources_location" value="${eclipse-home}/tips_tests_resources"/>
+
+	<!-- This target holds all initialization code that needs to be done for -->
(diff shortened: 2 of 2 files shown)


akurtakov:
No one did that in years probably so only way to figure it out is to push and if things fail investigate. Sorry for the late reply, I just managed to get to this one today in the very long queue of things.

HeikoKlare:
> No one did that in years probably so only way to figure it out is to push and if things fail investigate. Sorry for the late reply, I just managed to get to this one today in the very long queue of things.

Thanks for the feedback, no matter how "late"! This is not urgent at all and I just kept it open to see if someone has some input. But I will do as you propose (probably next week): just merge and see investigate whatever happens.

jukzi:
@HeikoKlare if you do not want to have this PR merged in the current form please convert to draft or set a negative review

HeikoKlare:
Thanksf or the reminder!
I just forgot to further process this PR. From my side, we can merge it. Effects (and potential issues with it) will only be visible in I-Builds.

jukzi (approved):

akurtakov:
Please include it in the jar via build.properties, otherwise it would not be shipped and thus has no effect.  Once this is done we can push this patch as it would have no effect until there is change to explicitly invoke the tests as can be seen in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/45bfffcb3d5f7a67842b2ae5353193433d21e3aa/production/testScripts/configuration/sdk.tests/testScripts/test.xml#L1859

HeikoKlare:
> Once this is done we can push this patch as it would have no effect until there is change to explicitly invoke the tests as can be seen in

Thanks for that pointer! So that should be the main `test.xml` to which the wiki tries to point here: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Automated-Testing#appendix
I will provide a fix of the wiki link.

I have amended this PR to include the `test.xml` in the jar. I will provide a follow-up PR to the platform aggregator adding this test project to the referenced main `test.xml`.

Issue #1148, opened 2024-01-25
Broken FinalTip#hashCode() / random failing TipProviderTest#testGetNextTip()


HeikoKlare:
The test case `org.eclipse.tips.core.TipProviderTest#testGetNextTip()` is randomly failing, e.g., in this run: https://github.com/eclipse-platform/eclipse.platform/pull/1147/checks?check_run_id=20823928039

```
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at org.eclipse.tips.core.TipProviderTest.testGetNextTip(TipProviderTest.java:106)
```

The reason is that the tips manager uses the hashcode of tips to identify whether they have been read, but the `FinalTip` creates a new creation data (used for the hashcode) on every access, thus being treated as unread if the system time has changed since it has been marked as read.

PR #1149, opened 2024-01-25
Fix FinalTip#hashcode() and TipProviderTest#testGetNextTip() #1148

HeikoKlare:
The test case org.eclipse.tips.core.TipProviderTest#testGetNextTip() randomly fails because it checks that the read state of a tip after being set can properly be retrieved. This functionality relies on the hash code of tips staying the same, but the one for the FinalTip currently changes if the system time has changed, as it re-calculates the creation time used for the calculation of the hash code on every access.

With this change, the creation time for the FinalTip is stored such that its hash code does not change. This also fixes the according test case.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1148

### DIFF
--- a/ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
+++ b/ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
@@ -31,17 +31,16 @@ public class FinalTip extends Tip implements IHtmlTip {
 	private static final String EH1 = "</h1>"; //$NON-NLS-1$
 	private static final String H1 = "<h1>"; //$NON-NLS-1$
 
-	/**
-	 * Constructor.
-	 */
+	private final Date creationDate;
+
 	public FinalTip(String providerId) {
 		super(providerId);
+		this.creationDate = Calendar.getInstance().getTime();
 	}
 
 	@Override
 	public Date getCreationDate() {
-		Calendar instance = Calendar.getInstance();
-		return instance.getTime();
+		return creationDate;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
```diff
@@ -31,17 +31,19 @@ public class FinalTip extends Tip implements IHtmlTip {
 	private static final String EH1 = "</h1>"; //$NON-NLS-1$
 	private static final String H1 = "<h1>"; //$NON-NLS-1$
 
+	private Date creationDate;
```
jukzi, line 34: final

ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
```diff
@@ -31,17 +31,19 @@ public class FinalTip extends Tip implements IHtmlTip {
...
+
 	/**
```
jukzi, line 36: i vote to remove such useless comments "Constructor" on the fly

ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
HeikoKlare: ade919d7d02308c4a8bf6a086269963a94f47612

ua/org.eclipse.tips.core/src/org/eclipse/tips/core/internal/FinalTip.java
HeikoKlare: ade919d7d02308c4a8bf6a086269963a94f47612

PR #1152, opened 2024-01-26
Ant tests don't need org.eclipse.pde.core

akurtakov:


akurtakov:
Prereq for https://github.com/eclipse-pde/eclipse.pde/pull/1060

Issue #1155, opened 2024-01-29
ant BuildTests fail since I20240126-1800


jukzi:
6 org.eclipse.ant.tests.ui.BuildTests fail on all platforms since
https://download.eclipse.org/eclipse/downloads/drops4/I20240126-1800/testResults.php

for example
Incorrect number of messages logged for build. Should be 8. Was 9 expected:<8> but was:<9>

```
java.lang.AssertionError: Incorrect number of messages logged for build. Should be 8. Was 9 expected:<8> but was:<9>
at org.eclipse.ant.tests.ui.BuildTests.testXmlLoggerListener(BuildTests.java:169)
```

akurtakov:
This is on me and caused by https://github.com/eclipse-platform/eclipse.platform/pull/1152 . Tests work successfully in Maven. I managed to figure out that in workspace if pde.core is not on the classpath everything is fine, but if pde.core is there exactly the same failures happen. I have to figure out how/why I-builds test runner includes it and stop this from happening.

PR #1156, opened 2024-01-29
Do not override eclipse.buildScript in ant.tests.ui

akurtakov:
This prevents the message for overriding existing ant task that causes failures in BuildTests.
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1155


akurtakov:
Unrelated failures.

PR #1157, opened 2024-01-29
fix "The parameter text should not be assigned"

jukzi:


jukzi:
@laeubi now the ECJ warning go berserk showing new Warnings that are indeed API warnings

laeubi:
> @laeubi now the ECJ warning go berserk showing new Warnings that are indeed API warnings

Nothing is going "beserk" please see
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1751
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1557#issuecomment-1913029297

PR #1165, opened 2024-02-01
Suppress deprecation: java.util.Observable

jukzi:
There is no plan to change the implementation

### DIFF
--- a/ua/org.eclipse.help.base/src/org/eclipse/help/internal/base/BookmarkManager.java
+++ b/ua/org.eclipse.help.base/src/org/eclipse/help/internal/base/BookmarkManager.java
@@ -33,6 +33,7 @@ import org.osgi.service.prefs.BackingStoreException;
  *
  * @since 3.1
  */
+@SuppressWarnings("deprecation") // java.util.Observable since 9;
 public class BookmarkManager extends Observable {
 	// all bookmarks removed
 	public static final int REMOVE_ALL = 1;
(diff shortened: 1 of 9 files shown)


jukzi:
jenkins plugin shows warnings that are not shown in workspace:
image
may be related to `<code.ignoredWarnings>-warn:-deprecation,raw,unchecked</code.ignoredWarnings>` in pom.xml?

jukzi:
@akurtakov should "code.ignoredWarnings" in all projects be removed or replaced by
"<code.ignoredWarnings>${tests.ignoredWarnings}</code.ignoredWarnings>" in order to get the same result as in workspace?

laeubi:
Last time I though we wanted to simply use project settings, so maybe just remove that at all?

akurtakov:
I agree that it would be best if workspace and maven build warnings are the same.

jukzi:
the ecj jenkins plugin still (or again?) validiates against the wrong base:
```
[ReferenceFinder] Found reference build '#697' for target branch
[Eclipse ECJ] Obtaining reference build from reference recorder
[Eclipse ECJ] -> Found 'eclipse.platform » master #697'
[Eclipse ECJ] Using reference build 'eclipse.platform/master #696' to compute new, fixed, and outstanding issues
[Eclipse ECJ] Issues delta (vs. reference build): outstanding: 62, new: 0, fixed: 9
```
image

probably because 697 failed a test.
wonder why 698 is not use.

PR #1173, opened 2024-02-04
Revert "Handle return value 24 the same as 23"

merks:
Reverts eclipse-platform/eclipse.platform#1172

The changes in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622 avoid this problem.


laeubi:
Would it harm to keep this and make it more robust?

merks:
> Would it harm to keep this and make it more robust?

Yes, if it's a bug that the 24 surfaced, then it's confusing at best to have downstream code deal with something that should never happen and actually does never happen except during the brief period of time when the bug existed.

PR #1174, opened 2024-02-05
remove ignoredWarnings

jukzi:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1628


akurtakov:
It's fully expected to have more warnings with a patch that stops hiding them.

PR #1186, opened 2024-02-07
Eclipse_Project_Update_Sites wiki page migrated and updated

vogella:
Trying to make this page a little bit more timeless, by adding future udpate sites and explaining the general concept of the I-update sites

### DIFF
--- a/docs/Eclipse_Project_Update_Sites.md
+++ b/docs/Eclipse_Project_Update_Sites.md
@@ -1,0 +1,82 @@
+Eclipse Project Update Sites
+============================
+
+This site contains information about the artifacts created by the Eclipse project.
+
+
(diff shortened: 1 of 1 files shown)


akurtakov:
There is https://download.eclipse.org/eclipse/updates/I-builds/ which points to latest I-builds stream so maybe using it makes more sense.

vogella:
> There is https://download.eclipse.org/eclipse/updates/I-builds/ which points to latest I-builds stream so maybe using it makes more sense.

Sorry, saw this a bit to late, not used in getting feedback on these PR. I will update the documentation accordingly

akurtakov:
Feedback should be given only when needed :)

PR #1194, opened 2024-02-08
Use the module timestamp instead of the timestamp of the state object

laeubi:
Currently the state timestamp is used what can change if a bundle is resolved not only when it is installed/updated/uninstalled, but the Platform#getStateStamp() claims exactly this.

To restore the documented behavior this iterates over all bundles and uses its Module#getTimestamp instead, this also removes the reference to the  PlatformAdmin that currently requires the compatibility module and could therefore be removed from this once we have migrated PDE.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
@@ -18,6 +18,7 @@ package org.eclipse.core.internal.runtime;
 
 import java.io.File;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -57,6 +58,7 @@ import org.eclipse.equinox.internal.app.IBranding;
 import org.eclipse.equinox.log.ExtendedLogReaderService;
 import org.eclipse.equinox.log.ExtendedLogService;
 import org.eclipse.equinox.log.Logger;
+import org.eclipse.osgi.container.Module;
 import org.eclipse.osgi.container.ModuleContainer;
 import org.eclipse.osgi.framework.log.FrameworkLog;
 import org.eclipse.osgi.service.datalocation.Location;
@@ -538,8 +540,8 @@ public final class InternalPlatform {
 	}
 
 	public long getStateTimeStamp() {
-		PlatformAdmin admin = getPlatformAdmin();
-		return admin == null ? -1 : admin.getState(false).getTimeStamp();
+		return Arrays.stream(getBundleContext().getBundles()).map(bundle -> bundle.adapt(Module.class))
+				.filter(Objects::nonNull).mapToLong(Module::getLastModified).sum();
 	}
 
 	public Location getUserLocation() {
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
@@ -1166,20 +1166,24 @@ public final class Platform {
 	/**
 	 * Returns the platform administrator for this running Eclipse.
 	 * <p>
-	 * Note: This is an internal method and <em>must not</em>
-	 * be used by clients which are not part of the Eclipse Platform.
-	 * This method allows access to classes which are not Eclipse
-	 * Platform API but are part of the OSGi runtime that the Eclipse
-	 * Platform is built on. Even as the Eclipse Platform evolves
-	 * in compatible ways from release to release, the details of
-	 * the OSGi implementation might not.
-	 * </p><p>
-	 * Clients can also acquire the {@link PlatformAdmin} service
-	 * to retrieve this object.
+	 * Note: This is an internal method and <em>must not</em> be used by clients
+	 * which are not part of the Eclipse Platform. This method allows access to
+	 * classes which are not Eclipse Platform API but are part of the OSGi runtime
+	 * that the Eclipse Platform is built on. Even as the Eclipse Platform evolves
+	 * in compatible ways from release to release, the details of the OSGi
+	 * implementation might not.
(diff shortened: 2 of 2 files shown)


laeubi:
To prevent us from future surprise I have now also deprecated the Platform#getPlatformAdmin and removed all remaining references to it in PDE with:
- https://github.com/eclipse-pde/eclipse.pde/pull/1137

we then should finally remove this method altogether in the future (it is documented to not be used outside of platform anyways).

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -538,8 +540,8 @@ public IPath getStateLocation(Bundle bundle, boolean create) throws IllegalState
 	}
 
 	public long getStateTimeStamp() {
-		PlatformAdmin admin = getPlatformAdmin();
-		return admin == null ? -1 : admin.getState(false).getTimeStamp();
+		return Arrays.stream(getBundleContext().getBundles()).map(bundle -> bundle.adapt(Module.class))
+				.filter(Objects::nonNull).mapToLong(Module::getLastModified).max().orElse(-1);
```
HannesWell, line 544: Alternatively the following could be used.
```suggestion
		Bundle bundle = FrameworkUtil.getBundle(InternalPlatform.class);
		org.eclipse.osgi.container.Module module = bundle.adapt(org.eclipse.osgi.container.Module.class);
		if (module == null) {
			return 0;
		}
		Generation generation = (Generation) module.getCurrentRevision().getRevisionInfo();
		org.eclipse.osgi.storage.Storage storage = generation.getBundleInfo().getStorage();
		return storage.getModuleDatabase().getRevisionsTimestamp();
```
The only problem with that code is that it uses internal Equinox methods and this project is configured to mark such usages with errors.

Another alternative would be to add to Equinox' `ModuleContainer` a method:
```
public long getRevisionsTimestamp() {
	return moduleDatabase.getRevisionsTimestamp();
}
```
And then just call
```suggestion
		Bundle bundle = FrameworkUtil.getBundle(InternalPlatform.class);
		org.eclipse.osgi.container.Module module = bundle.adapt(org.eclipse.osgi.container.Module.class);
		if (module == null) {
			return 0;
		}
		return module.getContainer().getRevisionsTimestamp();
```
All of that would have the advantage that you don't have to iterate over all bundle in the runtime.

laeubi: I don't think this is really relevant, this method is actually nowhere used except in this place and one another on startup/shutdown so I'd like to make it as simple and understandable as possible without a chance to break because we interpret what "timestamp" means differently, e.g. the `ModuleDatabase.getRevisionsTimestamp()` says:

>  **For example:**
> 	 * <ul>
> 	 * <li>installing a module
> 	 * <li>updating a module
> 	 * <li>uninstalling a module
> 	 * <li>modifying the wirings
> 	 * </ul>
> 	 * <p>

so this example already contain something we don't want (change in wiring), and "for example" indicates to me that in the future maybe other things will become relevant.

laeubi: > The performance might be secondary, but the suggested approaches also ensure it works as before, because that is what was effectively called by the PlatformAdmin.

Yes an "works like before" is wrong, as the method only want to have a different timestmap if a bundle was installed/updated/uninstalled, nothing else.

HeikoKlare:
I am not sure why the test failed on Mac. The master build after merging this PR succeeded: https://github.com/eclipse-platform/eclipse.platform/actions/runs/7827313394/job/21354884845
I've also tried the patch locally on Mac and it also resolves the test failure there. So let's see if there will be failing actions runs.

Thanks for the fix, Christoph!

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -538,8 +540,8 @@ public IPath getStateLocation(Bundle bundle, boolean create) throws IllegalState
 	}
 
 	public long getStateTimeStamp() {
-		PlatformAdmin admin = getPlatformAdmin();
-		return admin == null ? -1 : admin.getState(false).getTimeStamp();
+		return Arrays.stream(getBundleContext().getBundles()).map(bundle -> bundle.adapt(Module.class))
+				.filter(Objects::nonNull).mapToLong(Module::getLastModified).sum();
```
HannesWell, line 544: Just for the record: Theoretically it would be possible that the sum is unchanged if bundles are uninstalled but the timestamp increase of the remaining bundles exactly compensate that removal.
But such situation is very unlikely and as far as I can tell `Platform::getStateStamp` also makes no guaranties that it can never happen that the same number is returned twice.

laeubi: Just for the record, this method now has only one single consumer left that seem to handle somehow if/when some welcome contributions are scanned again :-D

szarnekow: I'm not sure this is actually true? https://github.com/laeubi/eclipse.platform/blob/master/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java#L548 but yeah, probabilities that this is going to happen, are rather low.

laeubi: I synced my fork to reflect current master and there you will see this

https://github.com/laeubi/eclipse.platform/blob/e7069db92859e15a44852ff753b9edeec2c1d7ed/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java#L599

:-)

PR #1196, opened 2024-02-08
Use a composite key of all bundle ids and version to check the cache

laeubi:
Currently the ContentDescriptionManager relies on a rather vague "timestamp" to indicate that the set of bundles has changed, this leads to some hard to explain instabilities in tests.

This now uses a composite key of all bundle ids and their version that has two benefits:

1) it is deterministic what properties of a bundle set is compared
2) in the test we can assert that the state is stable and if not give a comparison list that can be used to determine what actually has changed

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
@@ -20,9 +20,16 @@ package org.eclipse.core.internal.resources;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HexFormat;
 import java.util.LinkedHashSet;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileStore;
 import org.eclipse.core.internal.events.ILifecycleListener;
@@ -57,6 +64,7 @@ import org.eclipse.core.runtime.content.IContentTypeMatcher;
 import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.osgi.util.NLS;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 
 /**
  * Keeps a cache of recently read content descriptions.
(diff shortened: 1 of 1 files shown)


laeubi:
okay it seems they stored data is limited to 2048 bytes .... so actually some kind of Hash would be required, for testing purpose I raised the limit to see if we can find if/what is failing the state compare here...

laeubi:
Alright, now large properties test fails but 2ndsession succeeds, I'll one more adjust this and then it hopefully is gone.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
```diff
@@ -65,6 +73,10 @@
  * @see IFile#getContentDescription()
  */
 public class ContentDescriptionManager implements IManager, IRegistryChangeListener, IContentTypeManager.IContentTypeChangeListener, ILifecycleListener {
+	private static final String EMPTY = ""; //$NON-NLS-1$
+	private static final String DELIMITER = System.lineSeparator();
+	private static final String SPACE = " "; //$NON-NLS-1$
```
HeikoKlare, line 78: One nitpicky design-related comment: these constants are a bit confusing, as they do not belong to the `ContentDescriptionManager` but to some kind of `PlatformStateDescription` that is handled by `getCurrentPlatformState()`, `getCachedPlatformState()` and the property setter in `shutdown()`. In order to separate concerns and clarify where these constants belong to, you could encapsulate this logic in some (inner, static) `PlatformStateDescription` class.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
```diff
@@ -480,6 +482,37 @@ public synchronized void invalidateCache(boolean flush, IProject project) {
...
+			return HexFormat.of().formatHex(digest.digest(ID.getBytes(StandardCharsets.UTF_8)));
+		} catch (NoSuchAlgorithmException e) {
+			return Integer.toHexString(ID.hashCode());
+		}
+	}
+
+	/**
+	 * @return the cached platform state, only public for testing purpose!
+	 */
+	public String getCachedPlatformState() {
```
HeikoKlare, line 508: Currently there is not test requiring this, so probably this method can be private?

HeikoKlare (commented):
The changes look sound, as the state description should now contain exactly what is required to identify a cache-relevant state change. Thanks for working on this!

I only have two proposals in terms of design, but I am also fine with the changes as they are.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
laeubi: Well actually this just to not make the code scattered with all these `//$NON-NLS-1$` stuff... I can't get why one ever would want to "translate" an empty String or a single space but I'm sure there is a reason I just don't understand :-P

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ContentDescriptionManager.java
laeubi: yes I missed to revert that...

HeikoKlare (approved):
Looks good, thanks!
Let's see if this finally solves the issue :-)

PR #1204, opened 2024-02-14
Suppress warnings about deprecated Preferences in legacy Code

jukzi:


akurtakov:
I don't understand what is won by  hiding warnings. Most important this would hide even other warnings in the future if they happen and go unnoticed. It would be time better spent to actually replace the deprecated Preferences usages with new API used.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocationValidator.java
```diff
@@ -98,6 +98,7 @@ public IStatus validateLinkLocation(IResource resource, IPath unresolvedLocation
 		return validateLinkLocationURI(resource, URIUtil.toURI(location));
 	}
 
+	@SuppressWarnings("deprecation") // org.eclipse.core.runtime.Preferences.getBoolean(String)
```
akurtakov, line 101: As you can see at https://github.com/eclipse-platform/eclipse.platform/blob/e7069db92859e15a44852ff753b9edeec2c1d7ed/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/PreferenceInitializer.java#L79 it's stored in DefaultScope so you can use DefaultScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES).getBoolean to actually move away of deprecated code.

jukzi:
The only benefit of IPreferences over Preferences is the default value handling. Legacy code does not have any problem with it. Its waste of time to refactor Preferences usage to IPreferences, and it is nontrivial. Those warnings encouarage todo such refactorings which might end in unnoticed error due to unnoticed changes.
It's not optimal that other deprecations may be hidden too but no deprecation is crucial as long as it is not for removal - which requires another suppress.

akurtakov:
Issue is that you're hiding EVERY possible deprecation in a method/class not only the current ones. Please don't make us have a discussion like https://github.com/eclipse-platform/eclipse.platform.ui/discussions/1644 again.

akurtakov:
If it's not clear I am acting as a PL here (as per https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines#committer-disagreement-resolution) .

akurtakov:
After https://github.com/eclipse-platform/eclipse.platform.ui/discussions/1644 I thought we had an understanding that sweeping things under that carpet is not gaining the project anything.

PR #1210, opened 2024-02-16
correct links after change of file names of markdown docs

jnerlich:
affected where those which included . and , in the file name

remove broken links


vogella:
Linux build failure seems unrelated, this commit only touches documentation.

Issue #1181, opened 2024-02-07
debugview "Terminate/Disconnect All" sometimes not working


jukzi:
* sometimes the Action is not enabled even though there are multiple launches that could be terminated
* sometimes the Action (when enabled) does not terminate all launches.

![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/8c4925be-d7e0-4f82-9f5e-ec4eb419ef0a)

at the same time the Terminate All in the launches View does not have  the same problems
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/24ba4df2-51b8-417e-ba71-3e008672611b)

reproduce: start multiple times:
```
public class Runner {
	public static void main(String[] args) throws InterruptedException {
		Thread.sleep(100000000L);
	}
}
```

jukzi:
same issue in the console window: Terminate/Disconnect All not available, while the single "Terminate" Button is available
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/42d7a929-788b-46b4-9695-d392cb056be6)

jukzi:
relates to https://bugs.eclipse.org/bugs/show_bug.cgi?id=578877

PR #1212, opened 2024-02-19
fix TerminateAllAction: enabled & terminate #1181

jukzi:
Action was only enabled if any launch was selected - but even if that was already terminated.
Did only terminate launches in the context.

https://github.com/eclipse-platform/eclipse.platform/issues/1181


iloveeclipse:
@jukzi : please update bundle versions if you merge at the begin of the release.
Same problem in https://github.com/eclipse-platform/eclipse.platform/pull/1211.

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/pull/1237

PR #1213, opened 2024-02-27
Fix NPE in TestRunSession#toString and use AtomicReference

laeubi:
Currently if one calls TestRunSession#toString (e.g. in a debugger) a
NPE can occur if the session is not yet started and the fStartTime is
null. Also it is possible that fStartTime is concurrently set.

This checks for this case adjust the message accordingly and also
replace the volatile field with an AtomicReference to ensure the start
time is only ever set by the first thread, alongside with that an
unnecessary synchronized is removed from the method
addTestSessionListener as ListenerList is already synchronized.

### DIFF
--- a/debug/org.eclipse.unittest.ui/src/org/eclipse/unittest/internal/model/TestRunSession.java
+++ b/debug/org.eclipse.unittest.ui/src/org/eclipse/unittest/internal/model/TestRunSession.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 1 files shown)


jukzi (changes requested):
There is no need for AtomicReference
"This class is immutable and thread-safe."
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html

the rest looks good.

laeubi:
> There is no need for AtomicReference "This class is immutable and thread-safe." https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html
> 
> the rest looks good.

The class is threadsafe, but not the access to its instance.

jukzi:
> The class is threadsafe, but not the access to its instance.

That was done by the volatile keyword.

laeubi:
I adjusted the commit message now.

PR #1215, opened 2024-02-27
Correct markdown formating issues in documentation

jnerlich:
Eclipse_Doc_Style_Guide, IRC_FAQ and Javadoc


vogella:
Only documentation so the freeze does not apply.

Thanks @jnerlich

PR #1219, opened 2024-02-29
ResourcesPlugin: Multithreaded lazy start

jukzi:
Reads all projects in parallel. As this was done during ResourcePlugin.start() it had to be deferred as multithreaded classloading during BundleActivator#start(BundleContext) is not supported.

All that happens while splash screen still shown.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
@@ -49,6 +49,7 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.stream.Collectors;
@@ -888,12 +889,61 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 			}
 			return;
 		}
-		IProject[] projects = ((IWorkspaceRoot) resource).getProjects(IContainer.INCLUDE_HIDDEN);
-		for (IProject project : projects)
-			if (project.isAccessible())
-				markerManager.restore(project, generateDeltas, monitor);
+		forEachProjectInParallel(monitor, project -> {
+			if (project.isAccessible()) {
+				markerManager.restore(project, generateDeltas, null);
+			}
+		});
 		if (Policy.DEBUG_RESTORE_MARKERS) {
 			Policy.debug("Restore Markers for workspace: " + (System.currentTimeMillis() - start) + "ms"); //$NON-NLS-1$ //$NON-NLS-2$
+		}
+	}
+
+	@FunctionalInterface
+	public interface CoreConsumer<T> {
+		/**
+		 * Performs this operation on the given argument.
+		 *
+		 * @param t the input argument
+		 */
+		void accept(T t) throws CoreException;
+	}
+
+	private void forEachProjectInParallel(IProgressMonitor m, CoreConsumer<IProject> consumer) throws CoreException {
+		IProject[] projects = workspace.getRoot().getProjects(IContainer.INCLUDE_HIDDEN);
+		if (projects.length == 0) {
+			return;
+		}
+		SubMonitor subMointor = SubMonitor.convert(m, projects.length);
+		IStatus[] stats;
+		// Never use a shared ForkJoinPool.commonPool() as it may be busy with other tasks, which might deadlock.
+		// Also use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
+		// potential SecurityManager, since the threads created by it get no permissions.
+		// See https://github.com/eclipse-platform/eclipse.platform/issues/294
+		ExecutorService executor = new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
+				pool -> new ForkJoinWorkerThread(pool) {
+					// anonymous subclass to access protected constructor
+				}, null, false);
+		try {
+			stats = executor.submit(() -> Arrays.stream(projects).parallel().map(project -> {
+				subMointor.split(1);
+				try {
+					consumer.accept(project);
+				} catch (CoreException e) {
+					return new ResourceStatus(IResourceStatus.FAILED_READ_METADATA, project.getFullPath(),
+							"Error with project " + project.getName(), e); //$NON-NLS-1$
+				}
+				return null;
+			}).filter(Objects::nonNull).toArray(IStatus[]::new)).get();
+		} catch (InterruptedException | ExecutionException e) {
+			List<Runnable> notExecuted = executor.shutdownNow();
+			throw new CoreException(Status.error("Error with " + notExecuted.size() + " projects left", e)); //$NON-NLS-1$//$NON-NLS-2$
+		} finally {
+			executor.shutdown();
+		}
+		if (stats.length > 0) {
+			throw new CoreException(new MultiStatus(ResourcesPlugin.PI_RESOURCES, IStatus.ERROR, stats,
+					"Error with " + stats.length + " projects", null)); //$NON-NLS-1$ //$NON-NLS-2$
 		}
 	}
 
@@ -926,15 +976,14 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 		if (Policy.DEBUG_RESTORE_METAINFO)
 			Policy.debug("Restore workspace metainfo: starting..."); //$NON-NLS-1$
 		long start = System.currentTimeMillis();
-		IProject[] roots = workspace.getRoot().getProjects(IContainer.INCLUDE_HIDDEN);
-		for (IProject root : roots) {
-			//fatal to throw exceptions during startup
-			try {
+		try {
+			forEachProjectInParallel(monitor, root -> {
 				restoreMetaInfo((Project) root, monitor);
-			} catch (CoreException e) {
-				String message = NLS.bind(Messages.resources_readMeta, root.getName());
-				problems.merge(new ResourceStatus(IResourceStatus.FAILED_READ_METADATA, root.getFullPath(), message, e));
-			}
+			});
+		} catch (CoreException e) {
+			// fatal to throw exceptions during startup
+			problems.merge(new ResourceStatus(IResourceStatus.FAILED_READ_METADATA, workspace.getRoot().getFullPath(),
+					"Could not read metadata", e)); //$NON-NLS-1$
 		}
 		if (Policy.DEBUG_RESTORE_METAINFO)
 			Policy.debug("Restore workspace metainfo: " + (System.currentTimeMillis() - start) + "ms"); //$NON-NLS-1$ //$NON-NLS-2$
@@ -1778,34 +1827,7 @@ public class SaveManager implements IElementInfoFlattener, IManager, IStringPool
 		// recurse over the projects in the workspace if we were given the workspace root
 		if (root.getType() == IResource.PROJECT)
 			return;
-		IProject[] projects = ((IWorkspaceRoot) root).getProjects(IContainer.INCLUDE_HIDDEN);
-		// Never use a shared ForkJoinPool.commonPool() as it may be busy with other tasks, which might deadlock.
-		// Also use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
-		// potential SecurityManager, since the threads created by it get no permissions.
-		// See https://github.com/eclipse-platform/eclipse.platform/issues/294
-		ForkJoinPool forkJoinPool = new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
-				pool -> new ForkJoinWorkerThread(pool) {
-					// anonymous subclass to access protected constructor
-				}, null, false);
-		IStatus[] stats;
-		try {
-			stats = forkJoinPool.submit(() -> Arrays.stream(projects).parallel().map(project -> {
-				try {
-					visitAndSave(project);
-				} catch (CoreException e) {
-					return e.getStatus();
-				}
-				return null;
-			}).filter(Objects::nonNull).toArray(IStatus[]::new)).get();
-		} catch (InterruptedException | ExecutionException e) {
-			throw new CoreException(Status.error(Messages.resources_saveProblem, e));
-		} finally {
-			forkJoinPool.shutdown();
-		}
-		if (stats.length > 0) {
-			throw new CoreException(new MultiStatus(ResourcesPlugin.PI_RESOURCES, IStatus.ERROR, stats,
-					Messages.resources_saveProblem, null));
-		}
+		forEachProjectInParallel(null, this::visitAndSave);
 	}
 
 	/**
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
@@ -27,13 +27,21 @@ import java.util.List;
 import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Plugin;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.IJobManager;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.osgi.service.datalocation.Location;
 import org.eclipse.osgi.service.debug.DebugOptions;
 import org.eclipse.osgi.service.debug.DebugOptionsListener;
-import org.osgi.framework.*;
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.util.tracker.ServiceTracker;
 import org.osgi.util.tracker.ServiceTrackerCustomizer;
 
@@ -419,7 +427,7 @@ public final class ResourcesPlugin extends Plugin {
 	 * @see #getSystemEncoding()
 	 */
 	public static String getEncoding() {
-		ResourcesPlugin resourcesPlugin = plugin;
+		ResourcesPlugin resourcesPlugin = getPlugin();
 		if (resourcesPlugin == null) {
 			return getSystemEncoding();
 		}
@@ -475,7 +483,21 @@ public final class ResourcesPlugin extends Plugin {
 	 * @return the single instance of this plug-in runtime class
 	 */
 	public static ResourcesPlugin getPlugin() {
-		return plugin;
+		if (plugin == null) {
+			return null;
+		}
+		return plugin.initialized();
+	}
+
+	private ResourcesPlugin initialized() {
+		ServiceTracker<Location, Workspace> t = instanceLocationTracker;
+		if (t != null && t.size() == 0) {
+			synchronized (t) {
+				// lazy initialize
+				t.open();
+			}
+		}
+		return this;
 	}
 
 	/**
@@ -491,10 +513,8 @@ public final class ResourcesPlugin extends Plugin {
 	 *         class.
 	 */
 	public static IWorkspace getWorkspace() {
-		ResourcesPlugin resourcesPlugin = plugin;
+		ResourcesPlugin resourcesPlugin = getPlugin();
 		if (resourcesPlugin == null) {
-			// this happens when the resource plugin is shut down already... or never
-			// started!
 			throw new IllegalStateException(Messages.resources_workspaceClosedStatic);
 		}
 		Workspace workspace = resourcesPlugin.workspaceInitCustomizer.workspace;
@@ -520,6 +540,7 @@ public final class ResourcesPlugin extends Plugin {
 		// save the preferences for this plug-in
 		getPlugin().savePluginPreferences();
 		plugin = null;
+		instanceLocationTracker = null;
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


jukzi:
```
org.opentest4j.AssertionFailedError: 

Expecting actual:
  ["instance", "configuration", "user", "default"]
to contain exactly (and in same order):
  ["project", "instance", "configuration", "user", "default"]
but could not find the following elements:
  ["project"]

	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at org.eclipse.core.tests.internal.preferences.PreferencesServiceTest.testLookupOrder(PreferencesServiceTest.java:236)
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -889,14 +889,56 @@ protected void restoreMarkers(IResource resource, boolean generateDeltas, IProgr
...
+		 */
+		void accept(T t) throws CoreException;
+	}
+
+	void runParallel(IProject[] projects, IProgressMonitor m, CoreConsumer<IProject> consumer) throws CoreException {
+		// Never use a shared ForkJoinPool.commonPool() as it may be busy with other tasks, which might deadlock.
+		// Also use a custom ForkJoinWorkerThreadFactory, to prevent issues with a
+		// potential SecurityManager, since the threads created by it get no permissions.
+		// See https://github.com/eclipse-platform/eclipse.platform/issues/294
+		ForkJoinPool forkJoinPool = new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
```
laeubi, line 917:
```suggestion
		ForkJoinPool forkJoinPool = Executors.newWorkStealingPool();
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -889,14 +889,56 @@ protected void restoreMarkers(IResource resource, boolean generateDeltas, IProgr
...
+					return e.getStatus();
+				}
+				return null;
+			}).filter(Objects::nonNull).toArray(IStatus[]::new)).get();
+		} catch (InterruptedException | ExecutionException e) {
+			throw new CoreException(Status.error(Messages.resources_saveProblem, e));
+		} finally {
+			forkJoinPool.shutdown();
+		}
+		if (stats.length > 0) {
```
laeubi, line 936: If length == 1 better not throw a multi status

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
```diff
@@ -475,7 +483,12 @@ private static String getSystemEncoding() {
 	 * @return the single instance of this plug-in runtime class
 	 */
 	public static ResourcesPlugin getPlugin() {
-		return plugin;
+		ResourcesPlugin p = plugin;
```
laeubi, line 486: This is wrong as it will prevent the workspace from become available unless someone calls this static method

laeubi (changes requested):
Please don't bind the workspace to static methods for its initialization.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
jukzi: why? i don't know the difference but the comment above explicitly states a custom ForkJoinWorkerThreadFactory had to be used.

laeubi: I can only see this bug reported against java 8 but we are using java 17 now, if it is still be present, it might be better to have a common method e.g. in `Platform` instead of implementing this pattern over and over again.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
jukzi: i added `new Thread(ResourcesPlugin::getPlugin, "Async ResourcesPlugin start").start();` to start() - ok?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
jukzi: Changing that is out of scope of this PR.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
jukzi: Extracases complicate sourcecode for things that should not happen anyway.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: Not really why do you do the init in this method and not in the activate that makes no sense to me at all...if you want to defer the tracking to happen in a different thread (for whatever reason) why not simply wrap the open call in a new thread right there?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
laeubi: What will not happen? That only one error occurs? You will find these "complications" all over the Eclipse codebase as Multistatus has a poor rendering in logfiles and dialogs so should be avoided if not needed.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
laeubi: > Changing that is out of scope of this PR.

You are changing it with this PR so please address it here as it is clearly in scope...

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
jukzi: Other plugins cal the static as soon as the Resourceplugin is started and a synchronize needs to wait for completion to finish.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
jukzi: That there is an Exception may happen only on error but it does not require a beauty full error log. The case that there is an exactly one error is even more rare.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
jukzi: No way. The custom ForkJoinWorkerThreadFactory was added because someone else had a serious problem not long ago and I am not about to change it on the fly.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: This is most likely not correct, it is the other way round. A plugin is starting, calling the static method and this triggers the activator to run and this then is automatically synchronized by the classloader lock already, so why one wants to break this behavior and instead add another synchronization.
These static access is actually bad and has lead to numerous problem in the past, so now requiring this for the startup of the plugin is wrong in many ways and you probably want to better sync on the project loading if projects are accessed.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
laeubi: You don't *change* anything you _add new_ code here. And copy the same problematic code over and over again does not help in fixing the issue, especially if its only because "someone has had a problem but we don't know if its still a problem".

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: Then this PR can't go in in this form. Relying on a static method call to init code that the Activator must perform to "fix" other plugins is simply not acceptable.

jukzi: You still don't understand this PRs changes: It only moves the existing initialization code into a static method which is nothing bad. If you know a better pattern please show that it works. There are enough tests that fail when you get it wrong.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
laeubi: > This PR does not add that code here or copies it around. It only moves the existing code into a method.

Okay, then still it looks strange why this is not a general purpose method somewhere in platform ...

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: The previous code has worked correctly and as you see test fail if you do it work, your are just hiding the problem here because test call that static method, so something is flawed here but moving the init into a completely unrelated static method is defiantly not the solution.

So this all circles back to the useage of the forkjoin pool as you obviously don't want to fork and **join(!)** what you so with .get() ... maybe you want to delay the computation and result retrieval to a later stage and the `CompletableFuture` might be a good choice who knows.

jukzi: Not that it changes anything but i made the initialized() a non static, just to please you.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
```diff
@@ -542,7 +562,7 @@ public void start(BundleContext context) throws Exception {
 				workspaceInitCustomizer);
 		plugin = this; // must before open the tracker, as this can cause the registration of the
 						// workspace and this might trigger code that calls the static method then.
-		instanceLocationTracker.open();
+		new Thread(this::initialized, "Async ResourcesPlugin start").start(); //$NON-NLS-1$
```
laeubi, line 566: This is still wrong, the contract of the Resources plugin is that it tracks the workspace in the activator, as the activation usually (but that's not guaranteed!) happens by a lazy-activation and in this case must be guarded by the classloader lock.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -888,15 +889,64 @@ protected void restoreMarkers(IResource resource, boolean generateDeltas, IProgr
...
+			stats = executor.submit(() -> Arrays.stream(projects).parallel().map(project -> {
+				subMointor.split(1);
+				try {
+					consumer.accept(project);
+				} catch (CoreException e) {
+					return new ResourceStatus(IResourceStatus.FAILED_READ_METADATA, project.getFullPath(),
+							"Error with project " + project.getName(), e); //$NON-NLS-1$
+				}
+				return null;
+			}).filter(Objects::nonNull).toArray(IStatus[]::new)).get();
```
laeubi, line 937: You must not call `get()` here as it blocks, and you don't want to be blocked. Instead you should either return a `Stream<..>` with an `onClose( ... )`, a `Future<..>` or similar or even better a `CompletableFuture<..>` then you can use that in an event driven non blocking way.

laeubi (changes requested):

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: This does not address the issues I mentioned here... If you want to do things async then perform them async.

jukzi: I don't understand any of your recent comments. The initialization does happens async and it fullfills all contracts. and of cause the result of the workspace read has to be joined before using it. Please stop wasting my time with arbitrary requests. Show that the implementation does fail somehow if you are convinced it can.

laeubi: > BTW the trick is that the ResourcesPlugin does not have a single real non-static public method. All methods call the static getPlugin() first.

And this is where your misunderstanding starts from, *any classload* will trigger the plugin and it has public exported classes (so any import/use of such class will trigger it) and it has extension points so any usage of extension points will trigger it, so your assumption that the only entrypoint are static methods is wrong.

jukzi: show me how that makes an issue.

laeubi: > show me how that makes an issue.

The Activator is the "synchronized initializer" of the Resources Bundle, synchronizing on the tracker open is prone to deadlocks, we have had this problem when migrating to the tracker approach and similar attempts to your where evaluated and caused problems. The dangerous part is that it *might* work when there is no workspace chooser, and suddenly fails when there is one, it might even work for a particular configuration (e.g. tests, a local workspace, ...) and deadlock once you install another plugin or use another combination. And as this is the most crucial part of Eclipse IDE I don't want to do any experiments just to save a few milliseconds if even not the problem is well described or understood.

So maybe it would better if you try to describe the issue you see or the problem you try to solve instead on insist on a hack.

jukzi: show me that THIS implementation makes an error, not yours. It does not even be a junit. just put breakpoints such that it fails if executed in a predefined order of your choice. But it boils down to that synchronize() just works: only one thread will do the init without waiting and all others wait.
The problem is simple that Eclipse on Windows needs >10 sec to start and >2 of them are caused by loading the projects.
image
loading them in parallel speeds that  up.

laeubi: > show me that THIS implementation makes an error, not yours. 

If you show me that MY implementation makes an error and not YOURS I might be convinced... longer loading times are not an error, you can for sure make the loading IO of `MarkerManager` parallel, but not by changing the Activator.

jukzi: Optimizing loading time of IDE is a valid and desired topic. Giving only spurious arguments does not help and contradict our code of conduct.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
iloveeclipse: I agree with Cristoph. The code expected to have workspace initialized after activator is done, **not** asynchronously after that.

Please put breakpoint at `org.eclipse.core.internal.resources.Workspace.open(IProgressMonitor)` and check that this is executed in context of `org.eclipse.core.resources.ResourcesPlugin.start(BundleContext)` call.

Initialization of ResourcesPlugin should be changed very carefully, there are different actors playing here and not everything can be validated via automated tests.

iloveeclipse (changes requested):

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ResourcesPlugin.java
laeubi: Claiming other wasting your time and their implementation is failed is also "contradict our code of conduct" .. I gave you several insight and ways to archive your goal differently so please don't claim any contradiction anytime someone does not agree with you.

iloveeclipse: @jukzi : startup performance and functional integrity of code are two different things.

Independently how fast this PR will make startup for some use case, **if** it will be broken for other use case, it is not acceptable.

And as you propose the change, it is your duty to proof it is safe and works, assuming current state is not broken (which is the case).

laeubi:
> I agree with Cristoph. The code expected to have workspace initialized after activator is done, not asynchronously after that.

Yes because of that any code that calls into resource plugin while the tracker.open is runs is not an indication of something is blocking, it means that the synchronization is actually effective, synchronize after that anything can happen (including it works, deadlock, get an exception) and you maybe just notice that randomly.

If one want to improve that there are some ways (and that was already performed by many components but still work in progress):

1. Do not call `ResourcePlugin` methods but pass the objects by reference, as in this special case of init is is allowed for internal code to "early access" these objects (_carefully_).
2. delay the access e.g do not call any code in static initializers of the class but on first access, this can be combined with (1)
3. decouple the code, e.g. by react to the workspace registered (one example is `CheckMissingNaturesListener`) and performing nay action the free of loader locks.
4. Create proxy object that themself only call when they are called the "real" object.
5. ... maybe more ...

jukzi:
> Yes because of that any code that calls into resource plugin while the tracker.open is runs is not an indication of something is blocking, it means that the synchronization is actually effective, synchronize after that anything can happen (including it works, deadlock, get an exception) and you maybe just notice that randomly.

It is not clear what you try say here. ||: Please give a concrete example how it can go wrong :||

vogella:
To continue here, you could ensure that the parallel reading finishes before the Activator is left @jukzi. This would still read in parallel but keep the contract.

jukzi:
> To continue here, you could ensure that the parallel reading finishes before the Activator is left

I would love todo, and i had no plan to not so, but it turns out it is not possible, because using the resourceplugin classes in another thread before the activator finished can (and will very often) cause a deadlock.
And i still see no reason why the workspace should have been fully read when plugin is started. Every method that can access that data will just wait on it and it works.
More over the behaviour of ResourcesPlulugin.start() to take that long not only violates the contract of org.osgi.framework.BundleActivator.start(BundleContext) `This method must complete and return to its caller in a timely manner.` but actually the Classloader does already detects that misbehavior with a timeout and does already skip to further wait on it. So in fact the current implementation is broken to not wait on loading finished, while this PR fixes that.

laeubi:
> So in fact the current implementation is broken to not wait on loading finished, while this PR fixes that.

This PR does not fix that it simply breaks out the classloader synchronization, another thread is not allowed to load *additional* classes ([because that's what lazy-activation is ensuring](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.lifecycle.html#i3270445)) that does not mean that the *current* thread is not allowed to offload some work to other threads.

And that's what I have already written here you can:

1. perform some work (e.g. I/O) in parallel threads from within the activator
2. refactor the caller of any static `ResourcesPlugin#getXXXXX` to something that either uses DI (e.g. SCR with reference to `IWorkspaceService`) or by passing it as a parameter instead.

both things will let you escape out of your classlaoderlock,improves the code and do not require to break the lazy-activation.

jukzi:
BTW the timeout (after 5 or 30sec) looks like
```
org.osgi.framework.BundleException: Unable to acquire the state change lock for the module: osgi.identity; type="osgi.bundle"; version:Version="3.20.200.qualifier"; osgi.identity="org.eclipse.core.resources"; singleton:="true" [id=110] STARTED [STARTED]
	at org.eclipse.osgi.container.Module.lockStateChange(Module.java:373)

```
Can be found for other plugins on stackoverflow aswell and the solution there was for example:

https://git.eclipse.org/r/c/sourceediting/webtools.sourceediting/+/154548/2/core/bundles/org.eclipse.wst.sse.core/src/org/eclipse/wst/sse/core/internal/SSECorePlugin.java

i.e starting a asynchronus thread in the BundleActivator.start() and now guess who create that commit?
So @iloveeclipse please explain why you think it is vaild there but not here.

iloveeclipse:
> please explain why you think it is vaild there but not here.

One should not compare apples with oranges.

Here we are talking about code contract that is as old as Eclipse is, in a bundle that one of the basic building blocks of the SDK with lot of consumers that relied on the existing behavior, implicitly or explicitly. 

Trying to gain some (not even quantified) startup improvement by ignoring possible compatibility break is not an option.

laeubi:
@jukzi As mentioned before you can fork a job for any loading in any component (if it is guaranteed that before the component access the results is joining on the job) and for example **CheckMissingNaturesListener does exactly that**, but here you do fork the whole workspace init into an arbitrary thread and because of that sync on arbitrary other threads.

Ans especially the case where lazy-activation thru calls to these methods happens because on classlaoding is a special case of activation, that is even problematic because it could result in "activation rush", but this is not to be fixed on the _Activator side_, it must be fixed on the _Caller site_!

So again, if any code calls `ResourcePlugin.getXXX` it is expected that before it can do so (that is where the clasloading starts) the activator is first run, and if an instance location is available it is initialized before the classlaoding call returns, that is because otherwise if the instance area is *NOT* set this is an error and must throw an exception. If one does not throw an exception it could happen that there are deadlocks of threads or even UI lockdowns, because of that it is discouraged to call the static methods.

jukzi:
> Here we are talking about code contract that is as old as Eclipse is

The instanceLocationTracker was introduced just 2 year ago. I Ionly added the the async initialisation thread to make that still work even though barely nobody uses it but only the static method are used. For example Eclipse IDE and all its tests work without it, because the Resourceplugin is started in the UI thread as one of the first plugins anyway.

> possible compatibility break

So what exactly do you think is breaking here?

jukzi:
> Resourceplugin is started in the UI thread

More exactly it is implicitly started in IDEApplication.start(IApplicationContext) as the IDEWorkbenchAdvisor uses IContainer  from org.eclipse.core.resources

laeubi:
> The instanceLocationTracker was introduced just 2 year ago.

Please look how the code was before it simply FAILED, but it never did anything async... thats the point and as said you have two options:

1. static access -> slow, discouraged, will run into classlaoder lock
2. "async" / service oriented, will not lock

PR #1228, opened 2024-03-02
Fix tips urls to use https protocol

Phillipus:
- See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1825


vogella:
LGTM, master is open, feel free to merge.

Phillipus:
> Please next time do not merge without updating bundle versions first. It is a pain to cleanup that.

Sorry, I did not know that. Next time, I'll let someone else merge.

iloveeclipse:
It's as simple as: if this is a first change in a release cycle, bundle version must be updated, otherwise we will ship "same" bundle version with different content.

iloveeclipse:
> Should I revert the commit?

No, I will push PR with bundle versions update.

Next time just always **rebase** your PR on master before merging, there is a check and it would fail your PR...

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/pull/1237

merks:
> It's as simple as: if this is a first change in a release cycle, bundle version must be updated, otherwise we will ship "same" bundle version with different content.

Just to be 100% clear and accurate.  There are actually two similar types of problems.  One problem is if literally nothing has changed in the git source of the bundle, but the build produces a jar with different content, e.g., if a new version of the compiler produces different byte code.  In this case the version is exactly the same but with different content.  The other type of change is where you actually change something in the git source such that a new version is actually produced by the build but that new version is different only for the qualifier.  The API tools generally want to avoid such a thing, because the consumer might want to create a bound that forces the version of the bundle with the committed change and one generally doesn't want to create lower bounds that specify a qualifier.  So yes, whenever you change something, you should look at the history of the version in the MANIFEST.MF to see if you need to change it (and look if there is a pom.xml which needs a matching change).   The segment to increment depends on the type of change, so either +100 on the micro segment or +1 on the minor segment (resting the micro to 0) if there change is visible in the API.

Phillipus:
> Next time just always rebase your PR on master before merging, there is a check and it would fail your PR...

Er, that's exactly what I did. There are some new commits even before mine. 😕

Phillipus:
If I get told that merging is OK, and then I rebase my PR and it is merged then I get told "you did it wrong!" then I think I won't bother next time.

iloveeclipse:
> If I get told that merging is OK

This was wrong 

> and then I rebase my PR 

That was done **before** master was opened for merge and **before** master switched to 4.32 version.

> and it is merged then I get told "you did it wrong!" 

If something is wrong, it is wrong.

> then I think I won't bother next time.

Next time:
1) Wait for master is opened
2) Rebase on latest master state **after** master is opened and check build results
3) In doubt, ask & merge

merks:
@Phillipus 

Please don't be offended.  ❤️   Doing everything exactly correctly is not as easy as it ought to be.   We all make mistakes and then we learn from them.

Phillipus:
> That was done before master was opened for merge and before master switched to 4.32 version.

Please take a look at this commit's position in the log - 30ef1d971f6a4fce6720986743fc3b21f8672662 

It comes *after* the "Update for release 4.32" commit and after three other commits. So how does that work?

laeubi:
> > Please next time do not merge without updating bundle versions first. It is a pain to cleanup that.
> 
> Sorry, I did not know that. Next time, I'll let someone else merge.

Ther is a little button in each PR that offers you to **reabse** (don't use the merge option!)the PR on current master if it is outdated, this will then show you any up-to-date issues.

Phillipus:
> Ther is a little button in each PR that offers you to reabse 

Yes, that will be the one I pressed.

iloveeclipse:
> Please take a look at this commit's position in the log - [30ef1d9](https://github.com/eclipse-platform/eclipse.platform/commit/30ef1d971f6a4fce6720986743fc3b21f8672662)
> 
> It comes _after_ the "Update for release 4.32" commit and after three other commits. So how does that work?

**github UI** created the commit by rebasing your original commit on top of master, see the commit date: 
image

What **you** were supposed to do is **before** hitting merge button **rebase** the commit on latest master state, so that tests are executed and all validation tasks would run.

Phillipus:
> Ther is a little button in each PR that offers you to reabse (don't use the merge option!)the PR on current master if it is outdated, this will then show you any up-to-date issues.

> What you were supposed to do is before hitting merge button rebase the commit on latest master state, so that tests are executed and all validation tasks would run.

Seems like conflicting advice.

iloveeclipse:
> Seems like conflicting advice.

All what is needed is to check if your commit is based on latest master or not. If not => rebase but not merge, wait for test results.

laeubi:
@Phillipus I'm lazy and found it convenient to use the github UI others might want to use the git-commandline or JGit, or...

grafik


Here you see the button on the left with rebase & merge, but also on the right the option to rebase it on current master first just in doubt. The best would be to rebase all PRs after open of the stream but currently there is no automatic way to do so.

PR #1233, opened 2024-03-04
[performance] do not normalize locations that are already normalized

jukzi:
When projects or links are added to the workspace the file location is normalized and saved to disk. Good. However it is not needed to do that for those already normalized paths that are loaded back from disk.

Saves some startup time on windows, where name capitalization of each path segment is expensive IO.


jukzi:
image

Issue #673, opened 2023-09-09
The  ContentType implementation of IContentType.isAssociatedWith is case insensitive and I find no case sensitive alternative


jantje:
The  ContentType implementation of IContentType.isAssociatedWith is case insensitive which means that if your content type contains .c and the presented file is a .C file ContentType.isAssociatedWith will return true.

There are 2 locations where ContentType.isAssociatedWith  uses equalsIgnoreCase 
https://github.com/eclipse-platform/eclipse.platform/blob/f385650ab42fb7275ab164643fe2d1db5040750e/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/FileSpec.java#L53
and
https://github.com/eclipse-platform/eclipse.platform/blob/f385650ab42fb7275ab164643fe2d1db5040750e/runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentType.java#L385

Unfortunately for me .c files are associated with c compilers and .C files are associated with C++ compilers which means CDT can not use ContentType.isAssociatedWith and chose to implemented a private method to provide the functionality.
https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.core/src/org/eclipse/cdt/internal/core/CContentTypes.java#L130

As I need a case sensitive implementation of  IContentType.isAssociatedWith I have the option to make the private method of CDT public or have a solution in (I)ContentType.
I'd prefer the second.

Note 1) I did not find any documentation that says the comparison is case insensitive.
Note 2) I think file name comparison is case insensitive as well.
Note 3) I can create a PR but I need some guidance on what the change should look like.

merks:
Why not use  .cpp which is pretty standard and works on case insensitive file systems like on Windows?

jantje:
Because it is not my call. I write a CDT extension not a cpp project.

merks:
Certainly you can't just make the comparison case-sensitive without breaking everyone relying on that not being the case-sensitive.  Given that Window's file system is case-**in**sensitive, relying on case seems like a poor design strategy in the first place.   (You have to ask yourself, why has this not been needed/implementation for the past 18 years?)

At best I could image specifying an additional option for whether to be case-sensitive or not, with not being the default...

laeubi:
[...]

PR #1235, opened 2024-03-04
Add a case sensitive variant of IContentType.isAssociatedWith #673

jantje:
As discussed in #673
Would something like this be acceptable?
I could not test the change so I do not know this will work yet so please do not yet accept this.
What is the advised way to add tests?


jukzi (changes requested):
> As discussed in #673
> Would something like this be acceptable?

maybe @jonahgraham can review as he was involved in the discussion?

> What is the advised way to add tests?

Copy and paste existing tests and adapt them. I typically find related tests in the call hierarchy.

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/runtime/content/IContentType.java
```diff
@@ -167,26 +169,56 @@ public interface IContentType extends IContentTypeSettings {
...
+	 * association is done ignoring case "c"=="C"
 	 *
 	 * @param fileName the file name
-	 * @return <code>true</code> if this content type is associated with
-	 * the given file name, <code>false</code> otherwise
+	 * @return <code>true</code> if this content type is associated with the given
+	 *         file name, <code>false</code> otherwise
 	 * @see #isAssociatedWith(String, IScopeContext)
 	 */
+	@Deprecated
```
merks, line 180: If you deprecate it, there should be javadoc explaining what to use instead.

I wonder though, why deprecate it instead of having a default method that calls the new method with true, which has been the behavior of the past?  But then, you've deprecated two methods and replaced them with a single method, so if I have a string but not a scope context, what will I do to fix the deprecation warning?

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/runtime/content/IContentType.java
```diff
@@ -167,26 +169,56 @@ public interface IContentType extends IContentTypeSettings {
...
 	 * @param fileName the file name
-	 * @param context a preference scope context
-	 * @return <code>true</code> if this content type is associated with
-	 * the given file name, <code>false</code> otherwise
+	 * @param context  a preference scope context
+	 * @return <code>true</code> if this content type is associated with the given
+	 *         file name, <code>false</code> otherwise
 	 * @since 3.1
 	 */
+	@Deprecated
```
merks, line 221: If you deprecate it, there should be javadoc explaining what to use instead.

merks (changes requested):
I'm not sure the deprecations in the API are necessary or helpful.  They're certainly not properly documented.

mickaelistria:
What would be responsible of defining whether we want `caseStrict` or not? Some global preference? A flag in the extension point?

merks:
> What would be responsible of defining whether we want `caseStrict` or not? Some global preference? A flag in the extension point?

That's also a very good question!  I see these callers:

image
image

But that's mostly in the type hierarchy.  It's not at all clear how the following usage would decide true versus false and in fact there isn't good non-deprecated method for it to call:

image

I would have expect the flag to be defined by the content type provider/definition such that this decision is not even something you see in the API.   I.e., just because CDT wants .c and .C to be the same, doesn't mean JDT should also .Java, .JaVa, and all variations there of...

runtime/bundles/org.eclipse.core.contenttype/src/org/eclipse/core/internal/content/ContentType.java
```diff
@@ -377,15 +392,15 @@ boolean hasBuiltInAssociations() {
 		return builtInAssociations;
 	}
 
-	boolean hasFileSpec(IScopeContext context, String text, int typeMask) {
+	boolean hasFileSpec(IScopeContext context, String text, int typeMask, boolean caseStrict) {
```
laeubi, line 395: As we already have a typeMask, why not having a new flag controlling the behavior?

PR #1241, opened 2024-03-07
Version bumps for 4.32 stream

merks:


merks:
The aggregation PR build is failing like this when trying to update the target platform:
> 11:17:45  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.6:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.core.runtime.feature: Only qualifier changed for (org.eclipse.core.runtime.feature.feature.jar/1.4.300.v20240307-0519). Expected to have bigger x.y.z than what is available in baseline (1.4.300.v20240215-1631) -> [Help 1]

PR #1245, opened 2024-03-08
SaveManager: extract forEachProjectInParallel

jukzi:
So that it can later be used by related loops too.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
@@ -49,6 +49,7 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -1778,33 +1779,56 @@ public void visitAndSave(final IResource root) throws CoreException {
...
 		// See https://github.com/eclipse-platform/eclipse.platform/issues/294
-		ForkJoinPool forkJoinPool = new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
+		ExecutorService executor = new ForkJoinPool(ForkJoinPool.getCommonPoolParallelism(),
 				pool -> new ForkJoinWorkerThread(pool) {
 					// anonymous subclass to access protected constructor
 				}, null, false);
-		IStatus[] stats;
 		try {
-			stats = forkJoinPool.submit(() -> Arrays.stream(projects).parallel().map(project -> {
+			stats = executor.submit(() -> Arrays.stream(projects).parallel().map(project -> {
```
laeubi, line 1811:
```suggestion
			stats = executor.submit(() -> Arrays.stream(projects).parallel().unordered().map(project -> {
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -1778,33 +1779,56 @@ public void visitAndSave(final IResource root) throws CoreException {
...
+				subMointor.split(1);
```
laeubi, line 1812: Split is usually not useful in a parallel execution, one should use `slice()` instead but need to call done on it after execution.

Issue #1236, opened 2024-03-05
Texts not externalised


BeckerWdf:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

Have a look at https://github.com/eclipse-platform/eclipse.platform/blob/a17ae73905f80cbd162cb17f96257c62fd63d2ed/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java#L5424C1-L5425C109

You see that the texts are not externalised and therefore cannot be translated.
This was overlooked in https://github.com/eclipse-platform/eclipse.platform/pull/731

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

lathapatil:
> @lathapatil: Can you provide a fix for this?

Yes. Can I create the those strings in ICompareUIConstants.java file ?

BeckerWdf:
> > @lathapatil: Can you provide a fix for this?
> 
> Yes. Can I create those strings in ICompareUIConstants.java file ?

No. Pls. have a look at https://github.com/eclipse-platform/eclipse.platform/blob/master/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.java#L20

PR #1248, opened 2024-03-13
Externalize the 'difference/s' text #1236

lathapatil:
Externalization of text used in "Textmergeviewer.java" related to number of differences shown in toolbar of compare editor.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1236

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
@@ -5421,8 +5421,7 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 		int differenceCount = fMerger.changesCount();
 		if (tbm != null) {
 
-			String label = differenceCount > 1 ? differenceCount + " Differences" //$NON-NLS-1$
-					: differenceCount == 1 ? differenceCount + " Difference" : "No Difference"; //$NON-NLS-1$ //$NON-NLS-2$
+			String label = MessageFormat.format(CompareMessages.TextMergeViewer_differences, differenceCount);
 			LabelContributionItem labelContributionItem = new LabelContributionItem(DIFF_COUNT_ID,
 					label);
 
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.properties
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.properties
@@ -81,6 +81,8 @@ TextMergeViewer_accessible_right={0} Right: {1}
 TextMergeViewer_atBeginning_message= Beginning of document reached. Continue from end?
 TextMergeViewer_accessible_ancestor=Ancestor: {0}
 
+TextMergeViewer_differences={0,choice,0#No Difference|1#1 Difference|1<{0} Differences}
+
 CompareNavigator_atEnd_title= End Reached
 CompareDialog_commit_button=C&ommit
 CompareDialog_error_title=Error Saving Changes
--- a/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.compare; singleton:=true
-Bundle-Version: 3.10.0.qualifier
+Bundle-Version: 3.10.100.qualifier
 Bundle-Activator: org.eclipse.compare.internal.CompareUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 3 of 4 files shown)


team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
```diff
@@ -5421,8 +5421,8 @@ private void updateToolbarLabel() {
 		int differenceCount = fMerger.changesCount();
 		if (tbm != null) {
 
-			String label = differenceCount > 1 ? differenceCount + " Differences" //$NON-NLS-1$
-					: differenceCount == 1 ? differenceCount + " Difference" : "No Difference"; //$NON-NLS-1$ //$NON-NLS-2$
+			String label = differenceCount > 1 ? differenceCount + CompareMessages.TextMergeViewer_differences
+					: differenceCount == 1 ? differenceCount + CompareMessages.TextMergeViewer_difference : CompareMessages.TextMergeViewer_no_difference; //$NON-NLS-1$ //$NON-NLS-2$
```
tomaswolf, line 5425: This is still not good for localization. There are languages in which the plural for two items is different from the plural of > 2 items. Just take a look at https://lingohub.com/blog/pluralization#cldr-overview . Pluralization is hard.

A better way to do this is to leave the choice to the translator. Using `MessageFormat.format` you could define a single message
```
TextMergeViewer_differences={0,choice,0#No differences|1#1 difference|1<{0} differences}
````
and then
```
String label = MessageFormat.format(CompareMessages.TextMergeViewer_differences, differenceCount);
```
This gives translators much more leeway in producing correct translations. It'll also format the number according to localized rules, for instance with thousands separators.

Comment (BeckerWdf) [reply team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java:5425]: I didn't know either that this was possible. Nice
@lathapatil: Can you adapt your PR accordingly?

Comment (lathapatil) [reply team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java:5425]: Sure, I will check on this.

Comment (tomaswolf) [reply team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java:5425]: It is possible, though standard Java MessageFormat still is not powerful enough to handle all cases. A translator would still struggle to get the right plurals for most slavic languages. ICU has full support for doing this right, but unfortunately Eclipse ditched ICU a while ago "because it was too big".

Comment (vogella) [reply team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java:5425]: The main reason why we ditched ICU was that maintenance of this library was not actively down. IIRC there was one part time person working on it in IBM with his focus on other things.

**Comment by BeckerWdf** [comment]
"Expected to have bigger x.y.z than what is available in baseline (3.10.0.v20240208-0728)"

You need to increase the version number of org.eclipse.compare to 3.10.100

**Review by BeckerWdf — CHANGES_REQUESTED** [review]
increase version number

**Comment by lathapatil** [comment]
> increase version number

ok. I have one basic question regarding version bump . When and why we have to increase the version now for this fix ?

**Comment by BeckerWdf** [comment]
> > increase version number
> 
> ok. I have one basic question regarding version bump . When and why we have to increase the version now for this fix ?

The version number has to be increase once each release cycle if changes are made to the code in the bundle.
If you only change the implementation but the API stays unchanged you only change the number after the 2nd dot.
If you change API (e.g. adding a new API method) you change the number after the 1st dot.
If you change API incompatible (e.g. removing API) you change the number in front of the 1st dot.

**Review by BeckerWdf — APPROVED** [review]

**Comment by BeckerWdf** [comment]
AllCompareTests org.eclipse.compare.tests.TextMergeViewerTest is failing. Might this be caused by your change?

**Comment by lathapatil** [comment]
> AllCompareTests org.eclipse.compare.tests.TextMergeViewerTest is failing. Might this be caused by your change?

Yes. Previously it was Uppercase 'D' in Differences  Now It is lowercase 'd'

Either I have to change in testcase or the  properties file .. which one is preferred ?

File: team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.properties [hunk 2]
```diff
@@ -81,6 +81,8 @@ TextMergeViewer_accessible_right={0} Right: {1}
 TextMergeViewer_atBeginning_message= Beginning of document reached. Continue from end?
 TextMergeViewer_accessible_ancestor=Ancestor: {0}
 
+TextMergeViewer_differences={0,choice,0#No difference|1#1 difference|1<{0} differences}
```
BeckerWdf, line 84:
```suggestion
TextMergeViewer_differences={0,choice,0#No Difference|1#1 Difference|1<{0} Differences}
```

BeckerWdf (changes requested):

BeckerWdf:
> > AllCompareTests org.eclipse.compare.tests.TextMergeViewerTest is failing. Might this be caused by your change?
> 
> Yes. Previously it was Uppercase 'D' in the text "Differences" and now It is lowercase 'd'
> 
> Either I have to change in testcase or the properties file .. which one is preferred ?

A it's a "title" / "caption" this should be title-case style.

BeckerWdf (approved):

BeckerWdf (approved):

PR #1254, opened 2024-03-15
MARGINS_DEFAULT leads to IllegalArgumentException

aLiEnHeAd13:
added missing break; before default case to avoid IllegalArgumentException when calling 
createGridLayout(int numColumns, PixelConverter converter, int margins)
with MARGINS_DEFAULT

Fixes https://github.com/JAndrassy/org.eclipse.team.cvs/issues/5

### DIFF
--- a/team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.team.ui; singleton:=true
-Bundle-Version: 3.10.300.qualifier
+Bundle-Version: 3.10.400.qualifier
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Don't get me wrong, but what's the point in duplicating this PR here: https://github.com/eclipse-platform/eclipse.platform/pull/1255 ?

JAndrassy:
> Don't get me wrong, but what's the point in duplicating this PR here: https://github.com/eclipse-platform/eclipse.platform/pull/1255 ?

my bad. this one was first and I missed its existence. sorry

iloveeclipse:
Regression from d2ed9d5ce7f06e0620f57537d835836555708a2a.

iloveeclipse:
Thanks for the patch.

prasanthreddy-git:
Would this patch be available in a eclipse update prior to 2024-06?

iloveeclipse:
No.

merks:
It will first be available for updating the products via this update site

https://download.eclipse.org/releases/milestone/

according to this schedule 

https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/2024-06.md

M1 is April 12.

PR #1263, opened 2024-03-21
The toURI method returns null if the path is not absolute.

pkriens:


iloveeclipse:
@pkriens : could you please add a line to the commit message with this content:

`Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1114`

This would automatically link this PR with the bug, fix the bug on merge and allow easy navigation from IDE to github?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/NullFileStore.java
```diff
@@ -32,11 +32,12 @@ public class NullFileStore extends FileStore {
 
 	/**
 	 * Creates a null file store
+	 * 
 	 * @param path The path of the file in this store
 	 */
 	public NullFileStore(IPath path) {
 		Assert.isNotNull(path);
-		this.path = path;
+		this.path = path.makeAbsolute();
```
laeubi, line 40:
```suggestion
		Assert.isLegal(path.isAbsolute());
```

pkriens: It is a null file system so you only reduce its input domain for no reason whatsoever.

laeubi: Passing a non absolute path seems illegal for this (internal) constructor, so it should be asserted but not fix caller errors automatically.

pkriens:
@iloveeclipse I did and get weird errors in the build now. Also had rather puzzling review from @laeubi 

You can take this as you please or close it but this is more time than I wanted to spend on this.

iloveeclipse:
The small fix should be fine, if it allows https://github.com/eclipse-platform/eclipse.platform/issues/1114 to be fixed. There aren't so many users of this, and @pkriens contribution makes sense for them.

pkriens:
@iloveeclipse thanks.

PR #1268, opened 2024-03-27
Fix tests in LaunchConfigurationTabGroupViewerTest

fedejeanne:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1075

- Remove invalid usages of `java.util.function.Predicate.not`
- Change assertions in `testOnlyDefaultTabInOtherConfigIsActivated` and reactivate the test
- Replace calls to assertj's `matches` method with calls to `isTrue`/`isFalse` to improve readability (also, some calls were wrong)

### DIFF
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/ui/LaunchConfigurationTabGroupViewerTest.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/ui/LaunchConfigurationTabGroupViewerTest.java
@@ -31,7 +31,6 @@ import org.eclipse.debug.ui.ILaunchConfigurationTab;
 import org.eclipse.debug.ui.ILaunchConfigurationTabGroup;
 import org.eclipse.swt.widgets.Display;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 public class LaunchConfigurationTabGroupViewerTest {
@@ -70,8 +69,8 @@ public class LaunchConfigurationTabGroupViewerTest {
 
 		final ILaunchConfigurationTab[] tabs = runOnDialog(createAndSelect1LaunchConfig);
 
-		for (int i = 0; i < tabs.length; i++) {
-			assertThat(((SpyTab) tabs[i])).withFailMessage("tab %s was not initialized", i).matches(SpyTab::isInitialized);
+		for (ILaunchConfigurationTab tab : tabs) {
+			assertThat(((SpyTab) tab)).matches(SpyTab::isInitialized, "should have been initialized");
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


fedejeanne:
Test failure is unrelated: #825

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/ui/LaunchConfigurationTabGroupViewerTest.java
```diff
@@ -71,7 +69,7 @@ public void testAllTabsAreInitializedByDefault() {
 		final ILaunchConfigurationTab[] tabs = runOnDialog(createAndSelect1LaunchConfig);
 
 		for (int i = 0; i < tabs.length; i++) {
-			assertThat(((SpyTab) tabs[i])).withFailMessage("tab %s was not initialized", i).matches(SpyTab::isInitialized);
+			assertThat(((SpyTab) tabs[i]).isInitialized()).withFailMessage("tab %s was not initialized", i).isTrue();
```
HeikoKlare, line 72: Why not keep using `matches`?
```suggestion
			assertThat((SpyTab) tabs[i]).matches(SpyTab::isInitialized, "is initialized");
```

fedejeanne: I find `isTrue` / `isFalse` more intuitive

HeikoKlare: I would argue against that but that's probably rather a matter of personal preference.
Drawback of manual boolean checks with `withFailMessage` is that you have to manually ensure that the object used for creating the failure message matches the object on which the assertion is defined. This is implicitly ensured when using `matches`.

fedejeanne: You mean the 2nd parameter  (the `i`) in the call to `withFailMessage`?

HeikoKlare: I mean the subject under test (i.e., the object passed to `assertThat`). In this case, the test class replicates logic for creating a descriptive name for the subject under test, but it would be better to simply define a proper `toString()` representation for `SpyTab`.

HeikoKlare (approved):
Changes look good. Thanks for correcting the test case!

I like the details provided by the added `toString()` implementation for getting useful information about the tab in case a test fails.

HeikoKlare:
Failing test is unrelated and documented in #128.

PR #1270, opened 2024-03-28
Increment the minor version of org.eclipse.help.ui

merks:
The version range change of the re-exported required bundle org.eclipse.ui mandates a minor version increment.


merks:
Without this change, I have this error in my workspace:

image

merks:
The Windows test failure appears unrelated.

Issue #219, opened 2022-10-11
contenttype.tests  and update.configurator.tests not run for I-builds


akurtakov:

akurtakov:
@HeikoKlare If you can find time to make these test bundles run on I-builds too it would be very nice.

HeikoKlare:
Sure, should be a rather easy thing to do with the experiences from the tips project.

PR #1272, opened 2024-03-28
Improve contenttype test source folder name #219

HeikoKlare:
Makes the org.eclipse.core.contenttype.tests use a different test resources folder than the org.eclipse.core.expressions.tests project.

The folder name is rather misleading than a real issue, as tests are not executed in parallel and the folder is cleaned before each test execution. Still, to avoid any kind of unanticipated issue, let's improve test isolation and avoid conflicting folders of different test projects.

Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/219


HeikoKlare:
Failing test unrelated and documented: #128

PR #1278, opened 2024-03-30
Migrate org.eclipse.update.configurator.tests to JUnit 5

HeikoKlare:
Replace test and test suite annotations as well as assertions with those from JUnit Platform / Jupiter.


akurtakov:
Thanks for the persistence you are showing in modernizing the test suites! It's thankless job but crucial for long term maintenance of the project.

PR #1277, opened 2024-03-29
Add support for "editor" content types for compare editor

iloveeclipse:
Extend `contentMergeViewers` extension point: add an attribute to consider "linked" editor content type associations. The code checks if contentMergeViewers contribution refers to such "linked" editor id in extension, and check if editor supports other content types. In case there are more content types supported by same "linked" editor, these would be automatically considered as valid contentTypes for the contentMergeViewers.

Updated CompareViewerSwitchingPane to consider more precise title calculations for ContentMergeViewer instances (where concrete type info is available after instantiation of a particular viewer).

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1747

PR #1283, opened 2024-04-04
Fix tests expecting TextMergeViewer class only

iloveeclipse:
TextMergeViewer was expected by SaveableCompareEditorInputTest but more specific implementation (subclass of it) - GenericEditorMergeViewer was seen by the test after changes made to allow more specific viewer to "win" in the compare editor for text content.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1747 
See https://github.com/eclipse-platform/eclipse.platform/pull/1277


akurtakov:
Thanks!

iloveeclipse:
The only issue I see is unrelated: `baseline and build for org/eclipse/compare/internal/Utilities.class have different contents`

PR #1285, opened 2024-04-04
Don't return null in JobManager::createMonitor

fedejeanne:
Revert unintentional change made in 9c3525cbd3fbd1e2618e8c9edf4871f5c8d70cb2 and don't return null when creating a progress monitor. This is part of the contract of the method.

Replaces #1282 

Contributes to
https://github.com/eclipse-platform/eclipse.platform/issues/664


iloveeclipse:
Sorry, this PR doesn't revert anything. Your original PR is already merged o master!

HeikoKlare:
You cannot amend commit in master. And there is no concept of automatically "replacing one commit with another" either, like the commit in this PR tries to do. If you rebase this commit on master, the commit will simply be removed (as it's then empty) and the the PR will be closed. A reverting PR needs to actively revert changes and not reapply the ones that shall remain in comparison to some other commit.

fedejeanne:
Replaced by #1286

Issue #1239, opened 2024-03-07
LaunchConfigurationTabGroupViewerTest fails on all platforms


jukzi:
I20240306-1800
for example https://download.eclipse.org/eclipse/downloads/drops4/I20240306-1800/testresults/html/org.eclipse.debug.tests_ep432I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html
```
java.lang.AssertionError:
Expecting actual:org.eclipse.debug.tests.ui.SpyTab$SpyTabA@20bc4c09to match 'is activated' predicate.
at org.eclipse.debug.tests.ui.LaunchConfigurationTabGroupViewerTest.testFirstTabIsActivatedByDefault(LaunchConfigurationTabGroupViewerTest.java:86)at java.base/java.lang.reflect.Method.invoke(Method.java:580)

```

may relate to https://github.com/eclipse-platform/eclipse.platform/issues/1075

iloveeclipse:
> may relate to #1075

No, it is a regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044.
Reverting SWT change fixes the test.

iloveeclipse:
Fixed in https://download.eclipse.org/eclipse/downloads/drops4/I20240307-1800/testResults.php

PR #46, opened 2022-07-15
Adapt to JUnit from Maven Central

mickaelistria:

PR #1287, opened 2024-04-04
testFirstTabIsActivatedByDefault is no more passing with SWT fix for issue #46

deepika-u:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1239

With the swt fix for [https://github.com/eclipse-platform/eclipse.platform.swt/issues/46](https://github.com/eclipse-platform/eclipse.platform.swt/issues/46) via pr https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044 we are seeing propagateTabActivation() is not being called on all platforms.

For more details please see -> https://github.com/eclipse-platform/eclipse.platform.swt/issues/46#issuecomment-2037102845

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
@@ -1180,10 +1180,7 @@ public class LaunchConfigurationTabGroupViewer {
 	 * @return if the dialog can launch in its current state
 	 */
 	public boolean canLaunch() {
-		if(fInitializingTabs) {
-			return false;
-		}
-		if (getWorkingCopy() == null) {
+		if (fInitializingTabs || (getWorkingCopy() == null)) {
 			return false;
 		}
 		try {
@@ -1271,11 +1268,7 @@ public class LaunchConfigurationTabGroupViewer {
 	 * @return the error message for the tab
 	 */
 	public String getErrorMesssage() {
-		if (fInitializingTabs) {
-			return null;
-		}
-
-		if (getWorkingCopy() == null) {
+		if (fInitializingTabs || (getWorkingCopy() == null)) {
 			return null;
 		}
 		try {
@@ -1452,12 +1445,14 @@ public class LaunchConfigurationTabGroupViewer {
 		fCurrentTabIndex = fTabFolder.getSelectionIndex();
 
 		ILaunchConfigurationTab[] tabs = getTabs();
-		if (previousTabIndex == fCurrentTabIndex || tabs == null || tabs.length == 0
+		if (tabs == null || tabs.length == 0
 				|| previousTabIndex > (tabs.length - 1)) {
 			return;
 		}
 
-		propagateTabDeactivation(previousTabIndex);
+		if (previousTabIndex != fCurrentTabIndex) {
+			propagateTabDeactivation(previousTabIndex);
+		}
 
 		propagateTabActivation();
 	}


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
```diff
@@ -1455,7 +1455,7 @@ protected void handleTabSelected() {
 		fCurrentTabIndex = fTabFolder.getSelectionIndex();
 
 		ILaunchConfigurationTab[] tabs = getTabs();
-		if (previousTabIndex == fCurrentTabIndex || tabs == null || tabs.length == 0
+		if (tabs == null || tabs.length == 0
```
tomaswolf, line 1458: There is the public API `setActiveTab()`. If that is ever called with the index of the already active tab, then your code change will first deactivate and then re-active that tab, which is probably not good.

fedejeanne: If I had to guess, I'd say the reason that this `if` returns `true` (_i.e._ the reason why `previousTabIndex == fCurrentTabIndex`) is because the change introduced in 5ca254b14421e64b122eed111b35621a1ef8e144 (which got reverted in 2a51e8c925354594e42fc758b332336db937e01c) is in direct conflict with the call in line 1680 of (`fTabFolder.setSelection(index);`)

And I agree with @tomaswolf : this change will have unintended (bad) implications

deepika-u: > There is the public API `setActiveTab()`. If that is ever called with the index of the already active tab, then your code change will first deactivate and then re-active that tab, which is probably not good.

Now that previousTabIndex and fCurrentTabIndex are same, that condition is updated. So accordingly when they are not same only we need to call the deactivation followed by activation. 

Updated as per your review comments, Thanks for your review.

fedejeanne: If I apply this PR then calling `setActiveTab(1)` twice would activate the 2nd tab twice without deactivating it in-between. IMO also not good

Also, if I apply this PR and re-apply https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044 then `testOnlyDefaultTabInOtherConfigIsActivated` breaks

I propose we define the expected behavior of `LaunchConfigurationTabGroupViewer` and its `setActiveTab(...)` and go from there. Once this is done, I can write some tests to check this behavior.

I'll start with:
1. All tabs are initialized upon creation of a `LaunchConfigurationTabGroupViewer`
2. The default tab is also activated upon creation
3. The default tab may come from the `CTabFolder`. If it does, it should be honored. If it doesn't, the 1st tab (index `0`) is the default tab
4. Taking the default tab from the `CTabFolder` should **not** lead to unnecessary activation/deactivation/initialization of tabs
5. Upon creation of a `LaunchConfigurationTabGroupViewer` only the following should occur
* All tabs are initialized (exactly once)
* The default tab (whether coming from `CTabFolder` or not) is activated exactly once
* No tab is deactivated
6. Activating the already active tab (_i.e._ calling `setActiveTab(0)` after creation) should neither deactivate the tab nor activate it again

Points 1, 2, 5 and 6 already hold before applying this PR + https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044 (tests are missing for 5 and 6 though)
Points 3 and 4 should not break the other ones.

WDYT?

fedejeanne (changes requested):
The current changes would break some existing functionality. I'd like to establish some expected behavior for `LaunchConfigurationTabGroupViewer` and for its `setActiveTab` method before moving forward with this PR and with the fix in SWT (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044)

deepika-u:
> The current changes would break some existing functionality. I'd like to establish some expected behavior for `LaunchConfigurationTabGroupViewer` and for its `setActiveTab` method before moving forward with this PR and with the fix in SWT ([eclipse-platform/eclipse.platform.swt#1044](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1044), [eclipse-platform/eclipse.platform.swt#1164](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1164))

Hi @fedejeanne 
I tried the below steps, can you try the same and confirm please? Sorry for the late reply as i was facing some workspace issue i took some time to get that sorted out. 

1. I have updated to latest eclipse(7th Apr build) and all the repos upto date
-------
Eclipse SDK
Version: 2024-06 (4.32)
Build id: I20240407-1800
OS: Windows 11, v.10.0, x86_64 / win32
Java vendor: Oracle Corporation
Java runtime version: 22+36-2370
Java version: 22
-------

2. Applied no prs here and seeing that "all the test cases are passing". 
-> 
image

3. Applied swt pr alone # 1164 and seeing below couple of failures. 
->
image

4. Applied swt pr # 1164 , platform pr #1287 and seeing "all test cases are passing". Then proposed the pr.
->
image

These are the only steps i tried and as the screenshots show, with both the prs in place i am seeing all the test cases are passing. Please confirm if the same works for you?

fedejeanne:
> Please confirm if the same works for you?

You're missing the point, it's important that no unnecessary (de)activations of tabs occur once you apply the fix (see my comment: https://github.com/eclipse-platform/eclipse.platform/pull/1287/files#r1553193527).

I added some tests to make sure of that: #1298 

Those test don't pass if you apply https://github.com/eclipse-platform/eclipse.platform.swt/issues/46 + #1287 

image

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
fedejeanne: Fixed by https://github.com/eclipse-platform/eclipse.platform/pull/1346

fedejeanne:
This was resolved in https://github.com/eclipse-platform/eclipse.platform/pull/1346

deepika-u:
Thanks fedejeanne.

Issue #1290, opened 2024-04-08
Deadlock in CharsetDeltaJob on workspace shutdown 


sratz:
We have been seeing increased amounts of deadlocks during workbench shutdown in our test suites in `org.eclipse.core.internal.resources.CharsetDeltaJob` (full [theaddump.txt](https://github.com/eclipse-platform/eclipse.platform/files/14903799/theaddump.txt)  attached).

The change in #631 seems to be the cause of this.

The relevant parts of the thread-dump are:
```
java.lang.Exception: ThreadDump for thread "Framework stop - Equinox Container: d9b0deee-892c-4a57-ba9c-02211f659f27"
	at java.base@17.0.10/java.lang.Object.wait(Native Method)
	at org.eclipse.core.internal.jobs.Semaphore.acquire(Semaphore.java:45)
	at org.eclipse.core.internal.jobs.JobManager.join(JobManager.java:994)
	at org.eclipse.core.internal.jobs.InternalJob.join(InternalJob.java:355)
	at org.eclipse.core.runtime.jobs.Job.join(Job.java:531)
	at org.eclipse.core.internal.resources.CharsetDeltaJob.shutdown(CharsetDeltaJob.java:243)
	at org.eclipse.core.internal.resources.CharsetManager.shutdown(CharsetManager.java:503)
	at org.eclipse.core.internal.resources.Workspace.shutdown(Workspace.java:2558)
	at org.eclipse.core.internal.resources.Workspace.close(Workspace.java:721)
	at org.eclipse.core.resources.ResourcesPlugin$WorkspaceInitCustomizer.removedService(ResourcesPlugin.java:599)
	at org.eclipse.core.resources.ResourcesPlugin$WorkspaceInitCustomizer.removedService(ResourcesPlugin.java:1)
	at org.osgi.util.tracker.ServiceTracker$Tracked.customizerRemoved(ServiceTracker.java:973)
	at org.osgi.util.tracker.ServiceTracker$Tracked.customizerRemoved(ServiceTracker.java:1)
	at org.osgi.util.tracker.AbstractTracked.untrack(AbstractTracked.java:342)
	at org.osgi.util.tracker.ServiceTracker.close(ServiceTracker.java:383)
	at org.eclipse.core.resources.ResourcesPlugin.stop(ResourcesPlugin.java:518)
	at org.eclipse.osgi.internal.framework.BundleContextImpl$3.run(BundleContextImpl.java:893)
[...]

PR #1302, opened 2024-04-12
fix CharsetDeltaJob Deadlock #1290

jukzi:
Unlike documented Workspace.prepareOperation() blocks on WorkManager.lock.acquire() even without ISchedulingRule

https://github.com/eclipse-platform/eclipse.platform/issues/1290

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
@@ -76,6 +76,7 @@ public class CharsetDeltaJob extends Job implements IContentTypeManager.IContent
 	public CharsetDeltaJob(Workspace workspace) {
 		super(Messages.resources_charsetBroadcasting);
 		this.workspace = workspace;
+		setRule(workspace.getRoot()); // make sure workspace.prepareOperation() does not block
 	}
 
 	private void addToQueue(ICharsetListenerFilter filter) {
@@ -237,10 +238,10 @@ public class CharsetDeltaJob extends Job implements IContentTypeManager.IContent
 
 	public void shutdown() {
 		try {
-			// try to prevent execution of this job to avoid prevent "already shutdown.":
+			// try to prevent execution of this job to avoid "already shutdown.":
 			cancel();
 			// if job is already running wait for it to finish:
-			join();
+			join(3000, null);
 		} catch (InterruptedException e) {
 			// ignore
 		}
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
@@ -2382,11 +2382,14 @@ public class Workspace extends PlatformObject implements IWorkspace, ICoreConsta
 	}
 
 	/**
-	 * Called before checking the pre-conditions of an operation.  Optionally supply
-	 * a scheduling rule to determine when the operation is safe to run.  If a scheduling
-	 * rule is supplied, this method will block until it is safe to run.
+	 * Called before checking the pre-conditions of an operation. Optionally supply
+	 * a scheduling rule to determine when the operation is safe to run. If a
+	 * scheduling rule is supplied, this method will block until it is safe to run.
+	 * Even if no scheduling is supplied this method blocks until no other workspace
+	 * operation concurrently runs.
 	 *
-	 * @param rule the scheduling rule that describes what this operation intends to modify.
+	 * @param rule the scheduling rule that describes what this operation intends to
+	 *             modify.
 	 */
 	public void prepareOperation(ISchedulingRule rule, IProgressMonitor monitor) throws CoreException {
 		try {


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
```diff
@@ -237,7 +238,7 @@ public void setDisabled(boolean disabled) {
 
 	public void shutdown() {
 		try {
-			// try to prevent execution of this job to avoid prevent "already shutdown.":
+			// try to prevent execution of this job to avoid "already shutdown.":
 			cancel();
 			// if job is already running wait for it to finish:
 			join();
```
iloveeclipse, line 244: Why not remove `join()`? The job checks for cancellation, so why should we try to block here?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
```diff
@@ -76,6 +76,7 @@ interface ICharsetListenerFilter {
 	public CharsetDeltaJob(Workspace workspace) {
 		super(Messages.resources_charsetBroadcasting);
 		this.workspace = workspace;
+		setRule(workspace.getRoot()); // make sure workspace.prepareOperation() does not block
```
iloveeclipse, line 79: I'm not sure if this is a good idea, just because I cannot foresee which code might be now affected by this job being waiting for a lock or other way around.

A more safe fix would be revert the part of the original change that introduced `join()`.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
jukzi: The job throws an exception when the workspace is shut down while the job is running

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
jukzi: While i am not 100% sure too effectivly the Jobs run() did wait on the workspace lock anyway. If cancel+join is removed we would be back to the old exception or - depending how far the job is  - the warning that a job is still running ... the plugin should cancel it before shutdown. The fix just did follow the guidelines :-)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
iloveeclipse: And how `join()` should help here? By avoiding to process further with shutdown before the job finishes? Shouldn't we just catch / ignore the error if the case?

jukzi: The join() ensures the job has finished (either canceled or completetd) before workspace is shutdown. Its pretty reproducibly with breakpoints that each of the commands helps to have an errorfree nonblocking shutdown.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetDeltaJob.java
iloveeclipse: My point is: I've never heard about issues on shutdown before, so this seem to be a corner case and from what I understand just of "cosmetic" kind. Not nice, but doesn't harm anyone.

To fix a minor issue, we added the join() and run in deadlocks. 
This is not minor issue anymore.

To fix deadlocks, we add more locks due the rule. 
This will most likely lead to more issues at runtime, as the job will run at different times due the rule and other jobs will run also at different times because waiting on this rule taken.

I'm not sure if all these changes with all the possible implications are worth to fix a minor issue logged at shutdown.

jukzi: It always was only a problem on junits, the deadlock only occur on junits too. The benefit to fix minor logging issues in tests is that the logfiles get rid of unrelated issues so that its easier to find the real issues.
https://github.com/eclipse-platform/eclipse.platform/blob/c5d730d4ae3fc16418df8165b7f46cb36daddf82/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java#L724
is very explict that the implementation was wrong. Keeping a wrong implementation can also mislead others to copy&paste wrong patterns.

iloveeclipse: That is not nice, and `cancel()` may help here. 
But `join()` to get rid of the message? It is simply not adequate solution, the side effects are worse as the benefit.

laeubi: Depending on how long one assumes the job should run a *timed join* might be better as it at least would exit somehow in case of deadlock situation.

jukzi: Even though there is no precedent (all oher stop/shutdowns like egit/jdt use cancel+join) a timeout seems to be a valid compromise. I updated the PR.

iloveeclipse: But with timeout we don't need the rule anymore. Could we please remove it?

jukzi: Without the rule it would needlessly wait the timeout time whenever it deadlocked before. The rule prevents that the timeout is ever reached (if the computer does not lag) and is used in similar Jobs (for example AutoBuildJob, ApiAnalysisMarkersJob, FlushJob ... ) as well. Please note that the CharsetDeltaJob is internal only and does nothing more then create Markers and update the int ResourceInfo.charsetAndContentId. I don't see any dangerous things or third party code that is called from there.

sratz: This looks good in general. However, in `AutoBuildJob` and `FlushJob` the rule is set in the Job constructor but also inside the `run()` method the same rule is used for the workspace operation. Should `CharsetDeltaJob` also do this?

jukzi: That looks unnecessary superfluous to me. I don't know why it is like that.

sratz (approved):
Looks good to me. Any objections to merge?

sratz:
> Yes, but they were mostly ignored so far.

I though they were addressed by @jukzi:

Adding the rule does not make things 'worse', but better:
- By telling the Job upfront what rule we need, we prevent it from being started/run too early and get stuck / deadlock *inside* its run method when acquiring the same rule anyways.
- When things are very quick (junit) a `cancel()` so early is then trivial, as this is done by the Job framework before the Job actually enters the running state and we don't risk a deadlock anymore
- If we can acquire the (outer) Job rule, we already have the same rule required inside the Job anyways. Plus, the what the Job does is self-contained, so no risk of any third-party code slipping in.

jukzi:
> I though they were addressed

proper summary.

PR #1305, opened 2024-04-12
Use diamond operator wherever applicable

HeikoKlare:
Applies the cleanup rule for using the diamond operator where applicable in all projects to be conform with defined save actions.


vogella:
LGTM

PR #1306, opened 2024-04-12
Use primitive types instead of wrapper types

HeikoKlare:
Applies the cleanup rule for using primitive instead of wrapper types in all projects to be conform with defined save actions.

### DIFF
--- a/ua/org.eclipse.help.base/src_demo/org/apache/lucene/demo/html/Entities.java
+++ b/ua/org.eclipse.help.base/src_demo/org/apache/lucene/demo/html/Entities.java
@@ -32,9 +32,9 @@ public class Entities {
 	start++;
 	radix = 16;
       }
-      Character c =
-	Character.valueOf((char)Integer.parseInt(entity.substring(start), radix));
-      return c.toString();
+      char c =
+	(char)Integer.parseInt(entity.substring(start), radix);
+      return Character.toString(c);
     } else {
       String s = decoder.get(entity);
       if (s != null)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/memory/renderings/RenderingsUtil.java
```diff
@@ -557,21 +557,21 @@ public static byte[] convertHexStringToByteArray(String str, int numBytes, int n
...
+				bytes[i] = myByte;
 			}
 			else
 			{
-				Byte myByte = Byte.valueOf(oneByte, 16);
-				bytes[i] = myByte.byteValue();
+				byte myByte = Byte.parseByte(oneByte, 16);
+				bytes[i] = myByte;
 			}
 		}
```
tomaswolf, line 576: Unrelated to the application of the automated cleanup: this looks like a very round-about way to convert a hex string to a byte. What's wrong with `bytes[i] = Integer.parseInt(oneByte, 16).byteValue()`?

HeikoKlare: > this looks like a very round-about way to convert a hex string to a byte

Yes, that's what this code seems to be about: it converts a substring containing a single byte representation to a byte. But I think I do not understand your proposal yet. What would be the benefit of replacing `Byte.parseByte(oneByte, 16)` with `Integer.parseInt(oneByte, 16).byteValue()`?

fedejeanne: > What's wrong with bytes[i] = Integer.parseInt(oneByte, 16).byteValue()

@tomaswolf  `Integer.parseInt` returns an `int` (primitive) so `Integer.parseInt(oneByte, 16).byteValue()` doesn`t even compile :-\

HeikoKlare: > `Integer.parseInt` returns an `int` (primitive) so `Integer.parseInt(oneByte, 16).byteValue()` doesn`t even compile :-\

You can box the returned `int` into an `Integer`.

fedejeanne: > > `Integer.parseInt` returns an `int` (primitive) so `Integer.parseInt(oneByte, 16).byteValue()` doesn`t even compile :-\
> 
> You can box the returned `int` into an `Integer`.

That would by an overkill (parsing + boxing) and would also skip the range validation since `Integer::byteValue` simply casts to `byte` while `Byte::parseByte` does validate the range (and throw a `NumberFormatException` if the parsed byze is out of range).

I'd go with the original form: `Byte.parseByte(oneByte, 16)`

tomaswolf: I meant the whole code looks way too complicated. I think one can replace all of this from line 560 to line 575 with just
`bytes[i] = (byte) Integer.parseInt(oneByte, 16).intValue();`.

HeikoKlare: > would also skip the range validation since `Integer::byteValue` simply casts to `byte` while `Byte::parseByte` does validate the range

The input already is a byte value, which is why the range check is not necessary.

> I meant the whole code looks way too complicated. I think one can replace all of this from line 560 to line 575

You are right. I first thought it would make a difference depending on whether the string to parse contains a signed or unsigned value, but that's not the case. So your proposal totally makes sense to me. So I would suggest to provide a PR to simplify this code. Do want to provide a PR or shall I, @tomaswolf?

tomaswolf: Go ahead; I currently have no development setup for Eclipse Platform and am too busy in other areas. Just noticed this by chance.

HeikoKlare: Thank you, @tomaswolf! I have created #1316 with the simplification. Could you please check if that properly reflects your proposal? I will not further process this PR until the other is merged, so that rebasing this on top will removing the cleanup for the code part that we have discussed.

PR #1309, opened 2024-04-12
Use boolean value rather than comparison against true/false

HeikoKlare:
Applies the cleanup rule to use boolean values in conditions directly instead of comparing them against true or false to all project in order to be conform with defined save actions.


vogella:
LGTM, thanks

PR #1310, opened 2024-04-14
Simplify lambda expression and method reference syntax

HeikoKlare:
Applies the cleanup action to simplify lambda expression and method reference syntax in all projects to conform to defined save actions. This includes the following:
* replace simple method calls in lambda expressions by method references
* remove unnecessary parentheses around lambda input argument


akurtakov:
Failing Mac test is unrelated.

PR #1311, opened 2024-04-14
Add missing SWT for Windows on Arm64 (WoA) fragment

chirontt:
Add missing SWT for Windows on Arm64 (WoA) fragment to the compare win32 plug-in.

### DIFF
--- a/team/bundles/org.eclipse.compare.win32/META-INF/p2.inf
+++ b/team/bundles/org.eclipse.compare.win32/META-INF/p2.inf
@@ -4,3 +4,8 @@ requires.0.name = org.eclipse.swt.win32.win32.x86_64
 #requires.0.range = [$version$,$version$]
 requires.0.filter = (&(osgi.os=win32)(osgi.ws=win32)(osgi.arch=x86_64))
 
+requires.1.namespace = org.eclipse.equinox.p2.iu
+requires.1.name = org.eclipse.swt.win32.win32.aarch64
+#requires.1.range = [$version$,$version$]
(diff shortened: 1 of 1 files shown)


HannesWell:
Because of the discussion in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577#issuecomment-2101533539, can you please split this into multiple parts, one for the addition of a `org.eclipse.core.filesystem.win32.aarch64` and one for a `org.eclipse.core.resources.win32.aarch64` so that we can discuss them separately?

The change of `org.eclipse.compare.win32/META-INF/p2.inf` can be applied now, so having that in a third PR would allow us to land it immediately.

chirontt:
> The change of `org.eclipse.compare.win32/META-INF/p2.inf` can be applied now, so having that in a third PR would allow us to land it immediately.

Yeah, I'm making this PR to be the third one of the three PRs.

HannesWell (approved):
Looks good, thank you.
But when reading through [Bug 361901](https://bugs.eclipse.org/bugs/show_bug.cgi?id=361901) I wonder if that p2.inf file is even necessary at all?
The SWT host bundle already has something like that so I actually wouldn't expect that to be necessary again?
@merks can you tell better?

Nevertheless I'll submit this now and we can still remove the entire file later if we conclude that it's obsolete.

merks:
I imagine each separate host bundle can benefit from filtered requirements on the known fragments.  Without this, one must rely on some feature encoding those requirements or replicating them somewhere downstream.

PR #1312, opened 2024-04-15
MT hardening ProjectDescription.natures

jukzi:
It is reported that sometimes ProjectDescription#hasNature(String) returns wrong values. While it is not clear how that happens chances are that this a concurrency problem: ProjectDescription can be used from arbitrary threads and yet totally missed any synchronization.
https://github.com/eclipse-pde/eclipse.pde/issues/1185

"when multiple threads share mutable data, each thread that reads or writes data must perform synchronization."
https://ahdak.github.io/blog/effective-java-part-10/

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
@@ -775,8 +775,8 @@ public class Project extends Container implements IProject {
 		projectDesc.setVariableDescriptions(internalDesc.getVariables());
 
 		//clear properties, markers, and description for the new project, because they shouldn't be copied.
-		info.description = null;
-		info.natures = null;
+		info.discardDescription();
+		info.discardNatures();
 		info.setMarkers(null);
 		info.clearSessionProperties();
 	}
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectDescription.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectDescription.java
@@ -19,13 +19,35 @@
 package org.eclipse.core.internal.resources;
 
 import java.net.URI;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import org.eclipse.core.filesystem.URIUtil;
 import org.eclipse.core.internal.events.BuildCommand;
 import org.eclipse.core.internal.utils.FileUtil;
-import org.eclipse.core.resources.*;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.resources.IBuildConfiguration;
+import org.eclipse.core.resources.ICommand;
+import org.eclipse.core.resources.IDynamicReferenceProvider;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectDescription;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
 
 public class ProjectDescription extends ModelObject implements IProjectDescription {
 	// constants
@@ -99,9 +121,16 @@ public class ProjectDescription extends ModelObject implements IProjectDescripti
 	protected HashMap<String, VariableDescription> variableDescriptions = null;
 
 	// fields
-	protected URI location = null;
-	protected String[] natures = EMPTY_STRING_ARRAY;
-	protected URI snapshotLocation = null;
+	private volatile URI location = null;
+	/**
+	 * synchronizes {@link #natures}
+	 */
+	private final Object naturesMutex = new Object();
+	/**
+	 * synchronized by {@link #naturesMutex}
+	 */
+	private String[] natures = EMPTY_STRING_ARRAY;
+	private volatile URI snapshotLocation = null;
 
 	public ProjectDescription() {
 		super();
@@ -428,9 +457,12 @@ public class ProjectDescription extends ModelObject implements IProjectDescripti
 	}
 
 	public String[] getNatureIds(boolean makeCopy) {
-		if (natures == null)
-			return EMPTY_STRING_ARRAY;
-		return makeCopy ? (String[]) natures.clone() : natures;
+		synchronized (naturesMutex) {
+			String[] n = natures;
+			if (n == null)
+				return EMPTY_STRING_ARRAY;
+			return makeCopy ? n.clone() : natures;
+		}
 	}
 
 	@Override
@@ -543,7 +575,7 @@ public class ProjectDescription extends ModelObject implements IProjectDescripti
 			return true;
 		if (!Arrays.equals(staticRefs, description.getReferencedProjects(false)))
 			return true;
-		if (!Arrays.equals(natures, description.getNatureIds(false)))
+		if (!Arrays.equals(getNatureIds(false), description.getNatureIds(false)))
 			return true;
 
 		HashMap<IPath, LinkedList<FilterDescription>> otherFilters = description.getFilters();
@@ -849,7 +881,10 @@ public class ProjectDescription extends ModelObject implements IProjectDescripti
 
 	@Override
 	public void setNatureIds(String[] value) {
-		natures = value.clone();
+		String[] n = Arrays.copyOf(value, value.length);
+		synchronized (naturesMutex) {
+			natures = n;
+		}
 	}
 
 	@Override
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectInfo.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectInfo.java
@@ -23,17 +23,17 @@ import org.eclipse.core.runtime.content.IContentTypeMatcher;
 
 public class ProjectInfo extends ResourceInfo {
 
-	/** The description of this object */
-	protected ProjectDescription description;
+	/** The description of this object, synchronized access */
+	private ProjectDescription description;
 
-	/** The list of natures for this project */
-	protected HashMap<String, IProjectNature> natures;
+	/** The list of natures for this project, synchronized access */
+	private HashMap<String, IProjectNature> natures;
 
 	/** The property store for this resource (used only by the compatibility fragment) */
-	protected Object propertyStore;
+	private volatile Object propertyStore;
 
 	/** The content type matcher for this project. */
-	protected IContentTypeMatcher matcher;
+	private volatile IContentTypeMatcher matcher;
 
 	/**
 	 * Discards stale natures on this project after project description
@@ -41,6 +41,10 @@ public class ProjectInfo extends ResourceInfo {
 	 */
 	public synchronized void discardNatures() {
 		natures = null;
+	}
+
+	synchronized void discardDescription() {
+		description = null;
 	}
 
 	/**
@@ -68,7 +72,7 @@ public class ProjectInfo extends ResourceInfo {
 	/**
 	 * Returns the description associated with this info.  The return value may be null.
 	 */
-	public ProjectDescription getDescription() {
+	public synchronized ProjectDescription getDescription() {
 		return description;
 	}
 
@@ -79,8 +83,7 @@ public class ProjectInfo extends ResourceInfo {
 		return matcher;
 	}
 
-	public IProjectNature getNature(String natureId) {
-		// thread safety: (Concurrency001)
+	public synchronized IProjectNature getNature(String natureId) {
 		HashMap<String, IProjectNature> temp = natures;
 		if (temp == null)
 			return null;
@@ -98,7 +101,7 @@ public class ProjectInfo extends ResourceInfo {
 	/**
 	 * Sets the description associated with this info.  The value may be null.
 	 */
-	public void setDescription(ProjectDescription value) {
+	public synchronized void setDescription(ProjectDescription value) {
 		if (description != null) {
 			//if we already have a description, assign the new
 			//build spec on top of the old one to ensure we maintain


laeubi:
I don't see how this relates to the issue or fixes any multithreading issues, do we assume that **the same `ProjectDescription`** object is reused for different projects? If yes this should be changed, otherwhise I don't see how multiple threads can ever make a difference here.

jukzi:
> I don't see how this relates to the issue or fixes any multithreading issues, do we assume that **the same `ProjectDescription`** object is reused for different projects? If yes this should be changed, otherwhise I don't see how multiple threads can ever make a difference here.

if there is no volatile/synchronize keyword other threads are not guaranteed to see the updated array nor are other threads guaranteed to see updated elements in the array.

szarnekow:
Isn't the project description supposed to be used as a snapshot from a single client? Quoting org.eclipse.core.resources.IProject.getDescription()
```
Returns the description for this project. The returned value is a copy and cannot be used to modify this project. The returned value is suitable for use in creating, copying and moving other projects.
```

laeubi:
But this is only relevant if multiple threads call this methods in a highly concurrent way, I can't see this is the case here. I also don't see that the array is always ever updated concurrently.

jukzi:
Where exactly in the implementation do see the used org.eclipse.core.resources.IProject.hasNature(String) is guarded to see the values set with org.eclipse.core.resources.IProjectDescription.setNatureIds(String[])? I only see a ton of reads that do not care about multithreading at all. Nor do I see any code that would guarantee to make other threads to see its writes. And that is not a problem of timing: if there is no guarantee that another thread will see it then they may never ever see it.

szarnekow:
>  is guarded to see the values set with org.eclipse.core.resources.IProjectDescription.setNatureIds(String[])

These values are set via a workspace modify operation which gives us a safe-point. But see the analysis in https://github.com/eclipse-pde/eclipse.pde/issues/1185 - the description is updated in an unsafe way.

jukzi:
The workspace operation only guarantee that code is not concurrently modifying resources it does not make any guarantee - and can not - that other threads see memory updates.

laeubi:
> Nor do I see any code that would guarantee to make other threads to see its writes. And that is not a problem of timing: if there is no guarantee that another thread will see it then they may never ever see it.

If that would be true almost no code in java using Threads would work at all. If one Thread _writes_ a value and another one _reads it afterwards_ it is guaranteed that it will see the value. I mentioned it before that an issue can only arise if two Threads concurrent _read and write_ the same variable and one Thread depends on the results of the second Thread.

So even if we assume that's the case here (what is not as it uses an implicit lock), how should it ever happens that one Thread see *more* data than another one that is never written? That's simply impossible, because what we see here is that it seems that a nature is detected that is actually not configured for the project.

jukzi:
> If that would be true almost no code in java using Threads would work at all

A missing guarantee to work does not mean it has to fail. In fact most programs work only by chance.  At some point in time an optimizing java hotspot compiler can  - and eventually will - reorder or even skip instructions that he thinks, are not needed for the current thread.
Guys i have no proof that this code is the actual problem for the issue, but i see its against the java specifications and should be fixed. So we should try out if it solves the issue.

laeubi:
> In fact most programs work only by chance. 

That's a very strong statement ...

> Guys i have no proof that this code is the actual problem for the issue, but i see its against the java specifications

Then please show at least where this "violates" the specification and improve the commit message. Just stating that it _might_ cause problems without any concrete example others can comprehend is highly confusing.

jukzi:
commit message updated

laeubi:
So where is the data mutated/shared by different Threads then? 

To be more concrete

https://github.com/eclipse-platform/eclipse.platform/blob/35820b0bdcdc3a46c28ea5e3812e9b5f79618cde/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IProject.java#L490-L507

explicitly says:
> The returned value is a copy and cannot be used to modify this project

So it is the property of the caller and I can't find a location where such callers share the description with multiple threads, setting the description also not retain the passed in objects, so I see literally no chance in anything happening due to "chances are that this a concurrency problem".

So the only thing that happens is that the object is written by one Thread, never modified outside a locking condition and then probably read by multiple threads (where?) what is always safe.

jukzi:
> "when multiple threads share mutable data, each thread that reads or writes data must perform synchronization."

Which part of this sentence don't you understand? Annoying to talk about such basics.

iloveeclipse:
> Which part of this sentence don't you understand? Annoying to talk about such basics.

Jörg, this is not appropriate discussion style. Consider to answer the question instead of offending people.

merks:
@laeubi 

This issue has diverged into multiple threads and there is useful information that is being overlooked.  I.e., I  think your most-recent question is answered by this:

https://github.com/eclipse-pde/eclipse.pde/issues/1185#issuecomment-2055756571

So the theory is that cloning could be happening on one thread while basicSetDescription is happening on another thread for the same description object.  We can turn this theory into facts.  As such, I try to reproduce that theory via breakpoints in org.eclipse.core.tests.resources.IProjectDescriptionTest.testBuildSpecBuilder() after adding code like this:
```
		new Thread() {
			@Override
			public void run() {
				try {
					project.getDescription();
				} catch (CoreException e) {
					fail("4.99", e);
				}
			}
		}.start();

		project.setDescription(desc, null);
```

If we let the main thread get this far, i.e, as far as updating the internal description:

image

And then on another thread, get the description and let it get as far as cloning:

image

then indeed, as @szarnekow theorized, the thread cloning the description is working with the same instance as the thread that is updating that description.

So now we have facts and could try to move forward from that...

laeubi:
And now how does calling a setter **on the clone** can influence the internal description?

I just wanted to put the code here that is to be used *in practice* is:

1. get a **clone** the ProjectDescription by calling IProject#getDescription
2. modify the cloned description
3. call IProject#setDescription

I didn't stated anywhere that the data is maybe not *read* concurrently (and therefore the clone *might* get a somehow outdated copy) but this is not addressed here in any way. Instead one singular method (setNatureIds) is changed under the assumption that multiple threads are concurrently modify that specific field what leads to natures appearing on the project that are nowhere configured.

So if we are really concerned about *that* case then one should synchronize on the instance while the clone or somehow modify the internal description and not put arbitrary synchronize primitives in the object itself.

As mentioned in [the reference](https://ahdak.github.io/blog/effective-java-part-10/)
>  In the multicore era, it is more important that ever not to over-synchronize. Synchronize your class internally only if there is a good reason to do so, and document your decision clearly.

Also from the issue it seems some user can reproduce and even *debug* the issue, what makes it even less likely a concurrency issue.

merks:
It seems to me that @szarnekow had a concrete (and simple) suggestion for how to move forward:

https://github.com/eclipse-pde/eclipse.pde/issues/1185#issuecomment-2055795515

Does something speak against such an approach?

Given the demonstrated and proven fact that two threads can be simultaneous reading and modifying the **same project description instance**, such that we might clone a **partially** updated/modified instance,  I think we must do something to prevent that.    I think those facts are established or is there still some disagreement?  There is no point in discussing a solution if everyone has not yet recognized the exact nature of the problem.

I ask this question because I have a feeling that @laeubi might have missed this fundamental point when asking "how does calling a setter on the clone can influence the internal description" because my information was exactly answering that question, in great detail, even before it was asked.  So to reiterate, point is that we are definitely calling setters on one thread on the exact same object being cloned **while** it is being cloned on a different thread.  Please look carefully at the screenshots.  This is fact.

merks:
Note that when you say "maybe it would help to not concentrate on facts that seem to strengten an own statement " you appear to suggest that I've made a statement that needs strengthening.    The statement fact is my first statement.  I've made no other statement to argue for or against anything.  Note that one would hate to take the statement " it would help to not concentrate on facts" to be taken out of context.  😱 

Perhaps you simply assume that I am arguing in favor of the changes suggested in this PR.  I am not arguing in favor of these changes.  In fact, I don't think this PR's changes will help address the problem I've shown. Moreover, I'm not even convinced there is any scenario where this PR's changes will fix a problem.   So I think we agree on that point.  If there is such a scenario, it would be good to factually demonstrated it in the way I demonstrated that cloning and updating might happen concurrently such that one thread sees a partial set of changes being set by another thread.  That's no doubt also **exceedingly unlikely** to happen in practice.

Looking at the existing code there are a bunch of synchronizes already.

image

Note that none of the ProjectDescription methods called by basicSetDescription are synchronized.

I think that if the natures thing is a problem, that it would be good to demonstrate the problem via breakpoints rather than simply adding more complexity and hope that it maybe solves what might be a problem.

laeubi:
> I demonstrated that cloning and updating might happen concurrently such that one thread sees a partial set of changes being set by another thread. 

@merks it is all fine and I agree that this might be a problem (clone gets outdated data), but I want to clarify two (distinct things):

1. Any (existing or added) synchronization won't fix that problem, because it does not synchronize on the object instance (!) but only one some dedicated methods, that means some data might already be cloned by _Thread A_, then a (synchronized or not) modification method is called by _Thread B_ and now _Thread A_ get a copy that is a **mixture** of old and new data. Next assume we have already perfectly synchronized everything even on the object that we get a consistent copy, this will still mean that our copy is immediately **outdated**.
This leads me to the conclusion that (even though it is not mentioned explicitly on the javadoc) if I ever need a consistent clone that is up-to date with the current project state I can not use a random Thread but need (as also mentioned by @szarnekow before) to operate under the _Workspace Lock_, what makes sense because why should I make a copy just to throw it away afterwards, so likely I want to modify it what requires a _Worskapace Lock_ anyways.
2. Looking [at the issues](https://github.com/eclipse-pde/eclipse.pde/issues/1185) there it is claimed that [hasNature(...) returned true](https://github.com/eclipse-pde/eclipse.pde/issues/1185#issuecomment-2053715406) (sadly I still see no reproducer or I would have debugged it myself) what was [the trigger for this change](https://github.com/eclipse-pde/eclipse.pde/issues/1185#issuecomment-2055618956) as it "might help" and that's something that pains me (and that was basically the only thing I asked for in the beginning and also @szarnekow agreed here):
- the code shows some (now invalid) access to obtain a java project from a nature and does not explain *where* this has returned true unexpectedly, they even changed their code an now it _seems to work_.
- even if we assume this is all only a multithreading problem what should be the "writer" that adds a java nature that is not on the project, but later does update it to something else, there must be _a sign in code_ that this can happen (for whatever reason) but I can't find any, nor was it tried to provide such thing, instead I got blamed to not "understand the basics" that almost all Java programs only ever work by chance.

merks:
@laeubi 

I fully agree with your analysis.  In addition, I can think of no possible reason why hasNature could ever return true for a description that never had that nature at some very recent point in the past.  If we are going to fix a problem, we should at least have a theory for how that problem arises.

merks:
@jukzi 

FYI, it looks like you pushed changes to this PR that are not related to the previous discussion on this PR.

jukzi:
> FYI, it looks like you pushed changes to this PR that are not related to the previous discussion on this PR.

@merks  I've hardly been ill for two weeks and forgot the git basics. Thanks for notification.
Could you now please review this PR instead of discussing 1185? That issue was the trigger for the fix but still this MT hardening should be done whether or not it would fix that. It does not make much sense to dig into related issues as long as the code violates the Java Specification.

merks:
@jukzi 

Unfortunately I think @szarnekow @laeubi  and I all agree that these changes appear not to help address any demonstrable problem.  In https://github.com/eclipse-platform/eclipse.platform/pull/1312#issuecomment-2082504590 I demonstrated how it's possible for the source object of a clone operation (the internal representation used and modified internally by the framework) to be used while concurrently another thread is directly modifying that identical object instance.  Even for that case, clear how the changes in this PR would help with the modifiy-while-cloning problem because the implementation just calling super.clone which operates directly on the fields as I understand it:
```
	public Object clone() {
		ProjectDescription clone = (ProjectDescription) super.clone();
```

Perhaps via some thought experiment you can explain what you believe the different threads might be doing that might lead to a problem.

Note that it has been asserted that any operations on the actual cloned result return by org.eclipse.core.resources.IProject.getDescription(), including modification operations, do not need "hardening" because this instance is a copy per call and that that returned object should not be used across threads.  Do you agree with that assertion?

jukzi:
You did proof the code can be accessed in different threads without any synchronization on the same instance. It is well known java best practice to have synchronization between threads. You deny that? The threads may just not see each others writes either because the compiler just drops writes that he thinks are unneeded or because the CPU cores just don't flush their memory with other CPUs or the reading thread may just use stale values because the CPU or the compiler thinks the value could not have changed in the meantime. It is beyond this PR to dig deep into the hotspot compiler what can and will go wrong on which compilation level on which architecture. Java code should just confirm to the Java specification. If it does not errors can and will happen - See for example https://bugs.eclipse.org/bugs/show_bug.cgi?id=58314 for similar errors in jdt 
or https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html for similar errors in java in general.
Don't try to be smarter then the JDK if you don't have synchronization there is no guarantee the code works.  Period. 
If you don't want this PR to get in then please leave a -1. otherwise a +1. please vote.

laeubi (changes requested):
-1 from me as explained before.

merks:
If you prefer not to have a discussion but simply want an executive summary then -1.  

If you don't care why, stop reading now.

---

I've spent several hours on this already in good faith trying to be constructive.  The result was "Could you now please review this PR instead of discussing 1185?", which I read as a complete dismissal of everything I've done or said so far as being effectively irrelevant.  That makes me feel poorly.

I tried again to be constructive nevertheless, pointing out that cloning directly accesses the ProjectDescription.natures field so none of the logic added to getNature/setNature kicks in to prevent problems.  In addition, I asked a simple question trying to determine if there is common understanding here:

> Note that it has been asserted that any operations on the actual cloned result return by org.eclipse.core.resources.IProject.getDescription(), including modification operations, do not need "hardening" because this instance is a copy per call and that that returned object should not be used across threads. Do you agree with that assertion?

The result, no answer to my question along with a lengthy generic lecture about synchronization that's so general that one could, without proper context, use it to argue for adding synchronized to ArrayList. 

To me this feels extremely and unnecessarily adversarial, leaving little room to discuss the actual technical details of the potential problems so as to arrive at the best fixes for those problems.   Quite frankly, for me personally this exchange process is upsetting so much so that I really have needed to take a deep breath to avoid sharing vocally my frustration with being poorly treated.  No doubt you too are extremely frustrated---I'm very sorry for anything I've done to stoke that---and are venting that frustration into every sentence that you write.  That behavior just goes nowhere fast and I'm at a loss for how to improve this situation.  I know that being over-sensitive is one of my numerous faults. so perhaps that is coloring the situation as well.

jukzi:
Thanks for expressing your vote. 
The frustration is on both sides since you keep trying to discuss unrelated question: clone() is not even executed when the random failing ProjectDescription.hasNature(String) is executed, as ProjectDescription.getNatureIds(makeCopy=**false**) is used. 
Sadly that way we will never get a fix for the random fails that no one can explain. It is not comprehensible why otherwise smart people disagree on such java basics.

merks:
Your last comment is **exactly** why I asked this specific question:

> Note that it has been asserted that any operations on the actual cloned result return by org.eclipse.core.resources.IProject.getDescription(), including modification operations, do not need "hardening" because this instance is a copy per call and that that returned object should not be used across threads. Do you agree with that assertion?

I.e., if you are calling ProjectDescription.hasNature on an object which you got from Project.getDescription then it's your own local copy used in your own local thread, never to be seen on any other thread.  

But the call to org.eclipse.core.internal.resources.Project.hasNature(String) is a different beast and indeed it's calling a method on the internal instance of a ProjectDescription.  This is much more questionable.   That is why I focus on what appears to me to be the more fundamental problem of how a Project interacts with (writes to and reads from) its internal ProjectDescription, because we see too that cloning is not safe either.  So to me, adding what appear to be band-aide solutions knowing that underlying problems nevertheless remain is seems questionable.

So this has nothing to do with disagreeing on Java basics, it's an issue of not sharing a common mental model of the problem space.  It may well be that I missed some fundamental insight.  I'm far from perfect.

Anyway, I'm no doubt flogging a dead horse.

jukzi:
I think i finally understand what you mean but still you are on the wrong track:
Take this example:
```
	@Test
	public void testConcurrentNatures() throws Exception {
		createInWorkspace(project);
		CountDownLatch l = new CountDownLatch(1);
		String natureName = ResourceTestPluginConstants.NATURE_SIMPLE;
		Thread thread = new Thread() {
			@Override
			public void run() {
				try {
					IProjectDescription descCopy = project.getDescription();
					descCopy.setNatureIds(new String[] { natureName });
					project.setDescription(descCopy, null);
				} catch (CoreException e) {
					e.printStackTrace();
				}
				l.countDown();
			}
		};
		thread.start();
		l.await(); // uncomment and use breakpoints instead
		boolean result = project.hasNature(natureName);
		thread.join();
		assertTrue(result);
	}
```
Any java newbee would expect result to be true and it will be in most cases. But java specification requires any memory synchronization between the threads.  By now there is no synchronization which can be proven by breakpoints in ProjectDescription.getNature() and ProjectDescription.setNature(). Verify that there is no "synchronize" nor "volatile" keyword in the code of the concurrent threads:
```
Thread [Thread-16] (Suspended (breakpoint at line 852 in ProjectDescription))	
	ProjectDescription.setNatureIds(String[]) line: 852	
	NatureManager.configureNatures(Project, ProjectDescription, ProjectDescription, MultiStatus) line: 200	
	Project.basicSetDescription(ProjectDescription, int) line: 144	
	Project.setDescription(IProjectDescription, int, IProgressMonitor) line: 1351	
	Project.setDescription(IProjectDescription, IProgressMonitor) line: 1378	
	NatureTest$2.run() line: 472	

Thread [main] (Suspended (breakpoint at line 431 in ProjectDescription))	
	owns: RunnableLock  (id=177)	
	ProjectDescription.getNatureIds(boolean) line: 431	
	ProjectDescription.hasNature(String) line: 468	
	Project.hasNature(String) line: 570	
	NatureTest.testConcurrentNatures() line: 481	
...
```
There is a workspace operation in Project.setDescription(IProjectDescription, int, IProgressMonitor) but that is not a memory barrier nor does the reader have any form of synchronization.
This PR would have added synchronization to both reader and writer. Without the result will remain undefined by java spec.

jukzi:
p.s. the code sample should be added to org.eclipse.core.tests.resources.NatureTest

laeubi:
> p.s. the code sample should be added to org.eclipse.core.tests.resources.NatureTest

This "code sample" breaks the API contract of `IProject#setDescription` that says

>> This method requires the {@link IWorkspaceRoot} scheduling rule.

> Any java newbee would expect result to be true and it will be in most cases.

So does this code actually fails or do you suspect it to fail?

> Without the result will remain undefined by java spec.

Please link the relevant part of the spec then, this code contains an implicit synchronization between Thread A and B (the count down latch) and therefore it a happens-before relation ship so the JVM is required to flush the memory here before performing a read.

merks:
Let's not let this spiral out of control again!  This is a civil technical discussion focused on technical details.  I think we can all whole heartily agree that the tangled mess of how project descriptions are managed and how natures are added/remove/updated is beyond complex; even org.eclipse.core.internal.resources.ProjectInfo.setNature(String, IProjectNature) plays a role.

I believe (as @laeubi  mentions) that with the latch, the Project.hasNature will always see the nature set.  Without the latch, the methods on the thread might not yet have been called, might be in the middle of being called, or might be finished, or some incomprehensible hotspot state in between.  All bets are off.

Note that your example uses Project.hasNature which I mentioned is different from purely ProjectDescription.hasNature because it used the internal ProjectDescription instance, so I'm not sure where exactly I went off track, but I think it matter not.  Let's just change your example slightly to call `project.getDescrption().hasNature(natureName)` instead of `project.hasNature(natureName)` because that is a valid track too.  What can we say about this?  It seems to me that here cloning kicks in which does not call getNature, so no synchronization.  So we can say something, but it will be a different something.

Taking a step back to the big picture, assuming that one thread is busy updating descriptions, i.e., is busy adding/removing natures, while another thread is inspecting those natures, and additionally assuming there exists no attempt to control the order of events across these threads, what results can we expect from such a system?   The reading thread might read arbitrarily before, after, or during, the writing threads updates.  So it might see natures in an old state, a new state, or some in-between state.   Some synchronization can help ensure that it never sees an in-between state.  But even then, whatever state it sees might be different immediately after its seen that state, so whatever action it takes as a result of the state it sees might be completely invalid nevertheless.  What can we really expect to come out of such a system and can we really expect to turn it into something sensible?

I think this is why we have workspace operations and such, i.e., we have a mechanism such that some thread doing a bunch of reading or updating can assume/know/guarantee it sees a consistent state not modified by some other thread doing some other arbitrary things at the same time.

---

It might be that I'm completely off track.  It just reminds me of the olden early days back at the big old company where EMF was a scapegoat and was pretty much the root of all evil.   EMF isn't thread safe they would declare with an accusing tone.  Neither is ArrayList I would bite back.  I was a little more hot-headed in the olden days.   Being right was the most important overriding concern.  In any case, there was always pressure to synchronize something, e.g, ResourceSet.getResource, as if just the right sprinkling of locks **within the framework** would make the world a better place.  I would argue that the application itself needs a high level policy of how it accesses the model instance and that locks within the framework would only get in the way of implementing such a policy. 

In this context too, I my sense is that workspace operations are such high-level/application-level controls.

laeubi:
That's why I asked for the **codepath** (!) not a synthetic example where this "bad" should happen because the API clearly states that any modification **must** be done under the workspace lock (== only ever one thread). As soon as one has this codepath, then even for https://github.com/eclipse-pde/eclipse.pde/issues/1185 one don't need mystic explanations *its simply a bug*, as it would be to modify the project state from different threads.

laeubi:
> Sadly that way we will never get a fix for the random fails that no one can explain. 

As you have linked https://github.com/eclipse-pde/eclipse.pde/issues/1265 as "random fails that no one can explain" there is a simple explanation that can be proofed by reproducible steps:
- https://github.com/eclipse-pde/eclipse.pde/issues/1265#issuecomment-2112450168

so in contrast of the lengthy report claiming it is **this** bug it can be shown without any assumption of memory effects or reordering issues.

So instead of claiming "others" don't understand the basics, its sometimes better to step back and just assume if another theory is probably better explaining the given effects.

Even the inital thing that triggered this here https://github.com/eclipse-pde/eclipse.pde/issues/1185 instantly becomes "explainable" once we identified the actual code path.

jukzi:
I agree project.getDescrption().hasNature(natureName) is a total different codepath then project.hasNature() so you still trying to talk about a total different thing.

> Please link the relevant part of the spec then, this code contains an implicit synchronization between Thread A and B (the count down latch) and therefore it a happens-before relation ship so the JVM is required to flush the memory here before performing a read.

https://docs.oracle.com/javase/specs/jls/se21/html/jls-17.html chapter "17.4. Memory Model"
"17.4.5. Happens-before Order" clearly states that only "actions" have a happens-before relationship. "actions" are defined in "17.4.2. Actions". reading/writing a NON-volatile  field is not an "action" in that sense. So the JVM is free to reorder reads/writes to those with the countdown latch. there is no guarantee the read/write would happen before or after the latch. From the 17.4 examples section:
 "However, compilers are allowed to reorder the instructions in either thread, when this does not affect the execution of that thread in isolation" 

i.e. 
```
latch.await(); // an action
n= projectDescription.natures; // not an action

```
can be executed as
```
n= projectDescription.natures; // not an action
latch.await(); // an action
```

laeubi:
That's not what the spec is writing, a reordering can only happen if the overall effect is the same, exchanging an assignment with a method call is not. To see that this is invalid, simply simplify your example:

```
long start = System.currentTimeMillis(); // not an action
latch.await(); // an action
long stop = System.currentTimeMillis(); // not an action
```

so any JVM that would 'reorder' that to 
```
long start = System.currentTimeMillis(); // not an action
long stop = System.currentTimeMillis(); // not an action
latch.await(); // an action
```

is badly broken.

jukzi:
> exchanging an assignment with a method call is not

your expectation != specification

> is badly broken.

that indeed matches the specs: `To some programmers, this behavior may seem "broken". ` (Table 17.4-B. Surprising results caused by statement reordering - valid compiler transformation)

System.currentTimeMillis() however probably counts as external action.

laeubi:
The only problem is that your example does not match any "valid compiler transformation", a reordering of *dependent* statements is simply not allowed.
You arte simply bend the sentence so it matches your argumentation.

PR #1322, opened 2024-04-19
Add org.eclipse.jdt.* filter to org.eclipse.javaDevelopment activity.

raghucssit:
Existing patterns does not filter out commands from JDT. This pattern will filter out all the commands.

see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1828


Bananeweizen:
You might be over-filtering with that pattern. Historically a lot of things have been developed in JDT and later moved to the general platform or workbench, without renaming their IDs. I remember things like the "Open with" menu not working anymore when I experimented with such more global JDT filters. In our company application I have therefore rather created another 10 or 20 jdt.foo activityPatternBindings.

PR #1323, opened 2024-04-20
Support Platform.getOS/WS/Arch() in non OSGi runtimes and use it in LocalFileSystem

HannesWell:
Unless the bundle's context has its own local configuration, `BundleContext.getProperty()` delegates to `System.getProperty()` anyways.
Use it in `LocalFileSystem`, which had it implemented again explicitly for non-OSGi environments.

Does anybody has concerns about this?

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/Convert.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/Convert.java
@@ -16,7 +16,6 @@ package org.eclipse.core.internal.filesystem.local;
 
 import java.io.UnsupportedEncodingException;
 import org.eclipse.core.runtime.Platform;
-import org.eclipse.osgi.service.environment.Constants;
 
 public class Convert {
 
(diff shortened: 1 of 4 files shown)


laeubi (approved):
Looks good, are there any specific use cases using this without OSGi?

laeubi:
> Unless the bundle's context has its own local configuration, `BundleContext.getProperty()` delegates to `System.getProperty()` anyways.

 `BundleContext.getProperty()`  delegates to the FrameworkProperties (!) what Equinox may to delegate to SystemProperties see `osgi.framework.useSystemProperties` here: https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fruntime-options.html

HannesWell:
> Looks good, are there any specific use cases using this without OSGi?

From code and comments, the `LocalFileSystem` seems to be intended to also run in non OSGi environments. But I don't know if this (still) works or to what extend this works. Personally I neither have nor now real applications for that.
Or maybe it should just work if the platform bundle is not available, but that's probably an even rarer case.

PR #1329, opened 2024-04-26
Add toString() for better resource debugging

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/issues/470

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/provider/FileInfo.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/provider/FileInfo.java
@@ -270,11 +270,33 @@ public class FileInfo implements IFileInfo {
 			this.linkTarget = value;
 	}
 
-	/**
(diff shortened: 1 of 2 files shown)


fedejeanne:
Could the calls to methods like `exists()` or `isSet(int)` become a bottleneck or are they mere bitwise comparisons?

iloveeclipse:
It's toString(), which bottleneck you mean? It's only used while debugging code.

fedejeanne:
I know, debugging is also a use-case for us developers, that's why I ask. Additionally, one can never completely rule out the possibility that production code is using `toString()` too (even if it's a bad idea). 

Let me rephrase :-)
I've noticed that `toString()` now call methods as opposed to just reading from fields. I trust they will not (significantly) degrade performance, even when debugging? I can live with a slight performance decrease when debugging if it helps to debug.

iloveeclipse:
No, they don't degrade performance in a way anyone would notice that.

iloveeclipse:
Test failis unrelated.

PR #1336, opened 2024-05-03
Add Hero Screen for Eclipse

Roncla:


iloveeclipse:
???

laeubi:
See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3963#note_2184357

laeubi:
> I assume this one should be closed? Or there is smth to be done? Having such jpg file stored here serves nothing.

I don't know the process so maybe ask at the helpdesk issue. I could imagine that:

1. we should store this somewhere so others can use it
2. make sure everything we use has gone through the IP process

akurtakov:
Helpdesk issue is closed, this file is not used anywhere. Unless someone tells this is actually needed and for what I'll close it after the vacation.

laeubi:
All I can tell is that it was requested there and I helped @Roncla to open the PR and the folder already contains similar images: https://github.com/eclipse-platform/eclipse.platform/tree/master/platform/org.eclipse.platform/splashscreens2024

laeubi:
I think everything is set and we can reopen if required.

PR #1337, opened 2024-05-06
Migrate session tests in org.eclipse.core.tests.runtime to JUnit 5 #903 

HeikoKlare:
Migrates the existing session tests in the org.eclipse.core.tests.runtime plug-in to JUnit 5 by using the new
SessionTestExtension. Also replaces some conditional fail statments in the test classes with assertions.


akurtakov (approved):
Looks good to me.

Issue #1338, opened 2024-05-07
ContentType file associations can't be extended by product customization


iloveeclipse:
As a product owner I've stumbled over the problem that I'm unable to modify existing content types (by adding new file associations) by using product customization.

My simple goal was to leverage tm4e syntax highlighting for file types that aren't by default configured with tm4e (because they have custom file extensions) but have the content supported by tm4e.

As a user I was able to configure that by adding new file patterns to the existing content type, 

![image](https://github.com/eclipse-platform/eclipse.platform/assets/964108/bb62affd-30a0-4b3c-ae24-6922c185c51f)

... but it turned out, all content type changes made by user are read from / persisted **to the workspace** and never read from product defaults, only from extension registry.

I'm not sure why it wasn't done but it turned out it seem to be easy to add support for product customization, so I will provide a PR for that.

mickaelistria:
> I'm not sure why it wasn't done but it turned out it seem to be easy to add support for product customization, so I will provide a PR for that.

When I worked on association as user preferences, at the beginning of the LS/TM era, the goal was to give a way for end-users to plug their LS & grammar in their IDE and get it to work for their files, all that only from UI.

For product, there is already an extension point dedicated to define extra content-type/pattern associations
```xml
<extension
      point="org.eclipse.core.contenttype.contentTypes">
   <file-association
         content-type="org.eclipse.tm4e.language_pack.cpp"
         file-extensions="cplus">
   </file-association>
</extension>
```

So I don't think a new way to configure it is so relevant.

mickaelistria:
Also, a more usual and better scaling pattern when it comes to product is to create a specialized content-type as child of some upstream type and define its association. Those files would then inherit all editor/LS associations from their parent.

> It allows you to do same without bundle changes.

I find product settings are way more messy and much harder to maintain than proper usage of extensions, so I would personally prefer updating a bundle and wouldn't recommend changing product preference for something that can be done in a tighter harness (plugin.xml) to anyone , but YMMV.

iloveeclipse:
The use case I have in mind is more dynamic: I have a product that is customized via product customization file located on network. So it is much easier to add "yet another" content type people want use via a single change of that file & push to network compared to changing a bundle plugin.xml, updating manifest version, building & deploying the site and finally users asked for update of the bundle via popup.

PR #1339, opened 2024-05-07
ContentType file associations can't be extended by product customization

iloveeclipse:
Allow products define custom file associations for existing content types, example below will link custom `*.cplus` files to the tm4e provided "C++" content type

`org.eclipse.core.runtime/content-types/org.eclipse.tm4e.language_pack.cpp/file-extensions=cplus`

image

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1338


iloveeclipse:
@mickaelistria : any objections?

mickaelistria:
No objection.

PR #1340, opened 2024-05-07
Migrate WorkspacePreferencesTest to JUnit 5 #903

HeikoKlare:
This makes the WorkspacePreferencesTest use the SessionTestExtension to be executed with JUnit 5. It adds a method to the
SessionTestExtensionBuilder to allow setting a workspace directory for a test class. In addition, this replaces assertions in the test class using numbers as messages with proper AssertJ assertions.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
@@ -14,17 +14,18 @@
  *******************************************************************************/
 package org.eclipse.core.tests.internal.resources;
 
+import static java.util.function.Predicate.not;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.eclipse.core.tests.harness.FileSystemHelper.getRandomLocation;
 import static org.eclipse.core.tests.resources.ResourceTestPluginConstants.PI_RESOURCES_TESTS;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createRandomString;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.removeFromFileSystem;
 
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import junit.framework.ComparisonFailure;
-import junit.framework.Test;
 import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.resources.WorkspacePreferences;
 import org.eclipse.core.resources.IWorkspace;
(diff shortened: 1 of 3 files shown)


runtime/tests/org.eclipse.core.tests.runtime/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Core Tests Runtime
 Bundle-SymbolicName: org.eclipse.core.tests.runtime; singleton:=true
-Bundle-Version: 3.21.500.qualifier
+Bundle-Version: 3.21.400.qualifier
```
akurtakov, line 5: Decreasing version looks totally wrong!

akurtakov: If the version bump made it to an I-build please don't revert at all as people that updated will never get changes after the revert.

HeikoKlare: Yes. Since the commit has been merged 1 hour ago, it is not part of an I-Build yet, so I think reverting is okay and  the right thing to do?

iloveeclipse: Yes

HeikoKlare:
> AutomatedResourceTests AllInternalResourcesTests WorkspacePreferencesTest ‑ test___cleanup

Removing now obsolete cleanup step is intended

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
```diff
@@ -243,48 +266,64 @@ public void testSetDescription() throws CoreException {
...
-		assertEquals(message + " - 4", description.isApplyFileStatePolicy(), expectedPreferences.getBoolean(ResourcesPlugin.PREF_APPLY_FILE_STATE_POLICY));
-		assertEquals(message + " - 5", description.getFileStateLongevity(), expectedPreferences.getLong(ResourcesPlugin.PREF_FILE_STATE_LONGEVITY));
-		assertEquals(message + " - 6", description.getMaxFileStates(), expectedPreferences.getInt(ResourcesPlugin.PREF_MAX_FILE_STATES));
-		assertEquals(message + " - 7", description.getMaxFileStateSize(), expectedPreferences.getLong(ResourcesPlugin.PREF_MAX_FILE_STATE_SIZE));
-		assertEquals(message + " - 8", description.getSnapshotInterval(), expectedPreferences.getLong(ResourcesPlugin.PREF_SNAPSHOT_INTERVAL));
-		assertEquals(message + " - 9", description.getMaxBuildIterations(), expectedPreferences.getLong(ResourcesPlugin.PREF_MAX_BUILD_ITERATIONS));
-		assertEquals(message + " -10", description.isKeepDerivedState(),
-				expectedPreferences.getBoolean(ResourcesPlugin.PREF_KEEP_DERIVED_STATE));
+	public void assertEquals(IWorkspaceDescription description, Preferences expectedPreferences)
```
fedejeanne, line 277: This could be made private in order to discourage inheritance and you could also flip the parameters so they follow the same logic as JUnit's `assertTrue` that is: first the **expected** value and then the **actual** value. Additionally, you could rename this method to something like `assertMatchesPreferences` or the likes so one can clearly see that this is a "custom" method and not a "JUnit" method (dealer's choice on the latter :-) ).

```suggestion
	private void assertEquals(Preferences expectedPreferences, IWorkspaceDescription description)
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
```diff
@@ -243,48 +266,64 @@ public void testSetDescription() throws CoreException {
...
-		assertEquals(message + " - 6", description1.getMaxFileStates(), description2.getMaxFileStates());
-		assertEquals(message + " - 7", description1.getMaxFileStateSize(), description2.getMaxFileStateSize());
-		assertEquals(message + " - 8", description1.getSnapshotInterval(), description2.getSnapshotInterval());
-		assertEquals(message + " - 9", description1.getMaxBuildIterations(), description2.getMaxBuildIterations());
-		assertEquals(message + " -10", description1.isKeepDerivedState(), description2.isKeepDerivedState());
-	}
-
-	public static Test suite() {
-		return new WorkspaceSessionTestSuite(PI_RESOURCES_TESTS, WorkspacePreferencesTest.class);
+	public void assertEquals(IWorkspaceDescription actualDescription, IWorkspaceDescription expectedDescription)
```
fedejeanne, line 304: Dito: private method, flip parameters (possibly rename)

```suggestion
	private void assertEquals(IWorkspaceDescription expectedDescription, IWorkspaceDescription actualDescription)
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
```diff
@@ -33,103 +34,126 @@
...
+	@AfterEach
+	public void restoreWorkspaceDescription() throws CoreException {
 		workspace.setDescription(Workspace.defaultWorkspaceDescription());
 	}
 
 	/**
 	 * Tests properties state in a brand new workspace (must match defaults).
 	 */
+	@Test
+	@Order(1)
```
fedejeanne, line 76: Why is the order of the tests relevant now?

fedejeanne (commented):
Thank you for migrating this! 
I only have some minor comments on the changes and the question about the newly enforced order of the tests.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
HeikoKlare: It has also been relevant before. The `WorkspaceSessionTestSuite` (that was used before) sorts the tests by default (based on a lexicographic order of test method names). This does not happen with the JUnit 5 `SessionTestExtension` anymore (as also documented in its Javadoc), which is both an active decision (as it improves comprehensibility when making the order of test cases explicit if relevant) and technically easier to realize with JUnit 5 extensions.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
HeikoKlare: Fixed with 63513a7.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
HeikoKlare: Completely agreed. I've renamed the `assertEquals` methods in 46f98ac. I've also renamed the other `assertEquals` method, as this naming produces recurring problems due to name clashes with static imports of JUnit's `assertEquals` method.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/WorkspacePreferencesTest.java
fedejeanne: It just occurred to me: if `WorkspaceSessionTestSuite` sorted the tests in lexicographical order then shouldn't you annotate the class with `@TestMethodOrder(MethodOrderer.Alphanumeric.class)`? That way new tests would follow the same order they would have followed in the old `WorkspaceSessionTestSuite`

fedejeanne (commented):
Looks good, I have one final question regarding the order of the tests.

HeikoKlare:
> It just occurred to me: if WorkspaceSessionTestSuite sorted the tests in lexicographical order then shouldn't you annotate the class with @TestMethodOrder(MethodOrderer.Alphanumeric.class)? That way new tests would follow the same order they would have followed in the old WorkspaceSessionTestSuite

Yes, you are right. I wanted to avoid relying on a lexicographic ordering here, as the test method are inappropriate for such an ordering (other test classes have `test1` `test2` etc.), but I attached numbers that differ from that order. Actually, the order in this class is completely irrelevant (there are no dependencies between the test cases), so I have now simply removed the ordering.

Nitpicky note: `MethodOrderer.Alphanumeric` is deprecated 😉

PR #1342, opened 2024-05-07
Replace JUnit 4 with JUnit 5 setup in runtime tests #903

HeikoKlare:
Applies the following migration steps from JUnit 4 to JUnit 5 in org.eclipse.core.tests.runtime:
* Replace JUnit 4 with JUnit 5 annotation (Test, After, Before, FixTestOrder)
* Replace TestName rules with TestInfo objects if necessary
* Replace TemporaryFolde rules with TempDir extension
* Remove RetryTestRule in favor of RepeatedTest


HeikoKlare:
> This pull request removes 7 and adds 61 tests. Note that renamed tests count towards both.

This is correct, since the 6 test cases in `OrderedLockTest` are executed 10 times each but now with the `RepeatedTest` functionality of JUnit 5 each test run is listed as an own test while all 10 runs were summarized with the JUnit 4 `RetryRule` that was used before.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/GithubBug_193.java
```diff
@@ -24,22 +24,18 @@
...
-import org.junit.Test;
-import org.junit.runners.MethodSorters;
+import org.junit.jupiter.api.Test;
 
 /**
  * Test for bug https://github.com/eclipse-platform/eclipse.platform/issues/193
  */
-@FixMethodOrder(MethodSorters.NAME_ASCENDING)
-public class GithubBug_193 extends TestCase {
+public class GithubBug_193 {
```
fedejeanne, line 38: You also removed the enforced order here. Was this intentional?

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/OrderedLockTest.java
```diff
@@ -53,7 +50,7 @@ private void createRunnables(ILock[] locks, int n, ArrayList<LockAcquiringRunnab
 		}
 	}
 
-	@Test
+	@RepeatedTest(value = REPETITIONS)
```
fedejeanne, line 53: Nit-pick: no need to write `value =` since it's a reserved word and it's the "default" attribute. If you wanted to set the `name` and/or `threshold` though then you would need to write `value =` too.

```suggestion
	@RepeatedTest(REPETITIONS)
```

fedejeanne (changes requested):
Thank you for migrating this to JUnit5.

I have only one question regarding the ordering of the method in `GithubBug_193`, hence my _request for changes_.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/OrderedLockTest.java
HeikoKlare: Good point, thank you! Improved in a886e5e.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/GithubBug_193.java
HeikoKlare: Yes, the class only contains a single test method, so there is no order that can be fixed, which makes it unnecessary to migrate the annotation to JUnit 5.

fedejeanne: 👍 
I totally missed that, he.

fedejeanne (approved):

PR #1347, opened 2024-05-08
Remove unnecessary variables in LaunchConfigurationTabGroupViewer

fedejeanne:


iloveeclipse:
@fedejeanne : please never merge anything without waiting for build results. Even a white space change can cause build errors.

PR #1348, opened 2024-05-09
Migrate ConfigurationSessionTestSuite to JUnit 5 #903

HeikoKlare:
This introduces a JUnit 5 replacement for the JUnit 3 ConfigurationSessionTestSuite that allows to define session tests using a custom workbench configuration with an explicit config.ini defining a custom-defined bundle set.

It is realized by the CustomSessionConfiguration class, which encapsulates all information and logic to set up a custom workbench configuration, and an additional parameterization of the JUnit 5 SessionTestExecution extension, which allows this extension to replace the default workbench configuration with the custom one. The change also adapts the way in which a workspace folder can be defined to use the same mechanism by providing a CustomSessionWorkspace extension to be passed to the SessionTestExtension.

The new capability of defining a custom workbench configuration for a JUnit 5 session test is applied to the PlatformURLSessionTest, which was based on the JUnit 3 ConfigurationSessionTestSuite so far.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/CustomSessionConfiguration.java
+++ b/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/CustomSessionConfiguration.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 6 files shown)


fedejeanne:
LGTM but my review was mostly centered on the coding practices since the technical part is a bit beyond my ken. But since I see no failed tests, I feel confident this is a good addition.

Thank you for contributing this!

HeikoKlare:
Thank you for your review! In my opinion, checking design and coding practices is most valuable here, since functional errors would be detected by failures of tests using this framework.

This PR currently also refactors some existing design for setting up the `SessionTestExtension` with custom workspaces (analogous to existing `WorkspaceSessionTests`), To keep things properly separated, I have extracted those changes to a separate PR with some further simplifications (#1358).

PR #1349, opened 2024-05-09
Remove empty org.eclipse.core.net.win32.x86_64 fragment

HannesWell:
The architecture specific `org.eclipse.core.net.win32.x86_64` fragment has been replaced by the architecture independent fragment `org.eclipse.core.net.win32` fragment over two years ago in [Bug 578341](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578341)

Similar as it was discussed in https://github.com/eclipse-equinox/equinox/pull/564#issuecomment-2018945338 there is no real advantage in extending the transition period because at one point references to the old fragment have to be replaced in features and products (which can only reference fixed versions of fragments).
If there is agreement to proceed I'll create a PR to remove the only reference in the SDK from the `org.eclipse.platform` (in the `platform.releng.aggregator` repo) and will check the SimRel aggregation for more references.


HannesWell:
@iloveeclipse or @akurtakov what do you think about this?

akurtakov:
Go for it.

HannesWell:
Created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2018 to remove it from the platform feature and checked the SimRel aggregation analyzer that at least there no other dependencies exist.
From the `Strict Requirement IUs` tab:

grafik

PR #1350, opened 2024-05-10
Auto-refresh fragment for Windows on Arm64 (WoA) platform.

chirontt:
The following new fragment for WoA is added to the project to support auto-refresh using native code:
```
org.eclipse.core.resources.win32.aarch64
```
To manually build the auto-refresh native library for WoA, on a WoA box, run the following commands:
```batch
cd resources\bundles\org.eclipse.core.resources\natives\win32
make.bat
```
and the following native library file for WoA is generated and saved in its relevant directory:
```
resources\bundles\org.eclipse.core.resources.win32.aarch64\os\win32\aarch64\win32refresh.dll
```
The existing 'make.bat' file mentioned above has been modified to generate correct binary for the current build environment (x64 or aarch64.)

As discussed [here](https://github.com/eclipse-platform/eclipse.platform/pull/1311#issuecomment-2102434757).

### DIFF
--- a/resources/bundles/org.eclipse.core.resources.win32.aarch64/.project
+++ b/resources/bundles/org.eclipse.core.resources.win32.aarch64/.project
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.core.resources.win32.aarch64</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.pde.PluginNature</nature>
+	</natures>
+</projectDescription>
--- a/resources/bundles/org.eclipse.core.resources.win32.aarch64/.settings/org.eclipse.core.resources.prefs
+++ b/resources/bundles/org.eclipse.core.resources.win32.aarch64/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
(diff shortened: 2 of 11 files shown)


chirontt:
> would you be interested in trying out a JNA alternative? If not I can look into that.

I don't have experience with JNA. Please try it yourself.

I could do it myself, but it will take time to get familiar with it first.

HannesWell:
> > would you be interested in trying out a JNA alternative? If not I can look into that.
> 
> I don't have experience with JNA. Please try it yourself.

No problem. I have just created https://github.com/eclipse-platform/eclipse.platform/pull/1394.
JNA is actually not that difficult (its main goal is to simplify native access). Although #1394 contains a few optimizations and it could be done even simple, I hope it is still not to difficult to understand.

HannesWell:
> But I have thought about this a bit more and I think we should evaluate if the refresh provider can be implemented through JNA without loosing too much performance, even though JNA itself is significantly slower than JNI. My 'hope' is that the native methods themself run long enough that the JNA overhead is not noticeable. In the best-case we could use it for all Windows architectures, in second best-case we avoid another native fragment for Windows on ARM.

My hope have been realized and https://github.com/eclipse-platform/eclipse.platform/pull/1394 just landed. With that, this is not necessary anymore and can be closed.

It would be great if you would try-out the up-coming I-build and verify if automatic refresh through native hooks now works under Windows on ARM (URL will probably be https://download.eclipse.org/eclipse/downloads/drops4/I20240612-1800/)

chirontt:
> It would be great if you would try-out the up-coming I-build and verify if automatic refresh through native hooks now works under Windows on ARM (URL will probably be https://download.eclipse.org/eclipse/downloads/drops4/I20240612-1800/)

I've used the following I-build:
https://download.eclipse.org/eclipse/downloads/drops4/I20240613-1800/

and download and run the Eclipse SDK for WoA, and the auto-refresh works fine in my WoA box, after I change the preference setting for it to `on`. A good thing about this auto-refresh feature is the quick detection of a new file in the workspace (quick in the sense of around 1 second by my informal perception), whereas the new file wouldn't be detected if this feature is off (which requires a F5 keystroke to force a manual refresh.)

Note that there was a delay/freeze of around 1 second of the UI when the feature was first turned on, as recorded in the error logs:
```
!ENTRY org.eclipse.ui.monitoring 2 0 2024-06-14 20:33:15.524
!MESSAGE UI freeze of 0.52s at 20:33:14.987
!SUBENTRY 1 org.eclipse.ui.monitoring 1 0 2024-06-14 20:33:15.524
!MESSAGE Sample at 20:33:15.326 (+0.339s)
Thread 'main' tid=1 (RUNNABLE)
```
My guess is this was due to the overhead of JNA initialization (and dynamic compiling?), after which there are no more delay/freeze even when the feature is turned off then back on again.

PR #1352, opened 2024-05-10
Stop using Lucene's deprecated IndexSearcher.doc method

akurtakov:
Fetch from the StoredFields instead.


akurtakov:
MacOS failure is unrelated.

PR #1356, opened 2024-05-10
Remove sorting parameter from ConfigurationSessionTestSuite #903

HeikoKlare:
The ConfigurationSessionTestSuite sorts the test methods by default. A field identifies whether sorting is activated, but the field cannot even be set but is always true. So, this parameterization is completely obsolete and only introduces unnecessary complexity to the session test framework.

This unnecessary complexity in the JUnit 3 implementation of session tests complicates their migration to JUnit 5. I propose to incrementally remove all unused functionality from the JUnit 3 session test implementation while migrating the actual test classes to the JUnit 5 equivalent and incrementally enhancing the JUnit 5 implementation with what is required. This applies one step of that incremental removal of unused functionality in the JUnit 3 implementation of session tests.

❗This removal becomes particularly useful when also #1354 removing shared session support is applied, as in combination even more code can be removed due to conditions becoming empty.


fedejeanne (approved):
LGTM

Issue #1098, opened 2024-01-18
InjectionOSGiTest.ensureJavaxIsNotAvailable() fails since I20240117-1800


jukzi:
org.eclipse.e4.core.internal.tests.di.extensions.InjectionOSGiTest.ensureJavaxIsNotAvailable()
on all platforms
```
java.lang.AssertionError: expected java.lang.ClassNotFoundException to be thrown, but nothing was thrown
at org.junit.Assert.assertThrows(Assert.java:1028)
at org.junit.Assert.assertThrows(Assert.java:981)
at org.eclipse.e4.core.internal.tests.di.extensions.InjectionOSGiTest.ensureJavaxIsNotAvailable(InjectionOSGiTest.java:120)

```
https://download.eclipse.org/eclipse/downloads/drops4/I20240117-1800/testresults/html/org.eclipse.e4.core.tests_ep431I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html

jukzi:
@HannesWell

jukzi:
there are no recent changes in this repository so it has to be regression from another library or build system.

HannesWell:
> there are no recent changes in this repository so it has to be regression from another library or build system.

Looking at the current target-platform I cannot find any non-optional dependency to a provider of the `javax.inject` package and if I replicate the test setup locally the test passes (one cannot use the default JUnit Plugin Test setup but has to modify it to only contain the test-plugin and add all required bundles automatically).

@laeubi did something change in Tycho regarding the assembly of the test-runtime? I.e. are optional dependencies now added too?

laeubi:
One can configure exclusions in the target:
https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#filters

Beside that there is no special handling of "optional" dependencies, if they are optional they might or might not be resolved depending on whats available.

HeikoKlare:
One additional observation: Note that the test only seems to fail in I-Builds. The GH workflow runs (including both GH actions and Jenkins builds) run fine.

jukzi:
What is the plan for this issue? Does it need to be fixed or test disabled?

HannesWell:
> And I don't think that having javax on the classpath is expected here, is it?

It is expected, that it is not on the classpath in order to test that the e4 injector works as expected without the javax annotations.

[...]

PR #1364, opened 2024-05-13
Disable InjectionOSGiTest if needed #1098

jukzi:
Convert assert to assume to disable false-positive in I-Builds. https://github.com/eclipse-platform/eclipse.platform/issues/1098

### DIFF
--- a/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/InjectionOSGiTest.java
+++ b/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/InjectionOSGiTest.java
@@ -18,8 +18,8 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
 
 import org.eclipse.e4.core.contexts.ContextInjectionFactory;
 import org.eclipse.e4.core.contexts.EclipseContextFactory;
@@ -33,6 +33,7 @@ import org.junit.Test;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleException;
+import org.osgi.framework.FrameworkUtil;
 
 import jakarta.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


jukzi:
That test is failing for 5 months now and i have seen no progress, tired to see it in every i build.

iloveeclipse:
> That test is failing for 5 months now and i have seen no progress, tired to see it in every i build.

Fully agree. 

However: as long as no real fix is in sight, instead of using "assume": could we add environment variable in Jenkinsfile & do not fail the test on official builds where this variable is not specified? 
This way we still can have the test failing in Jenkins in case same problem with the target platform will appear there.

HeikoKlare:
I totally agree that the persisting test failure is very annoying, but still I want to state that in my opinion this kind of temporary fix is _wrong_ (see the discussion in #1098). With this change, we will not notice when for whatever reasons also the GH Actions builds do not execute the tests properly anymore and thus we may end up in having no validation for the javax annotations without taking notice.

> However: as long as no real fix is in sight, instead of using "assume": could we add environment variable in Jenkinsfile & do not fail the test on official builds where this variable is not specified?

Couldn't we also do something like removing the test plug-in from the Ant specifications used by the Jenkins build? The test plug-in is currently useless in the Jenkins builds anyway (as it tests the same as the "ordinary" test plug-ins using the jakarta annotation).

jukzi:
If anyone is really interested in that test why shouldn't he use his time to fix it - instead of insisting it should not be disabled? 
5 month it is failing and no one cared. Who would care if it really fails instead of false-positive?

jukzi:
> ```
> CoreTestSuite org.eclipse.e4.core.internal.tests.di.extensions.InjectionOSGiTest ‑ ensureJavaxIsNotAvailable
> ```

this is intended, 
the failing ParallelBuildChainTest is unrelated,
macos: `Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.8-SNAPSHOT:verify (verify) on project org.eclipse.help.webapp: Execute ApiApplication failed: InvocationTargetException: Component 'Help System Webapp' in the baseline 'workspace' is disposed: Component was disposed here -> [Help 1]` is unrelated.

InjectionOSGiTest is still executed and green on jenkins CI:
image
https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-1364/2/testReport/CoreTestSuite%20org.eclipse.e4.core.internal.tests.di.extensions/InjectionOSGiTest/

If anybody wants another solution please work on it timely.

HeikoKlare:
> If anyone is really interested in that test why shouldn't he use his time to fix it - instead of insisting it should not be disabled?
> 5 month it is failing and no one cared. Who would care if it really fails instead of false-positive?

1. We are (or should) all be intestered in that test as it is part of the validation for the product we are all together responsible for. Do you think there is someone particularly reponsible for this test and its failure?
2. I think there is some misunderstanding regarding the meaning of this single test case: It is not about a single test but about the complete test plug-in. This single test case identifies whether all the other tests in that same plug-in make any sense at all (as it checks for the proper configuration in which the plug-in is run).
3. The test runs fine in GH actions builds.
4. Because of (3), no one says the test should not be disabled, but it should be disabled _selectively_ only where it is not configured properly (for whatever reasons).

jukzi:
> Do you think there is someone particularly reponsible for this test and its failure?

sure - if someone contributes a test he should keep it alife. in this case @HannesWell contributed it. Anyway as we can not force anybody to work on anything the typical way is to just disable false positives, as say do not add any value but disturb.

> The test runs fine in GH actions builds.

this PR does not change that.

> 4\. Because of (3), no one says the test should not be disabled, but it should be disabled _selectively_ only where it is not configured properly (for whatever reasons).

Then please do it.

HeikoKlare:
> sure - if someone contributes a test he should keep it alife

I completely disagree here. That would be a great incentive to _not_ provide tests. I have fixed so many tests that I did not contribute, no matter whether the initial contributor was still active or not. There have been several tests badly written, containing race conditions, missing isolatation etc., but I would have never thought about blaming anyone for contributing the test and not fixing it when it started to fail. In particular in this case, the problem is not even about the test itself but about the execution environment.

> > The test runs fine in GH actions builds.
> 
> this PR does not change that.

Depends on what you define as "running fine". I have explained the effects of this PR on the GH actions builds before: if those builds start to pull in unexpected annotations as well, no will notice after this PR as they will simply not be executed anymore.

HannesWell (changes requested):
I fully agree with Heiko here, so I won't repeat that. But just because there has not been a solution it doesn't mean there is progress or at least work happening. I simply didn't have the time to complete the work since the I build environment is not trivial to understand and I had a lot of other things to do.

But I can probably push a PR that selectively disables the configuration test case in I-builds as suggested by Heiko.

jukzi:
> But I can probably push a PR that selectively disables the configuration test case in I-builds as suggested by Heiko.

would be very welcome

HannesWell:
> > But I can probably push a PR that selectively disables the configuration test case in I-builds as suggested by Heiko.
> 
> would be very welcome

Please see https://github.com/eclipse-platform/eclipse.platform/pull/1365.

PR #1365, opened 2024-05-13
Disable InjectionOSGiTest.ensureJavaxIsNotAvailable in I-build tests

HannesWell:
Disable the `InjectionOSGiTest.ensureJavaxIsNotAvailable()` test-case failing in I-Build tests since `javax.inject` and `jakarta.inject` bundles are always both installed into the Eclipse installation running the tests.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1098

### DIFF
--- a/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/InjectionOSGiTest.java
+++ b/runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/InjectionOSGiTest.java
@@ -28,6 +28,7 @@ import org.eclipse.e4.core.di.annotations.Optional;
 import org.eclipse.e4.core.di.extensions.OSGiBundle;
 import org.eclipse.e4.core.internal.tests.CoreTestsActivator;
 import org.junit.After;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 import org.osgi.framework.Bundle;
@@ -109,6 +110,13 @@ public class InjectionOSGiTest {
 
 	@Test
 	public void ensureJavaxIsNotAvailable() {
+		// This entire test-plugin is not really useful if javax.inject and
+		// jakarta.inject is available and this test-case fails then.
+		// However due to the way I-build tests are set up (i.e. by using a built
+		// Eclipse installation) exactly that's the case.
+		// -> Disable this for I-build tests
+		Assume.assumeFalse("org.eclipse.ant.core.antRunner".equals(System.getProperty("eclipse.application")));
+
 		// Ensure that the providing bundles of the following classes are absent of the
 		// test-runtime and thus the mentioned classes cannot be loaded
 		assertThrows(ClassNotFoundException.class, () -> Class.forName("javax.inject.Inject"));


HannesWell:
@iloveeclipse or can't we just delete the `org.eclipse.e4.core.tests/test.xml` to prevent the entire test-plugin from running?
On the other hand, altough that test-plugin is intended to run without `javax.inject` it can still useful to test `jakarta.inject` annotations, even if `javax.inject` is present too.
@HeikoKlare didn't you meant that in https://github.com/eclipse-platform/eclipse.platform/pull/1364#issuecomment-2107758417 ?

runtime/tests/org.eclipse.e4.core.tests/src/org/eclipse/e4/core/internal/tests/di/extensions/InjectionOSGiTest.java
```diff
@@ -109,6 +110,13 @@ public void setUp() throws Exception {
 
 	@Test
 	public void ensureJavaxIsNotAvailable() {
+		// This entire test-plugin is not really useful if javax.inject and
+		// jakarta.inject is available and this test-case fails then.
+		// However due to the way I-build tests are set up (i.e. by using a built
+		// Eclipse installation) exactly that's the case.
+		// -> Disable this for I-build tests
+		Assume.assumeFalse("org.eclipse.ant.core.antRunner".equals(System.getProperty("eclipse.application")));
```
jukzi, line 118: How is that any better then my solution? It has the same downside that accidentally this condition may change in future on the various build types or when running the test locally.

HannesWell: Because with your suggestion in #1364 that case cannot fail at all and therefore becomes useless because it cannot indicate that `javax.inject` is present although it is expected to be absent:
https://github.com/eclipse-platform/eclipse.platform/pull/1364#issuecomment-2107758417

This solution only disables that case for I-builds since it cannot really be changed that `javax.inject` is present (at least I failed to implement the dynamic removal).

> It has the same downside that accidentally this condition may change in future on the various build types or when running the test locally.

Yes, but that's intended. If in the future I-builds use another application that condition has to be adapted. But I think that's the smaller evil. And in local builds the test should still fail if javax.inject is present to make developers aware that the current setup is not as it should be.

iloveeclipse:
> or can't we just delete the `org.eclipse.e4.core.tests/test.xml` to prevent the entire test-plugin from running?

Not sure I understood the proposal. If the patch in this PR works, let merge it, it is more "problem oriented" as changing entire test run configuration.

HeikoKlare (approved):
Thank you, @HannesWell! The proposed change implements the idea of a selective disablement of the test. Since (only) I-Builds execute the tests via Ant, checking the `eclipse.application` property for an according value looks sound to me. According to the logs, I would expect the property validation to work correct:
```
[echoproperties] classname=org.eclipse.e4.core.tests.CoreTestSuite
...
[echoproperties] eclipse.application=org.eclipse.ant.core.antRunner
```

I would be in favor of just merging this and checking next I-build for whether it was successful as expected.

jukzi:
I still think that that this PR is a dirty hack and we will never see a clean fix after this. But as long as it disables the test i am fine with it.

HeikoKlare:
> > or can't we just delete the `org.eclipse.e4.core.tests/test.xml` to prevent the entire test-plugin from running?
> 
> Not sure I understood the proposal. If the patch in this PR works, let merge it, it is more "problem oriented" as changing entire test run configuration.

I agree that removing the `test.xml` is probably not the best solution if we can basically achieve the same by checking for the ant runner (which is the only reading the `test.xml`) as proposed in this PR. Note that even if the local `test.xml` is removed, there will still be further configuration in the aggregator expecting that `test.xml` to exist (such as [in the global `test.xml`](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/415e32c0764b760b04f1a40bc19534c82dee93ab/production/testScripts/configuration/sdk.tests/testScripts/test.xml#L2325-L2329)), which then results in warnings/errors in the logs, empty test results etc.

iloveeclipse:
So let's merge it.

HeikoKlare:
> I still think that that this PR is a dirty hack and we will never see a clean fix after this. But as long as it disables the test i am fine with it.

If you see any better option, please explain or propose it. The only alternative we had so far ( #1364) had the same and even more drawbacks. You were asking for a quick fix of the test failure and now that @HannesWell took the time and proposed one, you complain about it. I really don't understand that.

HeikoKlare:
@HannesWell Unfortunately, it seems like this attempt was not successful, as the test still fails in today's I-Build. I guess even though the ant runner is for the overall execution, the tests will be individually run with the Eclipse application as defined in the `test.xml`. So for this plug-in it's the headless application `org.eclipse.test.coretestapplication` defined by using the [`core-test`](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/415e32c0764b760b04f1a40bc19534c82dee93ab/eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/src/main/scripts/library.xml#L179-L188) ant target.

Just as an idea: wouldn't it be possible to set an additional system property for the Ant build, which we can then evaluate in the test. In the best case, we could add this to the `test.xml` so that the workaround is encapsulated in the single test plug-in, but I am not sure if it is possible to set a system property there, because it probably has to be set in the `java` execution, so we might have to add it to the [global specification](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/415e32c0764b760b04f1a40bc19534c82dee93ab/eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/src/main/scripts/library.xml#L251). But I think that having a system property that allows to define an Ant build would not hurt, would it?

Issue #1197, opened 2024-02-09
Compare Tool misalign difference indicators if custom line spacing is used


xmichael91:
## Steps to reproduce

* Go to Preferences, and set some value for General->Editors->Text Editors->Line Spacing field (e.g. 25).
* Open Compare Tool for two different files/revisions (e.g. select two different files in your workspace and start the tool).
* Observe misaligned lines as on the attached screenshot.

![screenshot](https://github.com/eclipse-platform/eclipse.platform/assets/51908948/d3197590-994b-4a4d-9675-5223b8b8c246)

## Tested under this environment:
Eclipse IDE for RCP and RAP Developers (includes Incubating components)
Version: 2023-12 (4.30.0)
Build id: 20231201-2043
OS: Windows 10, v.10.0, x86_64 / win32
Java vendor: Eclipse Adoptium
Java runtime version: 17.0.10+7
Java version: 17.0.10

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

DenisUngemach:
Hello, 

in the TextMergeViewer in line 4530 the line height will be used directly, this seems to cause this error:

int lineHeight = tp.getSourceViewer().getTextWidget().getLineHeight();

The calculation of the line height is only the size of the textfont, and not the height with additional attributes like the line spacing.

I will check how to fix this.

Best regards,
Denis

DenisUngemach:
Hi, 

this commit fixes the problem: https://github.com/DenisUngemach/eclipse.platform
This is not perfect, because the real calculation of line heights should be inside the text widget and the renderer not outside in the compare tool.

iloveeclipse:
@DenisUngemach : I assume you mean this commit: https://github.com/DenisUngemach/eclipse.platform/commit/b4343794a49f718bcc1235fc3c1599527dc6354d (your link above doesn't work).

Could you please open a PR so the change could be reviewed properly?

DenisUngemach:
> Thanks Denis!

You're welcome. I will check the other issue you mentioned.

PR #1369, opened 2024-05-14
 fix: #1197 Compare Tool misalign

DenisUngemach:
Issue: Compare Tool misalign difference indicators if custom line spacing is used https://github.com/eclipse-platform/eclipse.platform/issues/1197

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
@@ -2238,7 +2238,6 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 
 	private Diff handleMouseInSides(Canvas canvas, MergeSourceViewer tp, int my) {
 
-		int lineHeight= tp.getSourceViewer().getTextWidget().getLineHeight();
 		int visibleHeight= tp.getViewportHeight();
 
 		if (! fHighlightRanges)
@@ -2258,8 +2257,8 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 					continue;
 
 				tp.getLineRange(diff.getPosition(leg), region);
-				int y= (region.x * lineHeight) + shift;
-				int h= region.y * lineHeight;
+				int y = getHeightBetweenLines(tp, 0, region.x) + shift;
+				int h = getHeightBetweenLines(tp, region.x, region.x + region.y);
 
 				if (y+h < 0)
 					continue;
@@ -2278,7 +2277,6 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 		if (! fSynchronizedScrolling)
 			return null;
 
-		int lineHeight= fLeft.getSourceViewer().getTextWidget().getLineHeight();
 		int visibleHeight= fRight.getViewportHeight();
 
 		Point size= canvas.getSize();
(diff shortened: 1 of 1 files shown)


team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
```diff
@@ -4232,6 +4232,16 @@ private boolean isAnySideEditable() {
 		return isLeftEditable() || isRightEditable();
 	}
 
+	/**
+	 * The line height does not only depend on the line height itself, but also on
+	 * the line spacing etc.
+	 */
+	private int getRendererLineHeight(MergeSourceViewer tp) {
+		return tp.getSourceViewer().getTextWidget().getLineHeight()
```
iloveeclipse, line 4240: 1) Could we avoid duplication and replace this with 

```
	private int getRendererLineHeight(MergeSourceViewer tp) {
		StyledText text = tp.getSourceViewer().getTextWidget();
		return text.getLineHeight() + text.getLineSpacing() + text.getLineVerticalIndent(0);
	}
```
2) `getLineVerticalIndent(0)` seem to be hard-coded to the first line. Do we assume it never changes?

DenisUngemach: 1. Sure.
2. I think this is actually a requirement with the StyledTextRenderer. 

In the class StyledTextRenderer in the method: getLineHeight(int lineIndex, boolean exact)
calculate the line height for a specific line will be calculated with:
getLineHeight() + getLineSpacing(lineIndex) + getLineVerticalIndent(lineIndex);

And in the paint methods where this method will be used, that what is actually wanted is not a line height, but the height from all lines before:
see for example Line 4438 in this file: 
	int y= (region.x * lineHeight) + shift;

here y is wanted, region.x are the number of lines.

I am currently not happy with this change, since you are right, each line could have different vertical indent in theory.

iloveeclipse (changes requested):
The change seem to be mixed with a complete code re-formatting. Please undo re.-formatting, a review is impossible now.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
DenisUngemach: 1. Done
2. I created a new method, which sums up for each line the indent and the spacing. This leads to an exact result.

DenisUngemach:
> The change seem to be mixed with a complete code re-formatting. Please undo re.-formatting, a review is impossible now.

I fixed this.
And sorry for the inconvenience.

DenisUngemach:
I think, now the change is ok.

iloveeclipse:
@DenisUngemach : thanks, looks much better now.

Could you please update commit message to follow https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations , to something like:

```
Misaligned range indicators in compare editor with non default line spacing

Compare editor difference range indicators were misaligned if custom line spacing is set under text editor preferences.
The reason is that original code didn't consider that editor lines can have (non-zero) spacing and indent, but these must be considered for the line height calculation.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1197
```

iloveeclipse:
One small cosmetic issue you could check (probably in other PR) is this (better to describe in a picture):
image

DenisUngemach:
> One small cosmetic issue you could check (probably in other PR) is this (better to describe in a picture)
sure.

iloveeclipse:
@DenisUngemach : I've updated commit message & rebased.

iloveeclipse (approved):

sratz:
> One small cosmetic issue you could check (probably in other PR) is this (better to describe in a picture): image

I have opened https://github.com/eclipse-platform/eclipse.platform.swt/issues/1286

PR #1371, opened 2024-05-15
Migrate non-serialization WorkspaceSessionTests to JUnit 5 #903

HeikoKlare:
This migrates all WorkspaceSessionTests that are not specializations of WorkspaceSerializationTest to JUnit 5, using the SessionTestExtension with the CustomSessionWorkspace. It extends the SessionTestExtension with functionality to set a system property as required for specific tests.


HeikoKlare:
> This pull request removes 21 tests.

This is (almost) expected as:
- In total, 25 test classes have been migrated
- 24 of them extended `WorkspaceSessionTest` so the now unnecessary "cleanup" dummy test was removed during migration (the other one did not extend that class, namely `TestBug94498`) 
- 2 of them defined their own test suite, so that the `cleanup` test of the super class was not executed, namely `Test1GALH44` and `TestBug294854`
- 1 of them was not even part of any test suite (`TestBug208833`). I will fix this in a separate PR.

PR #1372, opened 2024-05-15
Migrate WorkspaceSerializationTests to JUnit 5 #903

HeikoKlare:
This migrates all WorkspaceSerializationTests, which are specializations of WorkspaceSerializationTests to JUnit 5, using the SessionTestExtension with the CustomSessionWorkspace.


HeikoKlare:
> This pull request removes 15 tests.

This is as intended, as exactly 15 test classes have been migrated and the previous "cleanup" dummy test case is not necessary anymore.

Issue #1373, opened 2024-05-15
Debug "Prompter" handler dialogs may be hidden behind progress dialog


iloveeclipse:
Our users are sometimes killing Eclipse because they claim it is waiting forever with "Progress Information" on starting program execution.

This is a usability issue caused by a bad timing of dialogs. Precondition: launch config must have "launch in background" checkbox **disabled** like here:

![image](https://github.com/eclipse-platform/eclipse.platform/assets/964108/ddc7ebe9-d637-4c84-ab17-341dfa58aaff)

**Steps leading to the problem**

- Project has compilation errors.
- User ignores errors and starts debugging execution (because compile error is in unrelated hard to fix code).
- Launch validation checks take a bit longer, so progress dialog is shown asynchronously by system
- At same time, right before progress dialog is shown, launch validation shows "compilation error" dialog and waits for user to confirm if we should proceed with the launch.
- Progress dialog is a bit bigger then the error dialog and is shown over them, so the user can't see error dialog that waits for feedback.
- User only sees the progress dialog and waits for it to continue, the UI waits for user input...
- After some time user kills Eclipse

The `ProgressMonitorJobsDialog` that opens "Progress Information" dialog after 800 ms while waiting on `org.eclipse.debug.core.model.LaunchConfigurationDelegate.finalLaunchCheck()` opens right after the `org.eclipse.debug.internal.ui.launchConfigurations.CompileErrorProjectPromptStatusHandler.handleStatus()` opens the "Errors in Workspace" dialog.

I can almost always reproduce the bad timing with breakpoints.
- Conditional breakpoint in `org.eclipse.debug.core.model.LaunchConfigurationDelegate.finalLaunchCheck(ILaunchConfiguration, String, IProgressMonitor)` with `Thread.sleep(850);` "condition" right before `errors.add(0, configuration);` line.
- Line breakpoint at the line `dialog.open();` in `org.eclipse.ui.internal.progress.ProgressManager.scheduleProgressMonitorJob(ProgressMonitorJobsDialog)`
- Line breakpoint at line `int open = dialog.open();` in `org.eclipse.debug.internal.ui.launchConfigurations.CompileErrorProjectPromptStatusHandler.handleStatus(IStatus, Object)`

In that case if I return from debugger to Eclipse debuggee I see only the progress dialog, but not the error one (because it is hidden behind the progress).

Bad state:
[...]

PR #1374, opened 2024-05-15
Use proper shell for modal dialogs

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1373

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIPlugin.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIPlugin.java
@@ -373,6 +373,22 @@ public class DebugUIPlugin extends AbstractUIPlugin implements ILaunchListener, 
 	}
 
 	/**
+	 * Returns shell that can be used to get the best parenting possible for a modal
+	 * dialog, or, if there is no one, currently active workbench window shell or
+	 * <code>null</code> if none.
+	 *
+	 * @return the modal shell or currently active workbench window shell or
+	 *         <code>null</code>
+	 */
+	public static Shell getShellForModalDialog() {
+		Shell shell = PlatformUI.getWorkbench().getModalDialogShellProvider().getShell();
+		if (shell == null) {
+			shell = getShell();
+		}
+		return shell;
+	}
+
+	/**
 	 * Returns the default label provider for the debug UI.
 	 * @return the singleton {@link DefaultLabelProvider}
 	 */
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/WorkingDirectoryStatusHandler.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/WorkingDirectoryStatusHandler.java
@@ -33,7 +33,7 @@ public class WorkingDirectoryStatusHandler implements IStatusHandler {
 		DebugUIPlugin.getStandardDisplay().syncExec(() -> {
 			String title = DebugUIMessages.WorkingDirectoryStatusHandler_Eclipse_Runtime_1;
 			String message = DebugUIMessages.WorkingDirectoryStatusHandler_0;
-			result[0] = (MessageDialog.openQuestion(DebugUIPlugin.getShell(), title, message));
+			result[0] = (MessageDialog.openQuestion(DebugUIPlugin.getShellForModalDialog(), title, message));
 		});
 		return Boolean.valueOf(result[0]);
 	}
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/groups/GroupCycleHandler.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/groups/GroupCycleHandler.java
@@ -17,6 +17,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.debug.core.IStatusHandler;
 import org.eclipse.debug.internal.ui.DebugUIMessages;
+import org.eclipse.debug.internal.ui.DebugUIPlugin;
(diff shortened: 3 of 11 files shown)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/sourcelookup/Prompter.java
```diff
@@ -78,11 +90,59 @@ public Object handleStatus(final IStatus status, final Object source) throws Cor
...
 		if (exception[0] != null ) {
 			throw exception[0];
 		}
 		return result[0];
 	}
+
+	/**
+	 * Helper to activate possible opened and hidden prompt dialog
+	 */
+	private final class ForcePrompShellActivation implements Runnable {
```
trancexpress, line 105: This can be `static`.

iloveeclipse:
The reason I've added the change in the Prompter is that there is not only one handler that can open dialogs, it is defined by extension point and anyone can do whatever possible. So I would prefer to have a more generic solution.

The "borderless" dialog is the result of `SWT.ON_TOP` flag? Looks quite ... unusual :-)
Is there a more "convenient looking" way to achieve the same effect?

tomaswolf:
Wait a sec. Why exactly is this happening?

The progress dialog comes from `IProgressService.busyCursorWhile()`, which is defined not to open a progress dialog if there is a modal dialog. The "Errors in Workspace" dialog _is_ a modal dialog. And showing a dialog _must_ happen on the UI thread, right? So there can be exactly two outcomes:
* Error dialog comes first, and then no progress dialog is shown (yet).
* Or progress dialog comes first, and then error dialog is shown on top.

_However_: I do notice that the error dialog opening in `CompileErrorProjectPromptStatusHandler` uses the active workbench window's shell as parent for its message dialog. But modal dialogs should always use `PlatformUI.getWorkbench().getModalDialogShellProvider().getShell()`. I suspect #1373 is actually case (2): progress dialog comes first, but then error dialog comes later _but has the wrong parent shell_ and thus ends up somehow behind the progress dialog. (Besides, having two modal dialogs open simultaneously from the same parent shell quite often results in UI deadlocks in my experience.)

iloveeclipse:
> So there can be exactly two outcomes:

I believe there are three, and third one is: progress dialog is opened **after** error dialog.

iloveeclipse:
> > So there can be exactly two outcomes:
> 
> I believe there are three, and third one is: progress dialog is opened **after** error dialog.

Or four: focus / activation is given to the wrong shell - because of timing issues in SWT event handling, I believe I saw some SWT code about "fix shell focus" in GTK port.

tomaswolf:
> > So there can be exactly two outcomes:
> 
> I believe there are three, and third one is: progress dialog is opened **after** error dialog.

If so, the `IProgressService.busyCursorWhile()` is broken. Its javadoc clearly states that it does not open the dialog if there is already a modal dialog.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/deb75d4e977d767d3c5249e8a953fd2ef986baab/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/progress/IProgressService.java#L132-L133

I would really suggest to at least just try changing `CompileErrorProjectPromptStatusHandler` to use  `PlatformUI.getWorkbench().getModalDialogShellProvider().getShell()` instead of `DebugUIPlugin.getShell()` at https://github.com/eclipse-platform/eclipse.platform/blob/6cb771edfb0680235d2a371668f8e9d194458fba/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/CompileErrorPromptStatusHandler.java#L41

iloveeclipse:
I will play tomorrow with the code, sure.

iloveeclipse:
> I would really suggest to at least just try changing `CompileErrorProjectPromptStatusHandler` to use `PlatformUI.getWorkbench().getModalDialogShellProvider().getShell()` instead of `DebugUIPlugin.getShell()`

This works perfectly, many thanks Thomas for the hint! Now progress dialog can't even be activated. I will change related handlers to that.

merks:
This is such a good example of excellent teamwork!

iloveeclipse:
I've touched all `IStatusHandler` implementations in platform, the rest (JDT/PDE) will follow with similar changes.

trancexpress:
The change seems way more straightforward now.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIPlugin.java
```diff
@@ -372,6 +372,22 @@ public static Shell getShell() {
...
+	 * dialog, or, if there is no one, currently active workbench window shell or
+	 * <code>null</code> if none.
+	 *
+	 * @return the modal shell or currently active workbench window shell or
+	 *         <code>null</code>
+	 */
+	public static Shell getShellForModalDialog() {
+		Shell shell = PlatformUI.getWorkbench().getModalDialogShellProvider().getShell();
+		if (shell == null) {
+			shell = getShell();
```
tomaswolf, line 386: Why is this needed? Seems to me that the (default) IShellProvider `ProgressManagerUtil::getDefaultParent` returns a shell if there is one. If it returns null, there is no shell, and the `DebugUIPlugin` won't find any either.

iloveeclipse:
@tomaswolf , @trancexpress : any objections? If not, I would merge.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIPlugin.java
tomaswolf: Yes, it can return null. My point was: If it returns null, then `getShell()` won't find anything either.

tomaswolf:
No objections, but I still don't see the point of calling `DebugUIPlugin.getShell()` as a "fallback". It won't hurt, though.

Thanks!

PR #1375, opened 2024-05-15
Migrate SnapshotTest to JUnit 5 #903

HeikoKlare:
This migrates the SnapshotTest to JUnit 5, using the SessionTestExtension with the CustomSessionWorkspace. Also re-enables the test cases on MacOS.


HeikoKlare:

This is intended as the previous "cleanup" dummy test case is not necessary anymore.

fedejeanne (approved):

PR #1378, opened 2024-05-16
Fix deletion of configuration and workspace directory for session tests

HeikoKlare:
The CustomSessionConfigurationImpl and CustomSessionWorkspaceImpl use `deleteOnExit()` to remove the configuration and workspaces directories on JVM shutdown. However, this does not work recursively and thus usually fails when the directory is not empty. This change corrects the cleanup by adding a shutdown hook (like used by `deleteOnExit()`) that recursively deletes all contents of the directory and only uses Java NIO instead of Java IO.


HeikoKlare:
Affects only tests, so can be merged despite freeze period.

Issue #1318, opened 2024-04-17
Launch View: "Terminate and Relaunch" sometimes does not relaunch all processes of a Launch Group


jukzi:
In the launch View selecting a running launch group and choosing "Terminate and Relaunch" from the contet menu sometimes fails to relaunch the whole group or some of the processes. 
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/8c2901e5-2123-42fe-abae-61efe7d047d4)

After that the launch view shows two nodes for the launchgroup, both showing the same - terminated - PIDs:
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/891e5a26-6e92-461d-82a0-8082eeb7248e)

Reproducable with a simple Program like:

```
public class Main {
	public static void main(String[] args) {
		while (true) {}
	}
}
```
Create 2 launches for the Program (just create 1 and duplicate it).
Create 1 launchgroup to start both launches.
Set option "adopt if runing"
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/30f7d104-7a5a-4417-a097-f603ea8b3559)

"run" the launch group, 
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/f3548222-5b63-4613-8237-324446c376aa)

"Terminate and Relaunch" the launch group. Sometimes it is not fully relaunched:
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/4720f28e-48a7-4372-8c64-07a69b49b82f)

I have only seen it failing when option "adopt if runing" was used.

PR #1381, opened 2024-05-17
Fix "Terminate and Relaunch" of groups that use adoptIfRunning #1318

jukzi:
This was a race condition:
RuntimeProcess.fTerminated was only set asynchronously, hindering relaunching the process, thinking it was already running.

Happened only for GroupLaunchElement.adoptIfRunning == true

fixes
https://github.com/eclipse-platform/eclipse.platform/issues/1318


jukzi:
ignoring unrelated random fail ProcessConsoleManagerTests

PR #1382, opened 2024-05-17
Check 'ECLIPSE_I_BUILD_TEST' to disable 'ensureJavaxIsNotAvailable' case

HannesWell:
Check the newly introduced 'ECLIPSE_I_BUILD_TEST' environment variable to disable the InjectionOSGiTest.ensureJavaxIsNotAvailable()' test case in I-build tests.
The previous attempt to disable it failed.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1098 Requires
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2031


HeikoKlare (approved):
Thank you, @HannesWell! The combination of the two PRs looks good and sound. Even if it would not work as expected, I at least don't see a risk of breaking anything. So from my side, let's just merge and see what happens in next I-Build 😊

HannesWell:
> Even if it would not work as expected, I at least don't see a risk of breaking anything. 

Absolutely, then let's go. Thanks.

PR #1384, opened 2024-05-20
Add Ant test.xml for org.eclipse.core.tests.resources.saveparticipant

HeikoKlare:
This adds a test.xml for Ant-based execution of tests in org.eclipse.core.tests.resources.saveparticipant. It is a follow-up to the addition of these tests to the Maven build.


HeikoKlare:
This PR does nothing on its own but will only enable the saveparticipant tests to be run via an Ant build, as will be enabled in the aggregator via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2035

The changes are analogous to what we did in #1147 for the tips tests.

Only test are affected, thus we can merge this despite freeze period.

PR #1386, opened 2024-05-21
Remove unused WorkspaceSessionTest(Suite) #903

HeikoKlare:
The WorkspaceSessionTest and WorkspaceSessionTestSuite are now unused as all their previous consumers and implementations have been migrated to the JUnit 5 SessionTestExtension with the CustomSessionWorkspace extension. This completely removes both classes.

I have checked that no remaining usages exist in the Eclipse SDK (with platform, Equinox, JDT and PDE).


HeikoKlare:
Only tests affected, so can be merged during freeze period.

PR #1387, opened 2024-05-21
Reduce number of parallel builds in ParallelBuildChainTest #825

HeikoKlare:
The test case testIndividualProjectBuilds_WithManyProjects_ProjectRelaxedRule() in ParallelBuildChainTest fails randomly (in I-Builds) because not all jobs that are supposed to run build operations are scheduled and started in time. Since the test is not supposed to validate that a specific number of jobs can be started, the number of jobs/builds to be spawned must actually not be as high as currently defined. This change reduces the number of projects in the test case to avoid random test failures while keeping expressiveness of the test case.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/825


HeikoKlare:
Build failure is only because of new deprecation issue because of gson version increment: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2036

Triggered build for `master` reports the same new issue: https://ci.eclipse.org/platform/job/eclipse.platform/job/master/847/

Jenkins reference build has been updated but was not properly taken into account for this PR build, which is still compared against the previous master build.

Change only affects test, thus merging despite freeze period.

PR #1388, opened 2024-05-22
JsonReader.setStrictness(Strictness.LENIENT) instead of setLenient(true)

merks:
- JsonReader.setLenient latter is deprecated in 2.11.0 and setStrictness is new to 2.11.0.


merks:
@akurtakov 

There is no strong need for this, but someone mentioned a new warning so  I tracked it down and fixed it.  But it can wait for the next release cycle...

akurtakov (approved):
This is good to go now.

merks:
Test failures unrelated.

PR #1391, opened 2024-05-23
Force signing of split package bundles to use new certificate

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2044


merks:
All the PRs are created and each is marked as a draft to ensure that no one merges them.

PR #1397, opened 2024-05-29
Provide API for IFile Bulk read/write of all bytes/String/chars

jukzi:
Convenience methods like java.nio.file.Files.readAllBytes(Path) etc.

The new API could be used to substitute similar Util methods like org.eclipse.jdt.internal.core.util.Util.getResourceContentsAsCharArray (hotspot of jdt pasrsing)
Searching the workspace like
org.eclipse.search.internal.core.text.FileCharSequenceProvider.toShortString(IFile) is similar but not directly replaceable with this API.

Note that clients currently do have to query file.getCharset() and file.getContents() separately, possible opening the file twice to identify the UTF-BOM while the new API combines both calls, which would allow platform to open the file single time. While this PR does not implements such optimizations it is already faster to use the JDKs bulk methods instead of streaming.

Also this API avoids typical warnings of not closed Streams with the former stream api in client code.

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/IFileStore.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/IFileStore.java
@@ -15,12 +15,17 @@
 ******************************************************************************/
 package org.eclipse.core.filesystem;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
 import org.eclipse.core.filesystem.provider.FileStore;
 import org.eclipse.core.internal.filesystem.FileStoreUtil;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IAdaptable;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Status;
 
 /**
  * A file store is responsible for storage and retrieval of a single file in some file system.
@@ -413,6 +418,29 @@ public interface IFileStore extends IAdaptable {
 	public InputStream openInputStream(int options, IProgressMonitor monitor) throws CoreException;
 
 	/**
+	 * Returns the bytes contents of this file. This method is not intended for reading in
+	 * large files that do not fit in a byte array. Preferable over calling
+	 * {@code openInputStream(options, monitor).readAllBytes()}
+	 * @param options bit-wise or of option flag constants (currently only {@link EFS#NONE}
+	 * is applicable).
+	 * @param monitor a progress monitor, or <code>null</code> if progress
+	 *    reporting and cancellation are not desired
+	 * @return contents of this file as byte array.
+	 * @throws CoreException if this method fails.
+	 * @see #openInputStream(int, IProgressMonitor)
+	 * @since 1.11
+	 */
+	public default byte[] readAllBytes(int options, IProgressMonitor monitor) throws CoreException {
+		// Default implementation is meant to be overridden for local files
+		// as Files.readAllBytes is ~ 1.5 times faster
+		try (InputStream stream = openInputStream(options, monitor)) {
+			return stream.readAllBytes();
+		} catch (IOException e) {
+			throw new CoreException(Status.error("Error reading " + getName(), e)); //$NON-NLS-1$
+		}
+	}
+
+	/**
 	 * Returns an open output stream on the contents of this file. The number of
 	 * concurrently open streams depends on implementation and can be limited.
 	 * The caller is responsible for closing the provided stream when it is no longer
@@ -448,6 +476,31 @@ public interface IFileStore extends IAdaptable {
 	 * </ul>
 	 */
 	public OutputStream openOutputStream(int options, IProgressMonitor monitor) throws CoreException;
+
+	/**
+	 * Writes the contents to this file.
+	 * <p>
+	 * The {@link EFS#APPEND} update flag controls where
+	 * output is written to the file. If this flag is specified, content written
+	 * to the stream will be appended to the end of the file. If this flag is
+	 * not specified, the contents of the existing file, if any, is truncated to zero
+	 * and the new output will be written from the start of the file.
+	 * </p>
+	 *
+	 * @param content contents given as byte array
+	 * @param options bit-wise or of option flag constants ({@link EFS#APPEND}).
+	 * @param monitor a progress monitor, or <code>null</code> if progress
+	 *    reporting and cancellation are not desired
+	 * @see #openOutputStream(int, IProgressMonitor)
+	 * @since 1.11
+	 * **/
+	public default void write(byte[] content, int options, IProgressMonitor monitor) throws CoreException {
+		try (OutputStream out = openOutputStream(options, monitor)) {
+			out.write(content);
+		} catch (IOException e) {
+			throw new CoreException(Status.error(e.getMessage(), e));
+		}
+	}
 
 	/**
 	 * Writes information about this file to the underlying file system. Only
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
@@ -31,6 +31,7 @@ import java.nio.file.DirectoryNotEmptyException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
+import java.nio.file.StandardOpenOption;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.IFileStore;
@@ -447,6 +448,7 @@ public class LocalFile extends FileStore {
 		}
 	}
 
+	/** @see #readAllBytes(int, IProgressMonitor) */
 	@Override
 	public InputStream openInputStream(int options, IProgressMonitor monitor) throws CoreException {
 		try {
@@ -467,6 +469,28 @@ public class LocalFile extends FileStore {
 		}
 	}
 
+	/** @see #openInputStream(int, IProgressMonitor) */
+	@Override
+	public byte[] readAllBytes(int options, IProgressMonitor monitor) throws CoreException {
+		try {
+			return Files.readAllBytes(file.toPath());
+		} catch (IOException e) {
+			String message;
+			if (!file.exists()) {
+				message = NLS.bind(Messages.fileNotFound, filePath);
+				Policy.error(EFS.ERROR_NOT_EXISTS, message, e);
+			} else if (file.isDirectory()) {
+				message = NLS.bind(Messages.notAFile, filePath);
+				Policy.error(EFS.ERROR_WRONG_TYPE, message, e);
+			} else {
+				message = NLS.bind(Messages.couldNotRead, filePath);
+				Policy.error(EFS.ERROR_READ, message, e);
+			}
+			return null;
+		}
+	}
+
+	/** @see #write(byte[], int, IProgressMonitor) */
 	@Override
 	public OutputStream openOutputStream(int options, IProgressMonitor monitor) throws CoreException {
 		try {
@@ -483,6 +507,30 @@ public class LocalFile extends FileStore {
 				Policy.error(EFS.ERROR_WRITE, message, e);
 			}
 			return null;
+		}
+	}
+
+	/** @see #openOutputStream(int, IProgressMonitor) */
+	@Override
+	public void write(byte[] content, int options, IProgressMonitor monitor) throws CoreException {
+		try {
+			boolean append = (options & EFS.APPEND) != 0;
+			if (append) {
+				Files.write(file.toPath(), content, StandardOpenOption.CREATE, StandardOpenOption.WRITE);
+			} else {
+				Files.write(file.toPath(), content); // default uses StandardOpenOption.TRUNCATE_EXISTING
+			}
+		} catch (IOException e) {
+			checkReadOnlyParent(file, e);
+			String message;
+			String path = filePath;
+			if (file.isDirectory()) {
+				message = NLS.bind(Messages.notAFile, path);
+				Policy.error(EFS.ERROR_WRONG_TYPE, message, e);
+			} else {
+				message = NLS.bind(Messages.couldNotWrite, path);
+				Policy.error(EFS.ERROR_WRITE, message, e);
+			}
 		}
 	}
 
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -424,11 +424,8 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 	 * are not considered.
 	 */
 	private boolean descriptionChanged(IFile descriptionFile, byte[] newContents) {
-		//buffer size: twice the description length, but maximum 8KB
-		int bufsize = newContents.length > 4096 ? 8192 : newContents.length * 2;
-		try (
-			InputStream oldStream = new BufferedInputStream(descriptionFile.getContents(true), bufsize);
-		) {
+		try {
+			InputStream oldStream = new ByteArrayInputStream(descriptionFile.readAllBytes());
 			InputStream newStream = new ByteArrayInputStream(newContents);
 			//compare streams char by char, ignoring line endings
 			int newChar = newStream.read();
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -870,6 +870,19 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 	}
 
 	public InputStream read(IFile target, boolean force, IProgressMonitor monitor) throws CoreException {
+		IFileStore store = getFileStore(target, force);
+		try {
+			return store.openInputStream(EFS.NONE, monitor);
+		} catch (CoreException e) {
+			if (e.getStatus().getCode() == EFS.ERROR_NOT_EXISTS) {
+				String message = NLS.bind(Messages.localstore_fileNotFound, store.toString());
+				throw new ResourceException(IResourceStatus.RESOURCE_NOT_FOUND, target.getFullPath(), message, e);
+			}
+			throw e;
+		}
+	}
+
+	private IFileStore getFileStore(IFile target, boolean force) throws ResourceException, CoreException {
 		IFileStore store = getStore(target);
 		if (lightweightAutoRefreshEnabled || !force) {
 			final IFileInfo fileInfo = store.fetchInfo();
@@ -878,9 +891,10 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 				String message = NLS.bind(Messages.localstore_fileNotFound, store.toString());
 				throw new ResourceException(IResourceStatus.RESOURCE_NOT_FOUND, target.getFullPath(), message, null);
 			}
-			ResourceInfo info = ((Resource) target).getResourceInfo(true, false);
-			int flags = ((Resource) target).getFlags(info);
-			((Resource) target).checkExists(flags, true);
+			Resource resource = (Resource) target;
+			ResourceInfo info = resource.getResourceInfo(true, false);
+			int flags = resource.getFlags(info);
+			resource.checkExists(flags, true);
 			if (fileInfo.getLastModified() != info.getLocalSyncInfo()) {
 				asyncRefresh(target);
 				if (!force) {
@@ -889,8 +903,14 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 				}
 			}
 		}
+		return store;
+	}
+
+	/** @see #read(IFile, boolean, IProgressMonitor) **/
+	public byte[] readAllBytes(IFile target, boolean force, IProgressMonitor monitor) throws CoreException {
+		IFileStore store = getFileStore(target, force);
 		try {
-			return store.openInputStream(EFS.NONE, monitor);
+			return store.readAllBytes(EFS.NONE, monitor);
 		} catch (CoreException e) {
 			if (e.getStatus().getCode() == EFS.ERROR_NOT_EXISTS) {
 				String message = NLS.bind(Messages.localstore_fileNotFound, store.toString());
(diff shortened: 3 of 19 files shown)


mickaelistria:
Can you please find some example of consumers in actual platform code and include replacement of older code with newer one in this PR so we can easily see the benefit of this new API over legacy and more easily discuss/review it?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -903,6 +903,38 @@ public InputStream read(IFile target, boolean force, IProgressMonitor monitor) t
...
+	public byte[] readAllBytes(IFile target, boolean force, IProgressMonitor monitor) throws CoreException {
+		IFileStore store = getStore(target);
+		if (lightweightAutoRefreshEnabled || !force) {
+			final IFileInfo fileInfo = store.fetchInfo();
+			if (!fileInfo.exists()) {
+				asyncRefresh(target);
+				String message = NLS.bind(Messages.localstore_fileNotFound, store.toString());
+				throw new ResourceException(IResourceStatus.RESOURCE_NOT_FOUND, target.getFullPath(), message, null);
+			}
+			ResourceInfo info = ((Resource) target).getResourceInfo(true, false);
```
laeubi, line 916: Extract target into local variable avoids multiple casts

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -903,6 +903,38 @@ public InputStream read(IFile target, boolean force, IProgressMonitor monitor) t
 		}
 	}
 
+	/** @see #read(IFile, boolean, IProgressMonitor) **/
+	public byte[] readAllBytes(IFile target, boolean force, IProgressMonitor monitor) throws CoreException {
```
laeubi, line 907: Must this be public or could it be package private maybe?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1289,6 +1321,107 @@ public void write(IFile target, InputStream content, IFileInfo fileInfo, int upd
 		}
 	}
 
+	/**
+	 * like
+	 * {@link #write(IFile, InputStream, IFileInfo, int, boolean, IProgressMonitor)}
+	 * but with byte array content instead of stream
+	 */
+	public void write(IFile target, byte[] content, IFileInfo fileInfo, int updateFlags, boolean append,
```
laeubi, line 1329: package private?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -299,6 +300,17 @@ public InputStream getContents(boolean force) throws CoreException {
 		return getLocalManager().read(this, force, null);
 	}
 
+	/** like {@link #getContents(boolean)} **/
+	@Override
+	public byte[] readAllBytes() throws CoreException {
+		boolean force = true;
```
laeubi, line 306: Inline?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -371,6 +391,32 @@ public void setContents(InputStream content, int updateFlags, IProgressMonitor m
 			subMonitor.done();
 		}
 	}
+	@Override
+	public void setContents(byte[] content, int updateFlags, IProgressMonitor monitor) throws CoreException {
+		String message = NLS.bind(Messages.resources_settingContents, getFullPath());
```
laeubi, line 396: Inline?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -371,6 +391,32 @@ public void setContents(InputStream content, int updateFlags, IProgressMonitor m
...
+		SubMonitor subMonitor = SubMonitor.convert(monitor, message, 100);
+		try {
+			if (workspace.shouldValidate)
+				workspace.validateSave(this);
+			final ISchedulingRule rule = workspace.getRuleFactory().modifyRule(this);
+			SubMonitor newChild = subMonitor.newChild(1);
```
laeubi, line 402: Inline?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
```diff
@@ -340,6 +347,71 @@ public static final void transferStreams(InputStream source, OutputStream destin
 		}
 	}
 
+	public static char[] readAllChars(IFile file) throws CoreException {
```
laeubi, line 350: Better place then in IFile as well, then it could be overwritten for local files as well.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
```diff
@@ -340,6 +347,71 @@ public static final void transferStreams(InputStream source, OutputStream destin
...
+		byte[] content = file.readAllBytes();
+		Charset charset = getCharset(file, content);
+		return FileUtil.toCharArray(content, charset);
+	}
+
+	public static String readString(IFile file) throws CoreException {
+		byte[] content = file.readAllBytes();
```
laeubi, line 357: this can use `readAllChars()`

laeubi (changes requested):
In general I like the idea and it seems a good improvement, but currently I see the following problems:

- [ ] API tools complains about wrong since
- [ ] Compiler complains about usage of deprecated methods
- [ ] Instead of inventing new static util methods, these should be (default implemented) methods in `IFile` so they can be easily overridden by more efficient implementations.

jukzi:
> Can you please find some example of consumers in actual platform code and include replacement of older code with newer one in this PR so we can easily see the benefit of this new API over legacy and more easily discuss/review it?

platform uses itself mainly in tests and never char[]
There are some calls to org.eclipse.core.resources.IFile.getContents(boolean) with parameter false in the test. That has no new aequivalent. 
 have refactored the rest, which already gives various examples how it can be used to simplify sourcecode.

szarnekow (commented):
I appreciate the effort and this looks already like a nice improvement.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: must be public otherwise org.eclipse.core.internal.resources.File.readAllBytes() -> "The method readAllBytes(IFile, boolean, IProgressMonitor) from the type FileSystemResourceManager is not visible"

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
szarnekow: Though it could, I'd rather avoid converting the bytes to chars here such that we can benefit from compressed strings (represented as bytes internally) without the intermediate allocation of the char[]

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: s.o.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
jukzi: I prefer short lines and consistency with the other code around.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
jukzi: I  don't see the benefit over consistency with the other code around.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
jukzi: I don't understand what you mean since LocalFile is not in the type hierarchy of IFIle. I put that code in a Utils class so that the Interface is not polluted with code. Also i am not aware why it should be ever overridden.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
jukzi: That would introduce ephemeral memory allocation for the char array which is not needed otherwise. This is a hotspot and optimized for speed.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
```diff
@@ -447,6 +447,7 @@ private boolean isSameFile(File source, File destination) throws IOException {
 		}
 	}
 
+	/** like {@link #readAllBytes(int, IProgressMonitor)} */
```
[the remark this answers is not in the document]
jukzi: I am not used to @inheritDoc, but it seems unrelated, since i do not try to inherit the documentation from the overriden. Also see https://www.cmos.blog/sinn-unsinn-von-inheritdoc/
Can you please suggest a concrete rewording? The idea is that all those methods are very similar and just return another Datatype.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+	 * Reads the content as char array. Skips the utf BOM header if any. This method
+	 * is not intended for reading in large files that do not fit in a char array.
+	 * Preferable equivalent of calling {@code readString().toCharArray()}
+	 *
+	 * @return content String without UTF BOM header
+	 * @throws CoreException on error
+	 * @see #readString()
+	 * @since 3.21
+	 */
+	public default char[] readAllChars() throws CoreException {
```
HannesWell, line 1246: Is this method that returns the in-between of a byte-array and a String really necessary?
AFAICT operating on char-arrays is not that common and there are more high-level and saver APIs like StringBuilder.
E.g. nio-Files also just offers `readAllBytes()` and `readString()/readLines()`.

In order to not blow up the API unnecessarily I think potentiall callers should be checked if they can work well with other the other methods.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
```diff
@@ -486,6 +509,25 @@ public OutputStream openOutputStream(int options, IProgressMonitor monitor) thro
 		}
 	}
 
+	/** like {@link #openOutputStream(int, IProgressMonitor)} */
+	@Override
+	public void write(byte[] content, int options, IProgressMonitor monitor) throws CoreException {
+		try {
+			Files.write(file.toPath(), content);
```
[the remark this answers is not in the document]
jukzi: nice catch! lets see if we can handle EFS.APPEND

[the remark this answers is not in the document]
jukzi: How could append=true be used? As far as i see the only possibe call stack uses append=false
image and there is no public api to test append=true

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -903,6 +903,38 @@ public InputStream read(IFile target, boolean force, IProgressMonitor monitor) t
...
+			int flags = ((Resource) target).getFlags(info);
+			((Resource) target).checkExists(flags, true);
+			if (fileInfo.getLastModified() != info.getLocalSyncInfo()) {
+				asyncRefresh(target);
+				if (!force) {
+					String message = NLS.bind(Messages.localstore_resourceIsOutOfSync, target.getFullPath());
+					throw new ResourceException(IResourceStatus.OUT_OF_SYNC_LOCAL, target.getFullPath(), message, null);
+				}
+			}
+		}
```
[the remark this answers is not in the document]
jukzi: good idea

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -299,6 +300,17 @@ public InputStream getContents(boolean force) throws CoreException {
 		return getLocalManager().read(this, force, null);
 	}
 
+	/** like {@link #getContents(boolean)} **/
```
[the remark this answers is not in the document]
jukzi: will add `with parameter force=true`

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: done

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
jukzi: that is outlined to have symmetry with org.eclipse.core.internal.resources.File.getContents(boolean)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1289,6 +1321,107 @@ public void write(IFile target, InputStream content, IFileInfo fileInfo, int upd
...
+		boolean restoreHiddenAttribute = false;
+		if (fileInfo.exists() && fileInfo.getAttribute(EFS.ATTRIBUTE_HIDDEN)
+				&& Platform.getOS().equals(Platform.OS_WIN32)) {
+			fileInfo.setAttribute(EFS.ATTRIBUTE_HIDDEN, false);
+			store.putInfo(fileInfo, EFS.SET_ATTRIBUTES, subMonitor.split(1));
+			restoreHiddenAttribute = true;
+		} else {
+			subMonitor.split(1);
+		}
+		int options = append ? EFS.APPEND : EFS.NONE;
```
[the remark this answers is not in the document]
jukzi: not anymore, but still i think that path is never taken.

HannesWell (commented):
This is an important enhancement because currently reading/writing content from IFiles is more cumbersome than it has to be.

> Instead of inventing new static util methods, these should be (default implemented) methods in IFile so they can be easily overridden by more efficient implementations.

`IFile` is marked as `@noExtend` and `@noImplement`, so as long as default implementations don't save us code, I don't think defaults are necessary.

The nio-Files class also provides methods for reading/writings the lines of a (text)-file into/from a List/Stream. I find this often convenient and it has the advantage that not a large byte-array/String has to be created.

As a in principle different approach we could also think about providing a custom nio-`FileSystem` that reflects Eclipse's workspace file-system and corresponding IPath implementations. Maybe IPath could then just extend nio-Path.
This would have the advantage that we can just use the standard Java `Files` API and would benefit from potential future enhancements.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: JDT requires char array. I don't like that but JDT has voted against changing everything to String. Converting from String to char[] is a memory and performance overhead. Converting from byte[] to char[] requires the Charset to be known, but that is currently only known by platform.

jukzi:
> `IFile` is marked as `@noExtend` and `@noImplement`, so as long as default implementations don't save us code, I don't think defaults are necessary.

So far the theory. We know that not all clients obey such. As a default are an easy way there seems to be no need to prvent using default.

> The nio-Files class also provides methods for reading/writings the lines of a (text)-file into/from a List/Stream. I find this often convenient and it has the advantage that not a large byte-array/String has to be created.

I am not aware of any such code actually used in any hotspot, so i do not see a need for that. Still you could suggest such methods in a separate PR.

> As a in principle different approach we could also think about providing a custom nio-`FileSystem` that reflects Eclipse's workspace file-system and corresponding IPath implementations. Maybe IPath could then just extend nio-Path. This would have the advantage that we can just use the standard Java `Files` API and would benefit from potential future enhancements.

That would be really cool, but needs intensive performance testing and my first guess is that the IPath to Path conversions may be a performance or memory bottle neck. I do not plan to work on that by now.

jukzi:
> * [ ]  API tools complains about wrong since

i guess i have to fix that after master is open with appropriate bundle version updates.

> * [ ]  Compiler complains about usage of deprecated methods

That may not be avoidable as most of the code is copy and paste of existing code that already has such warnings. This PR is not about to change that but to keep code consistent with the existing implementation.

laeubi:
> That may not be avoidable as most of the code is copy and paste of existing code that already has such warnings. 

Please just fix that, if we "copy & paste" Warnings this does not help in any way, if there is no easy alternative the warning must be suppressed, in most cases it is simply that the name changed or another call should be used instead.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
HannesWell: Understand. Too bad.
Would it be possible to achieve the same performance by using only internals?
I'm asking because I'm not sure we should promote using char-arrays in general by adding corresponding APIs?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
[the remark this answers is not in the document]
Bananeweizen: Please avoid using inheritdoc, it makes comments much harder to explore both in the Eclipse IDE as well as in generated Javadoc (because both the Javadoc view as well as the doclet generate better information _without_ that tag).

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
laeubi: I won't mind that additional method, it does not hurt to have i in the API as it is default implemented.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+		// as Files.readAllBytes is ~ 1.5 times faster
+		try (InputStream stream = getContents(true)) {
+			return stream.readAllBytes();
+		} catch (IOException e) {
+			throw new CoreException(Status.error("Error reading " + getFullPath(), e)); //$NON-NLS-1$
+		}
+	}
+
+	/**
+	 * Reads the content as char array. Skips the utf BOM header if any. This method
```
merks, line 1237: Picky things.  It's UTF not utf.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+	 * is not intended for reading in large files that do not fit in a char array.
+	 * Preferable equivalent of calling {@code readString().toCharArray()}
```
merks, line 1239: Picky.  Sentences end in a period.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+	 *
+	 * @return content String without UTF BOM header
```
merks, line 1241: It does not return a String.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+	 * @throws CoreException on error
+	 * @see #readString()
+	 * @since 3.21
+	 */
+	public default char[] readAllChars() throws CoreException {
+		return FileUtil.readAllChars(this);
+	}
+
+	/**
+	 * Reads the content as String. Skips the utf BOM header if any. This method is
```
merks, line 1251: UTF.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: > promote using char-arrays

There is nothing bad about returning arrays (same as byte array).

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: thanks

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: thanks

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: thanks

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: thanks

jukzi:
CI complains with warnings like `Your build strictly depends on unit org.eclipse.core.resources.source 3.20.200.v20240513-1323 that is shadowed by a reactor project, this can lead to unexpected build results!`
Similar warnings did already exist. can that be ignored?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
HannesWell: > > promote using char-arrays
> 
> There is nothing bad about returning arrays (same as byte array).

Yes, the 'idea' was that for low level access there is the bytes method and otherwise the String.
But I understand the need for JDT and if everyone is fine with it, I'm fine as well.

jukzi:
All checks passed and i have solved all review comments as far as it made sense. Please finish your reviews.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/IFileStore.java
```diff
@@ -449,6 +477,16 @@ public interface IFileStore extends IAdaptable {
 	 */
 	public OutputStream openOutputStream(int options, IProgressMonitor monitor) throws CoreException;
 
+	/** @since 1.11 **/
```
iloveeclipse, line 480: Javadoc is missing but should be added for new API.

jukzi: ok, added

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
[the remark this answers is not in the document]
jukzi: ok, changed to @see

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1163,6 +1173,93 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 */
 	void setContents(IFileState source, int updateFlags, IProgressMonitor monitor) throws CoreException;
 
+	/**
+	 * Preferable equivalent of calling
```
HannesWell, line 1177: Can elaborate on this why it is preferable?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectPreferences.java
```diff
@@ -198,7 +197,7 @@ private static Properties loadProperties(IFile file) throws BackingStoreExceptio
 			Policy.debug("Loading preferences from file: " + file.getFullPath()); //$NON-NLS-1$
 		Properties result = new Properties();
 		try (
-			InputStream input = new BufferedInputStream(file.getContents(true));
+				ByteArrayInputStream input = new ByteArrayInputStream(file.readAllBytes());
```
HannesWell, line 200: Doesn't this increase the memory consumption and GC pressure noticeably if the preference files are larger here and below? For project files it's probably not a big problem since they are usually relatively small, but preference files are often larger.

jukzi: I am not aware of any preference file that has any significant size. Especially it is just a key value store that is completely read into memory anyway.  For the GC it's easier to clean up a single array that has the exact needed size instead of heaving to cleaning up multiple temp Buffers.

mickaelistria: FWIW, the biggest org.eclipse.jdt.core.prefs files that I see in my project are slightly less than 40k bytes. It's more a data point, I don't know if it qualifies as significant or not in this context.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/FileUtil.java
laeubi: there is now `IFile.readAllBytes()` and in the same it should have `IFile.readAllChars()` that the might delegate to the (LocalFile)Store for a maybe better alternative ... this does not really "pollutes" the interface instead everyone requiring to reference a Util method does + it require us to maintain the Utilclass as API.

Alternativly you can follwo the Aproach of the JDK, there each Path has a "Filesystem" (here each IFIle has a FileStore) and the Util method effectivly delegate to the FileSystem, that way the Filesystem defines how things are performed and you don't need any readAlBytes() in the File API. Then of course it would be bst to place the Util next to IFile and maybe name it `IFiles` instead of FileUtils...)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: done.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectPreferences.java
mickaelistria: But realistically, there is nothing in the API that discourages a .prefs file to be 100MB, so assuming all those .prefs files are slim seems too optimistic to me. I support @HannesWell 's impression that unless we have full control or know constraints about the file, going for a stream is safer.
Maybe there is a way to determine the best strategy according to the file size?

laeubi: I agree with @HannesWell here that this looks strange, why should it be better to first read all bytes from a file/stream, then wrap then into a stream and then read that stream again?

e.g. `readAllBytes` of java NIO is effectively creating a buffer array with the size of the file and reads it then from the stream, the same as `BufferedInputStream` would do, so choosing a suitable buffer size here would have the same effect.

jukzi: I removed that example from commits to get progress here.

Issue #1390, opened 2024-05-23
Ant Build Failed: buildExtraJAR


jukzi:
With a clean OOmphed platform workspace i straight ran into this logged exception:

```
org.eclipse.core.runtime.CoreException: C:\Users\jkubitz\platform-2024-05-23\git\eclipse.platform\resources\bundles\org.eclipse.core.resources\scripts\buildExtraJAR.xml:38: Compile failed; see the compiler error output for details.
	at org.eclipse.ant.core.AntRunner.handleInvocationTargetException(AntRunner.java:443)
	at org.eclipse.ant.core.AntRunner.run(AntRunner.java:375)
	at org.eclipse.ant.internal.launching.launchConfigurations.AntLaunchDelegate.runInSameVM(AntLaunchDelegate.java:299)
	at org.eclipse.ant.internal.launching.launchConfigurations.AntLaunchDelegate.launch(AntLaunchDelegate.java:261)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:805)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:716)
	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:711)
	at org.eclipse.core.externaltools.internal.model.ExternalToolBuilder.launchBuild(ExternalToolBuilder.java:188)
	at org.eclipse.core.externaltools.internal.model.ExternalToolBuilder.doBuildBasedOnScope(ExternalToolBuilder.java:176)
	at org.eclipse.core.externaltools.internal.model.ExternalToolBuilder.build(ExternalToolBuilder.java:95)
	at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:1077)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:296)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:352)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:441)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:444)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:555)
[...]

PR #1398, opened 2024-05-30
Fix buildExtraJAR.xml to jar the bin_ant content into resource-ant.jar

merks:
- Remove the tasks that try to compile/copy the src_ant because the build.properties already properly populates the bin_ant folder with the contents needed in resource-ant.jar.

https://github.com/eclipse-platform/eclipse.platform/issues/1390

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
+++ b/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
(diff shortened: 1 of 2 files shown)


laeubi:
Ant is included by default, otherwise one could simply use Tycho as well ;-)

HannesWell:
> Ant is included by default, otherwise one could simply use Tycho as well ;-)

Looking at the `build.properties`, that's what already happening at build time.
If requiring m2e would be ok, it might be sufficient to just add the m2e nature/builder and we could remove all extra ant configuration.

merks:
This same pattern is repeated in many places.  If "we' (by which I mean, someone other than me) are going to change it, we should change them all.

image

I have quite bad experience with m2e doing way more than I want it to do.  But that probably because it's poorly configured in those case.

The aggregator setup already requires m2e because it needed for the *.target...

merks:
There's no point in doing builds right now because they all just fail even without changes.

PR #1403, opened 2024-06-03
Pin inspect

mickaelistria:


iloveeclipse:
I have merge issues trying to rebase on master. Please fix & push.

jukzi (dismissed):
i tested this change with a simple example

```
package t;

public class Main {
	record S(String s) {};
		for (int i=0; i<10;i++) {
			final int j=i;
		new Thread() {
			S s;
			@Override
			public void run() {
				s=new S(""+j);
				s=null; // <- breakpoint
			}
		}.start(); 
		}
	}
}

```
and it just does not work: 
1. selecting another thread in the Debug view just closes even pinned pop up. 
2. pressing ctr+shift+D just closes pinned pop up

mickaelistria:
Here is 🎥 what I get locally

https://github.com/eclipse-platform/eclipse.platform/assets/469058/64266ae8-8e03-419a-a1e3-6fd96b324b3f

mickaelistria:
I can reproduce the issue you mention about switching to another stackframe *from the same execution* closing the dialog. I will look at it as the comparison of values from different threads makes as much sense as comparing values from different launches.

mickaelistria:
>  I can reproduce the issue you mention about switching to another stackframe from the same execution closing the dialog. 

I was wrong, I cannot reproduce this issue, all seems to work as expected, see 🎥


https://github.com/eclipse-platform/eclipse.platform/assets/469058/32ecac7f-d33b-4feb-ac64-89c9c53840d5

jukzi:
currently i can not reproduce closing by thread selection myself. odd. But reproducibly crt/shift/i crt/shift/d closes a pinned dialog instead of the last opened on the second invocation.

mickaelistria:
> But reproducibly crt/shift/i crt/shift/d closes a pinned dialog instead of the last opened on the second invocation.

I tried it but couldn't reproduce the issue. I guess there is a detail that differs here. Can you please share finer-grain steps to reproduce?

jukzi:
> I tried it but couldn't reproduce the issue. I guess there is a detail that differs here. Can you please share finer-grain steps to reproduce?

1. open a pop up A, pin it
2. open a pop up B, pin it
3. open a pop up C, DO NOT pin it
4. try to close pop up C with the key shortcut it prints
=> Popup A is closed.
5. again: try to close pop up C with the key shortcut it prints
=> Popup B is closed.

mickaelistria:
This new commit should fix the reported issue.

merks:
I think this would make a nice new and noteworthy entry...

PR #1407, opened 2024-06-04
[19+] fix warning about Thread.getID() 

jukzi:
'The method getId() from the type Thread is deprecated since version 19'


akurtakov:
As this is tests code it's fine to bump its BREE to 21 and land this one in. We already struggle with maintaining the PR queue so better to not have such as this one lurking around.

jukzi:
> As this is tests code it's fine to bump its BREE to 21

ok if i just bump all tests?

akurtakov:
I don't mind moving to Java 21 but mass moves tend to make things go unnoticed (e.g. many new warnings at once) .

akurtakov:
Another thing to consider is that moving tests to Java 21 makes it harder to test bundles against Java 17 if the bundle under test is still on Java 17.

PR #1408, opened 2024-06-04
fix warning about Field.isAccessible()

jukzi:
'The method isAccessible() from the type AccessibleObject is deprecated since version 9'

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/FieldRequestor.java
+++ b/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/FieldRequestor.java
@@ -44,12 +44,10 @@ public class FieldRequestor extends Requestor<Field> {
 		Object userObject = getRequestingObject();
 		if (userObject == null)
 			return false;
-		if (!field.isAccessible()) {
-			field.setAccessible(true);
-		}
 		try {
+			field.trySetAccessible();
 			field.set(userObject, value);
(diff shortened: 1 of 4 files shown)


runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/FieldRequestor.java
```diff
@@ -44,9 +44,7 @@ private boolean setField(Field field, Object value) throws InjectionException {
 		Object userObject = getRequestingObject();
 		if (userObject == null)
 			return false;
-		if (!field.isAccessible()) {
-			field.setAccessible(true);
-		}
+		field.trySetAccessible();
```
fedejeanne, line 47: Why don't you just replace `isAccessible()` with `canAccess()`? 

Using `trySetAccessible()` changes the logic: now there will no exception thrown if the security manager wouldn't allow to modify the method's accessibility (it did before).

How about:
* Replacing `isAccessible()` wit `canAccess()`
* Putting the `if (!canAccess())` inside the `try-catch` block too
* Catching also `InaccessibleObjectException` and `SecurityException` (they can be thrown by `setAccessible(boolean)`) so they are also rethrown inside the `InjectionException`. 

?

fedejeanne: Good point: `trySetAccessible(boolean)` also checks the permissions 👍 

Thank you!

PR #1413, opened 2024-06-10
Proposal: Opening/Closing Mechanism for Zip Files

Michael5601:
# Proposal: Opening/Closing Mechanism for Zip Files

## Background
The Eclipse IDE has no built in functionality to open Zip Files and read or manipulate their content. Because of this, other operations like searching inside of Zip Files or comparing two Zip Files were also not possible.

## Goal:
This pull request introduces a mechanism for handling Zip Files within the Eclipse workspace, enhancing the functionality to read and write Zip files. The primary goal is to provide a seamless experience for developers working with zip archives directly within Eclipse.

## Description:
Zip files must be opened manually within the workspace by using the new command "Open Zip File" in the menu when right clicking the zip file. It is also possible to open nested zip files. 

Zip Files are opened by replacing the file in the workspace with a linked folder that reads and writes the Zip File in the file system. By closing the Zip FIle, the linked folder will be deleted and the file can be seen in the workspace again. 

Please note that only ZIP Archives are supported in this current implementation. Other archive types can be added in future improvements. Also linked Zip Files can not be opened with this implementation because the Zip File must be local.

An additional PR for the repository **eclipse.platform.ui** that grants access to the open/close mechanism for zip files over UI can be found in the following:

https://github.com/eclipse-platform/eclipse.platform.ui/pull/1947

Co-Authored-By: CodeLtDave <david.erdoes@vector.com>


laeubi:
Given there is a [Zip Filesystem](https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) for local files in the JDK already that support read/write I wonder if one really needs the "Open Zip FIle" step? Also [the examples](https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/examples/org.eclipse.ui.examples.filesystem/src/org/eclipse/core/internal/filesystem/zip) contain such a ZipFile system for eclipse already.

So I wonder if you maybe can explain why this design decision was made with linking / extraction of the Zip?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+		for (int i = 0; i < entryCount; i++) {
+			names[i] = computeName(entries[i]);
+		}
+		return names;
+	}
+
+	/**
+	 * Computes the simple file name for a given zip entry.
+	 */
+	private String computeName(ZipEntry entry) {
```
Wittmaxi, line 144: Can this be made static?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+	private String getPluginId() {
+		return FrameworkUtil.getBundle(this.getClass()).getSymbolicName();
+	}
+
+	public IPath getPath() {
+		return path;
+	}
+
+	private boolean isNested() {
+		return this.rootStore instanceof ZipFileStore;
```
Wittmaxi, line 338:
```suggestion
		return rootStore instanceof ZipFileStore;
```

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+		} catch (IOException e) {
+			throw new CoreException(Status.error("Could not read file: " + rootStore.toString(), e)); //$NON-NLS-1$
+		}
+	}
+
+	@Override
+	public OutputStream openOutputStream(int options, IProgressMonitor monitor) {
+		// Creating a ByteArrayOutputStream to capture the data written to the
+		// OutputStream
+		ByteArrayOutputStream baos = new ByteArrayOutputStream() {
```
Wittmaxi, line 457:
```suggestion
		return new ByteArrayOutputStream() {
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ZipFileTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+	 */
+	public static void openZipFile(IFile file, IProgressMonitor monitor, boolean backgroundRefresh)
+			throws URISyntaxException, CoreException {
+		SubMonitor subMonitor = SubMonitor.convert(monitor, 20);
+		try (InputStream fis = file.getContents()) {
+			ZipFileUtil.checkFileForZipHeader(fis);
+			// Additional operations can continue here if header is correct
+		} catch (IOException e) {
+			// If the header is incorrect or there's an IO error, handle gracefully
+			throw new CoreException(new Status(IStatus.ERROR, "your.plugin.id", //$NON-NLS-1$
```
Wittmaxi, line 85: Is `"your.plugin.id"` the correct id?

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
mickaelistria: Note that unless there is a project rule emitting some warning here, it's all fine to keep `this.`.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/ZipFileUtil.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Utility class to determine if a file is an archive based on file header information.
+ * This class checks for known file signatures to identify if a given file is a ZIP archive
+ * or a format based on ZIP, such as EPUB, JAR, ODF, and OOXML.
+ *
+ * @since 1.11
+ */
+public class ZipFileUtil {
+
+	private static final Set<Integer> ARCHIVE_FILE_SIGNATURES = new HashSet<>();
```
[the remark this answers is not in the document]
mickaelistria: Maybe even better
```suggestion
	private static final Set<Integer> ARCHIVE_FILE_SIGNATURES = Set.of(0x504B0304, 0x504B0506, 0x504B0708);
```

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+
+	/**
+	 * Creates a new zip file store.
+	 */
+	public ZipFileStore(IFileStore rootStore, IPath path) {
+		this.rootStore = rootStore;
+		this.path = path.makeRelative();
+	}
+
+	private ZipEntry[] childEntries(IProgressMonitor monitor) throws CoreException {
```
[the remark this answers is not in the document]
mickaelistria: Note that APIs from newer Java versions have started to recommend and adopt method names that are not getters. I think it's fine to adopt this more modern style that seems recommended nowadays.

mickaelistria:
I think before diving in the code, @laeubi 's comment at https://github.com/eclipse-platform/eclipse.platform/pull/1413#issuecomment-2157982423 is most important to answer. IMO, if we're to include a Zip filesystem in Platform, it'd better be the one that is already in the example. UI also could be built on top of this filesystem in example. One benefit is that we could make progress with UI without being blocked by Platform.

laeubi:
> One benefit is that we could make progress with UI without being blocked by Platform.

Honestly I think we don't need any new UI but it seems I have missed the discussion about the requirements for this change, but take as an example the Classpath Container, if there is a jar file (what actually is a zip) it simply has a handle so I can expand it and it shows the content and I can double click files and open editors and so on, I would expect something similar for "something" that support open zip files without any need to open/close/extract/... UI because that's the whole point of a file-system abstraction that you can have "virtual" items pointing to something that is not a regular file.

iloveeclipse (changes requested):
> Co-Authored-By: CodeLtDave [david.erdoes@vector.com](mailto:david.erdoes@vector.com)

Here and on other PR: **ALL** contributors must sign ECA. 

If `david.erdoes@vector.com` has not signed ECA, we can't  merge this PR.

laeubi:
[eclipsefdn/eca](https://api.eclipse.org/git/eca/status/gh/eclipse-platform/eclipse.platform/1413) succeeds so I don't see an immediate issue here:
grafik

iloveeclipse:
> [eclipsefdn/eca](https://api.eclipse.org/git/eca/status/gh/eclipse-platform/eclipse.platform/1413) succeeds so I don't see an immediate issue here:

OK. Haven't noticed the check was done for the **other** person as the PR owner. 
Looks weird however to see only **one** author checked. What about @Michael5601 ECA agreement then?

resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
```diff
@@ -35,3 +35,9 @@ Service-Component: OSGI-INF/ResourceChangeListenerRegistrar.xml,
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.core.resources
 Import-Package: org.eclipse.osgi.service.datalocation;version="[1.3.0,2.0.0)"
+Service-Component: OSGI-INF/org.eclipse.core.internal.resources.CheckMissingNaturesListener.xml
```
jukzi, line 38: remove that. strange: i got that automatically added in a PR too

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
```diff
@@ -345,27 +345,48 @@ private boolean internalDeleteFolder(IFolder folder, int flags, IProgressMonitor
...
+				// if the IResource.CLOSE_ZIP_FILE flag is set, the file should not be deleted
+				// in the file system.
+				if ((flags & (IResource.CLOSE_ZIP_FILE)) != 0) {
+					return true;
+				}
+				IFile file = folder.getParent().getFile(IPath.fromOSString(folder.getName()));
+				IFileStore parentStore = localManager.getStore(file);
+				parentStore.delete(EFS.NONE, Policy.subMonitorFor(monitor, Policy.totalWork / 4));
+				return true;
+			} catch (Exception e) {
```
jukzi, line 362: avoid e.printStackTrace();

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ZipFileTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+	 * the children from being loaded.
+	 *
+	 * This method prevents opening linked zip files. zip files must be local to be
+	 * opened. Otherwise a CoreException is thrown.
+	 *
+	 * @param file    The file representing the zip file to open.
+	 * @param monitor monitor indicating the completion progress
+	 *
+	 */
+	public static void openZipFile(IFile file, IProgressMonitor monitor, boolean backgroundRefresh)
```
jukzi, line 77: opening a zip file only takes some milliseconds, not worth a ProgressMonitor.
"backgroundRefresh" is not documented.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/zip/CopyTest.java
```diff
@@ -0,0 +1,211 @@
...
+	public void setup() throws Exception {
+		ZipFileSystemTestSetup.defaultSetup();
+	}
+
+	@AfterEach
+	public void teardown() throws Exception {
+		ZipFileSystemTestSetup.teardown();
+	}
+
+	@ParameterizedTest
```
jukzi, line 44: Avoid "@ParameterizedTest" as long as you do not really need it as it makes trouble building a good test name within the build jobs

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/ZipFileUtil.java
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * Checks if the provided {@link InputStream} represents a ZIP archive
+	 * by reading its first four bytes and comparing them against known ZIP file signatures.
+	 * This method throws {@link IOException} if the file signature does not match any known ZIP archive signatures.
+	 *
+	 * @param fis The {@link InputStream} of the file to check.
+	 * @throws IOException If the file signature does not match known ZIP archive signatures
+	 *                     or an I/O error occurs during reading from the stream.
+	 */
+	public static void checkFileForZipHeader(InputStream fis) throws IOException {
```
jukzi, line 100: instead of manual ARCHIVE_FILE_SIGNATURES you can just try to open the zip file with std java, which esentially does the same.

mickaelistria:
Can someone (a project lead?) gives some opinions about the fact that we already have code for it in the codebase? IMO, this particular PR is not necessarily and we should only discuss whether we want to include the existing zip filesystem implementation in the main feature or not.

laeubi:
> Looks weird however to see only one author checked. What about @Michael5601 ECA agreement then?

I don't know the details @waynebeaton maybe can tell or we need a servicedesk ticket to clarify what / when is checked...

CodeLtDave:
> Given there is a [Zip Filesystem](https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) for local files in the JDK already that support read/write I wonder if one really needs the "Open Zip FIle" step? Also [the examples](https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/examples/org.eclipse.ui.examples.filesystem/src/org/eclipse/core/internal/filesystem/zip) contain such a ZipFile system for eclipse already.
> 
> So I wonder if you maybe can explain why this design decision was made with linking / extraction of the Zip?

I appreciate everyone's interest in our project. A frequent topic of discussion has been the existing [zip example](https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/examples/org.eclipse.ui.examples.filesystem/src/org/eclipse/core/internal/filesystem/zip) provided in Eclipse. I want to clarify that our pull request builds upon this example, adding more functionality and integrating it into the main feature.

As @mickaelistria mentioned, the key question is whether this functionality should be part of the main feature, and I firmly believe it should be.

Regarding the [nio Zip Filesystem](https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) available in the JDK, we utilize its capabilities extensively. Our pull request essentially incorporates the NIO Zip Filesystem into the existing example, resulting in significant performance improvements. Most of the logic can be found in the [ZipFileStore.java](https://github.com/CodeLtDave/eclipse.platform/blob/ZipFileImplementation/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java) where the nio Zip Filesystem is opened.

mickaelistria:
> I want to clarify that our pull request builds upon this example, adding more functionality and integrating it into the main feature.

OK, great. Thank you for clarifying!
Do you think it would be possible for you to create a dedicated PR for those improvements and additional tests, but keeping the code in the "example" bundle for the moment? Such a PR could be merged in a breeze and be a valuable iteration while we still discuss the best place to host the filesystem in the end.

laeubi:
@CodeLtDave can you explain why explicitly Open/Close is required? Can't we reuse the existing functionality (and possibly extend if something is missing here)?

## What we already have

To make it more clear I just choose this random project with some java files and it looks like this
grafik

Now I expand the handle and it looks like this:
grafik

So I have a `File (.java)` and it shows we there is maybe more to explore and after that I see some more items that are "inside" the file... 

As a user I would expect same/similar if it was not a java but a Zip-File for browsing the content similar to what I have for example with Plugin-Dependencies:

grafik

## What would be new

Now if I open one file in an editor, I would like to be able to save this (and maybe getting asked if I really want to update the zip archive).

Next when I choose the File Search Dialog I would maybe expect some new option to include searching inside archives as well:

grafik


So to summarize I don't see we really need any new Actions for that (load/unload) but should use whats already is offered here.

CodeLtDave:
> So I wonder if you maybe can explain why this design decision was made with linking / extraction of the Zip?

Regarding the functionality of the ZipFilesystem, as implemented in the example, we don't extract any files for opening or editing. Opening the Zip is akin to [walking down the File Tree](https://github.com/eclipse-platform/eclipse.platform/pull/1413/files#diff-1fb44518fdb735c3e099466ff46c543dacab29c6aa6846150ce8f6804bb05d3c:~:text=Files.walkFileTree%28zipRoot%2C%20EnumSet.noneOf%28FileVisitOption.class%29%2C%201%2C%20new%20SimpleFileVisitor%3CPath%3E%28%29%20%7B) and collecting all necessary file information.

When a file is modified, it's done by opening the Zip with NIO and using the respective operations.

Based on our testing, this approach is much more lightweight and performant than always extracting the entire Zip and repackaging it. For small changes, this saves significant time and reduces overhead. We didn't investigate the internal workings of NIO in detail, so it's possible that some extraction and repackaging might be happening in the background. However, since it's much faster than manual extraction and repackaging, we believe this is not the case.

Regarding the approach with a linked resource, this seems to be the best method for linking any virtual FileSystem. Since it was already implemented this way in the example project, we believe it is appropriate and fitting for our use case.

jukzi:
I wish ZIP is support from core. Currently i have to go windows file explorer to open zip files. awkward. Searching in zip files would be even better, but that can be a follow up step. At best JAR files are handled as ZIP files as well, because technically they are. The file ending is rather irrelevant.

laeubi:
I think then this description is confusing:

> Zip Files are opened by replacing the file in the workspace with a linked folder that reads and writes the Zip File in the file system. By closing the Zip FIle, the linked folder will be deleted and the file can be seen in the workspace again.

Why is it a *folder* then? (see example with java file that is not a folder at all)? As far as I understand the example it is done that way so you can link in _any external zip file_ but here we want to show additional content for a file *inside the workbench* so this indirection should not be required (see java files and classpath jar files).

CodeLtDave:
> I wish ZIP is support from core. Currently i have to go windows file explorer to open zip files. awkward. Searching in zip files would be even better, but that can be a follow up step. At best JAR files are handled as ZIP files as well, because technically they are. The file ending is rather irrelevant.

@jukzi 

The use case for searching in ZIP files is honestly the motivation for this project. Many colleagues, including myself, have often expressed frustration over this limitation. As you mentioned, JAR files can only be differentiated from ZIP files by their structure, specifically the presence of a META-INF/MANIFEST.MF file and possibly a Main-Class attribute.

Given that this is the platform repository, we have manually disabled support for JAR files. We plan to add JAR support, along with all the necessary functionality and tests, in the JDT (Java Development Tools) as suggested by @HeikoKlare.

CodeLtDave:
> I think then this description is confusing:
> 
> > Zip Files are opened by replacing the file in the workspace with a linked folder that reads and writes the Zip File in the file system. By closing the Zip FIle, the linked folder will be deleted and the file can be seen in the workspace again.

I think you are right, that sounds confusing. Essentially, the zip file is hidden as long as there is a linked folder (or opened Zip) accessing the zip file. When it's closed, the link is deleted and the zip file is visible again.

> 
> Why is it a _folder_ then? (see example with java file that is not a folder at all)? As far as I understand the example it is done that way so you can link in _any external zip file_ but here we want to show additional content for a file _inside the workbench_ so this indirection should not be required (see java files and classpath jar files).

I think I misunderstand your question. Every Zip/JAR is a compressed folder and is therefore displayed as such. We didn't change anything about that part from the example.

laeubi:
> JAR files can only be differentiated from ZIP files by their structure, specifically the presence of a META-INF/MANIFEST.MF file and possibly a Main-Class attribute

Jar files are Zip files, a manifest is not mandatory nor any special headers:
https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html#introduction

>> A JAR file is essentially a zip file that contains **an optional META-INF directory**

So I don't see why one should / would disable anything for them?

> We didn't change anything about that part from the example.

The example is for linking a filesystem as a *new item* into the workbench, here we want to show _additional_ childs for an _existing file in the workbench_. That's an important part I think and would avoid having to "shadow" an existing file, maybe such special resource must implement both `IFolder` and `IFile` as otherwise other parts of the system can no longer "see" the zip/jar as a file what could have some implications.

CodeLtDave:
> So to summarize I don't see we really need any new Actions for that (load/unload) but should use whats already is offered here. I haven't checked in detail how these functionality is implemented so maybe there are good reasons not to do it like this... but it would better match user expectation I think.

We share your opinion and are also not fully satisfied with the current UI usability for the ZipFileSystem. We have developed a version with a double-click handler. For now, we are relying on the implementation from the example. Implementing the existing UI features has proven to be very challenging, as the UI was not easily extensible for our needs. Since I was not involved in the UI development, I will try to put more focus on this aspect.

Given these challenges, this PR should perhaps be considered a draft. Our main goal is to gather feedback, such as yours, and discuss the possibility of moving the ZipFs into the main project.

laeubi:
I think moving the bare implementation (== File system provider) into core should be something that is quite easy, as it would just offer the new FS type but not changes anything.

mickaelistria:
> I think moving the bare implementation (== File system provider) into core should be something that is quite easy, as it would just offer the new FS type but not changes anything.

FWIW, I would also support this move of the plain filesystem at the moment. Everything consensual should be ideally provided separately and merged so we can then focus on the friction points.
I'm more reluctant about the particular additions to IResource (which I believe look like a good starting point but too intrusive for a good final solution).

CodeLtDave:
> > JAR files can only be differentiated from ZIP files by their structure, specifically the presence of a META-INF/MANIFEST.MF file and possibly a Main-Class attribute
> 
> Jar files are Zip files, a manifest is not mandatory nor any special headers: https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html#introduction
> 
> > > A JAR file is essentially a zip file that contains **an optional META-INF directory**
> 
> So I don't see why one should / would disable anything for them?

Since Eclipse is primarily used for Java projects, I understand your perspective. However, I still believe it is not suitable to include this in the platform repository. As mentioned, there are additional concerns for JAR files that require thorough testing. All this will be addressed in a planned PR for the JDT (Java Development Tools).

If you still believe that everyone using an Eclipse product should have the logic for JARs included, and if this is the consensus, we can reconsider. However, it is important to consider users who use Eclipse for projects other than Java. For example, similar logic could apply to other file types such as WAR, EAR, APK, NPK, XPI, CRX, or VSIX.
> 
> > We didn't change anything about that part from the example.
> 
> The example is for linking a filesystem as a _new item_ into the workbench, here we want to show _additional_ childs for an _existing file in the workbench_. That's an important part I think and would avoid having to "shadow" an existing file, maybe such special resource must implement both `IFolder` and `IFile` as otherwise other parts of the system can no longer "see" the zip/jar as a file what could have some implications.

Without fundamental changes to many parts of the platform, achieving this would be challenging. I spent two weeks exploring this approach and welcome any solutions, but from what I have seen, it cannot be done without very intrusive changes. The hardest part is determining an entry point. I couldn't find a place where I can dictate that this file should be treated with the ZipFileSystem as long as it has a URI for the local filesystem. There are many differences regarding the functional conditions of a LocalFileSystem and a virtual ZipFileSystem.

My main concern is that our PR adds new functionality for the user. If we were to treat the ZIP as a combination of file and folder, the behavior of a ZIP file would definitely change. I believe it must be a conscious decision to change the behavior from a file that cannot be opened or read to a folder.

jukzi:
I have not followed and understand everything written, but please keep in mind that a zip **file** needs to stay a IFile so that it can be for exampled read bytewise to copy it into another IFolder. On the other hand a zip archive **mounted** into the workspace has to be a IFolder as it can and should contain IFiles as children. As a special case there can be a Zip in a Zip recursively and at best a search would open all zips recursively.
Best wishes!
For me the best usecase of a mounted zip would be to compile all .class files directly into a zip/jar, so that compilation does not need to open so many small OS files. - Which would be a big performance gain on Windows. At best the java launcher would understand the archive is an archive and can be handled by java as classpath .jar instead of a directory. In a perfect world my whole workspace could be a single zip on the OS file system, but i doubt tools like "java" would understand a jar inside a zip :-)

iloveeclipse:
> Given these challenges, this PR should perhaps be considered a draft.

Yes, please convert to a draft.

What I would recommend also is to provide a detailed explanation 
- **what exactly** is the main goal of the change
- how the proposed solution works technically (highlighting most significant design decisions)
- how the proposed solution works from user point of view (highlighting most significant UI behavior changes)
- which other possible solutions were evaluated
- pros/cons of current/evaluated solutions from technical and user point of view.

Based on that we could discuss where it is all going.

Note from our application point of view: we have several custom zip based file formats (both encrypted/not encrypted) for which we have dedicated custom editors. So from our application point of view even if our files are zip files inside, no other editor should be opened by double click on them, no other editor should be used to modify them and no extra indexing/searching should be done for the content, especially because some of the files can be very huge (up to 4 GB compressed) and may contain many thousand of (plain text) child elements. So whatever is developed here should allow application deny any of the mentioned operations on specific file types.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/ZipFileUtil.java
```diff
@@ -0,0 +1,114 @@
...
+		IFileStore store;
+		try {
+			store = EFS.getStore(locationURI);
+		} catch (CoreException e) {
+			return false;
+		}
+		return isInsideOpenZipFile(store);
+	}
+
+	//TODO Implement this method
```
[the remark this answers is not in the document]
Michael5601: Thank you for the hint. I deleted the line in b388c72169f4228e5ce3598c740830f991a6abcc.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+			name = name.substring(0, name.length() - 1);
+		}
+		// creates last segment after last /
+		int lastIndex = name.lastIndexOf('/');
+
+		if (lastIndex != -1) {
+			return name.substring(lastIndex + 1);
+		}
+
+		return name; // Falls kein '/' gefunden wurde
```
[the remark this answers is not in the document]
Michael5601: thanks for the hint. This comment should obviously be english.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	private IFileInfo convertToIFileInfo(Path zipEntryPath, BasicFileAttributes attrs) {
```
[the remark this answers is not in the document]
Michael5601: No fetchInfo is also implemented but has a different purpose. convertToIFileInfo is used inside of fetchInfo.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+		return info;
+	}
+
+	/**
+	 * Creates a file info object corresponding to a given zip entry
+	 *
+	 * @param entry the zip entry
+	 * @return The file info for a zip entry
+	 */
+	private IFileInfo convertZipEntryToFileInfo(ZipEntry entry) {
```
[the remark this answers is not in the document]
Michael5601: I can't see the benefits of doing so. Can you elaborate? In my understanding the method is private so there is no point in  extracting it.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
Michael5601: yes this is cleaner. I added this change. Thank you!

resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
Michael5601: This is really strange. I did not include this in a commit. I deleted the lines in 97dadc19c97b4e7dcdc2ac5ac668611fc60886d8 but I needed to do it with the windows text editor because some configuration always added the lines when I safed the file.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
Michael5601: I improved the error handling. Thanks for the hint!

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ZipFileTransformer.java
Michael5601: I added documentation and removed the monitor in commit 91f9a51af1e29228cd3014a69be650a69518fe23. We had performance issues and used a ProgressDialog in UI and the ProgressMonitor here because of the performance issues. But now the issues are solved so I removed these changes.

Michael5601:
> > So to summarize I don't see we really need any new Actions for that (load/unload) but should use whats already is offered here. I haven't checked in detail how these functionality is implemented so maybe there are good reasons not to do it like this... but it would better match user expectation I think.
> 
> @laeubi
> 
> We share your opinion and are also not fully satisfied with the current UI usability for the ZipFileSystem. We have developed a version with a double-click handler. For now, we are relying on the implementation from the example. Implementing the existing UI features has proven to be very challenging, as the UI was not easily extensible for our needs. Since I was not involved in the UI development, I will try to put more focus on this aspect.
> 
> Given these challenges, this PR should perhaps be considered a draft. Our main goal is to gather feedback, such as yours, and discuss the possibility of moving the ZipFs into the main project.

Thank you for your valuable input. We appreciate every suggestion that helps us improve this PR. I'd like to address some additional questions you raised that @CodeLtDave did not cover.

Initially, we considered using the existing implementation for "opening" files, as you suggested. However, we encountered a challenge: this functionality is JDT-based rather than Platform-based (Please correct me if I am wrong). As noted in the example project referenced several times in our discussion, it also does not utilize this functionality. Instead, it hides the zip files in the workspace behind linked folders, which can be expanded to view the zip file contents. These folders only appear after manually pressing a button in the UI. 

In the long term I think we need to implement the JDT based approach and also the Platform based approach because these are different domains.

The biggest part of our research in the beginning of our project was to find a satisfying way to include files in the workspace that can be opened like folders. Every attempt failed except for the one we provided in this PR which is the same as in the example project. This implementation lead to more problems regarding linked resources but we managed to fix many of them. 

In the [PR](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1947) for platform.ui I provided some [screenshots](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1947#issuecomment-2160596872) for the workflow so it can be understood better.  

Lastly, I want to mention that the file search can immediately search inside zip files if they are manually opened before initiating the search. This does not require additional functionality since all the files are located in the workspace. The filter button you mentioned will be added in a future improvement, but there are still some considerations regarding the UI and future implementations. Our team has already discussed this and decided not to include a filter in the first version.

I hope I could clarify some questions and looking forward to additional input.

Michael5601:
> > Given these challenges, this PR should perhaps be considered a draft.
> 
> Yes, please convert to a draft.
> 
> What I would recommend also is to provide a detailed explanation
> 
> * **what exactly** is the main goal of the change
> * how the proposed solution works technically (highlighting most significant design decisions)
> * how the proposed solution works from user point of view (highlighting most significant UI behavior changes)
> * which other possible solutions were evaluated
> * pros/cons of current/evaluated solutions from technical and user point of view.
> 
> Based on that we could discuss where it is all going.
> 
> Note from our application point of view: we have several custom zip based file formats (both encrypted/not encrypted) for which we have dedicated custom editors. So from our application point of view even if our files are zip files inside, no other editor should be opened by double click on them, no other editor should be used to modify them and no extra indexing/searching should be done for the content, especially because some of the files can be very huge (up to 4 GB compressed) and may contain many thousand of (plain text) child elements. So whatever is developed here should allow application deny any of the mentioned operations on specific file types.

We will provide a detailed explanation addressing the open discussion points you mentioned. I hope this will clarify many of the issues raised in the discussion.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
CodeLtDave: made static in e528b7d4ea6f9572631fcd3b24d239f53323a342

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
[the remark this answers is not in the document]
CodeLtDave: made static in e528b7d4ea6f9572631fcd3b24d239f53323a342

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+		Map<String, String> env = new HashMap<>();
+		env.put("create", "false"); //$NON-NLS-1$ //$NON-NLS-2$
+
+		// Assuming the directory to create is represented by 'this.path'
+		try (FileSystem zipFs = FileSystems.newFileSystem(zipUri, env)) {
+			Path dirInZipPath = zipFs.getPath(this.path.toString());
+			if (Files.notExists(dirInZipPath)) {
+				Files.createDirectories(dirInZipPath);
+
+				// To ensure the directory is actually added to the ZIP, we
```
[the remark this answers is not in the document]
CodeLtDave: good point this is covered in org.eclipse.core.tests.filesystem.zip.CreateTest.testCreateFolderInsideOfZipFile()
25fa7bb55dd82b677254c6e5c50189ce6e2ccf1d

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/filesystem/ZipFileUtil.java
CodeLtDave: I agree that's more fitting here
acba1d89ec4414a199eda3b29e02b0f6294294ae

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/zip/ZipFileStore.java
```diff
@@ -0,0 +1,559 @@
...
+		this.path = path.makeRelative();
+	}
+
+	private ZipEntry[] childEntries(IProgressMonitor monitor) throws CoreException {
+		List<ZipEntry> entryList = new ArrayList<>();
+		String myName = path.toString();
+
+		try (FileSystem zipFs = openZipFileSystem()) {
+			Path zipRoot = zipFs.getPath(myName);
+			Files.walkFileTree(zipRoot, EnumSet.noneOf(FileVisitOption.class), 1, new SimpleFileVisitor<Path>() {
```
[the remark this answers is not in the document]
CodeLtDave: I agree that cleans up some space 
93b70fd15a4285478f9a58fc40832d8876544df1

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/ZipFileTransformer.java
Michael5601: changed in fd7727f12d9fd0c1a90fab8e6bd27c246a1308bb

waynebeaton:
> I don't know the details @waynebeaton maybe can tell or we need a servicedesk ticket to clarify what / when is checked...

Both contributors appear to have signed the ECA, so you should be good to go.

The ECA checker only checks the actual author of the commit. It does not check `Also-Authored-By` fields and similar. 

I'll use this opportunity to remind everybody that the ECA checker is just a tool and that its purpose is to _help_ committers understand when the contributors are covered by ECAs. No tool is perfect and we depend on committers to be our first line of defence in maintaining our IP regime. If you feel that something is out of place or missing, please engage with helpdesk (or EMO) so that we can help.

Finally... what the heck is the gender of a ZIP file?

jukzi:
I just became aware that there is an existing https://github.com/bndtools/bnd/blob/master/bndtools.jareditor/src/bndtools/jareditor/internal/JarFileSystem.java which is also used in the eclipse IDE.

Michael5601:
> > I don't know the details @waynebeaton maybe can tell or we need a servicedesk ticket to clarify what / when is checked...
> 
> Both contributors appear to have signed the ECA, so you should be good to go.
> 
> The ECA checker only checks the actual author of the commit. It does not check `Also-Authored-By` fields and similar.
> 
> I'll use this opportunity to remind everybody that the ECA checker is just a tool and that its purpose is to _help_ committers understand when the contributors are covered by ECAs. No tool is perfect and we depend on committers to be our first line of defence in maintaining our IP regime. If you feel that something is out of place or missing, please engage with helpdesk (or EMO) so that we can help.
> 
> Finally... what the heck is the gender of a ZIP file?

Thank you for the clarification. I used the term "gender" for zip files in commit f092e33aed648bd2b3329af3a4c0e39de1953d4a because of an issue with the synchronizeGender() method in [RefreshLocalVisitor](resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/RefreshLocalVisitor.java
). This method checks if a folder in the workspace is also a folder in the file system. For zip files, we use linked folders in the workspace, but the corresponding resource in the file system remains a file, causing synchronization problems in synchronizeGender(). Therefore, in this context, "gender" refers to whether the resource is a folder or a file.

The provided workaround is temporary, and we are working on a better solution.

merks:
> I just became aware that there is an existing https://github.com/bndtools/bnd/blob/master/bndtools.jareditor/src/bndtools/jareditor/internal/JarFileSystem.java which is also used in the eclipse IDE.

PDE makes that library available:

image

Issue #1411, opened 2024-06-07
NPE in Cache$Entry.unchain


jukzi:
probably ancient code that is somehow not threadsafe:
```
java.lang.NullPointerException: Cannot assign field "next" because "this.previous" is null
	at org.eclipse.core.internal.utils.Cache$Entry.unchain(Cache.java:122)
	at org.eclipse.core.internal.utils.Cache.getEntry(Cache.java:174)
	at org.eclipse.core.internal.utils.Cache.getEntry(Cache.java:165)
	at org.eclipse.core.internal.resources.ContentDescriptionManager.getDescriptionFor(ContentDescriptionManager.java:393)
	at org.eclipse.core.internal.resources.File.internalGetCharset(File.java:263)
	at org.eclipse.core.internal.resources.File.getCharset(File.java:220)
	at org.eclipse.core.internal.resources.File.getCharset(File.java:208)
	at org.eclipse.jdt.internal.core.CompilationUnit.getContents(CompilationUnit.java:673)
	at org.eclipse.jdt.internal.compiler.parser.SourceTypeConverter.getSource(SourceTypeConverter.java:719)
	at org.eclipse.jdt.internal.compiler.parser.SourceTypeConverter.convertAnnotations(SourceTypeConverter.java:689)
	at org.eclipse.jdt.internal.compiler.parser.SourceTypeConverter.convert(SourceTypeConverter.java:552)
	at org.eclipse.jdt.internal.compiler.parser.SourceTypeConverter.convert(SourceTypeConverter.java:193)
	at org.eclipse.jdt.internal.compiler.parser.SourceTypeConverter.buildCompilationUnit(SourceTypeConverter.java:99)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.accept(CompilationUnitProblemFinder.java:123)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.askForType(LookupEnvironment.java:381)
	at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.getTypeOrPackage(PackageBinding.java:276)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.getTypeOrPackage(Scope.java:3571)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.getType(Scope.java:3274)
	at org.eclipse.jdt.internal.compiler.ast.SingleTypeReference.getTypeBinding(SingleTypeReference.java:60)
[...]

PR #1414, opened 2024-06-10
Rework org.eclipse.core.internal.utils.Cache with standard java

jukzi:
* Threadsafe implementation
* Instead of predefined cache size use SoftReferences.

Should avoid random NullPointerException in concurrent use: https://github.com/eclipse-platform/eclipse.platform/issues/1411


mickaelistria (approved):
The code looks better. I think M1 is the right time to merge such refactorings so we have time to spot regressions if any.
So IMO, just merge now if you're ready.

jukzi:
To make sure the memory does not explode by this change i took heapdump on platform workspace during clean build. It looks really harmless (80kb):
image

PR #1415, opened 2024-06-10
Remove unused native code and assume unicode is always supported in Win32Natives refresh provider

HannesWell:
Unicode support is assumed to be present if the Windows-OS version is greater or equal five, which corresponds to Windows 2000 or later: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfow

For example the JDK binaries provided by Temurin and Oracle require at least Windows-8 for a JDK-17:
- https://www.oracle.com/java/technologies/javase/products-doc-jdk17certconfig.html
- https://adoptium.net/de/supported-platforms/

and given that even Windows-10 will be EOL in about a year I think we can confidently remove that support for pre Windows 2000 versions.

This was originally part of https://github.com/eclipse-platform/eclipse.platform/pull/1394.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/natives/make.bat
+++ b/resources/bundles/org.eclipse.core.resources/natives/make.bat
@@ -11,22 +11,20 @@
 @rem Contributors:
 @rem     IBM Corporation - initial API and implementation
 @rem ***************************************************************************
+@echo off
 REM build JNI header file
-cd ..\bin
-"C:\Program Files\Java\jdk1.8.0_65\bin\javah.exe" org.eclipse.core.internal.resources.refresh.win32.Win32Natives
-move org_eclipse_core_internal_resources_refresh_win32_Win32Natives.h ..\natives\ref2.h
+cd %~dp0\..\src
+
+"%JAVA_HOME%\bin\javac" -h . org\eclipse\core\internal\resources\refresh\win32\Win32Natives.java
+del org\eclipse\core\internal\resources\refresh\win32\Win32Natives.class
+move org_eclipse_core_internal_resources_refresh_win32_Win32Natives.h ..\natives\ref.h
 
 REM compile and link
+if "%MSVC_HOME%"=="" set MSVC_HOME=C:\Program Files\Microsoft Visual Studio\2022\Community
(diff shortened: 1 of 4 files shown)


jukzi:
This commit would rely for example on `FindFirstChangeNotificationW `which is documented to require Windows XP or Windows Server 2003

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstchangenotificationw

HannesWell:
> This commit would rely for example on `FindFirstChangeNotificationW `which is documented to require Windows XP or Windows Server 2003
> 
> https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstchangenotificationw

That's correct and the same applies for the removed code, for example `FindFirstChangeNotificationA()`:
 https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstchangenotificationa

Both, the A suffixed and W suffixed methods have in introduced when Unicode support was added in Windows XP:
https://en.wikipedia.org/wiki/Unicode_in_Microsoft_Windows#In_various_Windows_families

Besides that there is only a 64-bit fragment `org.eclipse.core.resources.win32.win32.x86_64` and no 32-bit fragment for `org.eclipse.core.resources`. When looking up the supported architecture I can see support respectively the requirement of 64-bit CPUs only since XP respectively Server 2003:
https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions#Personal_computer_versions

With all that `org.eclipse.core.resources.win32.win32.x86_64` cannot be used prior Windows XP/Server 2003 anyways and thus it isn't a changed minimally supported version.

jukzi:
If the "A" and "W" variants where introduced at the same time i don't get if that "A" code was ever used or needed.
Anyway i am OK with removal of "A" variant.

HannesWell:
> If the "A" and "W" variants where introduced at the same time i don't get if that "A" code was ever used or needed. Anyway i am OK with removal of "A" variant.

I'm not sure either, so I can only guess.

But then lets not further delay this and submit it.

PR #1416, opened 2024-06-10
Remove unused (native) code in Windows LocalFileNatives

HannesWell:
Unicode support was assumed to be present if the Windows-OS version is greater or equal five, which corresponds to Windows 2000 or later:
https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfow

For example the JDK binaries provided by Temurin and Oracle require at least Windows-8 for a JDK-17:
- https://www.oracle.com/java/technologies/javase/products-doc-jdk17certconfig.html
- https://adoptium.net/de/supported-platforms/

and given that even Windows-10 will be EOL in about a year I think we can confidently remove that support for pre Windows 2000 versions.
Similar to https://github.com/eclipse-platform/eclipse.platform/pull/1415


jukzi:
As far as i understand this PR removes support for older (non-unicode) Windows versions and eclipse IDE would then require at least Windows XP or  Windows Server 2003. That is OK for me (but i do not see a need to drop that support), but the PR and commit message should clearly state that. You should ask on mailing list if dropping that support is ok, as there are still recent bug reports on windows <=8 like https://github.com/eclipse-platform/eclipse.platform.swt/issues/1147

HannesWell:
> As far as i understand this PR removes support for older (non-unicode) Windows versions and eclipse IDE would then require at least Windows XP or Windows Server 2003. That is OK for me (but i do not see a need to drop that support), but the PR and commit message should clearly state that. 

This is similar like in https://github.com/eclipse-platform/eclipse.platform/pull/1415#issuecomment-2161611644, although the removed code calls `FindFirstFile()` instead of `FindFirstFileA()`. This is currently defined as the following type-def:
```
#ifdef UNICODE
#define FindFirstFile  FindFirstFileW
#else
#define FindFirstFile  FindFirstFileA
#endif // !UNICODE
```
So compiling it now would lead to either `FindFirstFileA` or `FindFirstFileW` be linked into the dll. And decompiling the currently shipped `localfile_1_0_0.dll` (last recompiled almost five years ago) shows that in fact `FindFirstFileA()` is linked.
See for example the Hey-Rays view of: https://dogbolt.org/?id=cf6fa0c6-416e-430a-ae5b-82d776c7a803#Hex-Rays=578

So again the dll already requires methods only available on Windows XP or later and is only available in 64-bit, which is also not supported before XP.
Therefore this does actually not change the required minimal version.

> You should ask on mailing list if dropping that support is ok, as there are still recent bug reports on windows <=8 like [eclipse-platform/eclipse.platform.swt#1147](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1147)

We are not talking about Win-7 in this case, we are not even talking about its predecessors Vista, XP or 2000, we are talking about versions like ME or Win98/95. Even if 32-bit artifacts where available that are linked against the 'real' `FindFirstFile` method, if possible at all it would probably be an academic exercise to make a JDK-17 and a recent Eclipse run on such a computer. I don't think there were a real user-base for that. But we are in the lucky position that this effectively does not change the requirements so we don't have to discuss that. :)

I have updated the commit message to describe all that.

jukzi:
Thanks for the explanation, i am fine with removal of the non-unicode code.

HannesWell:
> Thanks for the explanation, i am fine with removal of the non-unicode code.

You're welcome, then lets have this one now as well.

PR #1418, opened 2024-06-11
LaunchConfigurationFilteredTree: filter with '*' in front

jukzi:
implementation taken from
org.eclipse.pde.internal.ui.shared.FilteredCheckboxTree.getFilterString()

image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationFilteredTree.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationFilteredTree.java
@@ -67,6 +67,7 @@ public final class LaunchConfigurationFilteredTree extends FilteredTree {
 		fFilters = filters;
 		fPatternFilter = filter;
 		fTreeStyle = treeStyle;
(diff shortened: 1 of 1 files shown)


mickaelistria (changes requested):
Instead of repeating this logic here, the PatternFiltter received at construction should be configured with `patternFilter.setIncludeLeadingWildcard(true)`.

jukzi:
> patternFilter

ok, that works too

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationFilteredTree.java
```diff
@@ -67,6 +67,7 @@ public LaunchConfigurationFilteredTree(Composite parent, int treeStyle, PatternF
 		fFilters = filters;
 		fPatternFilter = filter;
 		fTreeStyle = treeStyle;
+		filter.setIncludeLeadingWildcard(true);
```
mickaelistria, line 70: I'm torn where is the best place to set it: is it here or in the same class as where PatternFilter is constructed? Anyway, it's really nitpicking and probably not profitable to debate this for a 1-liner that would be easy to change later if necessary.

mickaelistria (approved):

PR #1420, opened 2024-06-12
Provide API for IFile create(byte[], ...) and createOrReplace()

jukzi:
A typical used pattern to replace a File content is
```
	if (file.exists()) {
			file.setContents(ByteArrayInputStream, ...);
		} else {
			file.create(ByteArrayInputStream, ...);
		}
	}
```
This can be replaced by convenience createOrReplace. For performance reasons it takes the content as byte array like
java.nio.file.Files.write(Path, byte[], OpenOption...).

An example consumer is the performance hotspot in JDT to write .class files during build:

org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.writeClassFileContents(ClassFile,
IFile, String, boolean, SourceFile)

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
@@ -358,6 +358,50 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	void create(InputStream source, int updateFlags, IProgressMonitor monitor) throws CoreException;
 
 	/**
+	 * Creates the file with the given content as byte array. Same as calling
+	 * {@link #create(InputStream, int, IProgressMonitor)} with flags depending on
+	 * the boolean parameters and a {@code new ByteArrayInputStream(content)} as
+	 * {@code InputStream}. This method is preferably over the streaming API when
+	 * the content is available as byte array anyway.
+	 *
+	 * @param content the content as byte array
+	 * @param force   a flag controlling how to deal with resources that are not in
+	 *                sync with the local file system
+	 * @param derived Specifying this flag is equivalent to atomically calling
+	 *                {@link IResource#setDerived(boolean)} immediately after
+	 *                creating the resource
+	 * @param monitor a progress monitor, or <code>null</code> if progress reporting
+	 *                is not desired
+	 * @throws CoreException if this method fails or is canceled.
+	 * @since 3.21
+	 */
+	public default void create(byte[] content, boolean force, boolean derived, IProgressMonitor monitor)
+			throws CoreException {
+		int createFlags = (force ? IResource.FORCE : IResource.NONE) | (derived ? IResource.DERIVED : IResource.NONE);
+		create(content, createFlags, monitor);
+	}
+
+	/**
+	 * Creates the file with the given content as byte array. Same as calling
+	 * {@link #create(InputStream, int, IProgressMonitor)} and a
+	 * {@code new ByteArrayInputStream(content)} as {@code InputStream}. This method
+	 * is preferably over the streaming API when the content is available as byte
+	 * array anyway.
+	 *
+	 * @param content     the content as byte array
+	 * @param createFlags bit-wise or of flag constants ({@link IResource#FORCE},
+	 *                    {@link IResource#DERIVED}, and
+	 *                    {@link IResource#TEAM_PRIVATE})
+	 * @param monitor     a progress monitor, or <code>null</code> if progress
+	 *                    reporting is not desired
+	 * @throws CoreException if this method fails or is canceled.
+	 * @since 3.21
+	 */
+	public default void create(byte[] content, int createFlags, IProgressMonitor monitor) throws CoreException {
+		create(new ByteArrayInputStream(content), createFlags, monitor);
+	}
+
+	/**
 	 * Creates a new file resource as a member of this handle's parent resource.
 	 * The file's contents will be located in the file specified by the given
 	 * file system path.  The given path must be either an absolute file system
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.net.URI;
+import java.util.Objects;
 import org.eclipse.core.internal.utils.FileUtil;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
@@ -303,6 +304,13 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 * with a value of <code>true</code> immediately after creating the resource.
 	 * </p>
 	 * <p>
+	 * The {@link IResource#KEEP_HISTORY} update flag indicates that the history of this file is kept if any
+	 * </p>
+	 * <p>
+	 * The {@link IResource#REPLACE} update flag indicates that this resource is overridden if already existing.
+	 *
+	 * </p>
+	 * <p>
 	 * The {@link IResource#TEAM_PRIVATE} update flag indicates that this resource
 	 * should immediately be set as a team private resource.  Specifying this flag
 	 * is equivalent to atomically calling {@link IResource#setTeamPrivateMember(boolean)}
@@ -330,7 +338,9 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 * @param source an input stream containing the initial contents of the file,
 	 *    or <code>null</code> if the file should be marked as not local
 	 * @param updateFlags bit-wise or of update flag constants
-	 *   ({@link IResource#FORCE}, {@link IResource#DERIVED}, and {@link IResource#TEAM_PRIVATE})
+	 *   ({@link IResource#FORCE}, {@link IResource#DERIVED},
+	 *    {@link IResource#KEEP_HISTORY}, {@link IResource#REPLACE},
+	 *    and {@link IResource#TEAM_PRIVATE})
 	 * @param monitor a progress monitor, or <code>null</code> if progress
 	 *    reporting is not desired
 	 * @exception CoreException if this method fails. Reasons include:
@@ -390,7 +400,9 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 *
 	 * @param content     the content as byte array
 	 * @param createFlags bit-wise or of flag constants ({@link IResource#FORCE},
-	 *                    {@link IResource#DERIVED}, and
+	 *                    {@link IResource#DERIVED},
+	 *                    {@link IResource#KEEP_HISTORY},
+	 *                    {@link IResource#REPLACE}, and
 	 *                    {@link IResource#TEAM_PRIVATE})
 	 * @param monitor     a progress monitor, or <code>null</code> if progress
 	 *                    reporting is not desired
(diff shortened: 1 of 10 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -357,6 +358,86 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
...
+	 *                    existing file
+	 * @param keepHistory a flag indicating whether or not store the current
+	 *                    contents in the local history if the file did already
+	 *                    exist
+	 * @param monitor     a progress monitor, or <code>null</code> if progress
+	 *                    reporting is not desired
+	 * @throws CoreException if this method fails or is canceled.
+	 * @since 3.21
+	 */
+	public default void createOrUpdate(byte[] content, boolean force, boolean derived, boolean keepHistory,
```
laeubi, line 428: Maybe instead of a new method one could think about having a new `FLAG` instead like `IResource.CREATE_MISSING` that way it could be used with any of the existing methods as well and one don't need to add more methods for different variations. This is similar to https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardCopyOption.html and https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html in the JDK.

jukzi: Yea probably that would be possible - somehow. But the details are unclear:
Could that flag be used in create() or setContents() or both? 
If both: where is the desired difference then? The problem to me seems that unlike JDK we already have 2 different methods. Other languages that started with different methods like SQL also came up with explicit new functions like 'UPSERT' or 'CREATE OR REPLACE'.
So unless you come up with a clean solution i would prefer to merge as is.

jukzi: I gave it a try to allow the existing(!) IResource.REPLACE flag for create() as implementation. The outcome is that it is possible with minor changes. The implementation does not become simpler, since the main changes come form allowing a byte array.  It would be a little bit incompatible if create() would suddenly accept flags it never rejected and behave different - but not break the documented API. I don't care much which API we want to use. The change would look like this (on top of this PR):
```
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
index 5cc21b1..8a94d8a 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1275,7 +1275,7 @@
 					throw new ResourceException(IResourceStatus.NOT_FOUND_LOCAL, target.getFullPath(), message, null);
 				}
 			} else {
-				if (fileInfo.exists()) {
+				if (!BitMask.isSet(updateFlags, IResource.REPLACE) && fileInfo.exists()) {
 					String message = NLS.bind(Messages.localstore_resourceExists, target.getFullPath());
 					throw new ResourceException(IResourceStatus.EXISTS_LOCAL, target.getFullPath(), message, null);
 				}
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
index 5bf8598..c699cae 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
@@ -126,7 +126,7 @@
 			final ISchedulingRule rule = workspace.getRuleFactory().createRule(this);
 			try {
 				workspace.prepareOperation(rule, subMonitor.newChild(1));
-				checkCreatable();
+				checkCreatable(updateFlags);
 				workspace.beginOperation(true);
 				IFileStore store = getStore();
 				IFileInfo localInfo = create(updateFlags, subMonitor.newChild(40), store);
@@ -172,7 +172,7 @@
 			final ISchedulingRule rule = workspace.getRuleFactory().createRule(this);
 			try {
 				workspace.prepareOperation(rule, subMonitor.newChild(1));
-				checkCreatable();
+				checkCreatable(updateFlags);
 				workspace.beginOperation(true);
 				IFileStore store = getStore();
 				IFileInfo localInfo = create(updateFlags, subMonitor.newChild(40), store);
@@ -202,8 +202,10 @@
 		}
 	}
 
-	private void checkCreatable() throws CoreException {
-		checkDoesNotExist();
+	private void checkCreatable(int updateFlags) throws CoreException {
+		if ((updateFlags & IResource.REPLACE) == 0) {
+			checkDoesNotExist();
+		}
 		Container parent = (Container) getParent();
 		ResourceInfo info = parent.getResourceInfo(false, false);
 		parent.checkAccessible(getFlags(info));
@@ -230,7 +232,7 @@
 				}
 			}
 		} else {
-			if (localInfo.exists()) {
+			if (!BitMask.isSet(updateFlags, IResource.REPLACE) && localInfo.exists()) {
 				// return an appropriate error message for case variant collisions
 				if (!Workspace.caseSensitive) {
 					String name = getLocalManager().getLocalName(store);
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
index 3e1dcac..05c7a6c 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
@@ -60,6 +60,7 @@
 import org.eclipse.core.internal.refresh.RefreshManager;
 import org.eclipse.core.internal.resources.ComputeProjectOrder.Digraph;
 import org.eclipse.core.internal.resources.ComputeProjectOrder.VertexOrder;
+import org.eclipse.core.internal.utils.BitMask;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
 import org.eclipse.core.internal.utils.StringPoolJob;
@@ -1392,7 +1393,7 @@
 	 * be immediately made derived, hidden and/or team private
 	 */
 	public ResourceInfo createResource(IResource resource, int updateFlags) throws CoreException {
-		ResourceInfo info = createResource(resource, null, false, false, false);
+		ResourceInfo info = createResource(resource, null, false, BitMask.isSet(updateFlags, IResource.REPLACE), false);
 		if ((updateFlags & IResource.DERIVED) != 0)
 			info.set(M_DERIVED);
 		if ((updateFlags & IResource.TEAM_PRIVATE) != 0)
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
index 6910c2d..d4c030e 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
@@ -392,6 +392,8 @@
 	 * @param content     the content as byte array
 	 * @param createFlags bit-wise or of flag constants ({@link IResource#FORCE},
 	 *                    {@link IResource#DERIVED}, and
+	 *                    {@link IResource#KEEP_HISTORY}, and
+	 *                    {@link IResource#REPLACE}, and
 	 *                    {@link IResource#TEAM_PRIVATE})
 	 * @param monitor     a progress monitor, or <code>null</code> if progress
 	 *                    reporting is not desired
@@ -428,92 +430,100 @@
 	public default void createOrUpdate(byte[] content, boolean force, boolean derived, boolean keepHistory,
 		Objects.requireNonNull(content);
-		if (exists()) {
-			if (derived != isDerived()) {
-				setDerived(derived, null);
-			}
-			setContents(content, force, keepHistory, monitor);
-		} else {
-			create(content, force, derived, monitor);
-		}
+		create(content, (force ? IResource.FORCE : 0) | (derived ? IResource.DERIVED : 0)
+				| (keepHistory ? IResource.KEEP_HISTORY : 0) | IResource.REPLACE, monitor);
 	}
 
 	/**
-	 * Creates a new file resource as a member of this handle's parent resource.
-	 * The file's contents will be located in the file specified by the given
-	 * file system path.  The given path must be either an absolute file system
-	 * path, or a relative path whose first segment is the name of a workspace path
-	 * variable.
+	 * Creates a new file resource as a member of this handle's parent resource. The
+	 * file's contents will be located in the file specified by the given file
+	 * system path. The given path must be either an absolute file system path, or a
+	 * relative path whose first segment is the name of a workspace path variable.
 	 * <p>
 	 * The {@link IResource#ALLOW_MISSING_LOCAL} update flag controls how this
 	 * method deals with cases where the local file system file to be linked does
 	 * not exist, or is relative to a workspace path variable that is not defined.
-	 * If {@link IResource#ALLOW_MISSING_LOCAL} is specified, the operation will succeed
-	 * even if the local file is missing, or the path is relative to an undefined
-	 * variable. If {@link IResource#ALLOW_MISSING_LOCAL} is not specified, the operation
-	 * will fail in the case where the local file system file does not exist or the
-	 * path is relative to an undefined variable.
+	 * If {@link IResource#ALLOW_MISSING_LOCAL} is specified, the operation will
+	 * succeed even if the local file is missing, or the path is relative to an
+	 * undefined variable. If {@link IResource#ALLOW_MISSING_LOCAL} is not
+	 * specified, the operation will fail in the case where the local file system
+	 * file does not exist or the path is relative to an undefined variable.
 	 * </p>
 	 * <p>
-	 * The {@link IResource#REPLACE} update flag controls how this
-	 * method deals with cases where a resource of the same name as the
-	 * prospective link already exists. If {@link IResource#REPLACE}
-	 * is specified, then the existing linked resource's location is replaced
-	 * by localLocation's value.  This does <b>not</b>
-	 * cause the underlying file system contents of that resource to be deleted.
-	 * If {@link IResource#REPLACE} is not specified, this method will
-	 * fail if an existing resource exists of the same name.
+	 * The {@link IResource#REPLACE} update flag controls how this method deals with
+	 * cases where a resource of the same name as the prospective link already
+	 * exists. If {@link IResource#REPLACE} is specified, then the existing linked
+	 * resource's location is replaced by localLocation's value. This does
+	 * <b>not</b> cause the underlying file system contents of that resource to be
+	 * deleted. If {@link IResource#REPLACE} is not specified, this method will fail
+	 * if an existing resource exists of the same name.
 	 * </p>
 	 * <p>
-	 * The {@link IResource#HIDDEN} update flag indicates that this resource
-	 * should immediately be set as a hidden resource.  Specifying this flag
-	 * is equivalent to atomically calling {@link IResource#setHidden(boolean)}
-	 * with a value of <code>true</code> immediately after creating the resource.
+	 * The {@link IResource#HIDDEN} update flag indicates that this resource should
+	 * immediately be set as a hidden resource. Specifying this flag is equivalent
+	 * to atomically calling {@link IResource#setHidden(boolean)} with a value of
+	 * <code>true</code> immediately after creating the resource.
 	 * </p>
 	 * <p>
 	 * Update flags other than those listed above are ignored.
 	 * </p>
 	 * <p>
-	 * This method synchronizes this resource with the local file system at the given
-	 * location.
+	 * This method synchronizes this resource with the local file system at the
+	 * given location.
 	 * </p>
 	 * <p>
-	 * This method changes resources; these changes will be reported
-	 * in a subsequent resource change event, including an indication
-	 * that the file has been added to its parent.
+	 * This method changes resources; these changes will be reported in a subsequent
+	 * resource change event, including an indication that the file has been added
+	 * to its parent.
 	 * </p>
 	 * <p>
-	 * This method is long-running; progress and cancellation are provided
-	 * by the given progress monitor.
+	 * This method is long-running; progress and cancellation are provided by the
+	 * given progress monitor.
 	 * </p>
 	 *
 	 * @param localLocation a file system path where the file should be linked
-	 * @param updateFlags bit-wise or of update flag constants
-	 *   ({@link IResource#ALLOW_MISSING_LOCAL}, {@link IResource#REPLACE} and {@link IResource#HIDDEN})
-	 * @param monitor a progress monitor, or <code>null</code> if progress
-	 *    reporting is not desired
-	 * @exception CoreException if this method fails. Reasons include:
-	 * <ul>
-	 * <li> This resource already exists in the workspace.</li>
-	 * <li> The workspace contains a resource of a different type
-	 *      at the same path as this resource.</li>
-	 * <li> The parent of this resource does not exist.</li>
-	 * <li> The parent of this resource is not an open project</li>
-	 * <li> The name of this resource is not valid (according to
-	 *    <code>IWorkspace.validateName</code>).</li>
-	 * <li> The corresponding location in the local file system does not exist, or
-	 * is relative to an undefined variable, and <code>ALLOW_MISSING_LOCAL</code> is
-	 * not specified.</li>
-	 * <li> The corresponding location in the local file system is occupied
-	 *    by a directory (as opposed to a file).</li>
-	 * <li> Resource changes are disallowed during certain types of resource change
-	 *       event notification.  See <code>IResourceChangeEvent</code> for more details.</li>
-	 * <li>The team provider for the project which contains this folder does not permit
-	 *       linked resources.</li>
-	 * <li>This folder's project contains a nature which does not permit linked resources.</li>
-	 * </ul>
+	 * @param updateFlags   bit-wise or of update flag constants
+	 *                      ({@link IResource#ALLOW_MISSING_LOCAL},
+	 *                      {@link IResource#REPLACE} and {@link IResource#HIDDEN})
+	 * @param monitor       a progress monitor, or <code>null</code> if progress
+	 *                      reporting is not desired
+	 * @exception CoreException              if this method fails. Reasons include:
+	 *                                       <ul>
+	 *                                       <li>This resource already exists in the
+	 *                                       workspace.</li>
+	 *                                       <li>The workspace contains a resource
+	 *                                       of a different type at the same path as
+	 *                                       this resource.</li>
+	 *                                       <li>The parent of this resource does
+	 *                                       not exist.</li>
+	 *                                       <li>The parent of this resource is not
+	 *                                       an open project</li>
+	 *                                       <li>The name of this resource is not
+	 *                                       valid (according to
+	 *                                       <code>IWorkspace.validateName</code>).</li>
+	 *                                       <li>The corresponding location in the
+	 *                                       local file system does not exist, or is
+	 *                                       relative to an undefined variable, and
+	 *                                       <code>ALLOW_MISSING_LOCAL</code> is not
+	 *                                       specified.</li>
+	 *                                       <li>The corresponding location in the
+	 *                                       local file system is occupied by a
+	 *                                       directory (as opposed to a file).</li>
+	 *                                       <li>Resource changes are disallowed
+	 *                                       during certain types of resource change
+	 *                                       event notification. See
+	 *                                       <code>IResourceChangeEvent</code> for
+	 *                                       more details.</li>
+	 *                                       <li>The team provider for the project
+	 *                                       which contains this folder does not
+	 *                                       permit linked resources.</li>
+	 *                                       <li>This folder's project contains a
+	 *                                       nature which does not permit linked
+	 *                                       resources.</li>
+	 *                                       </ul>
 	 * @exception OperationCanceledException if the operation is canceled.
-	 * Cancelation can occur even if no progress monitor is provided.
+	 *                                       Cancelation can occur even if no
+	 *                                       progress monitor is provided.
 	 * @see IResource#isLinked()
 	 * @see IResource#ALLOW_MISSING_LOCAL
 	 * @see IResource#REPLACE

```

jukzi: I updated the implementation to use IResource#REPLACE and renamed the new API to 'createOrReplace'.

HannesWell: What do you think about naming it just `write()`, similar like `Files.write()` from the JDK?
Semantically this could include both cases and the flag would also fit.

mickaelistria: I agree `doXOrDoY` API doesn't look so nice and efficient, so +1 for `write` or a variation of `setContents`. I'm also under the impression that the new `create` method with a boolean for derived could probably have been spared, as it increases the size of the API without adding new capability here. And the purpose of flags is exactly to prevent from growing the API methods for every new option.
In a comment yesterday, you mention allowing  `create` to allow the `REPLACE` flag to not fail in case resource is already existing. This seems like a good idea, why was it abandoned? That would require less new methods (only tweaking the existing `create` and adding a new one taking a byte[] as input instead of a stream).

jukzi: > you mention allowing `create` to allow the `REPLACE` flag to not fail in case resource is already existing. This seems like a good idea, why was it abandoned?

create does allow REPLACE now. But i would not use that for clients, as it does not give a warning that the resource bundle may not may support it, as it is only behaviour change but not new semanitc API.

> Files.write()

ok for me. i will create PR to rename it.

laeubi: >  as it does not give a warning that the resource bundle may not support it, as it is only behaviour change but not new semanitc API.

Adding a new flag is an API change that requires a version increment so everyone that use that requires to raise their minimal version, I'm not sure what kind of "warning" is meant here?

merks: It's not new, it just has an additional meaning/use-case:

image

merks:
@jukzi 

I think when changes are merged to some of the most central APIs in the entire Platform that it would be really a good idea to allow discussions to complete and to allow sufficient/reasonable time for folks to complete those discussion.  I know it's nice to finish a set of changes and move on to the next issue.  And of course it's nice  to do that quickly, but it's also an issue of respect for everyone involved in the project.

akurtakov:
@eclipse-platform/eclipse-platform-project-leads Please help in fixing the disagreement here.

laeubi:
Even though I hate to point to "the process" but I think this is clearly in violation of https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines#committer-disagreement-resolution what says

> Committing changes **in case of pending concerns** from committers without the above process **should not happen**. If they are committed by accident or because of a misunderstanding, these changes should be reverted and a new change with the original changes re-created so that the discussion can continue.

so following that this would require a a revert instead of a follow up PR.

It also feels a bit disturbing given that recently @jukzi requested a lot of changes on some of @HannesWell PRs (and others) and they where patiently addressed to "rush in" things just in the middle of discussion.

iloveeclipse:
> Please help in fixing the disagreement here.

Regarding the way how this PR was merged: it is not OK. 
The PR changes API and if there any concerns, they should be discussed not ignored, because API is what makes Eclipse strong.

@jukzi : **please** consider that for the future. It is not the first time we see problems with PR's merged too fast, but I hope it will be the last time.

@merks: could you please be more precise what part of the change needs to be discussed?

> so following that this would require a a revert instead of a follow up PR (as I assume it was by accident here).

I would first clarify the question above, which part of the PR is controversial. Depending on the severity of that we should decide to revert or not.

Note: even if I did many reverts in the past it doesn't mean I like this procedure. Better the PR's are not merged without common agreement.

merks:
My concern is not about any specific aspect of the change but rather a concern about what appears to be short circuiting or bypassing the normal review process. As you say. It’s just not okay.

vogella:
My understanding is that continuing to violate the rules of the PMC would lead to loosing the committership in the top-level Eclipse project. I'm surprised that this behavior happened again, maybe the consequences were not clearly explained the last time, so I hope my statement will help to close this information gap.

@merks or @akurtakov as I'm not a PMC member anymore please correct if my statement above it not correct (anymore).

merks:
@vogella indeed that would be a last resort but I don’t think anyone here would actually want to see us take such a drastic step.  I think we can all agree that @jukzi makes so many high quality contributions that it would be a tragedy for that to stop.  Really a tragedy.  Unfortunately there is also a pattern of social behavior that I personally find very disturbing even just as an observer so I can imagine how that feels when deeply involved in the exchange of ideas.  I tell my partner, oh dear, the children are misbehaving again, as kind of a sad joke.  I definitely have higher expectations and if they are not met, bad things will happen.

vogella:
@merks thanks for the clarification. I agree that loosing highly active committers would be very bad for the project but on the flip side, constant violations of the rules of the project can push other highly active committers out or even worse, let them also  start violating the rules (going the the train of though: why should I care, if others do not and get away with it).

So I would hope that such behavior stops after the right information; as in my opinion such a behavior hurts the project no matter if it has direct consequences or not.

laeubi:
@iloveeclipse The main "disagreement" is was about adding new API, as far as I know we should(?) (or must?) have at least one approving review for new API... this has not happened here and I also really dislike reverts I don't really see how to "heal" this (from process POV) without and actually reviews are not for the comport of the PR author but also to get feedback and maybe new ideas and get confidence from others working on the project that it is a "good thing".

Anyways beside that the following remarks where not yet addressed / discussed and maybe more has to be come if review was not disruptively ended by the merge:

- [ ] how much new API is really needed
- [ ] naming of new API methods
- [ ] reuse existing constants and give them a new meaning versus creating a new one
- [ ] make this general available for the `InputStream` based methods to get rid of the presented (anti-)pattern
- [ ] giving others some time to chime in, test, ...

nothing of that looks like something that can't be addressed and in general I agree that the current check-exits-then-create-or-update is a bad pattern and the change looks useful, but maybe wore would come after iterating reviews.

jukzi:
Hi, lets recap the rules:
This project is driven by the "eclipse-charter" [1].
Despite my repeated requests to PMC members to update the charter to the current practice they did not - and they are not permitted to do so their own [2].
Since the `overall practice is that committers push things without review when feeling comfortable about the stability of the patch` [3] i assume that we now act under section `code changes [...] may be committed first and changed as needed, with conflicts resolved by majority vote of the Committers of the Project or component` [4] - and that even PMC has no right to establish other rules for M1 (which we are in).
Nevertheless, my own goal is to wait 24h on business days for possible reviews for important changes and I did that on this PR as usual. If someone wants more time for review, then please indicate so. At least two committers did review and left a 0. There was no -1. The proposed change to rename the new method was done with  no -1 and three +1.
If someone has a disagreement, he must vote -1 [5]. Later claiming that there has been a disagreement without a -1 is not valid. 
Accusing me to violate our rules hurts my feelings. If anybody - even PMC- has own personal expectation or rules: they are irrelevant.

Nevertheless, i feel very sorry that i have hurt feelings. That was not intentional: I even implemented the wish of laeubi only slightly modified - despite my personal reservations.

So please let us calm down. Make your reviews timely and don’t forget to leave votes as dictated by the charter. I would really like to concentrate on the code.

[1] https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/eclipse-charter.php
[2] https://www.eclipse.org/lists/eclipse-pmc/msg04131.html
[3] https://www.eclipse.org/lists/platform-dev/msg03798.html
[4] https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/eclipse-charter.php#L262C58-L266C62
[5] https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/eclipse-charter.php#L197

mickaelistria:
> my own goal is to wait 24h on business days for possible reviews for important changes

I think this is usually too little for important API changes, as everyone is busy and we cannot expect such reactivity. I think the usual practice of most committers is to wait for a few days, and if there was no review after that time, drop a note on the PR to inform the intent to merge in a few more days, and then, unless there was a negative review or some requests for change, merge it.
All together, it makes about a week.

jukzi:
> I think the usual practice of most committers is to wait for a few days

Your own practice is to sometimes just remove reviews and submit few seconds later. Do you need examples form this week? Please don't put higher expectations on others then on yourself.

laeubi:
>  At least two committers did review and left a 0. There was no -1. The proposed change to rename the new method was done with no -1 and three +1. If someone has a disagreement, he must vote -1

I don't think leaving a comment (!) on a review can be considered a "vote" in that sense (@waynebeaton might can give guidance here) at least I won't consider this as abstain(!) as it definitely is a participation  but at least deserves (for me) to be considered, especially if something is suggest it is just good behavior to at least explain why it was not taken into account or one disagrees, just ignoring it and merge (whatever time frame is waited) feels quite impolite and leaves a bad feeling.

So the consequence would be to always give a -1 on each and every PR, what feels (for me) like really bad experience, I therefore try to avoid that if it is not really critical, but sure can change behavior here, it just don't feels "nice" for minor changes / ideas / suggestions that are maybe debatable and require discussions.

> my own goal is to wait 24h on business days for possible reviews for important changes and I did that on this PR as usual

It really feels like "kindergarden" (as @merks mentioned already) that I'm now need to look up the history and as @mickaelistria  mentions 24hrs are a really tough time-frame given that many people work on their free times here so maybe more have time outside "business days" but here you

1. force pushed the branch on 14.06 8:29 (CET)
2. merged it  on 14.06 10:47 (CET)

that are roughly **3 (!) hours**.

> Your own practice is ...

I think that's exactly what @vogella mentioned here, the worst thing that could happen is that we start to pick whatever we think others doing wrong and use that as a justification for doing the same...

HannesWell:
> So please let us calm down. Make your reviews timely and don’t forget to leave votes as dictated by the charter. I would really like to concentrate on the code.

I assume everyone here would like to focus on the code as well and I also think nobody wants to delay progress. But what timely means for everyone is highly individual. And even if one has no time for a review within a business day (for whatever reason) it can still be that one is interested in participating in the discussion of a change and can supply valuable feedback making the change better. So especially for such fundamental change, personally I think giving others more time would be good.
Furthermore if a discussion has started I think it should be concluded from both sides, even if that were just comments without a 'vote'. I don't think it should be necessary to put a -1/`Request Changes` on a PR for everything that's not an approval, just to get the time to complete a discussion.

But regardless of if behavior is in line with the Eclipse development process or not, I think we should not only try to fulfill the minimum but instead should try to make the discussions satisfactory for everyone who wants to participate.
In order to improve our mutual understanding of a good discussions, I suggest we join at the next Developer community call taking place at the Thursday 27th of June 16:00 CEST, as listed in the calendar I send around at the mailing lists last week, and discuss it.

waynebeaton:
> I don't think leaving a comment (!) on a review can be considered a "vote" in that sense (@waynebeaton might can give guidance here) at least I won't consider this as abstain(!) as it definitely is a participation but at least deserves (for me) to be considered, especially if something is suggest it is just good behavior to at least explain why it was not taken into account or one disagrees, just ignoring it and merge (whatever time frame is waited) feels quite impolite and leaves a bad feeling.

IMHO, a vote should be very clearly a vote. It's generally accepted that +1 signifies agreement.

Having said that, how actual votes are run are really a matter between the project and their PMC. From the EMO's perspective, as long as the open source rules of engagement are respected, we're happy.

PR #1422, opened 2024-06-12
[FileSystem] Replace DosHandler by a Handler calling native methods through JNA

HannesWell:
Calling `Path.toRealPath()` is a very expensive operation on Windows because for each segment of the path the actual name on the file-system is looked up (the Windows FS is case-insensitive). In a Windows JDK this is implemented by calling the native Windows function [`FindFirstFileW()`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilew) for each segment of path. Because we are just interested in the last segment's name, i.e. the filename, all calls except for the last segment are unnecessary.

With this change the `DosHandler` is optimized to call `FindFirstFileW()` only once for the last segment of the path and to read all file-infos from the returned `WIN32_FIND_DATAW` struct using JNA.

I have done some preliminary benchmarks locally and the results look very promising. Depending on the path length the `DosHandler` becomes faster by multiple factors.
I'll hand in the final benchmarks with the final implementation later.


HannesWell:
Now that all tests succeeded I re-run my benchmark with the current code to fetch the info of an existing file.
<details>
<summary>Windows file `NativeHandler` read performance benchmark code</summary>

The old code of the `DosHandler` is stripped down to the two methods that dominate its runtime into the method `readFileAttributesNIO()`.

```
package eclipse.resources.performance.testing;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.LinkOption;
import java.nio.file.Path;
import java.nio.file.attribute.DosFileAttributes;
import java.util.concurrent.TimeUnit;

import org.eclipse.core.internal.filesystem.local.LocalFileNatives;
import org.eclipse.core.internal.filesystem.local.nio.DosHandler;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;

@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
public class DosHandlerBenchmark {

	private static final String PATH = "C:\\foo\\bar\\car\\dar\\ear\\NEUESTEXTDöKUMENT.TXT";
	private static final Path PATH_SHORT = Path.of("C:\\foo");

	private static final Path _PATH = Path.of(PATH);

	@Benchmark
	public DosFileAttributes testReadFileInfoJava() throws IOException {
		DosFileAttributes attrs = readFileAttributesNIO(_PATH);
		return attrs;
	}

	@Benchmark
	public DosFileAttributes testReadFileInfoJava_shortPath() throws IOException {
		return readFileAttributesNIO(PATH_SHORT);
	}

	private static DosFileAttributes readFileAttributesNIO(Path path2) throws IOException {
		Path realPath = path2.toRealPath(LinkOption.NOFOLLOW_LINKS);
		return Files.readAttributes(realPath, DosFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
	}

	@Benchmark
	public Object testReadFileInfoJNI() {
		return LocalFileNatives.fetchFileInfo(PATH);
	}

	private static final DosHandler HANDLE = new DosHandler();

	@Benchmark
	public Object testReadFileInfoJNA() {
		return HANDLE.fetchFileInfo(PATH);

	}
}
```
</details>

The results are 

```
java -jar target\benchmarks.jar
# JMH version: 1.37
# VM version: JDK 21.0.3, OpenJDK 64-Bit Server VM, 21.0.3+9-LTS
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 5 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Result "eclipse.resources.performance.testing.DosHandlerBenchmark.testReadFileInfoJNA":
  27,082 ±(99.9%) 0,095 us/op [Average]
  (min, avg, max) = (26,822, 27,082, 27,299), stdev = 0,127
  CI (99.9%): [26,987, 27,178] (assumes normal distribution)

Result "eclipse.resources.performance.testing.DosHandlerBenchmark.testReadFileInfoJNI":
  27,206 ±(99.9%) 0,241 us/op [Average]
  (min, avg, max) = (26,804, 27,206, 27,937), stdev = 0,322
  CI (99.9%): [26,965, 27,447] (assumes normal distribution)

Result "eclipse.resources.performance.testing.DosHandlerBenchmark.testReadFileInfoJava":
  119,634 ±(99.9%) 0,262 us/op [Average]
  (min, avg, max) = (118,651, 119,634, 120,141), stdev = 0,350
  CI (99.9%): [119,372, 119,896] (assumes normal distribution)

Result "eclipse.resources.performance.testing.DosHandlerBenchmark.testReadFileInfoJava_shortPath":
  30,447 ±(99.9%) 0,473 us/op [Average]
  (min, avg, max) = (29,863, 30,447, 31,682), stdev = 0,632
  CI (99.9%): [29,974, 30,920] (assumes normal distribution)


# Run complete. Total time: 00:33:28

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

NOTE: Current JVM experimentally supports Compiler Blackholes, and they are in use. Please exercise
extra caution when trusting the results, look into the generated code to check the benchmark still
works, and factor in a small probability of new VM bugs. Additionally, while comparisons between
different JVMs are already problematic, the performance difference caused by different Blackhole
modes can be very significant. Please make sure you use the consistent Blackhole mode for comparisons.

Benchmark                                           Mode  Cnt    Score   Error  Units
DosHandlerBenchmark.testReadFileInfoJNA             avgt   25   27,082 ± 0,095  us/op
DosHandlerBenchmark.testReadFileInfoJNI             avgt   25   27,206 ± 0,241  us/op
DosHandlerBenchmark.testReadFileInfoJava            avgt   25  119,634 ± 0,262  us/op
DosHandlerBenchmark.testReadFileInfoJava_shortPath  avgt   25   30,447 ± 0,473  us/op
```

These results show that this change enhances the read performance of the `DosHandler`, for paths that have more than one element even significantly. They also confirm the anticipated dependency of the runtime regarding the number of elements in the path. So another advantage of this change is that it make the runtime of `DosHandler` independent of the number of path elements.

The the performance of `putFileInfo()` is not yet measured but when looking at the current implementation it can probably enhanced too because each of the set operations called reads the current value of the attribute and directly writes the value to the file-system if it was changed:
https://github.com/eclipse-platform/eclipse.platform/blob/6f2754b4d458e2fd18871d7af01fe530cc3cf21c/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/nio/DosHandler.java#L96-L100

This could be enhanced to a single native read of the attributes bit-set and a single write if that bit-set has changed after all attribute values were set. That's the way how it is done in the JNI based native implementation for Windows `localfile.c`.
But I don't know if `putFileInfo()` is called often respectively if it is a hotspot. On the other hand, whit everything that's available now it shouldn't be too much work to implement it.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/Convert.java
```diff
@@ -25,8 +25,8 @@ public class Convert {
 	/** Indicates if we are running on windows */
 	private static final boolean IS_WINDOWS = Platform.OS.isWindows();
 
-	private static final String WIN32_FILE_PREFIX = "\\\\?\\"; //$NON-NLS-1$
-	private static final String WIN32_UNC_FILE_PREFIX = "\\\\?\\UNC"; //$NON-NLS-1$
+	public static final String WIN32_LONG_PATH_PREFIX = "\\\\?\\"; //$NON-NLS-1$
```
jukzi, line 28: That this can be used for long paths is only an example
`For file I/O, the "\\?\" prefix to a path string tells the Windows APIs to disable all string parsing and to send the string that follows it straight to the file system. For example, if the file system supports large paths and file names, you can exceed the MAX_PATH limits that are otherwise enforced by the Windows APIs`

HannesWell: That's right, but I didn't found a technical term for it. Do you know one?
Unless we find a 'definition' from Microsoft I would continue with `WIN32_RAW_PATH_PREFIX`/`WIN32_UNC_RAW_PATH_PREFIX`. The current name basically says nothing.

HannesWell:
> The the performance of `putFileInfo()` is not yet measured but when looking at the current implementation it can probably enhanced too because each of the set operations called reads the current value of the attribute and directly writes the value to the file-system if it was changed:
> 
> https://github.com/eclipse-platform/eclipse.platform/blob/6f2754b4d458e2fd18871d7af01fe530cc3cf21c/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/nio/DosHandler.java#L96-L100
> 
> This could be enhanced to a single native read of the attributes bit-set and a single write if that bit-set has changed after all attribute values were set. That's the way how it is done in the JNI based native implementation for Windows `localfile.c`. But I don't know if `putFileInfo()` is called often respectively if it is a hotspot. On the other hand, whit everything that's available now it shouldn't be too much work to implement it.

The writing method `putFileInfo()` is now also optimized into a single native attribute read and, if something changed, a single native file-attribute write, all done through JNA.
Since this is now actually a complete rewrite of the `DosHandler` that also use java NIO only as a utility in two special-cases I have renamed it to `Win32Handler` and moved it out of the `filesystem.local.nio` package. Before this is submitted I'll squash the commits so that it will be a addition of the new `Win32Handler` and the removal of the previous `DosHandler`.

If there are no unresolved objections I plan to submit this tomorrow, Monday (Europe) evening.

Issue #1427, opened 2024-06-14
random Exception logged from CharsetManagerJob at end of junit tests


jukzi:
i ran org.eclipse.jdt.debug.tests.ui.OpenFromClipboardTests locally and got
```
!MESSAGE The workspace will exit with unsaved changes in this session.
[type: ERROR], [path: null], [message: Workspace was not properly initialized or has already shutdown.], [plugin: org.eclipse.core.resources], [exception: java.lang.Exception: Workspace was not properly initialized or has already shutdown.]

java.lang.Exception: Workspace was not properly initialized or has already shutdown.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Workspace.getWorkManager(Workspace.java:1890)
	at org.eclipse.core.internal.resources.Workspace.endOperation(Workspace.java:1554)
	at org.eclipse.core.internal.resources.CharsetManager$CharsetManagerJob.run(CharsetManager.java:131)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

!ENTRY org.eclipse.core.resources 4 566 2024-06-14 13:29:23.046
!MESSAGE Workspace was not properly initialized or has already shutdown.
!STACK 0
java.lang.Exception: Workspace was not properly initialized or has already shutdown.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Workspace.getWorkManager(Workspace.java:1890)
	at org.eclipse.core.internal.resources.Workspace.endOperation(Workspace.java:1554)
	at org.eclipse.core.internal.resources.CharsetManager$CharsetManagerJob.run(CharsetManager.java:131)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

PR #1428, opened 2024-06-14
clean CharsetManagerJob shutdown #1427

jukzi:
To prevent logging exception on shutdown at end of Junit tests: 'Workspace was not properly initialized or has already shutdown.'

https://github.com/eclipse-platform/eclipse.platform/issues/1427


jukzi:
if there is no objection i plan to submit this tomorrow.

PR #1434, opened 2024-06-18
Revert IFile.create(...REPLACE) - fixes #1433

### LINKED ISSUE #1433: IFile.write: IllegalArgumentException
Replacing IFile.setContents with IFile.write in org.eclipse.jdt.internal.core.Buffer.save(IProgressMonitor, boolean) causes a regression:

```
java.lang.IllegalArgumentException: Attempted to beginRule: P/P, does not match outer scope rule: L/P/Test.java
	at org.eclipse.core.runtime.Assert.isLegal(Assert.java:68)
	at org.eclipse.core.internal.jobs.ThreadJob.illegalPush(ThreadJob.java:144)
	at org.eclipse.core.internal.jobs.ThreadJob.push(ThreadJob.java:416)
	at org.eclipse.core.internal.jobs.ImplicitJobs.begin(ImplicitJobs.java:66)
	at org.eclipse.core.internal.jobs.JobManager.beginRule(JobManager.java:343)
	at org.eclipse.core.internal.resources.WorkManager.checkIn(WorkManager.java:124)
	at org.eclipse.core.internal.resources.Workspace.prepareOperation(Workspace.java:2402)
	at org.eclipse.core.internal.resources.File.create(File.java:174)
	at org.eclipse.core.resources.IFile.write(IFile.java:441)
	at org.eclipse.jdt.internal.core.Buffer.save(Buffer.java:401)
	at org.eclipse.jdt.internal.core.CommitWorkingCopyOperation.executeOperation(CommitWorkingCopyOperation.java:115)
	at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:739)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2452)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2477)
	at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:809)
	at org.eclipse.jdt.internal.core.CompilationUnit.commitWorkingCopy(CompilationUnit.java:398)
	at org.eclipse.jdt.core.tests.model.EncodingTests.test031(EncodingTests.java:599)
```

I will investigate. If someone has an idea: please share

jukzi:
Allows to use modifyRule for IFile.write() if the file already exits, so that every IFILE.setContents() can be replaced with IFile.create().

https://github.com/eclipse-platform/eclipse.platform/issues/1433

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1286,15 +1286,11 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 				}
 			}
 		}
-		if (BitMask.isSet(updateFlags, IResource.REPLACE)) {
-			// forced update of derived flag during IFile.write:
+		if (BitMask.isSet(updateFlags, IResource.DERIVED)) {
+			// update of derived flag during IFile.write:
 			Resource resource = (Resource) target;
 			ResourceInfo info = resource.getResourceInfo(false, false);
-			if (BitMask.isSet(updateFlags, IResource.DERIVED)) {
-				info.set(ICoreConstants.M_DERIVED);
-			} else {
-				info.clear(ICoreConstants.M_DERIVED);
-			}
+			info.set(ICoreConstants.M_DERIVED);
 		}
 		// add entry to History Store.
 		if (BitMask.isSet(updateFlags, IResource.KEEP_HISTORY) && fileInfo.exists()
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1286,6 +1286,16 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 				}
 			}
 		}
+		if (BitMask.isSet(updateFlags, IResource.REPLACE)) {
+			// forced update of derived flag during IFile.write:
+			Resource resource = (Resource) target;
+			ResourceInfo info = resource.getResourceInfo(false, false);
+			if (BitMask.isSet(updateFlags, IResource.DERIVED)) {
+				info.set(ICoreConstants.M_DERIVED);
+			} else {
+				info.clear(ICoreConstants.M_DERIVED);
+			}
+		}
 		// add entry to History Store.
 		if (BitMask.isSet(updateFlags, IResource.KEEP_HISTORY) && fileInfo.exists()
 				&& FileSystemResourceManager.storeHistory(target))
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1275,7 +1275,7 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 					throw new ResourceException(IResourceStatus.NOT_FOUND_LOCAL, target.getFullPath(), message, null);
 				}
 			} else {
-				if (!BitMask.isSet(updateFlags, IResource.REPLACE) && fileInfo.exists()) {
+				if (fileInfo.exists()) {
 					String message = NLS.bind(Messages.localstore_resourceExists, target.getFullPath());
 					throw new ResourceException(IResourceStatus.EXISTS_LOCAL, target.getFullPath(), message, null);
 				}
@@ -1285,12 +1285,6 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 							null);
 				}
 			}
-		}
-		if (BitMask.isSet(updateFlags, IResource.DERIVED)) {
-			// update of derived flag during IFile.write:
-			Resource resource = (Resource) target;
-			ResourceInfo info = resource.getResourceInfo(false, false);
-			info.set(ICoreConstants.M_DERIVED);
 		}
 		// add entry to History Store.
 		if (BitMask.isSet(updateFlags, IResource.KEEP_HISTORY) && fileInfo.exists()
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
@@ -414,7 +414,7 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 *                    in sync with the local file system
 	 * @param derived     Specifying this flag is equivalent to atomically calling
 	 *                    {@link IResource#setDerived(boolean)} immediately after
-	 *                    creating the resource or non-atomically setting the
+	 *                    creating the resource or atomically setting the
 	 *                    derived flag before setting the content of an already
 	 *                    existing file
 	 * @param keepHistory a flag indicating whether or not store the current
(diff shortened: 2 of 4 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -438,8 +428,14 @@ public default void create(byte[] content, int createFlags, IProgressMonitor mon
 	public default void write(byte[] content, boolean force, boolean derived, boolean keepHistory,
 			IProgressMonitor monitor) throws CoreException {
 		Objects.requireNonNull(content);
-		create(content, (force ? IResource.FORCE : 0) | (derived ? IResource.DERIVED : 0)
-				| (keepHistory ? IResource.KEEP_HISTORY : 0) | IResource.REPLACE, monitor);
+		if (exists()) {
+			if (derived != isDerived()) {
+				setDerived(derived, null);
+			}
+			setContents(content, force, keepHistory, monitor);
```
mickaelistria, line 435: Usually, methods in resource are supposed to be atomic (ie grouped in a single operation), here those 2 methods would lead to 2 operations side by side but not grouped.
What you can try to remediate that would be something like
```java
if (derived == isDerived()) {
	setContents(...);
} else {
	SubMonitor progress = SubMonitor.convert(monitor, 100).checkCanceled();
	// need to group operation
	final ISchedulingRule groupRule = new MultiRule[] { workspace.getRuleFactory().derivedRule(this), workspace.getRuleFactory().modifyRule(this) };
	try {
		workspace.prepareOperation(rule, progress.split(1)); // locks
		workspace.beginOperation(true);
		setDerived(derived, progress.split(1));
		setContents(content, force, keepHistory, progress.split(98))
	} catch (OperationCanceledException e) {
		workspace.getWorkManager().operationCanceled();
		throw e;
	} finally {
		progress.done();
		workspace.endOperation(rule, true); // this is extremely important of workspace is locked
	}
```

mickaelistria: API-wise it's totally fine. See nested comment about the workspace operation should be atomic.

jukzi: Thanks for the suggestion! I have tried your implementation and it would work but can not be done in default method of IFile due to references to internal methods. but based on the idea i think i have found a easier solution with less overhead. please review again. Junit also tests for atomic operations now.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1286,6 +1286,16 @@ private void prepareWrite(IFile target, IFileInfo fileInfo, int updateFlags, boo
 				}
 			}
 		}
+		if (BitMask.isSet(updateFlags, IResource.REPLACE)) {
+			// forced update of derived flag during IFile.write:
+			Resource resource = (Resource) target;
+			ResourceInfo info = resource.getResourceInfo(false, false);
+			if (BitMask.isSet(updateFlags, IResource.DERIVED)) {
+				info.set(ICoreConstants.M_DERIVED);
```
mickaelistria, line 1294: This could probably be even happening without REPLACE.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -414,7 +414,7 @@ public default void create(byte[] content, int createFlags, IProgressMonitor mon
 	 *                    in sync with the local file system
 	 * @param derived     Specifying this flag is equivalent to atomically calling
 	 *                    {@link IResource#setDerived(boolean)} immediately after
-	 *                    creating the resource or non-atomically setting the
+	 *                    creating the resource or atomically setting the
```
mickaelistria, line 417: Given the pain introduced by "atomically setting the derived flag before setting the content of an already" which as far as I understand is not a very interesting case (derived usually matters at creation), maybe we should just restrict it and make it "createAsDerived" and ignore it when resource already exists.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
mickaelistria: >  in default method of IFile due to references to internal methods. 

There is nothing mandating to create a default method here. IFile is annotated as "noextends/noimplements", so we can add new interface methods in here and have the implementation in File.

jukzi: So far the therory. In practice there are other implementations even within the IDE (PDE's ContributionDataFile) :-(

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: If we remove the REPLACE condition we would set derived=false on every normal setContent(FLAG). That is not what we want.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
mickaelistria: While it's an annoyance, I think keeping the resource code as maintainable as posible is more important than keeping e4 tools alive. IMO, it's best to go for a regular interface method and impl in File, and to adapt PDE then.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
mickaelistria: I was speaking of this particular case (setting DERIVED to true) doesn't require REPLACE. Here REPLACE would only be useful in order to un-set derived.
But as mentioned earlier, the case of changing value and derived at the same time seems not so useful and not so trivial to implement properly. We could probably just ditch it by restraining how `derived` is used in `write`

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: The single most important usage of this API still is org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.writeClassFile(ClassFile, SourceFile, boolean) where the derived flag is updated also for existing files. Thats ancient code and i do not know why. But its legit. 
As a compromise we could only change (i.e. set) derived flag if derived==true is given as argument. In that case we would also not need the REPLACE. I pushed a implemementation for that: "A value of false will not update the derived flag of an existing file."

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: ok, i think i implemented that now  se below.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
jukzi: sure, that would be good. But i have no idea why it is like that in PDE and how it could be solved. So as long as we can implement a default its  much easier to proceed.

jukzi:
@mickaelistria can you please review again - i would like to merge today

mickaelistria (changes requested):
My previous comments still stand, and I even support it stronger: it's better to just adopt the usual pattern of decl in interface + impl in File to not be bothered with REPLACE in lower-level API, even if it requires changing PDE also. This will ultimately lead to better result.
The over-usage of the REPLACE field in the lower level FileSystemResourceManager is worrying me, it's not the right grain where to hook this complexity. This is IMO a kind of design leak from a higer-level utility method into its low-level backend, and we shouldn't allow that unneeded change in the backend just for introduction of a convenience method downstream.

jukzi:
> The over-usage of the REPLACE field in the lower level FileSystemResourceManager is worrying me

@mickaelistria i already solved that as you recommended. Please read the code again.
It feels wrong to reject a fix only because PDE did wrong.

mickaelistria:
> i already solved that as you recommended. Please read the code again.

To be clear, I think that the introduction of `write` should not change FileSystemResourceManager at all, and I even think that everything about REPLACE should also remained the same old. This patch still modifies it.
I also don't see File implementing write directly as we discussed ealier.

> It feels wrong to reject a fix only because PDE did wrong.

The thing is that I barely see your proposal here as a fix but more as some higher level concepts leaking into lower-level API, thus placing complexity at the wrong level which will utlimately hurt.
I suggest you just fully ignore PDE at the moment, and just make the default method an empty shell doing nothing (until PDE is fixed) and then allow yourself to implement a write in File that is consistent with all other operations; without need to modify FileSystemResourceManager.

jukzi:
> To be clear, I think that the introduction of `write` should not change FileSystemResourceManager at all,

ok, i moved the code from FileSystemResourceManager to File - it can be done with the default method to not produce error in PDE.

mickaelistria (approved):
Thanks, I think the current proposal is a clear improvement.

PR #1440, opened 2024-06-22
Merge native org.eclipse.core.net fragments into host plug-in

HannesWell:
And fix warnings stricter in the target plugin.

This avoids the need to maintain the code in multiple projects and having an extra jar in Windows and Linux installations.
The drawbacks are that for Mac a few more classes are contained in the `org.eclipse.core.net` jar and that `org.eclipse.core.net` now has an extra dependency to JNA. But even for just an Eclipse SDK JNA is already included in mac installations.
For Linux and Windows the overhead of having a separate jar file probably outweighs the few extra classes for the other OS anyway and the fragments already depended on JNA.


team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/linux/ProxyProviderLinux.java
```diff
@@ -11,7 +11,7 @@
  * Contributors:
  *     Red Hat Inc. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.core.net;
+package org.eclipse.core.net.proxy.linux;
```
akurtakov, line 14: What do you think about making this internal package? I don't believe this is smth that should be exposed as API.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/win32/ProxyBypass.java
```diff
@@ -12,7 +12,7 @@
  * 	compeople AG (Stefan Liebig) - initial API and implementation
  *  IBM Corporation - Add proxy providers layer on the top of ProxyManager (bug 255616)
  *******************************************************************************/
-package org.eclipse.core.internal.net.proxy.win32.winhttp;
+package org.eclipse.core.net.proxy.win32;
```
akurtakov, line 15: Make it internal package. Definetely not smth we want anyone to consider API.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/win32/ProxyProviderUtil.java
```diff
@@ -13,7 +13,7 @@
  * 	IBM Corporation - bug 246072 - adding IProxyData.source support
  *  IBM Corporation - Add proxy providers layer on the top of ProxyManager (bug 255616)
  *******************************************************************************/
-package org.eclipse.core.internal.net.proxy.win32.winhttp;
+package org.eclipse.core.net.proxy.win32;
```
akurtakov, line 16: Internal package please. Shouldn't be API.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/win32/ProxyProviderWin32.java
```diff
@@ -11,7 +11,7 @@
  * Contributors:
  *     Rolf Theunissen <rolf.theunissen@gmail.com> - initial API and implementation
  *******************************************************************************/
-package org.eclipse.core.net;
+package org.eclipse.core.net.proxy.win32;
```
akurtakov, line 14: Internal package please.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/win32/StaticProxyConfig.java
```diff
@@ -13,7 +13,7 @@
  *  IBM Corporation - handling URI without a scheme by select(URI, List) (bug 246065)
  *  IBM Corporation - Add proxy providers layer on the top of ProxyManager (bug 255616)
  *******************************************************************************/
-package org.eclipse.core.internal.net.proxy.win32.winhttp;
```
akurtakov, line 16: Internal package please.

akurtakov (commented):
This is a move in the right direction. We have went many steps, every OS with native fragment, some via JNA other native, all via JNA and this is the natural next step - merge fragments in the main bundle. One day we will even drop JNA to make it even more straightforward.
Only request is to move arch specific parts into internal packages as no one should consider them API in any way.

HannesWell:
> This is a move in the right direction. We have went many steps, every OS with native fragment, some via JNA other native, all via JNA and this is the natural next step - merge fragments in the main bundle. One day we will even drop JNA to make it even more straightforward. 

Absolutely. I'm looking forward to the day when we can use the Java FFM API to do the native calls.

> Only request is to move arch specific parts into internal packages as no one should consider them API in any way.

I considered that as well, but didn't do it since these packages are not exported at all and therefore not accessible and API.
But I'm also fine to make it even more explicit that they are not accidentally exported one day.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/win32/ProxyBypass.java
HannesWell: Done.

team/bundles/org.eclipse.core.net/src/org/eclipse/core/net/proxy/linux/ProxyProviderLinux.java
HannesWell: There packages where not exported, but it can definitely be made more explicit.
Implemented it.

laeubi:
> But I'm also fine to make it even more explicit that they are not accidentally exported one day.

I just wanted to note that not exporting the package is the right way, "internal" is nothing specified or mandatory by OSGi and actually a lot of "internal" packages are exported in Eclipse platform so I don't think the naming "internal" it helps in any way.

HannesWell:
> > But I'm also fine to make it even more explicit that they are not accidentally exported one day.
> 
> I just wanted to note that not exporting the package is the right way, "internal" is nothing specified or mandatory by OSGi and actually a lot of "internal" packages are exported in Eclipse platform so I don't think the naming "internal" it helps in any way.

That's all correct. But at the same time it does not hurt to add an `internal` element in the package name since we can name it as we like. Since placing non API elements in an internal package is a common naming-convention, especially in areas that don't have a power-full module system like OSGi that can really prevent access. So at least for some it makes it simpler to understand the elements in the bundle and their structure.
But before this leads to a fundamental discussion:
Because both is done, package named 'internal' and not exported, everyone should be happy :)

___


Before this is submitted the removed fragments should be removed from the `org.eclipse.platform-feature`:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/cc85a7a285e7e893345763cc6f8df51957546a47/eclipse.platform.releng/features/org.eclipse.platform-feature/feature.xml#L72-L80

The `org.eclipse.core.net.linux` fragment is also listed in some of the p2 products. But since they are not build and many of the listed bundles are already gone I don't see it as a strict prerequisite to remove it there  (although a clean-up of them would be good in general).

I have also checked the SimRel aggregator and nothing else depends on these fragments.

grafik

merks:
I expect the installer to be affected but that’s totally not a problem. It will be a simplification and that’s a good thing.

HannesWell:
Unless there are objections, I plan to submit this tomorrow (European) evening.

HannesWell:
> Before this is submitted the removed fragments should be removed from the `org.eclipse.platform-feature`:
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/cc85a7a285e7e893345763cc6f8df51957546a47/eclipse.platform.releng/features/org.eclipse.platform-feature/feature.xml#L72-L80

This is now done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2156.

As soon as that PR is submitted, I'll continue and submit this.

PR #1441, opened 2024-06-22
[FileSystem] Simplify and modernize NativeHandlers

HannesWell:


jukzi (changes requested):
The title "modernize" is to vague. Please explicitly describe what you have done - and why. i don't even get why Path.of() should be any better.
The implicit exists() check seems to be good!

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/nio/DefaultHandler.java
```diff
@@ -35,13 +37,12 @@ public class DefaultHandler extends NativeHandler {
...
 	public FileInfo fetchFileInfo(String fileName) {
-		Path path = Paths.get(fileName);
+		Path path = Path.of(fileName);
 		FileInfo info = new FileInfo();
-		boolean exists = Files.exists(path);
-		info.setExists(exists);
 
 		try {
 			BasicFileAttributes readAttributes = Files.readAttributes(path, BasicFileAttributes.class);
+			info.setExists(true);
```
jukzi, line 45: please add a junit test if this behaves like before for hidden files as the code contains comments that opening hidden files can lead to FileNotFoundException. See uses of org.eclipse.core.filesystem.EFS.ATTRIBUTE_HIDDEN

HannesWell:
I'll have a look at the tests. 

> i don't even get why Path.of() should be any better.

`Paths.get()` was only introduced because back then static methods on interfaces were not supported. When it was possible `Path.of` was added, as a method that is easier to discover/remember/read. And since both are technical equal I don't see a reason to not use it.

jukzi:
> And since both are technical equal I don't see a reason to not use it.

I would also use Path for new code, but is it worth to enlarge the git history with rewriting it?

akurtakov:
> I would also use Path for new code, but is it worth to enlarge the git history with rewriting it?

As the PR is not just about that and what is questioned removes one useless method call from the chain - it is worth it.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/nio/DefaultHandler.java
jukzi: you can look at https://github.com/eclipse-platform/eclipse.platform/pull/1444/files#diff-81300c5bdd85b0fca987936ae34590fbf4bbf835119e1efd68698461d230b817R252 how to create a hidden file. 
But to me it is not clear how/if the Defaulthandler is tested in CI

vogella:
@HannesWell can you please resolve the conflicts  in this one? Would be nice to merge this early next release.

Issue #1443, opened 2024-06-24
IFile.write performance overhead


jukzi:
IFile.write invokes org.eclipse.core.filesystem.provider.FileStore.fetchInfo() several times:
![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/6bd86ec2-7f8f-491c-92bf-ebd396742708)
That should not be needed. - similar to https://github.com/eclipse-platform/eclipse.platform/issues/1442 for the read case

PR #1446, opened 2024-06-25
[performance] IFile.write(byte[]): reduce store.fetchInfo()

jukzi:
Optimistically assume the parent folder already exists on local File System when it exists in workspace. So do not explicitly check it. If it did not exist file write will fail and can be retried after creating the parent in File System - which can only happen when the workspace is out of sync with local File System.
This optimization is only implemented for byte[] content in the not appending case:
 * InputStream content would need a reset.
 * In append mode it is not obvious if something was already appended.

https://github.com/eclipse-platform/eclipse.platform/issues/1443

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1212,7 +1212,7 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 		try (content) {
 			Resource targetResource = (Resource) target;
 			IFileStore store = getStore(target);
-			prepareWrite(target, fileInfo, updateFlags, append, targetResource, store);
+			prepareWrite(target, fileInfo, updateFlags, append, targetResource, store, false);
 
 			int options = append ? EFS.APPEND : EFS.NONE;
 			try {
@@ -1254,7 +1254,7 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 	}
 
 	private void prepareWrite(IFile target, IFileInfo fileInfo, int updateFlags, boolean append,
-			Resource targetResource, IFileStore store) throws CoreException {
+			Resource targetResource, IFileStore store, boolean assumeParentDirectoryExists) throws CoreException {
 		if (fileInfo.getAttribute(EFS.ATTRIBUTE_READ_ONLY)) {
 			String message = NLS.bind(Messages.localstore_couldNotWriteReadOnly, target.getFullPath());
 			throw new ResourceException(IResourceStatus.FAILED_WRITE_LOCAL, target.getFullPath(), message, null);
@@ -1303,7 +1303,7 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 			// never move to the history store, because then the file is missing if write
 			// fails
 			getHistoryStore().addState(target.getFullPath(), store, fileInfo, false);
-		if (!fileInfo.exists()) {
+		if (!assumeParentDirectoryExists && !fileInfo.exists()) {
 			IFileStore parent = store.getParent();
 			IFileInfo parentInfo = parent.fetchInfo();
 			if (!parentInfo.exists()) {
@@ -1338,7 +1338,7 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 		SubMonitor subMonitor = SubMonitor.convert(monitor, 4);
 		Resource targetResource = (Resource) target;
 		IFileStore store = getStore(target);
-		prepareWrite(target, fileInfo, updateFlags, append, targetResource, store);
+		prepareWrite(target, fileInfo, updateFlags, append, targetResource, store, true);
 
 		// On Windows an attempt to open an output stream on a hidden file results in
 		// FileNotFoundException.
@@ -1353,7 +1353,25 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 			subMonitor.split(1);
 		}
 		int options = append ? EFS.APPEND : EFS.NONE;
-		store.write(content, options, subMonitor.split(1));
+		try {
+			store.write(content, options, subMonitor.split(1));
+		} catch (CoreException e) {
+			if (append) {
+				throw e;
+			}
+			if (e.getStatus().getCode() != EFS.ERROR_WRITE) {
+				throw e;
+			}
+			IFileStore parent = store.getParent();
+			IFileInfo parentInfo = parent.fetchInfo();
+			if (parentInfo.exists()) {
+				throw e;
+			}
+			// create missing folders:
+			parent.mkdir(EFS.NONE, null);
+			// try again:
+			store.write(content, options, null);
+		}
 		if (restoreHiddenAttribute) {
 			fileInfo.setAttribute(EFS.ATTRIBUTE_HIDDEN, true);
 			store.putInfo(fileInfo, EFS.SET_ATTRIBUTES, subMonitor.split(1));
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IFileTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IFileTest.java
@@ -46,6 +46,7 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
+import org.eclipse.core.internal.resources.ResourceException;
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFileState;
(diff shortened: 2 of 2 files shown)


jukzi:
if there are no objections i plan to merge tomorrow

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1243,7 +1243,7 @@ public void write(IFile target, InputStream content, IFileInfo fileInfo, int upd
 	}
 
 	private void prepareWrite(IFile target, IFileInfo fileInfo, int updateFlags, boolean append,
-			Resource targetResource, IFileStore store) throws CoreException {
+			Resource targetResource, IFileStore store, boolean optimistic) throws CoreException {
```
HannesWell, line 1246: The name of the parameter should be more precise. For example `assumeParentDirectoryExists`.
Then you don't have to add a comment below about what the method is optimistic about :)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1342,7 +1345,23 @@ public void write(IFile target, byte[] content, IFileInfo fileInfo, int updateFl
...
+			if (e.getStatus().getCode() != EFS.ERROR_WRITE) {
+				throw e;
+			}
+			IFileStore parent = store.getParent();
+			IFileInfo parentInfo = parent.fetchInfo();
+			if (!parentInfo.exists()) {
+				parent.mkdir(EFS.NONE, null);
+				// try again
+				store.write(content, options, null);
+			}
```
HannesWell, line 1363: Would it be possible to handle this within the store implementation?
This would save throwing and catching the CoreException. But this would probably require an extra options-flag.

According to this article creating exceptions is still surprisingly expensive:
https://www.baeldung.com/java-exceptions-performance

When I compare it to the numbers I found for https://github.com/eclipse-platform/eclipse.platform/pull/1422#issuecomment-2184161944 I see a difference of three magnitudes.

I have to admit that I would not have expected that throwing an exception is almost a thousand times slower than fetching file infos from the drive. So I'm not sure if I miss-interpret something and of curse comparing benchmarks from different machines is usually not meaningful.
But I think it would be good to have some form of benchmark if the new solution is net better than the old approach in an average workspace.
I assume that if I for example run a CLEAN+FULL_BUILD that all folders in the classpath output folder are deleted and re-created when the first class file was compiled? Overall these are many folders that are then missing on first write.

HannesWell: If throwing/catching exceptions is indeed that expensive, can't we just reuse the fetched file-info instead?

laeubi: > According to this article creating exceptions is still surprisingly expensive:

I also mentioned that last time with the "NIO vs native" discussion, the NIO is*only* noticeable slower when ist came to non exiting files because there Exceptions are used and if you have a large number of exception it makes a difference, but in general one should not use exceptions to control the flow anyways

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: done.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
jukzi: > According to this article creating exceptions is still surprisingly expensive:
> https://www.baeldung.com/java-exceptions-performance

Which numbers do you read? Be careful: that Article measures a loop of LIMIT invocations (which is bad practice) and gives numbers of that. It also does not mention that the time actually highly depends on the depth on the stacktrace elements, So it just can't give absolute numbers.

> I assume that if I for example run a CLEAN+FULL_BUILD that all folders in the classpath output folder are deleted and re-created when the first class file was compiled? Overall these are many folders that are then missing on first write.

Put a breakpoint in the catch clause and run the action you mention => it never stops. Because it would require the absurd conditions given in the added testcase: The folder must exit in workspace but not exist in FileSystem. That only happens if the folder is deleted without using the resource API.
So this is a non-issue for good plugins using the resource API only.

jukzi: > Would it be possible to handle this within the store implementation?

Only by changing that API which is not the plan for this PR and would have use only for the unlikely use case.

HannesWell: > > According to this article creating exceptions is still surprisingly expensive:
> > https://www.baeldung.com/java-exceptions-performance
> 
> Which numbers do you read? Be careful: that Article measures a loop of LIMIT invocations (which is bad practice) and gives numbers of that. It also does not mention that the time actually highly depends on the depth on the stacktrace elements, So it just can't give absolute numbers.

I read the number of the `ExceptionBenchmark.throwAndCatchException` benchmark.
But you are right, this is done 10_000 times in a loop. No clue why this is done since JMH can measure in µs.

> 
> > I assume that if I for example run a CLEAN+FULL_BUILD that all folders in the classpath output folder are deleted and re-created when the first class file was compiled? Overall these are many folders that are then missing on first write.
> 
> Put a breakpoint in the catch clause and run the action you mention => it never stops. Because it would require the absurd conditions given in the added testcase: The folder must exit in workspace but not exist in FileSystem. That only happens if the folder is deleted without using the resource API. So this is a non-issue for good plugins using the resource API only.

OK, I verified that. But please be more descriptive in your PR and commit message what you mean with 'normal case'. Because it is not clear that it can only happen if the the workspace resource store out of sync with the file-system and since the parent existence check for the resources system happens somewhere else also not obvious from the changed code.

Then I'm fine with this.

sratz:
The `assumeParentDirectoryExists` optimization does not play well with the Semantic File System, https://git.eclipse.org/r/plugins/gitiles/e4/org.eclipse.e4.resources/+/refs/heads/master/bundles/org.eclipse.core.resources.semantic/src/org/eclipse/core/internal/resources/semantic/SemanticFileStore.java

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.resource.Resource.save(java.util.Map)" because "resource" is null
	at org.eclipse.core.internal.resources.semantic.SemanticMetadataPersistenceManager.saveSemanticDB(SemanticMetadataPersistenceManager.java:132)
	at org.eclipse.core.internal.resources.semantic.SemanticFileSystem.saveSemanticDB(SemanticFileSystem.java:598)
	at org.eclipse.core.internal.resources.semantic.SemanticFileSystem.requestFlush(SemanticFileSystem.java:122)
	at org.eclipse.core.internal.resources.semantic.SemanticFileStore.openOutputStream(SemanticFileStore.java:713)
	at org.eclipse.core.filesystem.IFileStore.write(IFileStore.java:498)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.write(FileSystemResourceManager.java:1349)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.internalWrite(FileSystemResourceManager.java:719)
	at org.eclipse.core.internal.resources.Project.writeDescription(Project.java:1466)
	at org.eclipse.core.internal.resources.Project.writeDescription(Project.java:1447)
	at org.eclipse.core.internal.resources.Project.create(Project.java:331)
	at org.eclipse.core.internal.resources.Project.create(Project.java:291)
```

The SFS implementation should probably be made more robust here and handle the case that the parent does not yet exist internally.

Nevertheless, when creating a new project, `Project.writeDescription()` will now **always** run into this _`CoreException` + retry logic_ branch at least once.

Is there a way to avoid this?

jukzi:
> Is there a way to avoid this?

Interesting observation: creating a projects should create directory and .project file! Something like this should avoid that (works in org.eclipse.core.tests.resources.IProjectTest):

```
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
index 872223f..3d7dd0b 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Project.java
@@ -328,6 +328,7 @@
 					workspace.getMetaArea().writePrivateDescription(this);
 				} else {
 					// write out the project
+					getLocalManager().getStore(this).mkdir(EFS.NONE, null);
 					writeDescription(IResource.FORCE);
 				}
 			} catch (CoreException e) {

```

sratz:
> Something like this should avoid that (works in org.eclipse.core.tests.resources.IProjectTest):

Sounds reasonable, however, reading the Javadoc of
`org.eclipse.core.internal.resources.Project.writeDescription(IProjectDescription, int, boolean, boolean)`
and
`org.eclipse.core.internal.localstore.FileSystemResourceManager.internalWrite(IProject, IProjectDescription, int, boolean, boolean)`
and seeing that they are apparently somewhat 'special', I guess the defensive approach would be to ensure directory is created **inside** the `internalWrite` method.

jukzi:
> is created **inside** the `internalWrite` method.

It is .. just in the exceptional case. it can be avoided by creating it before.  That is not a contradiction. Also all of that is just internal stuff.

PR #1447, opened 2024-06-25
Add id_ed25519 to list of SSH private key defaults

HannesWell:
For SSH I usually use `ed25519` type keys if possible. Therefore one of my first actions when using a fresh installation (without Oomph) is to add `id_ed25519` to the list of private SSH keys.

According to Stack-Exchange https://security.stackexchange.com/a/144044 ed25519 is currently the recommended key type, therefore I think it is fine to add it to the list of defaults.
Given it is the recommended type, it could even be added to the head of the list to be searched first.
Any opinions on that?


vogella:
Thanks for this, I would also like to see it at the first element in the search first.

vogella (approved):
LGTM, thanks for this.

PR #1452, opened 2024-06-28
LocalFileCopy

jukzi:
used during
org.eclipse.jdt.internal.core.builder.BatchImageBuilder.cleanOutputFolders ()


jukzi:
-1 is slower on windows then the eclipse version :-(

Issue #1458, opened 2024-07-02
Reset preference should set the encoding to UTC-8


vogella:
If a project does not have an explicit encoding set, we can a warning. But I reset the preference to default, I get encoding from the container. This feels wrong, I suggest we change our default to use UTC-8 encoding.

Current behavior:

Before reset:

![image](https://github.com/eclipse-platform/eclipse.platform/assets/139910/6421ab34-4331-4734-b40a-ff7bdf75ac8b)

After reset:

![image](https://github.com/eclipse-platform/eclipse.platform/assets/139910/1e3a089c-9d10-4979-a17e-1deab636fd03)

vogella:
I think the basic issue was this one https://github.com/eclipse-platform/eclipse.platform.resources/issues/154

vogella:
Also this one set the UTF-8 default for new workspaces. https://github.com/eclipse-platform/eclipse.platform.resources/pull/51/files but restore defaults still changes it to inherit from container.

jukzi:
+1 **reset** workspace should always be the same state as in **new** workspace

vogella:
Not sure there the default is set, /org.eclipse.core.resources/src/org/eclipse/core/internal/resources/PreferenceInitializer.java seems to not involved it the preference page setting. I think we should also set in this page UTF-8 as default but that has no direct impact on the preference page default.

vogella:
https://github.com/eclipse-platform/eclipse.platform/pull/1462

PR #1462, opened 2024-07-03
Using UTF_8 as default in core.resources.PreferenceInitializer

vogella:
https://github.com/eclipse-platform/eclipse.platform.resources/pull/51/ implemented logic to set UTF-8 as standard.
I suggest to also set the default preference value to UTF-8.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/PreferenceInitializer.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/PreferenceInitializer.java
@@ -15,9 +15,12 @@
  *******************************************************************************/
 package org.eclipse.core.internal.resources;
 
+import java.nio.charset.StandardCharsets;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.preferences.*;
+import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer;
+import org.eclipse.core.runtime.preferences.DefaultScope;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 
 /**
  * @since 3.1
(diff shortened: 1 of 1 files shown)


merks:
@iloveeclipse 

Sorry for the reviewer bounce.  

I have a feeling this is not correct.  Looking at the  logic here, the default depends on a bunch of things such as the system property or the encoding preference:

https://github.com/eclipse-platform/eclipse.platform/blob/0f64336d2bc4c785643f38d429bd36bdbae0337e/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java#L2339-L2383

But the initializer logic of this PR hard codes the value.

vogella:
@merks I do not get what is different in your review request? Can you explain? 

image

merks:
> @merks I do not get what is different in your review request? Can you explain?

I **accidentally** removed him and then added him back.

iloveeclipse:
Thanks Ed! Lars, it os not as trivial as one could think, I recommend to read the code Ed referenced and related tickets.
I will be able ti review this later, hopefully on Friday, as I'm two days mostly not on my workstation.

vogella:
> > @merks I do not get what is different in your review request? Can you explain?
> 
> I **accidentally** removed him and then added him back.

Thanks for the clarification

vogella:
> Thanks Ed! Lars, it os not as trivial as one could think, I recommend to read the code Ed referenced and related tickets. I will be able ti review this later, hopefully on Friday, as I'm two days mostly not on my workstation.

See also https://github.com/eclipse-platform/eclipse.platform/issues/1458, currently the reset of the preference sets the encoding to an undesired value. This is *NOT* changed by this PR, I could not find the responsible preference initializer.

merks:
@vogella 

My concern is that if the preference is set, then the logic at line 2357 will find a non-blank value and will not longer consider the explicit file.encoding system property:

https://github.com/eclipse-platform/eclipse.platform/blob/0f64336d2bc4c785643f38d429bd36bdbae0337e/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java#L2354-L2357

When I debugged this I see we do get to org.eclipse.core.internal.resources.PreferenceInitializer.initializeDefaultPreferences() before we get to org.eclipse.core.internal.resources.Workspace.setExplicitWorkspaceEncoding() so this is definitely of concern.

In the preferences page, these things are all of interest:

- org.eclipse.ui.ide.dialogs.ResourceEncodingFieldEditor.findDefaultEncoding()
- org.eclipse.ui.WorkbenchEncoding.getWorkbenchDefaultEncoding()

When pressing the Restore Defaults:

- org.eclipse.ui.ide.dialogs.ResourceEncodingFieldEditor.loadDefault()

Unfortunately when debugging it sets it to UTF-8 which is not what I see in actual IDE when I do that...

image

Anyway, more debugging is required.  We need to be cautious that this seemingly innocent change doesn't actually break the logic that @iloveeclipse  has carefully implemented and tested...

PR #1463, opened 2024-07-04
FileSystemResourceManager: Create project directory before .project file

sratz:
Follow-up on #1446 to avoid `mkdir()`-retry and minimize incompatibility with non-standard `IFileSystem` implementations.


jukzi (commented):
would be OK for me, but mkdir better placed at line 701 (before workspace.createResource(descriptionFile, false);) Avoids additional condition check. The condition is almost the same, just in case of
 Project.move(... REPLACE) in org.eclipse.core.tests.resources.IProjectTest.testReplaceLocation()
or org.eclipse.core.internal.resources.ResourceTree.standardDeleteProject(IProject, int, IProgressMonitor) during org.eclipse.core.tests.resources.WorkspaceTestRule.after()
 the folder already exists and you call mkdir needlessly.

sratz:
> would be OK for me, but mkdir better placed at line 701 (before workspace.createResource(descriptionFile, false);) Avoids additional condition check. 

Done

jukzi (approved):

PR #1474, opened 2024-07-21
Various minor code clean-ups and modernizations in Resources plugin

HannesWell:
- Use `Arrays.sort()` instead of custom quick-sort implementation
- Rely on monitor accepting methods that a null-monitor can be handled
- `Platform.getOS()` can now be used in non-OSGi environments

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/UnifiedTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/UnifiedTree.java
@@ -22,14 +22,28 @@ package org.eclipse.core.internal.localstore;
 
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.regex.Pattern;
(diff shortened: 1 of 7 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/UnifiedTree.java
```diff
@@ -221,11 +228,11 @@ protected void addChildren(UnifiedTreeNode node) {
 			addChildrenMarker();
 	}
 
-	protected void addChildrenFromFileSystem(UnifiedTreeNode node, IFileInfo[] childInfos, int index) {
+	protected void addChildrenFromFileSystem(UnifiedTreeNode node, List<IFileInfo> childInfos, int index) {
 		if (childInfos == null)
 			return;
-		for (int i = index; i < childInfos.length; i++) {
-			IFileInfo info = childInfos[i];
+		for (int i = index; i < childInfos.size(); i++) {
```
merks, line 234: A counted loop with n calls to size doesn’t look like a nice cleanup.

HannesWell: That's right. The 'problem' here is that the loop starts at index and not zero.
But we can instead create a `subList()` here:
`for (IFileInfo info : childInfos.subList(index, childInfos.size())) {`

PR #1476, opened 2024-07-22
[FS] Remove JNI based LocalFileHandler for Windows and its filesystem.win32 fragment

HannesWell:
The new native `Win32Handler` is as fast as the JNI based implementation (see https://github.com/eclipse-platform/eclipse.platform/pull/1422#issuecomment-2184161944), but simpler to maintain because it only uses JNA.
Therefore this suggests to remove the more complex JNI based `LocalFileHandler` and the `org.eclipse.core.filesystem.win32.x86_64` fragment that carries the compiled native library.

One can already test the resulting behavior by launching the latest I-build with the VM argument
`-Declipse.filesystem.useNatives=false`.

When this is submitted, the removed fragment should also be removed in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/084430982c2c860916b3ca7e9324481ccb4511ea/eclipse.platform.releng/features/org.eclipse.platform-feature/feature.xml#L220-L224


HannesWell:
Although I would like to have the `Win32Handler` in use for M2 to have more testers, I'm not 100% confident it can handle all corner cases well. And because M2 is soon (on Friday) and I don't want to break that mile-stone, having the old implementation still around but disabled by default seems to be a good compromise for M2. Therefore I have created https://github.com/eclipse-platform/eclipse.platform/pull/1481.
If everything goes well I would like to submit this shortly after M2.

What do you others think?

HannesWell:
Created now https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2200 to remove this fragment from the platform-feature.
Now that M2 has passed I plan to submit this tomorrow, Tuesday evening.

HeikoKlare:
Just a short note: if I am not mistaken, we do not have an M2 build and no M2-tagged commits/I-Builds yet. I am not sure whether or when M2 will be there, but currently the process of "forcing" people to use `Win32Handler` with an opt-in for the old native fragment during M3 period does not seem to be possible yet.

HannesWell:
> Just a short note: if I am not mistaken, we do not have an M2 build and no M2-tagged commits/I-Builds yet. 

Eclipse/Equinox SDK M2 has already been contributed to the SimRel via https://github.com/eclipse-simrel/simrel.build/pull/464.
But I don't know why there was neither an announcement nor a tag being added? All milestones are now 'lightweigt' but I assumed that both is still done.

HeikoKlare:
Thank you for the pointer! I have only looked at tags and the project download page (which still lists M1), as both is still done even with lightweight milesones. Still, the simrel.build repo is a good source for identifying which builds are contributed, so I'll check that in the future.

PR #1481, opened 2024-07-23
Use Win32Handler native file-system handler by default on Windows

HannesWell:
This is done in preparation for the removal of the JNI based native `LocalFileHandler` for Windows. This increases the number of users (and therefore testers) but still makes it easily possible to revert to the old implementation if regressions are found.

With https://github.com/eclipse-platform/eclipse.platform/pull/1476 I proposed to remove the JNI based `LocalFileHandler` and the associated fragment for Windows that carries the compiled native library.
But while I would like to have the Win32Handler in use for M2 to have more testers, I'm not 100% confident it can handle all corner cases well. And because M2 is soon (on Friday) and I don't want to break that mile-stone, having the old implementation still around but disabled by default seems to be a good compromise for now.

If desired, one can revert this change by specifying the VM argument `-Declipse.filesystem.useNatives=true`.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_530868.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_530868.java
@@ -16,7 +16,6 @@ package org.eclipse.core.tests.resources.regression;
 import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createTestMonitor;
 import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeFalse;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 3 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/Bug_530868.java
```diff
@@ -64,8 +63,6 @@ public void setUp() throws Exception {
 	public void testMillisecondResolution() throws Exception {
 		assumeFalse("not relevant on Mac, as it does not have milliseconds resolution", OS.isMac());
 		try {
-			assertTrue("can only run if native provider can be enabled", LocalFileNativesManager.setUsingNative(true));
-
```
HannesWell, line 68: For Windows the new Win32Handler supports millisecond resolution.
And according to this answer https://stackoverflow.com/a/24806116/14542697 respectively https://bugs.openjdk.org/browse/JDK-8177809, millisecond resolution should now also be available through java.nio on Linux.

HannesWell:
Since there have been no remarks, I'll submit this so we have this at least one day before M2.

merks:
It’s great to see things being simplified and streamlined. 🏆

PR #1486, opened 2024-07-30
Simplify SymLink creation in eclipse.core.tests

HannesWell:
As discussed in https://github.com/eclipse-platform/eclipse.platform/pull/755#discussion_r1649782934.


vogella:
@HannesWell I did the update of your PR, I hope that is ok. Your change looked just to good to be lost.

HannesWell:
> I did the update of your PR, I hope that is ok. Your change looked just to good to be lost.

Yes, of course. Thank you for updating it.

Unfortunately there are (still) failing tests with this and I didn't had the time to investigate the reasons (which is why this staled).
If you can have a look, any help is appreciated.

PR #1490, opened 2024-08-05
Add ResourcesPlugin.FAMILY_SNAPSHOT API for DelayedSnapshotJob

iloveeclipse:
DelayedSnapshotJob may in parallel to any other workspace job and triggers resource changes. Clients that rely on resource change events (like JDT core) should have a chance to wait for possible pending resource changes coming from this job.

Added a family to the job & a constant in the ResourcesPlugin.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2716#issuecomment-2268875660


laeubi:
@iloveeclipse should something similar maybe done for API tools as well? I noticed that sometimes it seem to run in conflict with resource change updates.

PR #1492, opened 2024-08-06
Compare View for big files with Line Spacing freezes eclipse

DenisUngemach:
For big files with many differences, the compare view freezed eclipse, if the line spacing was set. The reason were the necessary resources to calculate the spacing in the method getHeightBetweenLines. We have to save the heights in a list and use them to increase the performance.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2143

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
@@ -29,6 +29,7 @@ import java.io.UnsupportedEncodingException;
 import java.lang.reflect.InvocationTargetException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -473,6 +474,12 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 	private boolean copyOperationInProgress = false;
 	private IUndoableOperation copyUndoable = null;
 	private IOperationHistoryListener operationHistoryListener;
+
+	// https://github.com/eclipse-platform/eclipse.platform.ui/issues/2143
+	// calling w.getLineHeight(line) + w.getLineSpacing() +
+	// w.getLineVerticalIndent(line); again and again for the same line is too slow
+	// we therefore introduce a line height cache by viewer
+	private final Map<MergeSourceViewer, List<Integer>> lineHeightsByViewer = new HashMap<>();
 
 	/**
 	 * Preference key for highlighting current line.
@@ -4001,6 +4008,16 @@ public class TextMergeViewer extends ContentMergeViewer implements IAdaptable {
 
 	@Override
 	protected void handlePropertyChangeEvent(PropertyChangeEvent event) {
+
+		// Property changes might change the height of lines,
+		// which means the heights cache should be cleared.
+		// But the actual height change of lines in StyledText does not happen in this
+		// method.
+		// There are other property change listeners for example:
+		// AbstractTextEditor.PropertyChangeListener,
+		// and these handle the height changes of StyledText.
+		lineHeightsByViewer.clear();
+
 		String key= event.getProperty();
 
 		if (key.equals(CompareConfiguration.IGNORE_WHITESPACE)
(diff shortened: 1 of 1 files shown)


team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
```diff
@@ -474,6 +475,12 @@ private static RGB interpolate(RGB fg, RGB bg, double scale) {
 	private IUndoableOperation copyUndoable = null;
 	private IOperationHistoryListener operationHistoryListener;
 
+	// https://github.com/eclipse-platform/eclipse.platform.ui/issues/2143
+	// calling w.getLineHeight(line) + w.getLineSpacing() +
+	// w.getLineVerticalIndent(line); again and again for the same line is too slow
+	// we therefore introduce a line height cache by viewer
```
iloveeclipse, line 481: Should it be a field, or can it be local variable?

Users can edit & save files in compare editor & so may add/remove lines.
Would this map **as field* mean, the line height will be wrong for the same compare editor after changes?

DenisUngemach: It must be a field, else the recalculation as local variable need too much time. 

Because of textChanges, there is a listener in the method: getHeightBetweenLines which reacts on text changes and clears this field completely.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/structuremergeviewer/DiffTreeViewer.java
```diff
@@ -670,6 +673,35 @@ private void internalSetSelection(TreeItem ti, boolean fireOpen) {
 		}
 	}
 
+	private void expandCollapseAction(DoubleClickEvent event) {
```
iloveeclipse, line 676: What is the purpose of this change?

DenisUngemach: I have deleted that sorry for that.
I hate github....

DenisUngemach: This is the next issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1838 
There i create a new pull request.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
```diff
@@ -4204,11 +4211,31 @@ private void updateLines(IDocument d) {
 	 * @param line
 	 * @return
 	 */
-	private int getHeightBetweenLines(MergeSourceViewer tp, int fromLine, int toLine) {
+	int getHeightBetweenLines(MergeSourceViewer tp, int fromLine, int toLine) {
```
BeckerWdf, line 4214:
```suggestion
	private int getHeightBetweenLines(MergeSourceViewer tp, int fromLine, int toLine) {
```

tobiasmelcher (changes requested):
I think the newly introduced cache lineHeightsByViewer needs to be cleared in method handlePropertyChangeEvent for the scenario when the line spacing settings are changed in the preferences.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
DenisUngemach: Sure.

DenisUngemach:
> I think the newly introduced cache lineHeightsByViewer needs to be cleared in method handlePropertyChangeEvent for the scenario when the line spacing settings are changed in the preferences.

This recommendation works: if i change the spacing in the Text Editor preferences, the line drawings will be updated properly.
But the call of handlePropertyChangeEvent doesn't change the spacing in the StyledTexts, , this happens in a completely different layer.

Which means this works, because the call handlePropertyChangeEvent in TextMergeViewer happens before StyledText.setLineSpacing(int) .

The actual change of the line spacing comes from AbstractTextEditor.PropertyChangeListener.propertyChange(PropertyChangeEvent)

In theory this could lead to line drawing bugs, but i wouldn't expect it.

Is this ok for everybody? 
It is also possible set the spacing and other line height properties directly in the TextMergeViewer.handlePropertyChangeEvent , but this is bad design, because then the StyledText-Properties will be modified twice.

It seems there is no StyledText-Listener where i can check whether the properties were changed and clear  lineHeightsByViewer  afterwards.

iloveeclipse:
> Is this ok for everybody?

OK for me. May be a comment could be added to the current listener?

DenisUngemach:
Ok, done.

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
BeckerWdf: "private" is still missing.

BeckerWdf (changes requested):

BeckerWdf (approved):

tobiasmelcher (approved):
thanks a lot Denis. Looks fine from my side.

PR #1510, opened 2024-08-14
Replace custom stream comparison assertion with library call #903

HeikoKlare:
The ResourceTestUtil provides a compareContents() method for comparing the contents of two streams. This method is used in assertTrue() statements to check for equality of contents of two stream. Modern assertion frameworks already provide that functionality and, in addition, also provide proper error messages when the comparison yields differences.

This change removes the compareContents() method and replaces all calls with either direct string comparison if the indirection over a stream became unnecesary or using assertThat().hasContent() or assertThat().hasSameContentAs().

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
@@ -13,11 +13,11 @@
  *******************************************************************************/
 package org.eclipse.core.tests.filesystem;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
 import static org.eclipse.core.tests.harness.FileSystemHelper.getTempDir;
 import static org.eclipse.core.tests.internal.localstore.LocalStoreTestUtil.createTree;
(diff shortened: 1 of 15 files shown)


HannesWell:
Thanks Heiko for this.
For the cases where the method used to obtain the file content is not relevant, you could also use the new `IFile.readString()` method. Then a standard JUnit `assertEquals()` would be sufficient.
I looked a bit into this some time ago, but didn't complete it yet. But if you want to apply this here it would be nice as well.

HeikoKlare:
Thank you, Hannes! Totally makes sense to simply compare strings where possible. I found the changes in the following classes to be exchangeable with string comparisons and updated the PR accordingly:
* SaveableCompareEditorInputTest
* Bug_025457
* LinkedResourceWithPathVariableTest

I think there are more places we might simplify, but at several places `IFile::getContents(false)` is used (i.e., an out-of-sync file is not refreshed) while `IFile::readString()` forces a refresh. Every place needs to be investigated whether this is relevant or not, so I would propose to do this in a separate PR, if desired.

HannesWell:
> I think there are more places we might simplify, but at several places `IFile::getContents(false)` is used (i.e., an out-of-sync file is not refreshed) while `IFile::readString()` forces a refresh. Every place needs to be investigated whether this is relevant or not, so I would propose to do this in a separate PR, if desired.

Sounds good. :)
And yes we have to be careful in this area to not change the meaning of a test. 
Some cases might even explicitly test the `getContent()` method and one should not accidentally replace that in a batch change.

Issue #1488, opened 2024-07-31
LaunchTests.testProcessLaunchWithLongWorkingDirectory fails on Linux since I20240730-1800


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20240730-1800/testresults/html/org.eclipse.debug.tests_ep433I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

Most likely regression from https://github.com/eclipse-platform/eclipse.platform/pull/1477

```
Exception occurred executing command line.

org.eclipse.core.runtime.CoreException: Exception occurred executing command line.
at org.eclipse.debug.core.DebugPlugin.exec(DebugPlugin.java:1015)
at org.eclipse.debug.tests.launching.LaunchTests.startProcessAndAssertOutputContains(LaunchTests.java:155)
at org.eclipse.debug.tests.launching.LaunchTests.testProcessLaunchWithLongWorkingDirectory(LaunchTests.java:150)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
Caused by: java.io.IOException: Cannot run program "java" (in directory "/home/jenkins/agent/workspace/AutomatedTests/ep433I-unit-cen64-gtk3-java17/tmp/junit9299054611657995980/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name/subfolder-with-relativly-long-name"): error=2, No such file or directory
at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1143)
at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1073)
at java.base/java.lang.Runtime.exec(Runtime.java:594)
at org.eclipse.debug.core.DebugPlugin.exec(DebugPlugin.java:1011)
... 3 more
Caused by: java.io.IOException: error=2, No such file or directory
at java.base/java.lang.ProcessImpl.forkAndExec(Native Method)
at java.base/java.lang.ProcessImpl.<init>(ProcessImpl.java:314)
at java.base/java.lang.ProcessImpl.start(ProcessImpl.java:244)
at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1110)
... 6 more
```

iloveeclipse:
@HannesWell : could you please check?

[...]

PR #1511, opened 2024-08-14
Fix test LaunchTests.testProcessLaunchWithLongWorkingDirectory on Linux

HannesWell:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1488 (hopefully).

My only other solution would be to disable the test on Linux.

### DIFF
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchTests.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchTests.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
+import java.nio.file.Files;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/launching/LaunchTests.java
```diff
@@ -146,6 +147,7 @@ public void testProcessLaunchWithLongWorkingDirectory() throws CoreException, IO
 		String[] segments = Collections.nCopies((400 - rootLength) / subPathElementsName.length(), subPathElementsName).toArray(String[]::new);
 		File workingDirectory = tempFolder.newFolder(segments);
 		assertTrue(workingDirectory.toString().length() > 300);
+		Files.createDirectories(workingDirectory.toPath());
 
 		startProcessAndAssertOutputContains(List.of("java", "--version"), workingDirectory, false, "jdk");
```
jukzi, line 152: Please note that "java --version" is wrong a as eclipse also supports j9. See org.eclipse.jdt.internal.launching.StandardVMType.CANDIDATE_JAVA_FILES. org.eclipse.jdt.internal.launching.LibraryInfo.getVersion() can be used to get the version

HannesWell: The idea of this is to just launch any sub-process in a working directory with a path longer than Window's MAX_PATH length limit (~260 chars).
I choose `java --version` since it's a quite short running process and I assume java will be available by default on all test machines. But thinking more about it, that's probably the actual cause of the failure, not the working directory not being existent.

PR #1512, opened 2024-08-14
[FS] Always copy file-attributes using Eclipse's Native(File)Handler

HannesWell:
Files.copy seems to copy file-attributes differently than Eclipse NativeHandler (for files) do on Mac. In order to stay consistent always use the Eclipse way.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1504 (hopefully).

Can someone with access to a Mac verify this in advance together with https://github.com/eclipse-platform/eclipse.platform/pull/1502 reverted locally (otherwise one needs a file larger than 1MiB?

PR #1520, opened 2024-08-19
Migrate FileStoreTest to JUnit-5

HannesWell:
Required for https://github.com/eclipse-platform/eclipse.platform/pull/1512

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2000, 2023 IBM Corporation and others.
+ *  Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


HannesWell:
@HeikoKlare FYI, just in case you wanted to migrate that test-class too.

HannesWell:
> ## Test Results
> 
>  1 653 files   -  81   1 653 suites   - 81   1h 27m 38s ⏱️ - 2m 0s  3 820 tests  - 153   3 793 ✅  - 158   22 💤 ±0  0 ❌ ±0   5 🔥 + 5  12 057 runs   - 459  11 885 ✅  - 468  161 💤  - 2  0 ❌ ±0  11 🔥 +11 
> 
> For more details on these errors, see [this check](https://github.com/eclipse-platform/eclipse.platform/runs/28965423005).

Looks like https://github.com/eclipse-tycho/tycho/issues/630 is still an issue.
Let's see we can just continue to use JUnit-4's `org.junit.Assume` (I think so, since the JUnit-5 pendant `org.junit.jupiter.api.Assumptions` is not referenced anywhere).

HeikoKlare (approved):
Thank you, Hannes!

The changes look fine. The calculation of a random unique file store seems a bit complicated (why not simply use one temp directory per test method and create the few required stores with explicit, test-method-unique names?), but it's also fine as is :-)

As you mentioned, assumptions can, unfortunately not be migrated to JUnit 5 yet. I did not invest time in Tycho support for that so far. Replacing the assumptions should be a simple follow-up task, once I/we/someone finds the time to contribute the support to Tycho.

HannesWell:
> The changes look fine. The calculation of a random unique file store seems a bit complicated (why not simply use one temp directory per test method and create the few required stores with explicit, test-method-unique names?), but it's also fine as is :-)

Yes, that would probably be better. My main goal was to make it work as JUnit-5 parameterized tests for https://github.com/eclipse-platform/eclipse.platform/pull/1512.
Please don't hesitate to create follow-up improvements.

> 
> As you mentioned, assumptions can, unfortunately not be migrated to JUnit 5 yet. I did not invest time in Tycho support for that so far. Replacing the assumptions should be a simple follow-up task, once I/we/someone finds the time to contribute the support to Tycho.

Whilte waiting for the other PR to complete I'm currently looking a bit into it, but the setup in tycho-surefire quite convoluted.

PR #1521, opened 2024-08-21
Adapted the client to use new Browser API

amartya4256:
The contribution allows the client to utilize the new API from SWT Browser to check if the url is the custom text url of the browser.

Contributes to #213

Must be merged only after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1409 is merged


amartya4256:
Closing beacause the API change is not needed anymore

Issue #1524, opened 2024-08-23
Attribute transfer in FileStore can lead to issues when copying file contents into another FileStore


Michael5601:
The commit found at https://github.com/eclipse-platform/eclipse.platform/pull/1475 introduced a change in FileStore's behavior, specifically affecting how file attributes are transferred during file copying operations.

Previously, the OutputStream was closed before transferring file attributes. However, the refactoring has altered this sequence, allowing file attributes to be transferred before the OutputStream is closed. As a result, when copying a file, the FileStore may only create the file after closing the OutputStream. This means that, under the new behavior, there is a risk that file attributes could be transferred to a file that does not yet exist.

A PR to fix this is provided in https://github.com/eclipse-platform/eclipse.platform/pull/1525.

jukzi:
@HannesWell please take a look.

HannesWell:
Thanks a lot @Michael5601 for this report and the associated fi in #1525. Nice work.

PR #1525, opened 2024-08-23
Move transferAttributes outside of try-with-resource-block to prevent premature file closure

Michael5601:
This commit refactors the code by moving the transferAttributes method call outside of the try-with-resources block. Closing the OutputStream prematurely could result in the destination FileStore not creating the file before the file attributes are transferred. Additionally, transferring attributes does not require the streams to remain open.

A TestFileStore and a matching test is provided.

This commit is an adjustment to https://github.com/eclipse-platform/eclipse.platform/pull/1475 and fixes https://github.com/eclipse-platform/eclipse.platform/issues/1524.

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
@@ -55,6 +55,7 @@ import org.eclipse.core.internal.filesystem.local.LocalFileSystem;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.tests.resources.util.WorkspaceResetExtension;
 import org.eclipse.osgi.util.NLS;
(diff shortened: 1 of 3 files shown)


jukzi:
Added tests need to be referenced in a suite so that they run on the build system.

HannesWell:
Good catch @Michael5601, thanks for the report and this fix.
I'll review this in detail this evening but suggest you add the test tests to the existing `FileStoreTest`.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/FileStoreTest.java
```diff
@@ -724,4 +728,30 @@ private boolean verifyTree(IFileStore[] tree) {
...
+					.getStore(new org.eclipse.core.runtime.Path(sourceFilePath.toString()));
+
+			CopyBugFileStore destFileStore = new CopyBugFileStore(destFilePath.toString());
+			sourceFileStore.copy(destFileStore, 0, new NullProgressMonitor());
+
+			byte[] sourceContent = Files.readAllBytes(sourceFilePath);
+			byte[] destContent = Files.readAllBytes(destFilePath);
+
+			assertArrayEquals("The destination file content should match the source file content.", sourceContent,
+					destContent);
```
HeikoKlare, line 751: Can you use `Files.readString()` here instead and then simply use `assertEquals()` to compare the contents? That would also slightly improve error messages, as you get a string comparison instead of a byte array comparison. In my opinion, you could also leave out the message, as it is kind of self explaining when reading the code.

HeikoKlare (commented):
Thank you for the fix, @Michael5601!

Overall, the change looks good to me. The test serves a proper regression test for the case. I have minor comments.

HannesWell (approved):
I have now reviewed this in detail and the fix itself looks perfectly fine.

Just for the test case multiple smaller things could have been done a bit more elegant, for details please see my follow-up commit I just pushed on your branch.
Because we are late in the release process this should be fixed ASAP and therefore I applied the enhancements by myself. If you are interested in details please don't hesitate to ask.

Besides moving the code a bit around to have it closer together, some of the changes are
- Use Path.of() instead of Paths.get()
- Use static `IPath` fatories instead of `org.eclipse.core.runtime.Path`
- Avoid intermediate file-system path Strings
- Use `Files.readString/writeString()` to write Strings without dealing bytes.

As part of that I have also addressed @HeikoKlare's remarks, thanks for them.

Unless there are objections, I'll squash this into one commit and submit this tomorrow before the noon I-build.

merks:
+1  For getting this into a build as soon as possible.

HannesWell:
> +1 For getting this into a build as soon as possible.

Just squashed the two commits without further changes. Because of the latter we don't have to await the build results, which were all green before and can submit this immediately to have it in the I-build about to start in 10min.

PR #1527, opened 2024-08-26
[Debug] Show elapsed Time in Console & Properties

jukzi:
Elapsed duration formated as H:MM:SS.mmm

While process is running console header is updated once per second:
image
image
image


jukzi:
ignoring known failing test AllTests TocLinkChecker.testPdeUserGenerate

akurtakov:
I would appreciate if failing tests are being looked instead of simply ignored.

PR #1529, opened 2024-08-27
CharsetManager.internalGetCharsetFor reuse calculated key

jukzi:
to reduce memory allocations.


jukzi:
idea came from recorded memory allocations during file search:
image

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/CharsetManager.java
```diff
@@ -369,23 +369,25 @@ private String internalGetCharsetFor(Preferences prefs, Preferences derivedPrefs
...
 		// try to find the encoding in regular and then derived preferences
+		String key1 = getKeyFor(resourcePath);
 		if (prefs != null)
-			charset = prefs.get(getKeyFor(resourcePath), null);
+			charset = prefs.get(key1, null);
 		// derivedPrefs may be not null, only if #isDerivedEncodingStoredSeparately returns true
 		// so the explicit check against #isDerivedEncodingStoredSeparately is not required
 		if (charset == null && derivedPrefs != null)
-			charset = derivedPrefs.get(getKeyFor(resourcePath), null);
+			charset = derivedPrefs.get(key1, null);
```
laeubi, line 378: Maybe better put into a method (372-378) then one can use the same call and don't need "key1" and "key2" ...

jukzi: done

jukzi:
will ignore known failing TocLinkChecker.testPdeUserGenerated

PR #1531, opened 2024-08-27
ElementTree#includes: relax synchronized access

jukzi:
lookupCache is declared volatile and it's type DataTreeLookup has all fields final, so they can be used outside the synchronized block. Sadly same pattern does not work for getElementData() as the returned Object is not guaranteed to be thread-safe but some checks can be moved outside the synchronized block.

ElementTree.includes() is heavily used for example during file search.


jukzi:
Without this change Flighrecorder shows 4 sec locks during filesearch in platform workspace:
image

With this change this reduces to 2 sec:
image

However the perfromance benefit its hardly noticable by the user (only a few milli seconds less)

jukzi:
ignoring unrelated failing test "TocLinkChecker.testPdeUserGenerated" - already failed in [IBuild](https://download.eclipse.org/eclipse/downloads/drops4/I20240904-1800/)

PR #1544, opened 2024-09-07
Revert tycho-packaging-plugin config in help.base

akurtakov:
It looks unneeded but might be what causes about.mappings comparator issues. Weirdly enough this is not handled by tycho but instead is manual maven-resources-plugin call in eclipse-platform-parent.


akurtakov:
Pushing directly and starting a new I-build as I won't be able to spend more time on it during the weekend and should hopefully get us back stable builds.

PR #1546, opened 2024-09-09
Move to ubuntu-latest image

laeubi:
cent-os is deprecated now see: https://github.com/eclipse-cbi/jiro-agents/blob/master/README.md


akurtakov:
What do you think about removing the commit changing the test?

laeubi:
> What do you think about removing the commit changing the test?

I have no objections.

akurtakov:
Closing in favor of https://github.com/eclipse-platform/eclipse.platform/pull/1558 which has only first commit.

Issue #1536, opened 2024-09-02
TextMergeView Unhandled IndexOutOfBoundsException


MikOfClassX:
What steps will reproduce the problem?
1. open java class into team/ src compare pane
2. click on src code (even mouse over works)
3. src compare should highlight code "lines" connecting old and new code. But we get an error.

-- Error Details --
eclipse.buildId=4.33.0.20240829-0648
java.version=22.0.1
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=it_IT
Framework arguments:  -product org.eclipse.epp.package.java.product
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -product org.eclipse.epp.package.java.product

This is a continuation of log file C:\eclipse_workspace_rel-8_0\.metadata\.bak_0.log
Created Time: 2024-09-02 09:46:19.979

Exception:
java.lang.IndexOutOfBoundsException: Index 13 out of bounds for length 13
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Unknown Source)
	at java.base/java.util.Objects.checkIndex(Unknown Source)
	at java.base/java.util.ArrayList.get(Unknown Source)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.getHeightBetweenLines(TextMergeViewer.java:4242)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.getDiffUnderMouse(TextMergeViewer.java:2310)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.handleMouseMoveOverCenter(TextMergeViewer.java:2514)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer$HoverResizer.mouseMove(TextMergeViewer.java:2449)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:229)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4326)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1174)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4124)
[...]

Issue #1542, opened 2024-09-06
ERROR: org.eclipse.compare.contentmergeviewer.TextMergeViewer.getHeightBetweenLines(MergeSourceViewer, int, int)


JVS94:
![image](https://github.com/user-attachments/assets/42b02e19-a181-4f5e-b060-5cb3401eede1)
![image](https://github.com/user-attachments/assets/c06c7c38-1fc3-4518-82b0-8f34a8064061)
![image](https://github.com/user-attachments/assets/133ed8a7-bfcb-4612-8162-da40918b8add)

```
Exception Stack Trace:

java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4903)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.custom.StyledText.getLineVerticalIndent(StyledText.java:3788)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.getHeightBetweenLines(TextMergeViewer.java:4211)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.paintSides(TextMergeViewer.java:4441)
	at org.eclipse.compare.contentmergeviewer.TextMergeViewer.lambda$4(TextMergeViewer.java:2147)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:247)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4285)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1160)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1184)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1169)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1540)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4769)
	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:340)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5050)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3671)
[...]

PR #1547, opened 2024-09-09
TextMergeView Unhandled IndexOutOfBoundsException at lines height calc

DenisUngemach:
In the TextMergeViewer the method
getHeightBetweenLines(MergeSourceViewer tp, int fromLine, int toLine) can be called with a higher toLine value than the number of lines in the TextWidget. In this case a default handling has to be used for getting the line height. This must be checked with:
https://github.com/eclipse-platform/eclipse.platform/issues/1542 which is the compare of java methods.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1536

### DIFF
--- a/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.compare/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.compare; singleton:=true
-Bundle-Version: 3.11.100.qualifier
+Bundle-Version: 3.11.200.qualifier
 Bundle-Activator: org.eclipse.compare.internal.CompareUIPlugin
 Bundle-Vendor: %providerName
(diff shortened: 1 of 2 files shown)


tobiasmelcher:
Fix works fine on my machine, IndexOutOfBoundsException from https://github.com/eclipse-platform/eclipse.platform/issues/1536 no longer occurs.

DenisUngemach:
If you only wish to prevent the runtime errors this works.
But i think the painting will fail in complex compare cases.

The actual problem is that there can be a  DocumentRegionOffset, which won't be handled before the PR and after this handling is not clean.

Example:

image


Here the method: getHeightBetweenLines will be called with fromLine: 5, toLine: 6 and the line count is 3. Of course this dumps.

This shift must be taken into consideration, if we want to calculate the height of lines.

DenisUngemach:
I think now this should work.

Christopher-Hermann:
> How is org.eclipse.debug.tests related to compare?

I think a rebase should do the trick

Christopher-Hermann (approved):
Changes look good for me, let's see if the build is okay

BeckerWdf:
@DenisUngemach: Thanks for fixing

PR #1549, opened 2024-09-10
API: add IWorkspace.write(Map<IFile, byte[]> ...)

jukzi:
to create multiple IFile in a batch.

For example during clean-build JDT first deletes all output folders and
then writes one .class file after the other. Typically many files are
written sequentially. However they could be written in parallel if there
would be an API.

This change keeps all changes to the workspace single threaded but
forwards the IO of creating multiple files to multiple threads.

The single most important use case would be JDT's
AbstractImageBuilder.writeClassFileContents()

The speedup on windows is ~ number of cores, when they have hyper
threading.

OutOfMemory is not to be feared as the caller has full control how many
bytes he passes.

---
discussion welcome.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
+++ b/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.21.100.qualifier
+Bundle-Version: 3.22.0.qualifier
 Bundle-Activator: org.eclipse.core.resources.ResourcesPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
@@ -21,6 +21,13 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.filesystem.IFileInfo;
 import org.eclipse.core.filesystem.IFileStore;
@@ -38,6 +45,7 @@ import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.QualifiedName;
@@ -203,7 +211,7 @@ public class File extends Resource implements IFile {
 		}
 	}
 
-	private void checkCreatable() throws CoreException {
+	void checkCreatable() throws CoreException {
 		checkDoesNotExist();
 		Container parent = (Container) getParent();
 		ResourceInfo info = parent.getResourceInfo(false, false);
@@ -211,7 +219,7 @@ public class File extends Resource implements IFile {
 		checkValidGroupContainer(parent, false, false);
 	}
 
-	private IFileInfo create(int updateFlags, SubMonitor subMonitor, IFileStore store)
+	IFileInfo create(int updateFlags, IProgressMonitor subMonitor, IFileStore store)
 			throws CoreException, ResourceException {
 		String message;
 		IFileInfo localInfo;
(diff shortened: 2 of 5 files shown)


jukzi:
CI measured a good performance gain (factor 7) also for linux:
```
!ENTRY org.eclipse.core.tests.resources 1 0 2024-09-11 06:47:50.781
!MESSAGE [_testWritePerformanceBatch_(org.eclipse.core.tests.resources.IFileTest)] setUp
parallel write took:60ms
sequential write took:443ms
```
does the CI really run with so many cores?

jukzi:
Windows verification build 2x:

```
parallel write took:147ms
sequential write took:335ms
```

macOS verification build 3x:

```
parallel write took:94ms
sequential write took:267ms

```
Linux verification build 5x:

```
parallel write took:40ms
sequential write took:206ms
```

HannesWell:
I'm sorry to say this but I wonder if this feature is really worth to become an API and to maintain that.
Are there more places in the SDK besides JDT that would significantly benefit from it? Or if this is something that can't be implemented externally with the same performance?
Couldn't one simply batch the changes with a workspace-runnable and just parallelize the writes within it at own discretion, e.g. by using a parallel stream or latest javas structured concurrency or simple threads?

jukzi:
If you can show how it is possible to actually do two workspace operations at once  - even though they are mutal exclusive and the aliasmanager is not even threadsafe - please do. You can take
https://git.eclipse.org/r/c/platform/eclipse.platform.resources/+/179934 `
Tries to write some files in parallel (two different projects).
Shows that files are NOT written in parallel
`
as a basis of a test to proof it.

mickaelistria:
I think the expectation would be that calling `file.setContents(...)` concurrently would in most cases allow to write files in parallel if they are exclusive.
Instead of a new API method, aren't there some internal bits we could change to allow that?

jukzi:
it's one of the fundamentals that each operation is mutual exclusive:
```
	at org.eclipse.core.internal.jobs.OrderedLock.acquire(OrderedLock.java:87)
	at org.eclipse.core.internal.resources.WorkManager.checkIn(WorkManager.java:125)
	at org.eclipse.core.internal.resources.Workspace.prepareOperation(Workspace.java:2398)
```
And that's appropriate as the listeners can be sure that one operation happens after another and that they are informed in the thread of the caller. Changing that would break the API. 
Also think of a single .java file compiles into multiple .class files - all in the same folder. So it's also the same parent Container that is changed multiple times.

laeubi:
P2 Repository API has a `IMetadataRepository.executeBatch(IRunnableWithProgress, IProgressMonitor)` method that allows batching successive writes into one big operation. I must confess I would have exspected such thing actually happen from WokrspaceRule/Job/...

In any case the speedup might not be because of many threads writing to the same device but more because of the batching of operations as you have noticed that the speedup is not related to the CPU threads.

So maybe something similar would be useful here so one can perform many updates (maybe in parallel threads) and the notifications are then just performed at the end (like in your parallel write API). This would then even not matter what operations are performed, if stream or byte[] are used and so on...

jukzi:
I agree that it would be cool if everything would just work in parallel or to have a Stream Collector like parallelStream().colllect(Workspace::collectFileContents) but as far as i know that is not possible. On the other hand the proposed API is easy to understand, simple to implement, useful and sufficient for the usecase.
It could also be used for other mass file creations like extracting all files of  an archive to workspace but i don't plan to work on such as there is no known hotspot in our daily use.

merks:
I suppose the entire design pattern for this would look pretty much the same (analogous) if in the end it called 
```
FileSystemResourceManager.write(IFile, InputStream, IFileInfo, int, boolean, IProgressMonitor)
```
instead of 
```
FileSystemResourceManager.write(IFile, byte[], IFileInfo, int, boolean, IProgressMonitor)
```
It just feels somehow unfortunate that the more general streaming APIs are not supported in favor of a must narrower use case.

mickaelistria:
> it's one of the fundamentals that each operation is mutual exclusive:

`prepareOperation`s are exclusive, but they're mostly asking/waiting for the workspace whether it's ready to support an operation for the given rule. It returns almost immediately if the requested rule is ready to be processed. It's not blocking the whole workspace while the underlying filesystem operation is running, it's only blocking the requested rule.
So unless rules are conflicting, `prepareOperation` returns promptly and further filesystem changes can be running in parallel. So while internal workspace model changes are exclusive, the filesystem operations are supposed to be able to run in parallel.

Here is an example
```java
	@Test
	public void testParallelWorkspaceOperations() throws Exception {
		IProject project = ResourcesPlugin.getWorkspace().getRoot().getProject("testParallel");
		project.create(null);
		project.open(null);
		List<IFile> _30files = new ArrayList<>(30);
		for (int i = 0; i < 30; i++) {
			IFile file = project.getFile("file" + i);
			file.create(new byte[0], 0, null);
			_30files.add(file);
		}
		Instant before = Instant.now();
		List<CompletableFuture<?>> futures = new ArrayList<>(30);
		long timeToProcessMs = 2000;
		for (IFile file : _30files) {
			CompletableFuture<?> future = CompletableFuture.runAsync(() -> {
				try {
					file.getWorkspace().run(monitor -> {
						try {
							Thread.sleep(timeToProcessMs);
						} catch (Exception ex) {
							throw new RuntimeException(ex);
						}
					}, file, 0, null);
				} catch (Exception ex) {
					throw new RuntimeException(ex);
				}
			});
			futures.add(future);
		}
		CompletableFuture.allOf(futures.toArray(CompletableFuture[]::new)).join();
		Duration duration = Duration.between(before, Instant.now());
		System.err.println(duration.getSeconds());
		assertTrue(duration.toMillis() < timeToProcessMs * _30files.size());
	}
```
You'll see that things perform well: the `Thread.sleep()` are running in parallel, on my machine the test takes 10s for 60s of actual wait. The `Thread.sleep()` could equally be a `Files.write()` operationt taking 2s. The workspace can parallelize operation, the important part is the scheduling rule.
A similar implementation can be achieved with WorkspaceJobs which allow a scheduling rule too.

JDT could be improved to take advantage of this without a new API. A new API might still be convenient, but I don't think it's necessary for this case. Moreover, the suggested implementation uses a MultiRule of all files, which would prevent any file from being modified is one is already referenced by the active rule. Going finer grain with the smallest possible rule (1 file) would be less blocking than creating and locking bigger rules.

jukzi:
> FileSystemResourceManager.write(IFile, InputStream,

@merks ok, so you suggest to have a IWorkspace.write(Map<IFile, InputStream> ..) as well? Would be doable - But who would use that? If the client cant convert the InputStreaminto a byte array because of memory issues why he should try to write mutliple files at once?

mickaelistria:
Thanks for trying with actual file.write.

> IFile.write does NOT forward the IO in parallel!

IMO, this is where investigation should be focused. Do you know what prevents parallel IO works here (now that we've established that begin/prepareOperation are not to blame)?

jukzi:
> begin/prepareOperation are not to blame

Each individual write has its's own prepareOperation. Only with a batch API that could be solved. see  https://github.com/eclipse-platform/eclipse.platform/pull/1549/files#diff-1fb86c69fb8845944ed1e086cf8b5b3b6ee725ae01d5bfb3ad8070cdc87e69bdR2852

mickaelistria:
> Each individual write has its's own prepareOperation

OK, I see that with `workspace.run(..., rule, ...)` there is a `getWorkManager().beginUnprotected();` which allows to release the lock taken by prepareOperation/checkIn; and that's why file.setContents(...) cannot be running in parallel.
So the underlying workspace tree requires a lock is actually not concurrent, despite existence of scheduling rules ensuring that changes are not conflicting... That would be a good, but challenging, thing to improve!

jukzi:
So if nobody comes up with a better alternative i would like to continue with this.

jukzi:
Do you miss Streaming of the IFile (Stream) or Streaming of the content (InputStream)- or both? it would be simple to provide an API
`IWorkspace.write(Stream<Map.Entry<IFile, InputStream>> contentStream, ... )`
As it can be converted to call the proposed API. However i don't know how it could be efficiently implemented in parallel, as to get the workspace lock it needs to know all IFile in advance within the calling thread.
Streaming only the content as in `IWorkspace.write(Map<IFile, InputStream> ..)` could be easily implemented as in the proposed API.

merks:
I feel a bit sheepish because I know you understand this stuff much more deeply than do I and have invested way more effort than I do with my superficial observations...

Mostly my observation is that these are the ways to update the contents:

image

And that the new API supports only parallelizing the byte[] versions but not the InputStream versions.  (I've never noticed the IFileState versions before.)    I think supporting InputStream would be sensible and somehow more uniform and as you say, it looks very similar to what you already did.  I don't know how others feel about that...

In any case, I will not block whatever decision you make.  And thanks for all the effort to make Eclipse faster!

jukzi:
I would like to get feedback if resources should have it's own ExecutorService for the parallelism or if that should be passed as parameter. And if resources creates an ExecutorService on it's own if it should be kept static or created per request. 

Pros for passing as parameter: 
* the caller can decide how many threads should be used (JDT for example has other threads running in parallel and would not like them to be thwart

Cons for having a static ExecutorService: 

- the calling thread might have another priority then then the ExecutorService Threads, so writing the files might get too high or too low priority.

merks:
Everyone is so busy.  With respect to executors, there are indeed pros and cons of of course.   If one needs to pass it in, then each framework that uses this will create their own executor and then one might end up with many, which would be kind of wasteful.   Creating one on each request is of course additional overhead and potentially costly per call, but no long term waste.  I didn't even think of the priority thing.   I don't have a good sense of what's best.  😞

PR #1553, opened 2024-09-15
Bump version to match jar signature of org.eclipse.compare

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2335


merks:
Windows testing appears to be stalled.  Given all the tests pass elsewhere and that tests are highly unlikely to be OS and bundle version dependent, I will merge this now.

PR #1557, opened 2024-09-17
ProcessConsole: skip milliseconds while running.

jukzi:
it's updating only once a second anyway
image
image


HannesWell (approved):
Agree.
Having milliseconds made updates feel a bit noisy.

Thanks.

PR #1566, opened 2024-09-23
Support multiple versions of annotation classes in E4 Injector

### LINKED ISSUE #1565: @PostConstruct not working
`org.eclipse.ui.tests.api.workbenchpart.DependencyInjectionViewTest.testDependencyInjectionLifecycle()` fails locally.
Enabling debug output in `org.eclipse.e4.core.internal.di.InjectorImpl.processAnnotated(AnnotationProxy, Object, Class<?>, PrimaryObjectSupplier, PrimaryObjectSupplier, List<Class<?>>)` reveals
annotated with
```
 "jakarta.annotation-api:3.0.0:jakarta.annotation.PostConstruct" but was looking for 
 "jakarta.annotation-api:2.1.1:jakarta.annotation.PostConstruct or jakarta.annotation-api:1.3.5:javax.annotation.PostConstruct"
```

I think it should not matter which API version is used. Only the qualified name` jakarta.annotation.PostConstruct` should matter

@HannesWell ?

HannesWell:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1565


HannesWell:
@merks is there a reason to not update to [`jakarta.annotation` version 3](https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api/3.0.0) respectively to provide both to allow easy migration?

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/e912d230f7dc8715d5616232236aa60c5bab7898/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target#L805-L810

Of course an application ideally only uses one version and I don't know what happens if two versions are available respectivly if the E4 injector really works then.

merks:
Supporting a wider range is good. I don’t see why both 2.x and 3.x need to be in the tp. Some other 3rd party dependency might need the 2.x. I’ll have to check with the aggregation analyzer.

laeubi:
> Of course an application ideally only uses one version and I don't know what happens if two versions are available respectivly if the E4 injector really works then.

Having multiple versions is supported by OSGi and the OSGi resolver should take care of it but the injector will only be able to handle "compatible" bundles then so all have to use 3.x or 2.x you can't mix them (in the current form). So if currently a "client" uses a narrow version range but the injector is bound to a higher range it wont work.

Eclipse Sisu semm to use a quite interesting aproach where they just look at the annotation class name and accept everything that ends with Inject e.g. `com.google.Inject`, `javax.annotation.Inject` and `jakarta.annotation.Inject` can be used together in any version.

merks:
There are a lot of upper bounds that exclude 3.0.0 which is to be expected because goodness knows what 3.0.0 breaks, which is nothing for this use case, but one doesn't know that before there is a 3.0.0.  

image

So I expect the downstream ecosystem will be in a similar state.  Certainly better if only the package name and the package version specified for it via its exporting bundle mattered, regardless of the bundle symbolic name and bundle version.

jukzi (dismissed):
This does not fix DependencyInjectionViewTest for me locally. Instead it crashes with:
```
MESSAGE bundle org.eclipse.pde.ds.tck:1.0.0.qualifier (397) Component descriptor entry 'OSGI-INF/testServiceSingleton.xml' not found
WARNING: Annotation classes from the 'javax.inject' or 'javax.annotation' package found.
It is recommended to migrate to the corresponding replacements in the jakarta namespace.
The Eclipse E4 Platform will remove support for those javax-annotations in a future release.
To suppress this warning, set the VM property: -Declipse.e4.inject.javax.warning=false
To disable processing of 'javax' annotations entirely, set the VM property: -Declipse.e4.inject.javax.disabled=true


!ENTRY org.eclipse.e4.core.di 2 0 2024-09-24 09:19:11.740
!MESSAGE Possbible annotation mismatch: method "void org.eclipse.e4.ui.internal.workbench.ResourceHandler.init()" annotated with "jakarta.annotation-api:2.1.1:jakarta.annotation.PostConstruct" but was looking for "jakarta.annotation-api:3.0.0:jakarta.annotation.PostConstruct
 or jakarta.annotation-api:1.3.5:javax.annotation.PostConstruct"

!ENTRY org.eclipse.e4.ui.workbench 4 0 2024-09-24 09:19:11.744
!MESSAGE Unable to load resource platform:/plugin/org.eclipse.platform/LegacyIDE.e4xmi
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.emf.ecore.resource.ResourceSet.getResource(org.eclipse.emf.common.util.URI, boolean)" because "this.resourceSet" is null
	at org.eclipse.e4.ui.internal.workbench.ResourceHandler.getResource(ResourceHandler.java:287)
	at org.eclipse.e4.ui.internal.workbench.ResourceHandler.loadResource(ResourceHandler.java:263)
	at org.eclipse.e4.ui.internal.workbench.ResourceHandler.loadMostRecentModel(ResourceHandler.java:178)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Application.loadApplicationModel(E4Application.java:368)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Application.createE4Workbench(E4Application.java:244)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:568)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.pde.internal.junit.runtime.NonUIThreadTestApplication.start(NonUIThreadTestApplication.java:58)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:271)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:668)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:605)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1454)

```

instead of having the Annotations as dependency for using `java.lang.reflect.AnnotatedElement.isAnnotationPresent(Class<? extends Annotation>)`
we could simply check the name like:
`Arrays.stream(element.getDeclaredAnnotations()).anyMatch(a -> a.annotationType().getName().equals(annotationClass.getName()))`
in `org.eclipse.e4.core.internal.di.AnnotationLookup.AnnotationProxy.isPresent(AnnotatedElement)` - which solves the test for me

HannesWell:
> This does not fix DependencyInjectionViewTest for me locally. Instead it crashes with:

Looks like what I have suspected in https://github.com/eclipse-platform/eclipse.platform/issues/1565#issuecomment-2368924990.



> instead of having the Annotations as dependency for using `java.lang.reflect.AnnotatedElement.isAnnotationPresent(Class<? extends Annotation>)`
> we could simply check the name like:
> `Arrays.stream(element.getDeclaredAnnotations()).anyMatch(a -> a.annotationType().getName().equals(annotationClass.getName()))`
> in `org.eclipse.e4.core.internal.di.AnnotationLookup.AnnotationProxy.isPresent(AnnotatedElement)` - which solves the test for me

Yes, almost. We just have to use `getAnnotations()` instead of `getDeclaredAnnotations()`. The method `AnnotatedElement.isAnnotationPresent(Class<? extends Annotation>)` also operates on the former.

I just pushed an update.

> Eclipse Sisu semm to use a quite interesting aproach where they just look at the annotation class name and accept everything that ends with Inject e.g. `com.google.Inject`, `javax.annotation.Inject` and `jakarta.annotation.Inject` can be used together in any version.

I have to say that I wouldn't be comfortable with that since then even more false positives are possible. This might be only hypothetical, but on the other hand it wouldn't be too difficult to support more annotations if really desired. 

> There are a lot of upper bounds that exclude 3.0.0 which is to be expected because goodness knows what 3.0.0 breaks, which is nothing for this use case, but one doesn't know that before there is a 3.0.0.

Thanks for the analysis. That's what I have expected and I'm actually happy that proper version-ranges are used, even if this makes migration in this case a bit harder than it would have been without.

laeubi:
> I have to say that I wouldn't be comfortable with that since then even more false positives are possible. This might be only hypothetical, but on the other hand it wouldn't be too difficult to support more annotations if really desired.

What means "even more"? Currently we have not nay false positive at all (only false negatives).

runtime/bundles/org.eclipse.e4.core.di/META-INF/MANIFEST.MF
```diff
@@ -15,7 +15,7 @@ Export-Package: org.eclipse.e4.core.di;version="1.7.0",
  org.eclipse.e4.core.internal.di.osgi;x-internal:=true,
  org.eclipse.e4.core.internal.di.shared;x-friends:="org.eclipse.e4.core.contexts,org.eclipse.e4.core.di.extensions.supplier"
 Require-Bundle: org.eclipse.e4.core.di.annotations;bundle-version="[1.4.0,2.0.0)";visibility:=reexport
-Import-Package: jakarta.annotation;version="[2,3)",
+Import-Package: jakarta.annotation;version="[2,4)",
```
jukzi, line 18: Do we need any import at all? Should be possible to use class names only as String.

HannesWell: Initially I wanted to keep this change minimal, but if this is started it should be done completely.
So in theory yes, but while implementing it I noticed that obtaining the values of `jakarta/javax.inject.Named` is not trivial because I think we want a more performant solution than using reflection all the time.
I think I'll try it with cached MethodHandles or alike.

laeubi: > I think we want a more performant solution than using reflection all the time

Have you measured the performance impact? Injection frameworks always require reflection and the JVM should be quite good at that so I'm curious if it is really a performance impact (in practical application).

merks: My suspicion would be that in the overall scheme of things the performance here is not relevant (especially base I've debugged all the injection stuff that happens a framework like Xtext that's using guice).

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,14 +42,15 @@ public class AnnotationLookup {
 	private AnnotationLookup() {
 	}
 
-	public static record AnnotationProxy(List<Class<? extends Annotation>> classes) {
-		public AnnotationProxy {
-			classes = List.copyOf(classes);
+	static record AnnotationProxy(List<Class<? extends Annotation>> classes, List<String> classNames) {
```
jukzi, line 45: Has to be public to make `org.eclipse.e4.core.internal.services.MessageFactoryServiceImpl.processPostConstruct(Object, Class<?>)` compile.
- i would like to see this PR merged soon. Even if it may not be the final solution it still helps to solve the current situation

HannesWell: Yes.

> Even if it may not be the final solution it still helps to solve the current situation

The problem with the first approach was that it only worked in multiple versions for some kind of annotations and not for all kind of annotations processed by the injector. And while partial progress is certainly better than no progress inconsistent behavior can be really bad.

But I think I have found a good approach now that works with all kind of annotation and I hope to complete it within the next days.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,129 +48,130 @@ public class AnnotationLookup {
...
-	private static List<Class<?>> getAvailableClasses(Class<?> jakartaClass, Supplier<? extends Class<?>> javaxClass) {
-		List<Class<?>> classes = new ArrayList<>();
-		classes.add(jakartaClass);
-		if (javaxClass != null) {
-			loadJavaxClass(() -> classes.add(javaxClass.get()));
-		}
-		return classes;
+	private static final Map<Class<? extends Annotation>, Function<Annotation, String>> NAMED_ANNOTATION2VALUE_GETTER2 = new ConcurrentHashMap<>();
+	private static final Set<String> NAMED_ANNOTATION_CLASSES = Set.of("jakarta.inject.Named", "javax.inject.Named"); //$NON-NLS-1$//$NON-NLS-2$
+	// TODO: warn about the javax-class?
```
jukzi, line 161: remove TODO

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,129 +48,130 @@ public class AnnotationLookup {
...
+
+	private static List<String> getAvailableClasses(String jakartaClass, String javaxClass) {
+		return javaxClass != null && canLoadJavaxClass(javaxClass) //
```
jukzi, line 164: why is it needed to check if javax is loadable?

HannesWell: This methods checks if the javax-classes are even available in the application in any version and if it's therefore worth to check for them. And, maybe even more important, it emits the warning about javax annotations being 'deprecated'.

But in the past it was asked to improve the implementation because the message is to coarse. So this can definitively improved. OTOH, if support for javax annotation does not require their classes in the build-system we can probably support it much easier and longer and it could be consider to un-deprecate their support.

merks: If it can be supported by reflection for very little ongoing maintenance overhead, that seems like a good thing for the consumers.

HannesWell:
From https://github.com/eclipse-platform/eclipse.platform/pull/1566#issuecomment-2370138739:
> Eclipse Sisu semm to use a quite interesting aproach where they just look at the annotation class name and accept everything that ends with Inject e.g. `com.google.Inject`, `javax.annotation.Inject` and `jakarta.annotation.Inject` can be used together in any version.

I have now implemented the approach purely look at only the fully qualified names and Provider Instances are now created using `java.lang.reflect.Proxy`.

Plus this also removes the warning about `javax.inject`/`javax.annotation` being present in the application and the associated deprecation warning. Supporting the javax-annotations is now just a matter of listing their names as Strings and the actual classes don't have to be provided anymore. Therefore I think we don't have to remove support for it in the (foreseeable future). And as far as I know, it just annoyed most people as in face it can be really difficult to remove the javax-annotations entirely.

An important question for me is: Should we submit this (given everyone is fine with this change) at this point in the release cycle, since the Eclipse TLPs are now moved to jakarta.annotations 3 (via https://github.com/eclipse-platform/eclipse.platform.ui/issues/3948) or should this better wait until next development cycle?

merks:
I'd rather see this finished sooner given we are now shipping a newer version of jakarta.annoation-api.

@akurtakov 

Do you have an opinion?

akurtakov:
I haven't reviewed the changes but the idea is sound so +1 from me to continue on that path.

HannesWell:
> I'd rather see this finished sooner given we are now shipping a newer version of jakarta.annoation-api.

Agree. So unless there are objections or requests for changes, I'll plan to submit this Monday (European) evening.

laeubi:
+1 for doing it now - otherwise it would possibly break users that are using a `[2,3)` version range (what was legal already).

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,154 +46,103 @@ public class AnnotationLookup {
...
-		@SuppressWarnings({ "rawtypes", "unchecked" })
-		List<Class<? extends Annotation>> annotationClasses = (List) classes;
-		return new AnnotationProxy(annotationClasses);
-	}
+	static final AnnotationProxy INJECT = createProxyForClasses(//
+			"jakarta.inject.Inject", "javax.inject.Inject"); //$NON-NLS-1$ //$NON-NLS-2$
+	static final AnnotationProxy SINGLETON = createProxyForClasses(//
+			"jakarta.inject.Singleton", "javax.inject.Singleton"); //$NON-NLS-1$//$NON-NLS-2$
+	static final AnnotationProxy QUALIFIER = createProxyForClasses( //
+			"jakarta.inject.Qualifier", "javax.inject.Qualifier"); //$NON-NLS-1$//$NON-NLS-2$
```
laeubi: Just wondering since you have mentioned the sisu approach, should this not be

```
AnnotationProxy INJECT = createProxyForClasses("Inject");
AnnotationProxy SINGLETON = ...
```

HannesWell: Sorry, I didn't read your comment in detail again with my old memory from about a year ago I just remembered the (fully-qualified) name based approach.
To me just looking at the simple name seems too lax. And I wonder if there would be a strong benefit?
Besides the jakarata and javax annotations I'm only aware of the Guice pendants.
If we want to support them, this they can be added easily (I'd suggest subsequently).
I'm not sure if we need, want or should support more or even anything custom?

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,154 +46,103 @@ public class AnnotationLookup {
...
-			// Attempt to load the class early in order to enforce an early class-loading
-			// and to be able to handle the NoClassDefFoundError below in case
-			// javax-Provider is not available in the runtime:
-			factory.create(null, null, null);
-		} catch (NoClassDefFoundError e) {
-			factory = ProviderImpl::new;
-		}
-		PROVIDER_FACTORY = factory;
+	private static final Set<String> PROVIDER_TYPES = Set.of( //
+			"jakarta.inject.Provider", "javax.inject.Provider"); //$NON-NLS-1$//$NON-NLS-2$
```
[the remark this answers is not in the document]
HannesWell: The flag `-Declipse.e4.inject.javax.disabled` was mainly intended to allow consumers to test their application's behavior after a future removal of javax-support and to spot remaining usages easier.
But since javax-support is now much easier, it will probably stay for longer/ever.

And since the warning is now removed, `-Declipse.e4.inject.javax.warning=false` is effectively enabled all the time.
I assume almost everybody is happy about one less warning.

So this should be fine as it is.

runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/AnnotationLookup.java
```diff
@@ -42,154 +46,103 @@ public class AnnotationLookup {
...
+			"jakarta.inject.Singleton", "javax.inject.Singleton"); //$NON-NLS-1$//$NON-NLS-2$
+	static final AnnotationProxy QUALIFIER = createProxyForClasses( //
+			"jakarta.inject.Qualifier", "javax.inject.Qualifier"); //$NON-NLS-1$//$NON-NLS-2$
 
-	private static final List<Class<?>> PROVIDER_TYPES = getAvailableClasses(jakarta.inject.Provider.class,
-			() -> javax.inject.Provider.class);
+	static final AnnotationProxy PRE_DESTROY = createProxyForClasses( //
+			"jakarta.annotation.PreDestroy", "javax.annotation.PreDestroy"); //$NON-NLS-1$//$NON-NLS-2$
+	public static final AnnotationProxy POST_CONSTRUCT = createProxyForClasses( //
+			"jakarta.annotation.PostConstruct", "javax.annotation.PostConstruct"); //$NON-NLS-1$//$NON-NLS-2$
```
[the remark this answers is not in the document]
HannesWell: I'm not aware of a relatively easy way to test this without adding a lot of (test-) code.
But since the code relies only on class names I think it's 'functional by design'.
Plus the javax tests already verify the injector can handle annotations from packages that are not imported/required.
Additionally I'm also working on removing even the imports of `jakarta.annotation` and `jakarta.inject` (which is currently only due to java-doc references and one pseudo link (which hopefully can be avoided):
https://github.com/eclipse-platform/eclipse.platform/blob/7c138cf3bfbd4bba353d217bb7c38f40d041b3f1/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/ClassRequestor.java#L32-L34

HeikoKlare:
+1 for the change and for doing it now from my side as well.

If I am not mistaken, we have the duplicated test bundle (`org.eclipse.e4.core.tests` and `org.eclipse.e4.core.javax.tests`) for testing both jakarta and javax annotations. The javax copy was intended to be dropped once javax support is removed. Now as the support will stay (and is even extended to further annotation types), should we remove the duplicated bundle? It of course ensures that javax annotations still work, but we do not test support for any other annotations either, and the question is when we would explicitly say that javax annotations are not different from any other kind of annotations such that explicit testing is not required anymore and the duplicated bundle can be removed. Maybe instead of duplicating the complete bundle with every single test, a single regression test for support of different kinds of annotations would be sufficient?
That's of course something we can do indepedent from this PR (and the PR can be merged without an agreement on that), but in my opinion it would be the right time to clean up the duplicated test bundle.

HannesWell:
> If I am not mistaken, we have the duplicated test bundle (`org.eclipse.e4.core.tests` and `org.eclipse.e4.core.javax.tests`) for testing both jakarta and javax annotations. The javax copy was intended to be dropped once javax support is removed. Now as the support will stay (and is even extended to further annotation types), should we remove the duplicated bundle? [...] Maybe instead of duplicating the complete bundle with every single test, a single regression test for support of different kinds of annotations would be sufficient?

Yes, that's a good point. I'll follow up on that.
At the moment I would say that the javax variant should stay as test bundle to ensure dependency separation, but should be stripped to a minimum (one smoke test per annotation).
Let's discuss the details in a corresponding PR.


> Now as the support will stay (and is even extended to further annotation types

To be precise, at the moment the same annotation (packages) are supported as before, only in multiple versions. But it would now be simpler to extend the list of supported packages in the future, if we want to. 

Since there are only approvals, this is ready. 
Thanks everyone for your participation.

Issue #1461, opened 2024-07-03
Improve File delete performance


jukzi:
Currently Project/Clean... takes much time to only delete the output. Hopefully that can be improved - somehow

![image](https://github.com/eclipse-platform/eclipse.platform/assets/51790620/c0deaf84-7f5b-4eba-bf0a-eba2884ca7b6)

HannesWell:
I wonder why RemoveDirectory is called if the element to be deleted is actually a file?

In general if the JDK implementation is too generic and therefore too slow we could consider adding a native (JNA, later FFM based) implementation by adding a `delete()` method to the NativeHandler? Maybe we can avoid some of the native calls done in nio.

jukzi:
i have made delete parallel and don't plan further improvements.

PR #1569, opened 2024-09-26
[performance] LocalFile: delete Files of Directories in parallel #1461

jukzi:
There is no honest progress report as before. Just fixed the
InfiniteProgress to be thread safe and honestly using it.

https://github.com/eclipse-platform/eclipse.platform/issues/1461


jukzi:
review  welcome

Issue #102, opened 2022-04-21
Use Declarative Services annotations in ResourceChangeListenerRegistrar


laeubi:
Currently there is one manually crafted DS component, we should convert this (and probably further DS components) to use the annotation support, this makes it much more convenient and clear:

- no need to leave the java editor to change the component
- no need for an additional editor
- xml is auto generated
- no risk of breaking in case of refactoring
- content sensitive error messages

PR #1573, opened 2024-09-30
Generate all OSGi Declarative Services component files by PDE-DS/Tycho

HannesWell:
Remove the now generated OSGi Declarative Services component-xml files from git.

This also fixes https://github.com/eclipse-platform/eclipse.platform/issues/102.

@laeubi do you want to review this? In some cases I set a few explicit configuration properties to match the existing component file, even when I wasn't certain it is necessary.


HannesWell:
This also fixes one build warning about Tycho not being able to check wildcards in DS component declarations in manifests.

PR #1574, opened 2024-09-30
Remove unnecessary specific entries in build.properties

HannesWell:


debug/org.eclipse.debug.examples.core/build.properties
```diff
@@ -22,6 +22,5 @@ bin.includes = plugin.xml,\
                .,\
                readme.html,\
                samples/,\
-               OSGI-INF/l10n/bundle.properties,\
```
Bananeweizen, line 25: BTW: I have a huge amount of regex based checkstyle rules at work for detecting this and similar bad things in eclipse specific files. If anyone is interested in applying them here, ping me.

laeubi: It is for sure obsolete but I won't call it bad :-)

In any case if you like feel free to open PRs to cleanup things, I think there is no need to specifically ask for it.

HannesWell: > It is for sure obsolete but I won't call it bad :-)

Technically it's not bad, but for the maintenance of the code I have the opinion that every thing that's unnecessary is bad, even if it's not changed. If it exists developers will spend their precious time wondering why it's there from time to time. 

> In any case if you like feel free to open PRs to cleanup things, I think there is no need to specifically ask for it.

Absolutely :)

PR #1575, opened 2024-10-01
Add constant for riscv64 architecture in Platform class

HannesWell:
As @merks pointed out, we should add a `riscv64` constant in the Platform class, that that this architecture is now supported on Linux.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2310
@yuzibo FYI

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
@@ -410,6 +410,15 @@ public final class Platform {
 	public static final String ARCH_AARCH64 = "aarch64";//$NON-NLS-1$
 
 	/**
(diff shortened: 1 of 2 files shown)


runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
```diff
@@ -409,6 +409,15 @@ public static boolean isMac() {
 	 */
 	public static final String ARCH_AARCH64 = "aarch64";//$NON-NLS-1$
 
+	/**
+	 * Constant string (value {@code aarch64} indicating the platform is running on
```
HannesWell, line 413: The value of course should have been `riscv64`. This is fixed in
- https://github.com/eclipse-platform/eclipse.platform/pull/1578

PR #1576, opened 2024-10-01
Deprecate ARCH_X86 and WS_WPF constants and constant for ppc64le architecture in o.e.core.runtime.Platform

HannesWell:
As far as I can tell x86 32-bit is not supported anymore for any OS just like WPF on Windows.

Or can any tell if it's different?

Additionally show constant values in `@code` elements.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
@@ -418,6 +418,14 @@ public final class Platform {
 	public static final String ARCH_AARCH64 = "aarch64";//$NON-NLS-1$
 
 	/**
+	 * Constant string (value {@code ppc64le}) indicating the platform is running on
+	 * an little-endian PowerPC 64bit based architecture.
+	 *
+	 * @since 3.32
+	 */
+	public static final String ARCH_PPC64LE = "ppc64le"; //$NON-NLS-1$
+
+	/**
 	 * Constant string (value {@code riscv64} indicating the platform is running on
 	 * an RISC-V 64bit-based architecture.
 	 *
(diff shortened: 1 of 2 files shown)


HannesWell:
Besides deprecating the not anymore supported arch `x86` and window-system `wpf`, this now also adds the missing constant for the supported `ppc64le` arch.

Furthermore this also updates the list of known arch and WS returned by the Platform to the values that are actually supported nowadays.
@akurtakov or anybody else interested in this, do you want to have a look at this?

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java
```diff
@@ -410,7 +412,15 @@ public static boolean isMac() {
 	public static final String ARCH_AARCH64 = "aarch64";//$NON-NLS-1$
 
 	/**
-	 * Constant string (value {@code aarch64} indicating the platform is running on
+	 * Constant string (value {@code ppc64le}) indicating the platform is running on
+	 * an AARCH64bit-based architecture.
```
akurtakov, line 416: I don't think this AARCH64

HannesWell: Yes, that's a mistake from https://github.com/eclipse-platform/eclipse.platform/pull/1575#discussion_r1785000606.

akurtakov: Just to be sure you fixed the value but added AARCH64bit-based (wrong!) where it should be PPC64LEbit-based in this commit.

HannesWell: Ah, yes. I oversaw that. Thanks.
Fixed it now.

Bananeweizen:
Just wondering if even such fundamental constants should be tagged "forRemoval", with a longer time span. Do we have some guidance why that was not done here, or is it just forgetting that this is meanwhile possible? :)

PR #1582, opened 2024-10-06
[Oomph-Setup] Add eclipse.platform configuration setup

HannesWell:
Additionally add a styled and drag&drop-able Oomph Configuration button.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2430

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,11 @@ Update provides Java interfaces like `IPlatformConfiguration`. `IPlatformConfigu
 
 Contributions are most welcome. There are many ways to contribute, from entering high quality bug reports, to contributing code or documentation changes.
 
-For a complete guide, see the https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md.
+For a complete guide, see the [CONTRIBUTING](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md) page.
 
+[![Create Eclipse Development Environment for Eclipse Platform](https://download.eclipse.org/oomph/www/setups/svg/Eclipse_Platform.svg)](
+https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-platform/eclipse.platform/master/releng/org.eclipse.platform.setup/PlatformConfiguration.setup&show=true
+"Click to open Eclipse-Installer Auto Launch or drag into your running installer")
 
 ## Issue Tracking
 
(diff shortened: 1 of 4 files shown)


merks:
Probably the new project will have a warning about not having an explicit coding...

Note that it's **essential** that the configuration be referenced from something that is in the setup Index.  Otherwise it will not work very well, i.e., the badge will not be generated:

https://ci.eclipse.org/oomph/job/setup-archiver/

Also what's in the archive is unzipped to this location:

https://download.eclipse.org/justj/?file=oomph/www/setups/

And that's what makes it possible to show the configuration's documentation on the page:

https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-platform/eclipse.platform/master/releng/org.eclipse.platform.setup/EclipsePlatformConfiguration.setup&show=true

We can't pull it directly from GitHub because of cross-site scripting restrictions.

And finally, if the platform repository is going to have its own configuration, the project setup for it should probably be factored out and also put in that git repository like is already the case for some of the other platform projects:

image

There should then be an Eclipse Ini task with a redirection so one can modify/use/test the setup in the repository being set up.  This would also be the good place from which to reference the configuration.

HannesWell:
> Probably the new project will have a warning about not having an explicit coding...

Good point, I just added the corresponding setting.



> Note that it's **essential** that the configuration be referenced from something that is in the setup Index. Otherwise it will not work very well, i.e., the badge will not be generated:

Thank you for the reminder and the elaboration, that was very helpful!

Added the reference via 
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2437




> And finally, if the platform repository is going to have its own configuration, the project setup for it should probably be factored out and also put in that git repository like is already the case for some of the other platform projects:

That's something I also wanted to discuss, but want to leave it for later. First I would like each active repo to have it's own configuration and 'Oomph-setup button' in order to make it visible for contributors.
In a second step I can then move around the setups to be managed in their repository.
In https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2437 I have now referenced this configuration using a complete URL.

With that, this should be ready for submission. The verification build does not test this at all.

HannesWell:
The README looks as expected:

grafik


And the website it opens looks also fine:
https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-platform/eclipse.platform/master/releng/org.eclipse.platform.setup/PlatformConfiguration.setup&show=true

PR #1587, opened 2024-10-14
Use RetentionPolicy CLASS in new 'ServiceContextKey' OSGi property type

HannesWell:
And permit only types as target.

If the `RetentionPolicy` is only `SOURCE` Tycho does not recognize the annotation because the underlying BND library only operates on compiled classes and not on sources.

`RetentionPolicy.CLASS` is the default, but I still declared it to be specific in this case.

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/contexts/IContextFunction.java
+++ b/runtime/bundles/org.eclipse.e4.core.contexts/src/org/eclipse/e4/core/contexts/IContextFunction.java
@@ -15,8 +15,10 @@
 
 package org.eclipse.e4.core.contexts;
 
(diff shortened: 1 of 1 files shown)


laeubi:
Probably PDE should warn if a property type is using source retention even though [the spec only says](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-component.property.types)

> To be used in this way, the component property type must be annotated with the [ComponentPropertyType](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#org.osgi.service.component.annotations.ComponentPropertyType) meta-annotation so that, at build time, the annotation is recognized as a component property type.

there is really no benefit in using `SOURCE` retention type and it actually limits the usefulness with other tools (e.g. bnd/tycho).

PR #1588, opened 2024-10-14
Use @ServiceRanking component property type annotation in e4.core.tests

HannesWell:


laeubi:
@vogella see https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e39295

runtime/tests/org.eclipse.e4.core.javax.tests/META-INF/MANIFEST.MF
```diff
@@ -26,6 +26,7 @@ Import-Package: javax.annotation;version="[1.3.0,2.0.0)",
  org.junit.runners;version="[4.13.0,5.0.0)",
  org.osgi.framework;version="[1.10.0,2.0.0)",
  org.osgi.service.component;version="1.3.0",
+ org.osgi.service.component.propertytypes;version="[1.4.0,2.0.0)",
```
laeubi, line 29: Imports should actually not required for this

Issue #1592, opened 2024-10-21
"Problems encountered while deleting files" due "Thread limit exceeded replacing blocked worker"


iloveeclipse:
Few test fails on Jenkins because of failing delete resources operation.

This is regression from 6d514be2aca4b9be131184b63570fd4351c47c77

[Test Result](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/2/testReport/) (5 failures / +5)
[UiTestSuite org.eclipse.ui.tests.dialogs.FilteredResourcesSelectionDialogTestSuite org.eclipse.ui.tests.dialogs.ResourceSelectionFilteringDialogTest.testMatch](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/2/testReport/junit/UiTestSuite%20org.eclipse.ui.tests.dialogs.FilteredResourcesSelectionDialogTestSuite%20org.eclipse.ui.tests.dialogs/ResourceSelectionFilteringDialogTest/testMatch/)
[UiTestSuite org.eclipse.ui.tests.internal.InternalTestSuite.testMultipleFilesWithSharedParentFolders](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/2/testReport/junit/UiTestSuite%20org.eclipse.ui.tests.internal/InternalTestSuite/testMultipleFilesWithSharedParentFolders/)
[UiTestSuite org.eclipse.ui.tests.internal.InternalTestSuite.testMultipleFilesWithNameCollisionAndEditorSplit](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/2/testReport/junit/UiTestSuite%20org.eclipse.ui.tests.internal/InternalTestSuite/testMultipleFilesWithNameCollisionAndEditorSplit/)
[UiTestSuite org.eclipse.ui.tests.internal.InternalTestSuite.testMultipleFilesWithSharedParentFoldersButNotAllShareTheSameParentFolders](https://ci.eclipse.org/platform/job/eclipse.platform.ui/job/PR-2431/2/testReport/junit/UiTestSuite%20org.eclipse.ui.tests.internal/InternalTestSuite/testMultipleFilesWithSharedParentFoldersButNotAllShareTheSameParentFolders/)
[...]

PR #1593, opened 2024-10-22
Fix for RejectedExecutionException during delete operations

iloveeclipse:
If the tasks scheduled by ForkJoinPool are all busy and new task is added, the pool would throw RejectedExecutionException if the "saturate" Predicate is not specified.

To avoid that, supply Predicate that is always true - with that, no exception will be thrown on busy pool and we would not reject new delete operations.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1592


iloveeclipse:
@akurtakov :  I plan to trigger new IBuild just to make sure this fix (or original issue) doesn't have any relationship to OOM's in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432
Any objections ?

akurtakov:
Go for it.

iloveeclipse:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.34/93/

PR #1596, opened 2024-10-25
Improve RSA key generation bit size to 2048

### LINKED ISSUE #1464: [ssh] Generated Key are too weak
If I go to Preferences > SSH > Key Management there is an option "Generate RSA key", this does (without an option to choose otherwise) generate a RSA-1024 key.

https://www.ssh.com/academy/ssh/keygen#choosing-an-algorithm-and-key-size says

> A key size of at **least 2048 bits** is recommended for RSA; 4096 bits is better.

axmanalad:
Generating an RSA key originally gave a bit size of 1024, which is not the recommended bit size of 2048.
This change allows an RSA-2048 key to be generated.

Fixes #1464


HannesWell:
I'm sorry I submitted this accidentally, actually I just wanted to rebase it on top of master. My mistake.
Reverted it for now in https://github.com/eclipse-platform/eclipse.platform/pull/1597. 

Please just create new PR for this so that we can do the usual review.

PR #1598, opened 2024-10-26
Avoid duplicated IFileInfo fetches when creating resources

HannesWell:
@jukzi I assume this didn't show up in your recent performance analysis of the resources handling so I assume it won't have a big impact, but avoiding subsequent duplicated file-fetches is probably still beneficial.


iloveeclipse:
Isn't that **intended extra** check for case insensitive file systems?

HannesWell:
> Isn't that **intended extra** check for case insensitive file systems?

Where exactly? And how does it help to fetch the info of the same store twice?
Sorry if I missed that discussion.

iloveeclipse:
No, I just wondered that this extra check was behind extra "if" condition for case insensitive file systems. May be that is not a mistake but intention. I would check in git blame when it was added, for sure it is something Windows specific.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Folder.java
```diff
@@ -47,9 +48,10 @@ protected void assertCreateRequirements(IFileStore store, IFileInfo localInfo, i
 		if (!force && localInfo.exists()) {
 			//return an appropriate error message for case variant collisions
 			if (!Workspace.caseSensitive) {
-				String name = getLocalManager().getLocalName(store);
+				String name = localInfo.getName();
 				if (name != null && !store.getName().equals(name)) {
-					String msg = NLS.bind(Messages.resources_existsLocalDifferentCase, IPath.fromOSString(store.toString()).removeLastSegments(1).append(name).toOSString());
+					String msg = NLS.bind(Messages.resources_existsLocalDifferentCase,
+							Path.of(store.toString()).resolveSibling(name));
 					throw new ResourceException(IResourceStatus.CASE_VARIANT_EXISTS, getFullPath(), msg, null);
```
jukzi, line 55: CASE_VARIANT_EXISTS makes only sense when name is tested against the actual name in the filesystem. This codepath is only taken if localInfo.exists(), so never in the performance critical case where where the directory is clean.

PR #1600, opened 2024-10-27
Improve DSA and RSA key generation bit size.

### LINKED ISSUE #1464: [ssh] Generated Key are too weak

axmanalad:
Generating a DSA or RSA key originally gave a bit size of 1024, which is not the recommended bit size of 2048 nor the stronger bit size of 4096.
This change allows a DSA-3072 key or an RSA-4096 key to be generated.
Note: DSA-4096 is not possible with the current algorithm of the DSA key generator. Therefore, DSA-3072 is implemented instead of DSA-4096.

Original PR from #1596
Fixes #1464

### DIFF
--- a/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
+++ b/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
@@ -89,6 +89,8 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
 		implements IWorkbenchPreferencePage{
 
 	private static final String SSH2_PREFERENCE_PAGE_CONTEXT="org.eclipse.jsch.ui.ssh2_preference_page_context"; //$NON-NLS-1$
+	private static final int RSA_KEY_SIZE = 4096;
+	private static final int DSA_KEY_SIZE = 3072;
 
 	private Label ssh2HomeLabel;
 	private Label privateKeyLabel;
@@ -494,10 +496,11 @@ protected Control createContents(Composite parent){
 
 					final KeyPair[] _kpair=new KeyPair[1];
 					final int __type=type;
+					int keySize = type == KeyPair.RSA ? RSA_KEY_SIZE : DSA_KEY_SIZE;
 					final JSchException[] _e=new JSchException[1];
 					BusyIndicator.showWhile(getShell().getDisplay(), () -> {
 						try {
-							_kpair[0] = KeyPair.genKeyPair(getJSch(), __type);
+							_kpair[0] = KeyPair.genKeyPair(getJSch(), __type, keySize);
 						} catch (JSchException e1) {
 							_e[0] = e1;
 						}
@@ -508,7 +511,7 @@ protected Control createContents(Composite parent){
 					kpair=_kpair[0];
 
 					ByteArrayOutputStream out=new ByteArrayOutputStream();
-					kpairComment=_type+"-1024"; //$NON-NLS-1$
+					kpairComment = _type + "-" + keySize; //$NON-NLS-1$
 					kpair.writePublicKey(out, kpairComment);
 					out.close();
 					publicKeyText.setText(out.toString());


HannesWell (changes requested):
Thank you for creating this follow-up.

Can you please first make sure that you don't revert changes made before, i.e. your commit `Reverted merge changes` is removed. If you have trouble compiling, you probably have to update your target-platform. You can do that for example by re-running your Oomph setup in your workspace.

After that, please squash all your changes into one commit with a meaningful message, In the commit message body you ideally also reference the issue fixed with this. E.g. using
`Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464`.
If you squash your change please don't include the change the eclipse-platform-bot has pushed to this branch. This should stay separated.

If you subsequently apply changes from the review, please just amend your commit and force-push to the branch.

axmanalad:
Thank you for the feedback. I was able to remove the `Reverted merge changes` commit and squashed my changes into one commit.
> After that, please squash all your changes into one commit with a meaningful message, In the commit message body you ideally also reference the issue fixed with this. E.g. using `Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464`. If you squash your change please don't include the change the eclipse-platform-bot has pushed to this branch. This should stay separated.

As someone who is still learning about Git, this comment was very helpful. Let me know if there are any more changes needed to be made.

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -508,7 +512,11 @@ else if(e.widget==keyGenerateRSA){
 					kpair=_kpair[0];
 
 					ByteArrayOutputStream out=new ByteArrayOutputStream();
-					kpairComment=_type+"-1024"; //$NON-NLS-1$
+					if (__type == KeyPair.RSA) {
+						kpairComment = _type + "-4096"; //$NON-NLS-1$
+					} else {
+						kpairComment = _type + "-3072"; //$NON-NLS-1$
+					}
```
HannesWell, line 519:
```suggestion
					kpairComment = _type + "-" + keySize; //$NON-NLS-1$
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -497,7 +497,11 @@ else if(e.widget==keyGenerateRSA){
 					final JSchException[] _e=new JSchException[1];
 					BusyIndicator.showWhile(getShell().getDisplay(), () -> {
 						try {
-							_kpair[0] = KeyPair.genKeyPair(getJSch(), __type);
+							if (__type == KeyPair.RSA) {
+								_kpair[0] = KeyPair.genKeyPair(getJSch(), __type, 4096);
+							} else {
+								_kpair[0] = KeyPair.genKeyPair(getJSch(), __type, 3072);
+							}
```
HannesWell, line 504: As suggested in https://github.com/eclipse-platform/eclipse.platform/issues/1464#issuecomment-2439600327, I think the key-size for RSA and DSA should be specified in a static final int constant at the top of this class:
```
private static final String SSH2_PREFERENCE_PAGE_CONTEXT="org.eclipse.jsch.ui.ssh2_preference_page_context"; //$NON-NLS-1$
private static final int RSA_KEY_SIZE = 4096;
private static final int DSA_KEY_SIZE = 3072;
```

Then you can create a variable in this method, before `BusyIndicator.showWhile()` that holds the actual keySize used for the specified key type:
```
int keySize = type == KeyPair.RSA ? RSA_KEY_SIZE : DSA_KEY_SIZE;
final JSchException[] _e=new JSchException[1];
BusyIndicator.showWhile(getShell().getDisplay(), () -> {
```

and then this if-else-block can stay a one-liner:
```suggestion
							_kpair[0] = KeyPair.genKeyPair(getJSch(), __type, keySize);
```

as well as the block below to create the comment.

HannesWell (changes requested):
>> After that, please squash all your changes into one commit with a meaningful message, In the commit message body you ideally also reference the issue fixed with this. E.g. using Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464. If you squash your change please don't include the change the eclipse-platform-bot has pushed to this branch. This should stay separated.
> As someone who is still learning about Git, this comment was very helpful. Let me know if there are any more changes needed to be made.

You are welcome and thanks for the update:)

But with referencing `Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464` I meant to mention it at the end of the commit message body and not only in the headline. The commit headline should still be a descriptive (half) sentence, just like this PR's name/headline. The following body can then contain more details and the mentioned reference.
The overall commit message could be for example (feel free to further adjust this):
```
[SSH] Increase bit-size of generated RSA and DSA keys

Generate RSA keys with 4096bit and DSA keys with 3072 (the underlying algorithm doesn't support higher values)

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464
```

Regarding the implementation I have a two comments below.

axmanalad:
> But with referencing Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1464 I meant to mention it at the end of the commit message body and not only in the headline. The commit headline should still be a descriptive (half) sentence, just like this PR's name/headline. The following body can then contain more details and the mentioned reference.

I see there was a misunderstanding between my thought and your intention. Thank you for the clarification; I changed how the format of the commit message should look like.

> As suggested in [#1464 (comment)](https://github.com/eclipse-platform/eclipse.platform/issues/1464#issuecomment-2439600327), I think the key-size for RSA and DSA should be specified in a static final int constant at the top of this class:
> 
> ```
> private static final String SSH2_PREFERENCE_PAGE_CONTEXT="org.eclipse.jsch.ui.ssh2_preference_page_context"; //$NON-NLS-1$
> private static final int RSA_KEY_SIZE = 4096;
> private static final int DSA_KEY_SIZE = 3072;
> ```
> 
> Then you can create a variable in this method, before `BusyIndicator.showWhile()` that holds the actual keySize used for the specified key type:
> 
> ```
> int keySize = type == KeyPair.RSA ? RSA_KEY_SIZE : DSA_KEY_SIZE;
> final JSchException[] _e=new JSchException[1];
> BusyIndicator.showWhile(getShell().getDisplay(), () -> {
> ```

I really like this implementation of both comments as it looks organized and easier for readability. I readjusted the changes as you have suggested. Thank you for the suggestion.

HannesWell (approved):
Thank you for the adjustments.

Looks good and works well now. Nice work. :)

Issue #1601, opened 2024-10-28
remove org.eclipse.ant.internal.core.AntSecurityManager.getInCheck() ?


jukzi:
`method has been removed in JDK 10`

```
java.lang.NoSuchMethodException: java.lang.SecurityManager.getInCheck()
	at java.base/java.lang.Class.getMethod(Class.java:2227)
	at org.eclipse.ant.internal.core.AntSecurityManager.getInCheck(AntSecurityManager.java:399)
	at org.eclipse.ant.tests.core.tests.AntSecurityManagerTest.test_getInCheck(AntSecurityManagerTest.java:16)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:767)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)

```
https://download.eclipse.org/eclipse/downloads/drops4/I20241028-0700/testresults/ep434I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.ant.tests.core.AutomatedAntSuite.txt

akurtakov:
Sounds like the right thing to do . All the other removed in Java 10 should reb removed too IMO.

akurtakov:
@jukzi Thanks for spotting that. IMO things like that are exactly what prevents us from having good overlook of the test suites as there is way too much useless things that accumulated over the years that need pruning very badly.

PR #1602, opened 2024-10-28
Stop AntSecurityManager from impl methods removed in Java 10

akurtakov:
As the projects are at Java 17 level now these are plain useless but generate warnings in "needless" test runs.
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1601


jukzi:
thanks. good job.

Issue #958, opened 2023-12-08
Replace or remove org.eclipse.jsch plugins?


HannesWell:
## Suggestion

The JSch library seems to be abandoned, the last uploads to [source forge](https://sourceforge.net/projects/jsch/) and [Maven-Central](https://mvnrepository.com/artifact/com.jcraft/jsch) were in Nov 2018.
There is a fork available at https://github.com/mwiede/jsch, which regularly publishes updates.
Nevertheless I think it would be good to get rid of outdated dependencies especially in security sensitive areas (I'm surprised there is not CVE listed yet).

The only user in the Eclipse SDK seems to be `org.eclipse.jsch.core` and `org.eclipse.jsch.ui`, which seem to provide the ability to create password protected SSH connections, but I have not yet analyzed its full usage and capabilities.

Depending on the usages I see two options for 
1. Deprecate this Plugin for removal without replacement
2. Replace it by an API not bound to a specific third party library and e.g. use MINA SSHD under the hood (which also has proper OSGi metadata so a simplification for SimRel-Orbit). This API would then probably be under another package name (in which ever plugin it is).

## Community

- [x] I understand suggesting an enhancement doesn't mandate anyone to implement it. Other contributors may consider this suggestion, or not, at their own convenience. The most efficient way to get it fixed is that I implement it myself and contribute it back as a good quality patch to the project.

HannesWell:
I also noticed that `org.eclipse.jsch.tests` is basically empty.

laeubi:
Last time I asked about jsch upgrade on jgit they told that jgit no longer support this but uses MINA SSHD instead.

merks:
There seem to be quite a few dependencies on this in SimRel:

![image](https://github.com/eclipse-platform/eclipse.platform/assets/208716/662ef37a-1505-494f-b78b-104a2ccaa921)

Bananeweizen:
@tomaswolf You probably know this best from the jgit side, so can you give advice, please?

tomaswolf:
JGit can still use JSch. Bundle `org.eclipse.jgit.ssh.jsch` still exists, but we don't actively maintain it anymore. Within Eclipse, JGit as used via EGit exclusively uses Apache MINA sshd via bundle `org.eclipse.jgit.ssh.apache`.

[...]

PR #1608, opened 2024-11-09
[SSH] Added key size selection

axmanalad:
This implementation allows to select the key size of the generated key ranging from 2048-4096 bits, which is using a Spinner to increment by 1024 bits. (DSA is limited to 3072 bits)

Refer to https://github.com/eclipse-platform/eclipse.platform/issues/1464

### DIFF
--- a/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/messages.properties
+++ b/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/messages.properties
@@ -89,6 +89,7 @@ CVSSSH2PreferencePage_144=Key Exchange &Methods
 CVSSSH2PreferencePage_145=MA&C Methods
 CVSSSH2PreferencePage_146=&SSH Agent
 CVSSSH2PreferencePage_147=Select preferred SSH Agent
+CVSSSH2PreferencePage_148=Key size:
 UserInfoPrompter_0=SSH2 Message
 UserInfoPrompter_1=SSH2 Message
 KeyboardInteractiveDialog_0=Keyboard Interactive authentication for {0}: {1}
--- a/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
+++ b/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
@@ -53,6 +53,7 @@ import org.eclipse.swt.custom.CTabItem;
 import org.eclipse.swt.custom.TableEditor;
 import org.eclipse.swt.events.FocusEvent;
 import org.eclipse.swt.events.FocusListener;
+import org.eclipse.swt.events.KeyListener;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.graphics.Image;
@@ -66,6 +67,7 @@ import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.FileDialog;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Spinner;
 import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableColumn;
 import org.eclipse.swt.widgets.TableItem;
@@ -89,8 +91,11 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
 		implements IWorkbenchPreferencePage{
 
 	private static final String SSH2_PREFERENCE_PAGE_CONTEXT="org.eclipse.jsch.ui.ssh2_preference_page_context"; //$NON-NLS-1$
-	private static final int RSA_KEY_SIZE = 4096;
-	private static final int DSA_KEY_SIZE = 3072;
+	private static final int RSA_MAX_KEY_SIZE = 15360;
+	private static final int DSA_MAX_KEY_SIZE = 3072;
+	private static final int INITIAL_KEY_SIZE = 4096;
+	private static final int MIN_KEY_SIZE = 2048;
+	private static final int KEY_SIZE_INCREMENT = 1024;
 
 	private Label ssh2HomeLabel;
 	private Label privateKeyLabel;
@@ -101,6 +106,7 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
 	private Button ssh2HomeBrowse;
 	Button keyGenerateDSA;
 	Button keyGenerateRSA;
+	private Spinner keySizeValue;
 	private Button keyLoad;
 	private Button keyExport;
 	Button saveKeyPair;
@@ -305,6 +311,13 @@ protected Control createContents(Composite parent){
 		gd.horizontalSpan=1;
 		keyLoad.setLayoutData(gd);
 
+		final Label keySizeValueLabel = new Label(group, SWT.NONE);
+		keySizeValueLabel.setText(Messages.CVSSSH2PreferencePage_148);
+		keySizeValue = new Spinner(group, SWT.BORDER);
+		int maxKeySize = Math.max(DSA_MAX_KEY_SIZE, RSA_MAX_KEY_SIZE);
+		keySizeValue.setValues(INITIAL_KEY_SIZE, MIN_KEY_SIZE, maxKeySize, 0, KEY_SIZE_INCREMENT, KEY_SIZE_INCREMENT);
+		keySizeValue.addKeyListener(KeyListener.keyPressedAdapter(e -> e.doit = false));
+
 		publicKeylabel=new Label(group, SWT.NONE);
 		publicKeylabel.setText(Messages.CVSSSH2PreferencePage_39);
 		gd=new GridData();
@@ -485,6 +498,9 @@ protected Control createContents(Composite parent){
 					if(e.widget==keyGenerateDSA){
 						type=KeyPair.DSA;
 						_type=IConstants.DSA;
+						if (keySizeValue.getSelection() > DSA_MAX_KEY_SIZE) {
+							keySizeValue.setSelection(DSA_MAX_KEY_SIZE);
+						}
 					}
 					else if(e.widget==keyGenerateRSA){
 						type=KeyPair.RSA;
@@ -496,7 +512,7 @@ protected Control createContents(Composite parent){
 
 					final KeyPair[] _kpair=new KeyPair[1];
 					final int __type=type;
-					int keySize = type == KeyPair.RSA ? RSA_KEY_SIZE : DSA_KEY_SIZE;
+					int keySize = keySizeValue.getSelection();
 					final JSchException[] _e=new JSchException[1];
 					BusyIndicator.showWhile(getShell().getDisplay(), () -> {
 						try {
--- a/team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jsch.ui;singleton:=true
-Bundle-Version: 1.5.500.qualifier
+Bundle-Version: 1.5.600.qualifier
 Bundle-Activator: org.eclipse.jsch.internal.ui.JSchUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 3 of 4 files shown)


team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/messages.properties
```diff
@@ -89,6 +89,7 @@ CVSSSH2PreferencePage_144=Key Exchange &Methods
 CVSSSH2PreferencePage_145=MA&C Methods
 CVSSSH2PreferencePage_146=&SSH Agent
 CVSSSH2PreferencePage_147=Select preferred SSH Agent
+CVSSSH2PreferencePage_148=Key Size to Generate:
```
HannesWell, line 92: Let's keep it simple here:
```suggestion
CVSSSH2PreferencePage_148=Key size:
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -101,6 +108,7 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
 	private Button ssh2HomeBrowse;
 	Button keyGenerateDSA;
 	Button keyGenerateRSA;
+	private Spinner keyGenerateSize;
```
HannesWell, line 111: Better name it `keySizeValue` or just `keySize`, but then you have to rename the existing local variable with the same name to `keySizeValue` or alike or you just inline `keySize.getSelection()`
```suggestion
	private Spinner keySizeValue;
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -305,6 +313,19 @@ private Control createKeyManagementPage(Composite parent){
...
+		keyGenerateSize=new Spinner(group, SWT.BORDER);
+		keyGenerateSize.setValues(MIN_KEY_SIZE, MIN_KEY_SIZE, MAX_KEY_SIZE, MAX_KEY_SIZE_DIGIT_COUNT,
+				KEY_SIZE_INCREMENT, KEY_SIZE_INCREMENT);
+
+		keyGenerateSize.addKeyListener(new KeyAdapter() {
+			@Override
+			public void keyPressed(KeyEvent e) {
+				e.doit = false;
+			}
+		});
```
HannesWell, line 327: This can be simplified using `KeyListener.keyPressedAdapter()`
```suggestion
		keySizeValue.addKeyListener(KeyListener.keyPressedAdapter(e -> e.doit = false));
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -485,6 +506,8 @@ public void widgetSelected(SelectionEvent e){
 					if(e.widget==keyGenerateDSA){
 						type=KeyPair.DSA;
 						_type=IConstants.DSA;
+						if (keyGenerateSize.getSelection()>DSA_KEY_SIZE)
+							keyGenerateSize.setSelection(DSA_KEY_SIZE);
```
HannesWell, line 510: Please always use braces:
```suggestion
						if (keyGenerateSize.getSelection()>DSA_KEY_SIZE) {
							keyGenerateSize.setSelection(DSA_KEY_SIZE);
						}
```

HannesWell (changes requested):
Thank you for this follow-up. Nice work, it works well.
I just have a few remarks below.

In general what do you think about setting the initial value to the recommended size of 4kBi for RSA and 3kBi for DSA and just allow lowering the value?
Are higher values possible for RSA too? If yes, I think we should also make that possible.

Additionally please make sure that you have formatted all new lines according to the formatter settings (unfortunately the existing code does not adhere to them). E.g. there should be a space before and after a equal-sign.

axmanalad:
> In general what do you think about setting the initial value to the recommended size of 4kBi for RSA and 3kBi for DSA and just allow lowering the value?

I think this is a good idea as it initially sets it to the more secured key size than 2048 bits for developers to acknowledge. I changed the initial value to 4096 bits.

> Additionally please make sure that you have formatted all new lines according to the formatter settings (unfortunately the existing code does not adhere to them). E.g. there should be a space before and after a equal-sign.

I have fixed the formatting. Thank you for clarification.

> Are higher values possible for RSA too? If yes, I think we should also make that possible.

Higher values for RSA are possible. The current RSA algorithm does not seem to have a limit for the max key size it can generate. However, the main drawbacks with generating larger RSA key lengths include [CPU overhead and performance issues](https://www.javamex.com/tutorials/cryptography/rsa_key_length.shtml). As you generate a larger key length, CPU runtime increases dramatically. A high-end computer could easily generate larger keys, but I would not recommend it for low-end computers.

Because of the overhead and performance, it is best to recommend using an [ECC encryption](https://ww1.microchip.com/downloads/en/DeviceDoc/00003442A.pdf) for keys larger than 4096 bits since it provides an equivalent level of encryption strength as RSA with smaller key sizes and faster performance. For example, an ECC key size of 512 bits is equivalent to an RSA key size of 15360 bits.

I changed the max RSA key size you can generate to 15360 bits as it is equivalent to 256 security bits (really strong in security encryption); generally, higher than 15360 bits is not recommended for the sake of runtime.

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -89,8 +91,13 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
 		implements IWorkbenchPreferencePage{
 
 	private static final String SSH2_PREFERENCE_PAGE_CONTEXT="org.eclipse.jsch.ui.ssh2_preference_page_context"; //$NON-NLS-1$
-	private static final int RSA_KEY_SIZE = 4096;
+	private static final int RSA_KEY_SIZE = 15360;
 	private static final int DSA_KEY_SIZE = 3072;
```
HannesWell, line 95: Let's add a `MAX` segment in these names to reflect that these are the maximum values:
```suggestion
	private static final int RSA_MAX_KEY_SIZE = 15360;
	private static final int DSA_MAX_KEY_SIZE = 3072;
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -89,8 +91,13 @@ public class PreferencePage extends org.eclipse.jface.preference.PreferencePage
...
+	private static final int INITIAL_KEY_SIZE = 4096;
+	private static final int MIN_KEY_SIZE = 2048;
+	private static final int MAX_KEY_SIZE = RSA_KEY_SIZE;
+	private static final int MAX_KEY_SIZE_DIGIT_COUNT = 0;
```
HannesWell, line 99: These constants are not needed.
```suggestion
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -305,6 +313,13 @@ private Control createKeyManagementPage(Composite parent){
 		gd.horizontalSpan=1;
 		keyLoad.setLayoutData(gd);
 
+		final Label keySizeValueLabel = new Label(group, SWT.NONE);
+		keySizeValueLabel.setText(Messages.CVSSSH2PreferencePage_148);
+		keySizeValue = new Spinner(group, SWT.BORDER);
+		keySizeValue.setValues(INITIAL_KEY_SIZE, MIN_KEY_SIZE, MAX_KEY_SIZE, MAX_KEY_SIZE_DIGIT_COUNT,
+				KEY_SIZE_INCREMENT, KEY_SIZE_INCREMENT);
```
HannesWell, line 320: This can be structured a bit more to prevent an over-long line and two constants:
```suggestion
		int maxKeySize = Math.max(DSA_MAX_KEY_SIZE, RSA_MAX_KEY_SIZE);
		keySizeValue.setValues(INITIAL_KEY_SIZE, MIN_KEY_SIZE, maxKeySize, 0, KEY_SIZE_INCREMENT, KEY_SIZE_INCREMENT);
```

team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/preference/PreferencePage.java
```diff
@@ -485,6 +500,9 @@ public void widgetSelected(SelectionEvent e){
 					if(e.widget==keyGenerateDSA){
 						type=KeyPair.DSA;
 						_type=IConstants.DSA;
+						if (keySizeValue.getSelection() > DSA_KEY_SIZE) {
+							keySizeValue.setSelection(DSA_KEY_SIZE);
```
HannesWell, line 504: Adapt to new constant names
```suggestion
						if (keySizeValue.getSelection() > DSA_MAX_KEY_SIZE) {
							keySizeValue.setSelection(DSA_MAX_KEY_SIZE);
```

HannesWell (commented):
Thanks for the update and sorry for the deplayed reply.
I just have a few minor style remarks below.
Technically this is fine. 👍🏽 

But we are currently in the quiet-period before the December release and have to await the start of the next development cycle, which should start in one to two weeks, before this can be submitted. But in the meantime we can make it ready.

> Because of the overhead and performance, it is best to recommend using an [ECC encryption](https://ww1.microchip.com/downloads/en/DeviceDoc/00003442A.pdf) for keys larger than 4096 bits since it provides an equivalent level of encryption strength as RSA with smaller key sizes and faster performance. For example, an ECC key size of 512 bits is equivalent to an RSA key size of 15360 bits. I am not sure how implementing ECC would work but it is only a thought I think would improve performance in generating keys.

It would be great if we could also generated more modern keys and if you are interested in looking into it, you are more than welcome.
I'm not sure, but I fear that the currently used library is not capable of these newer algorithms.
In fact we actually would like to replace JSCH with something that's maintained better and more up-to-date, but havn't done it yet due to a lack of known simple replacements, see
- https://github.com/eclipse-platform/eclipse.platform/issues/958

But if you are interested in that please don't hesitate to look into it. :)

axmanalad:
No worries, it is understandable. I reformatted the variables as requested.

> It would be great if we could also generated more modern keys and if you are interested in looking into it, you are more than welcome. I'm not sure, but I fear that the currently used library is not capable of these newer algorithms. In fact we actually would like to replace JSCH with something that's maintained better and more up-to-date, but havn't done it yet due to a lack of known simple replacements, see
> 
> * [Replace or remove org.eclipse.jsch plugins? #958](https://github.com/eclipse-platform/eclipse.platform/issues/958)
> 
> But if you are interested in that please don't hesitate to look into it. :)

While looking in the #958, I believe replacing the JSch library is a major change to be made. I lack the knowledge to knowing how to make these replacements however. In the meantime, I looked deeper into the codebase to find that there exists an ECC algorithm in the JSch library known as **ECDSA** before the library became outdated. This is a type of ECC algorithm that does exactly what I previously stated and is found within "*com.jcraft.jsch.KeyPairECDSA*". Maybe after this PR, I can make a third contribution to add the feature of generating ECDSA keys in a new PR, which is why I added a "TODO" comment on Line 509.

HannesWell:
Sorry, I somehow missed the notification about your latest update, will look into it as soon as possible.

> > It would be great if we could also generated more modern keys and if you are interested in looking into it, you are more than welcome. I'm not sure, but I fear that the currently used library is not capable of these newer algorithms. In fact we actually would like to replace JSCH with something that's maintained better and more up-to-date, but havn't done it yet due to a lack of known simple replacements, see
> > 
> > * [Replace or remove org.eclipse.jsch plugins? #958](https://github.com/eclipse-platform/eclipse.platform/issues/958)
> > 
> > But if you are interested in that please don't hesitate to look into it. :)
> 
> While looking in the #958, I believe replacing the JSch library is a major change to be made. I lack the knowledge to knowing how to make these replacements however. In the meantime, I looked deeper into the codebase to find that there exists an ECC algorithm in the JSch library known as **ECDSA** before the library became outdated. This is a type of ECC algorithm that does exactly what I previously stated and is found within "_com.jcraft.jsch.KeyPairECDSA_". Maybe after this PR, I can make a third contribution to add the feature of generating ECDSA keys in a new PR, which is why I added a "TODO" comment on Line 509.

That would be very nice. But we should be cautious and should check how well that implementation works. Having a flawed implementation of a per-se save algorithm would pretend false security, which would be really bad.

HannesWell (approved):
This looks perfectly fine now.
Thanks for the update and sorry for the delayed reply. I lost track of this during the hectic rush before Christmas.

Issue #1609, opened 2024-11-11
[2024-09 regression] Restart is needed to refresh shared launch configurations


rhoehener:
Since 2024-09, we need to restart eclipse to refresh our shared launch configurations. Otherwise Eclipse will still show launch configs that have been changed by someone else (= the change happens via souce control updates, not via launch config UI).

I assume this might also affect adding or renaming of launch configurations.

To reproduce, delete a shared launch configuration on the file system and try to refresh / update the project in the workspace -> the launch configuration should still be visible in the list.

jukzi:
@rhoehener thanks for your report, can you please contribute a fix or at least point to a commit that broke the behaviour?

rhoehener:
@jukzi i'm sorry i'm just a long-time user of eclipse. but please let me know if you cannot reproduce the problem or need more information.

merks:
Perhaps this has some unintended impact?

https://github.com/eclipse-platform/eclipse.platform/commit/8d5bfec6e9f2e6c6136413e0a82062d667029dbc

iloveeclipse:
> Perhaps this has some unintended impact?
> 
> [8d5bfec](https://github.com/eclipse-platform/eclipse.platform/commit/8d5bfec6e9f2e6c6136413e0a82062d667029dbc)

Yes, https://github.com/eclipse-platform/eclipse.platform/pull/1487 looks related.
@HannesWell : could you please check? This isn't a nice regression ...

iloveeclipse:
@trancexpress : could you please bisect if this is from https://github.com/eclipse-platform/eclipse.platform/pull/1487 and if yes, provide a fix **or** revert? Would be nice  for RC1.

trancexpress:
I can't reproduce with latest I-build, on Linux:

```
Eclipse SDK
Version: 2024-12 (4.34)
Build id: I20241111-1800
```

I tried:

1. Create a new Java project.
2. Create some Java class with a `main` snippet.
3. Run the snippet.
4. Adjust the launch to be shared, shared launch file is in the root of the project.
5. Delete the launch file on command line.
6. Refresh the project, the launch configuration is gone.

I also tried restarting Eclipse between steps 4. and 5.. 

I also tried with the Eclipse 4.33 release:

```
Eclipse SDK
Version: 2024-09 (4.33)
Build id: I20240903-0240
```

@rhoehener  please provide specific reproduction steps, the type of launch, where its stored and so on.

[...]

PR #1616, opened 2024-11-13
Revert "Make LaunchManager's IResourceDeltaVisitors more light weight"

trancexpress:
This reverts commit 8d5bfec6e9f2e6c6136413e0a82062d667029dbc.

Fixes: #1609

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -38,6 +38,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


merks:
I was going to suggest that it might be polite to given @HannesWell a chance to comment/review, but that's water under the bridge now...

HannesWell:
> I was going to suggest that it might be polite to given @HannesWell a chance to comment/review, but that's water under the bridge now...

Yes. But I would have appreciated it even more if instead of doing a full revert just the parts that are not trivial would have been reverted. When you are already familiar with the reproducer it will probably relatively easy to filter out the change that caused the regression. Now both of us have to setup a reproducer.

iloveeclipse:
The constraints we had:

- we wanted to fix this regression for RC1
- the original change was not a significant feature that would be worth to wait for RC2 but a refactoring
- pocking around in a code that was seemingly "trivial" but had a subtle bug could cause another regression
- we had no time & recources (I'm offline)
- we had no guarantee Hannes will be there with a fix before RC2

Under these conditions doing a revert was the safest approach.

trancexpress:
> Yes. But I would have appreciated it even more if instead of doing a full revert just the parts that are not trivial would have been reverted. When you are already familiar with the reproducer it will probably relatively easy to filter out the change that caused the regression. Now both of us have to setup a reproducer.

In future, please separate clean-ups from actual changes (with different commits). I would have reverted only the actual changes that cause the bug, if this was the case.

I checked the diff but had no time to divine what actually changed.

HannesWell:
> Under these conditions doing a revert was the safest approach.

Ok, that's understandable, it would just have been nice to have, anyways. Thanks for fixing that issue.
I have now created a second PR with split commits:
- https://github.com/eclipse-platform/eclipse.platform/pull/1618

@trancexpress if you still have the reproducer at hand, it would be nice if you would have a look. If not I'll check if myself.

 

> > Yes. But I would have appreciated it even more if instead of doing a full revert just the parts that are not trivial would have been reverted. When you are already familiar with the reproducer it will probably relatively easy to filter out the change that caused the regression. Now both of us have to setup a reproducer.
> 
> In future, please separate clean-ups from actual changes (with different commits). I would have reverted only the actual changes that cause the bug, if this was the case.

Usually that's what I usually do. I assume in that case I assume that the change is so small and save. But I'm not 100% sure anymore since I did the change four month ago.

PR #1618, opened 2024-11-14
Clean-up LaunchManager attempt 2

HannesWell:
Second attempt of https://github.com/eclipse-platform/eclipse.platform/pull/1487 after it was reverted in https://github.com/eclipse-platform/eclipse.platform/pull/1616.

This PR is split into two commits, the first one only containing clean-ups that should not change any logic. The second commit defers obtaining the `IResource` of a delta as long as possible.

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
@@ -542,7 +542,7 @@ public class LaunchManager extends PlatformObject implements ILaunchManager, IRe
 			return false;
 		}
 		if (0 != (delta.getFlags() & IResourceDelta.OPEN)) {
-			if (delta.getResource() instanceof IProject project) {
+			if (delta.getFullPath().segmentCount() == 1 && delta.getResource() instanceof IProject project) {
 				if (project.isOpen()) {
 					LaunchManager.this.projectOpened(project);
 				} else {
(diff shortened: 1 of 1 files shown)


trancexpress:
The problem is still seen. This "fixes" the bug:

```
diff --git a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
index 80a62dfea1..35228b29ad 100644
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/LaunchManager.java
@@ -551,8 +551,8 @@ public class LaunchManager extends PlatformObject implements ILaunchManager, IRe
                        }
                        return false;
                }
-               if (LAUNCH_CONFIG_FILE_EXTENSIONS.contains(delta.getFullPath().getFileExtension())) {
-                       if (delta.getResource() instanceof IFile file) {
+               if (delta.getResource() instanceof IFile file) {
+                       if (LAUNCH_CONFIG_FILE_EXTENSIONS.contains(delta.getFullPath().getFileExtension())) {
                                ILaunchConfiguration handle = new LaunchConfiguration(file);
                                switch (delta.getKind()) {
                                        case IResourceDelta.ADDED:
```

I'm not exactly sure why, probably the `return false` statement in the outer `if` block now does something differently.

Here are the reproduction steps:

1. Create a Java project.
2. Create a folder `.launch` in the project.
3. Create a snippet with a `main()` method.
4. Run the snippet.
5. Go into the launch config dialog for the new launch (of the snippet created above).
6. Go to the `Common` tab, make the launch shared, the directory of the shared launch file is `.launch`.
7. Delete the shared launch file on command line.
8. Refresh the project.
9. Observe that the launch config dialog still lists the launch for the snippet, it should not.


Also recorded:

[platform_gh1618_reproduction_steps.webm](https://github.com/user-attachments/assets/b4988563-34cf-478a-a85d-ad1e39a5e90e)

iloveeclipse:
May be this PR should have a test case?

PR #1622, opened 2024-11-20
Adds forRemoval to long deprecated methods in Platform

vogella:
Mark long time deprecated constants and methods for removal so that we have the option to remove them after the deprecation period ends.

This is planned to be merged beginning of the 2025-03 release.


vogella:
@HannesWell here I also do not see the version increase commit from the bot. Can you have a look?

PR #1625, opened 2024-11-20
show diff in "replace with local history" dialog on selection change

tobiasmelcher:
The "Replace With" -> "Local History ..." option in the source editor context menu launches a dialog that allows users to revert to a previous version. Initially, this dialog displays a list of all available revisions:

replace_with_local_history

Previously, selecting an entry from the revision list did not trigger any action. Users had to double-click to view the differences.

This pull request enhances the behavior by displaying the differences upon a single click/selection.

### DIFF
--- a/team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.team.ui; singleton:=true
-Bundle-Version: 3.10.500.qualifier
+Bundle-Version: 3.11.0.qualifier
 Bundle-Activator: org.eclipse.team.internal.ui.TeamUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/ui/history/HistoryPageCompareEditorInput.java
+++ b/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/ui/history/HistoryPageCompareEditorInput.java
@@ -31,6 +31,7 @@ import org.eclipse.swt.widgets.Display;
 import org.eclipse.team.internal.ui.TeamUIMessages;
 import org.eclipse.team.internal.ui.Utils;
 import org.eclipse.team.internal.ui.history.DialogHistoryPageSite;
+import org.eclipse.team.internal.ui.history.LocalHistoryPage;
 import org.eclipse.team.ui.PageCompareEditorInput;
 import org.eclipse.ui.part.IPage;
 import org.eclipse.ui.part.Page;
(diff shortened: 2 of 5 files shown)


tobiasmelcher:
Issue was detected in an internal UX review. UX experts were asking us why the UI is not updated when an entry in the history list is selected. Question is absolutely valid from my point of view.

tobiasmelcher:
Following code block is needed to fix layouting issue on macOS. See screencast below:

```
@Override
public void run(boolean fork, boolean cancelable, IRunnableWithProgress runnable) throws InvocationTargetException, InterruptedException {
	// Avoid using super.run() on macOS as it causes layout issues.
	// The compare viewer controls appear small and do not fill the entire dialog.
	runnable.run(new NullProgressMonitor());
}
```

https://github.com/user-attachments/assets/e2c269ab-e903-43ea-ab61-1207bbdc0835

team/bundles/org.eclipse.team.ui/src/org/eclipse/team/ui/history/HistoryPageCompareEditorInput.java
```diff
@@ -79,11 +80,23 @@ protected void handleDispose() {
 
 	@Override
 	protected IPage createPage(CompareViewerPane parent, IToolBarManager toolBarManager) {
+		return createPage(parent, toolBarManager, true);
+	}
+
+	/**
+	 * @since 3.10
```
BeckerWdf, line 87:
```suggestion
	 * @since 3.11
```

team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.team.ui; singleton:=true
-Bundle-Version: 3.10.500.qualifier
+Bundle-Version: 3.10.600.qualifier
```
BeckerWdf, line 5: CI-Build says:

The *minor* version should be incremented in version 3.10.600, since new APIs have been added since version 3.10.500

So I think 3.11.000 would be the correct new version not. 3.10.600.

BeckerWdf (changes requested):

team/bundles/org.eclipse.team.ui/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.team.ui; singleton:=true
-Bundle-Version: 3.10.500.qualifier
+Bundle-Version: 3.11.000.qualifier
```
BeckerWdf, line 5:
```suggestion
Bundle-Version: 3.11.0.qualifier
```

BeckerWdf (changes requested):

runtime/features/org.eclipse.core.runtime.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.core.runtime.feature"
       label="%featureName"
-      version="1.4.600.qualifier"
+      version="1.4.700.qualifier"
```
BeckerWdf, line 5: why did the bot not increase the version of the feature?

BeckerWdf (approved):

runtime/features/org.eclipse.core.runtime.feature/feature.xml
BeckerWdf: @HannesWell: Can you answer this question?

HannesWell: I assume it's because the included `org.eclipse.equinox.core.feature`, which is maintained in the equinox repository was changed due to comparator-errors:
- https://github.com/eclipse-equinox/equinox/pull/707

And since the content changed, the feature needs a version bump, but it looks like this was already done and rebasing would let the change vanish:
- https://github.com/eclipse-platform/eclipse.platform/pull/1639

BeckerWdf: Thanks @HannesWell.

BeckerWdf:
Failures are unrelated. Merging this.

PR #1650, opened 2024-12-12
Move TextMatcher from Platform UI to Platform Core Resources

ptziegler:
The TextMatcher class is used in the UI component, despite not depending on any UI classes. By moving it to the platform, it can be used in both mixed and pure E4 applications.


ptziegler:
Prerequisite to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567

vogella (commented):
Should this class be used from other plugins? In this case please use a non internal package

ptziegler:
> Should this class be used from other plugins? In this case please use a non internal package
That's the plan, at least. But I haven't found any good place to put it... Perhaps this really needs to be pushed down to Equinox.

Issue #1572, opened 2024-09-30
ConfigurationActivator should lazy load the PlatformConfiguration


laeubi:
Currently `ConfigurationActivator` eagerly loads and creates `PlatformConfiguration`, as this involves scanning a lot of jar files I noticed this becomes quite slow on Windows.

As the field is only read when calling `ConfigurationActivator.getBundleGroups()` the creation and initialization of the object should be handled there to shift the penalty to the place where it is actually required, callers can then for example benefit from the UI and inform the user that something is in progress instead of blocking the OSGi Startup in general, whether or not the information is used at all.

This might not be noticeable in a regular IDE if the windows defender exclusion is in place or might depend on other factors but seems easy enough to improve the situation for example custom RCP application.

IamLRBA:
@laeubi I believe since  PlatformConfiguration is being eagerly created, it is initialized as soon as the ConfigurationActivator is activated, even if it is not immediately needed. Just as you have described. This causes a delay in OSGi startup due to scanning many JAR files, especially on Windows.
I would like to be assigned to this issue to work on it please.
Thanks!

laeubi:
@IamLRBA as this affects some (very old) core eclipse component I would not recommend this task for newcomers but I'm currently working on this!

PR #1651, opened 2024-12-15
Mark IPlatformConfiguration/Factory for removal

laeubi:
These interfaces are currently only used by ConfiguratorUtils that is already marked for removal on 2024-03

See
- https://github.com/eclipse-platform/eclipse.platform/issues/1572

### DIFF
--- a/update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfiguration.java
+++ b/update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfiguration.java
@@ -36,7 +36,7 @@ import java.net.URL;
  * @deprecated The org.eclipse.update component has been replaced by Equinox p2.
  * This API will be deleted in a future release. See bug 311590 for details.
  */
-@Deprecated
+@Deprecated(forRemoval = true, since = "2025-03")
 public interface IPlatformConfiguration {
 
 	/**
--- a/update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfigurationFactory.java
+++ b/update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfigurationFactory.java
@@ -27,7 +27,7 @@ import java.net.URL;
  * @deprecated The org.eclipse.update component has been replaced by Equinox p2.
  * This API will be deleted in a future release. See bug 311590 for details.
  */
-@Deprecated
+@Deprecated(forRemoval = true, since = "2025-03")
 public interface IPlatformConfigurationFactory {
 	/**
 	 * Returns the current platform configuration.
--- a/update/org.eclipse.update.configurator/META-INF/MANIFEST.MF
+++ b/update/org.eclipse.update.configurator/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.update.configurator; singleton:=true
-Bundle-Version: 3.5.600.qualifier
+Bundle-Version: 3.5.700.qualifier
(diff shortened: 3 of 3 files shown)


vogella:
LGTM

vogella:
@laeubi looks like this adds additional warnings (for example The type IPlatformConfiguration.IFeatureEntry has been deprecated and marked for removal) to the build and fails the quality gate. Can you have to look? IIRC mark for removal should  be done in M1

laeubi:
@vogella this is JDT bug:

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3402

so we can not really do much than to accept the new warnings and hope for the best :-)

vogella:
> @vogella this is JDT bug:
> 
> [eclipse-jdt/eclipse.jdt.core#3402](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3402)
> 
> so we can not really do much than to accept the new warnings and hope for the best :-)

In this case I suggest to merge 

(I assume this will update also the quality gate and future PR will use the new warning number created by this JDT bug ).

vogella:
Can this be merged?

update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfiguration.java
```diff
@@ -36,7 +36,7 @@
  * @deprecated The org.eclipse.update component has been replaced by Equinox p2.
  * This API will be deleted in a future release. See bug 311590 for details.
  */
-@Deprecated
+@Deprecated(forRemoval = true, since = "2025-03")
```
HannesWell, line 39:
```suggestion
@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```

update/org.eclipse.update.configurator/src/org/eclipse/update/configurator/IPlatformConfigurationFactory.java
```diff
@@ -27,7 +27,7 @@
  * @deprecated The org.eclipse.update component has been replaced by Equinox p2.
  * This API will be deleted in a future release. See bug 311590 for details.
  */
-@Deprecated
+@Deprecated(forRemoval = true, since = "2025-03")
```
HannesWell, line 30:
```suggestion
@Deprecated(forRemoval = true, since = "2025-03 (removal in 2027-03 or later)")
```

HannesWell (commented):
Could you please extend the deprecation note a bit.
We discussed this topic a while ago at the PMC, but I haven't had the time to send the note around yet.

merks:
It's 2025-06 now...

laeubi:
I hereby grant anyone interested in this change to apply whatever seems suitable... I just wanted to make it for removal now its pending for month due to different things and the build constantly fails some validations that I'm not sure I want to really invest more time here.

vogella:
Update here: https://github.com/eclipse-platform/eclipse.platform/pull/1867

HannesWell:
> I dislike the additional meta-data request.
> 
> The PMC might decide to reduce or extend the API deletion period so putting the deletion period in comment seems wrong to me.

In my opinion a once set deprecation period should not be shortened under normal circumstances, because downstream consumers might rely on it. If it's extended that also covered by the proposed change, as it has the `or later` suffix.
The main motivation is that probably not many downstream users are aware of the exact deprecation policy of Eclipse (TLP). So this additional text is intended to make them aware and set expectations in both directions: They can expect it stays until the specified date but also have to expect it's gone afterwards. And better awareness of the situation by more people is IMO better.

But anyways that's not yet a rule as it was not yet officially announced. I'll try to do that soon and also apply the update as you mentioned in: https://github.com/eclipse-platform/eclipse.platform/pull/1866#issuecomment-2866039709 Thanks for the pointer.

vogella:
Done via https://github.com/eclipse-platform/eclipse.platform/pull/1867

Issue #1652, opened 2024-12-15
Get rid of org.eclipse.update.configurator


laeubi:
The bundle `org.eclipse.update.configurator` only contains deprecated code and the only evidence that it is actually used anywhere I found in PDE that says:

```
if (bundles.get("org.eclipse.update.configurator") != null) { //$NON-NLS-1$
	// this argument is required as long as we support old target platforms containing o.e.update.configurator
	// otherwise both simpleconfigurator and update.configurator will try to install bundles, slowing down launches.
	properties.setProperty("org.eclipse.update.reconcile", "false"); //$NON-NLS-1$ //$NON-NLS-2$
}
```

This is from 2020 and the commit says this is needed to **disable** it. It therefore looks like this functionally is actually not used/needed/wanted anywhere, but the bundle is for example still included in `org.eclipse.rcp` feature and some launch configs.

Sadly this even seem to slowdown Eclipse  starts see 
- https://github.com/eclipse-platform/eclipse.platform/issues/1572

I'm currently working on this to lower the impact, but on the long run it seems better to deprecate the whole bundle and finally remove it completely from the platform.

laeubi:
Currently the whole purpose seem to be that is provides the service `IBundleGroupProvider` this then needs to be migrated, this currently also not working very well from a IDE launched product as only deployed features (aka bundle groups) are detected at all.

Phillipus:
If `org.eclipse.update.configurator` is removed our RCP product will stop working as it relies on it for the dropins folder. We use the `dropins` folder to add additional plugins.

laeubi:
Dropins folder is provided by P2, so we probably need adjustments there, but I can't find how it is used anywhere there

https://wiki.eclipse.org/Equinox/p2/Getting_Started#Dropins
https://github.com/eclipse-equinox/p2/tree/master/bundles/org.eclipse.equinox.p2.reconciler.dropins

beside this, using the drop-ins folder is not really something one would recommend these days as it has several problems.

Phillipus:
> beside this, using the drop-ins folder is not really something one would recommend these days as it has several problems.

As I said:

> If org.eclipse.update.configurator is removed our RCP [product](https://www.archimatetool.com/) will no longer support additional plug-ins as it relies on it for the dropins folder. We use the dropins folder to add additional plugins. I know of [another](https://github.com/ModelioOpenSource) RCP product that also relies on the dropins folder (also the Eclipse IDE currently supports the dropins folder).

[...]

PR #1653, opened 2024-12-15
Make the PlatformConfigurationFactory a component

laeubi:
Currently the activator created the factory (but only register it as a plain service) and on the other hand the BundleGroupComponent then access the activator in a static way.

This now decouples the Factory and BundleGroupComponent from the activator by making them components.

Due to
- https://github.com/eclipse-platform/eclipse.platform/issues/1652#issuecomment-2543480890

one can only see if it really works in a deployed product, so beside checking the basic things this also needs to be validated after the next I-build.

Relates to
- https://github.com/eclipse-platform/eclipse.platform/issues/1572


laeubi:
I now verified with a debugger that the configuration is only read when it is first requested (in this case by `BundleGroupComponent`)

PR #1656, opened 2024-12-16
Run jenkins build with Java 21

akurtakov:
Prereq for
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2654

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ pipeline {
 	}
 	tools {
 		maven 'apache-maven-latest'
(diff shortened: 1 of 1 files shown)


vogella:
So we are moving to Java 21 for the SDK? Should you also update the GH actions in this case?

akurtakov:
Already done via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2662

merks:
YOu mean like this;

PR #1659, opened 2024-12-17
Finish o.e.help.base move to Lucene 10 and Java 21

akurtakov:


merks:
This all makes good sense.  Thanks.  👍

merks:
FYI, note that changing the URL methods being called to fix the deprecation warnings is **really dangerous**.

PR #1661, opened 2024-12-17
Move transitive dependees of o.e.help.base to Java 21

akurtakov:
Reduces pde warnings about that.


akurtakov:
Jenkins reports 16 new warnings - all of them are usages of APIs deprecated in newer JVMs.

Issue #1660, opened 2024-12-17
41 AntUtilTests fail since I20241215-1800


jukzi:
for example
```
java.lang.AssertionError: Incorrect number of targets retrieved; should be 5 was: 1
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.ant.tests.ui.AntUtilTests.assertTargets(AntUtilTests.java:269)
at org.eclipse.ant.tests.ui.AntUtilTests.testGetIncludeTargetsComplexHierarchyAlias(AntUtilTests.java:152)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at org.eclipse.ant.tests.ui.testplugin.TestAgainExceptionRule$1.evaluate(TestAgainExceptionRule.java:39)
```
https://download.eclipse.org/eclipse/downloads/drops4/I20241215-1800/testresults/html/org.eclipse.ant.tests.ui_ep435I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

can be locally reproduced:
![image](https://github.com/user-attachments/assets/6777835c-027f-412d-842c-d2a816087818)

jukzi:
I could not figure out which commit it broke. But i see this may be related to one ore more commits of `[ANT] Only attempt to set SecurityManager when it is supported` across multiple repositories:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2623
@HannesWell

merks:
I think we are going to need adjustments now that we are running with Java 21.  I do see things like this in the log while running the test suite:
```
java.lang.UnsupportedOperationException: The Security Manager is deprecated and will be removed in a future release
	at java.base/java.lang.System.setSecurityManager(System.java:430)
	at org.eclipse.ant.internal.ui.model.AntModel.parseDocument(AntModel.java:367)
	at org.eclipse.ant.internal.ui.model.AntModel.reconcile(AntModel.java:302)
	at org.eclipse.ant.internal.ui.model.AntModel.getProjectNode(AntModel.java:1417)
	at org.eclipse.ant.internal.ui.model.AntModel.getProjectNode(AntModel.java:1425)
	at org.eclipse.ant.internal.ui.editor.text.AntFoldingStructureProvider.updateFoldingRegions(AntFoldingStructureProvider.java:92)
[...]

PR #1662, opened 2024-12-17
Guard all the places that ant sets the security manager

merks:
- This copies the code because org.eclipse.ant.internal.launching.remote.InternalAntRunner can't see (much of) anything else define by org.eclipse.ant.

https://github.com/eclipse-platform/eclipse.platform/issues/1660

### DIFF
--- a/ant/org.eclipse.ant.launching/remote/org/eclipse/ant/internal/launching/remote/InternalAntRunner.java
+++ b/ant/org.eclipse.ant.launching/remote/org/eclipse/ant/internal/launching/remote/InternalAntRunner.java
@@ -50,6 +50,7 @@ import org.apache.tools.ant.Target;
 import org.apache.tools.ant.Task;
 import org.apache.tools.ant.TaskAdapter;
 import org.apache.tools.ant.util.FileUtils;
+import org.apache.tools.ant.util.JavaEnvUtils;
 import org.eclipse.ant.internal.launching.remote.logger.RemoteAntBuildLogger;
 
 /**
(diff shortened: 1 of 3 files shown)


ant/org.eclipse.ant.launching/remote/org/eclipse/ant/internal/launching/remote/InternalAntRunner.java
```diff
@@ -58,6 +59,17 @@
  */
 public class InternalAntRunner {
 
+	private static final boolean IS_SECURITY_MANAGER_SUPPORTED = isSecurityManagerAllowed();
+
+	private static boolean isSecurityManagerAllowed() {
+		String sm = System.getProperty("java.security.manager"); //$NON-NLS-1$
+		if (sm == null) { // default is 'disallow' since 18 and was 'allow' before
+			return !JavaEnvUtils.isAtLeastJavaVersion("18"); //$NON-NLS-1$
```
laeubi, line 67: If we require Java 21 now it might be obsolete?!

merks: These tests are failing when running with Java 21.  Just because the IDE requires Java 21 doesn't mean an ant task requires Java 21 or even the Eclipse ant bundles for that matter. 

Anyway, it's a copy of this already-committed code:

https://github.com/eclipse-platform/eclipse.platform/blob/9636c7e790426c77388df64dc5f83ca37222ac86/ant/org.eclipse.ant.core/src_ant/org/eclipse/ant/internal/core/ant/InternalAntRunner.java#L86-L95

merks:
I think we need to move forward on this to stabilize and improve the builds.  If there are ways to improve these changes, with less duplication, we can do that as a follow up.

HannesWell:
> I think we need to move forward on this to stabilize and improve the builds. If there are ways to improve these changes, with less duplication, we can do that as a follow up.

Thanks for this! And for the follow-up, please have a look at:
- https://github.com/eclipse-platform/eclipse.platform/pull/1664

PR #1663, opened 2024-12-17
Infocenter webapp is Java 21 not 11

akurtakov:
It was wrongly claiming Java 11 compatibility while it was Java 17, with latest updates it's even Java 21.


akurtakov:
Warnings are unrelated to change in infocenter product.

PR #1664, opened 2024-12-17
[ANT] Remove obsolete comments and update ant.launching/remote.jar

HannesWell:
Also remove obsolete comments.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform/pull/1655
- https://github.com/eclipse-platform/eclipse.platform/pull/1662

Currently the [ANT remote.jar](https://github.com/eclipse-platform/eclipse.platform/blob/master/ant/org.eclipse.ant.launching/lib/remote.jar) pops-up as changed in my workspace and I didn't find an immediate hint that it's built in the CI.
So I guess it has to be updated and committed manually?


HannesWell:
The de-duplication didn't work because all the `InternalAntRunner` classes seem to go into separate jars.
Giving up on this.

merks:
> The de-duplication didn't work because all the `InternalAntRunner` classes seem to go into separate jars. Giving up on this.

Yes, welcome to my world!

merks:
Hmm.  I would have hoped the build would build them, but apparent not:

image

merks (approved):
Sorry I should have reverted the public thing after reuse failed.

vogella:
As the SDK requires now Java 21, maybe we could also update Ant to use Java 21 and remove the conditional check?

laeubi:
Only the help requires java 21

vogella:
> Only the help requires java 21

Yes and therefore the SDK requires Java 21 as it contains help. And we are open to move more components to Java 21, see recent PMC mail from @akurtakov.

vogella:
> This would mean users that have Java < 21 configured to run ant files will be broken.

I think that is similar to user who would like to run the SDK with Java < 21. This is also not possible anymore. 

Just to clarify: I don't feel strongly about updating Ant to Java 21 but from the discussion it sounded to me that supporting both is complex and time consuming so I wanted to add this option. As others are working on it I do not plan to mingle into their code changes.

merks:
Please let's not do unnecessary disruptive things!  Note that some of these things run inside an actual ant task running inside a separate user-configured JVM.   Note too that Ant itself still runs on Java 8:

image

iloveeclipse:
> I think that is similar to user who would like to run the SDK with Java < 21. This is also not possible anymore.

It is not! You can use Eclipse running on Java 21 but that doesn't mean all your products are supposed to use Java 21 now! This is the basics of IDE development!

HannesWell:
> Sorry I should have reverted the public thing after reuse failed.

Actually it was the other way round. I havn't removed it from my experiments but did it now.

> Hmm. I would have hoped the build would build them, but apparent not:
> 
As far as I can tell the current version of the remote.jar was compiled with `11.0.14+9 (Eclipse Adoptium)` (at least according to the MANIFEST.MF.)
Therefore I wonder for how long this jar has not been recompiled and if just relying on the rebuild from the workspace is the right thing. I guess the class files now target Java-17? And before probably targeted Java-11.

Can someone more familiar with ant respectively this part can check this advice the right thing to do?


> > I think that is similar to user who would like to run the SDK with Java < 21. This is also not possible anymore.
> 
> It is not! You can use Eclipse running on Java 21 but that doesn't mean all your products are supposed to use Java 21 now!

Yes.

iloveeclipse:
> I guess the class files now target Java-17? And before probably targeted Java-11.

I hope not, it should have been 1.7 or 1.8.
One can disassemble (javap or Bytecode Outline) class files to see to which JLS target they were compiled. JDK mentiomed in the Manifest doesn't mean anything as you can compile on Java 23 with target 1.8.

HannesWell:
> > I guess the class files now target Java-17? And before probably targeted Java-11.
> 
> I hope not, it should have been 1.7 or 1.8. One can disassemble (javap or Bytecode Outline) class files to see to which JLS target they were compiled. JDK mentiomed in the Manifest doesn't mean anything as you can compile on Java 23 with target 1.8.

That's right. I'll check that.

akurtakov:
This conflicts with master. Please rebase it.

akurtakov:
There doesn't seem to be interest in doing this one in half a year. Closing.

Issue #1668, opened 2024-12-19
API types to simplify typical tasks for launch configuration attributes


ruspl-afed:
During work with `ILaunchConfiguration` there are several typical tasks that are quite verbose and produce repeating code fragments, especially during Launch UI tabs programming and then handling the same launch attributes in `ILaunchDelegate` code.

The suggestion is to introduce a number of types to simplify these tasks.

PR is following

PR #1669, opened 2024-12-19
[#1668] API types to simplify work with launch configuration attributes

ruspl-afed:
* identify launch attribute
* connect it with preference metadata (to supply defaults/label/description)
* read attribute from configuration
* write attribute to configuration working copy

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeDefined.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeDefined.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2024 ArSysOp.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.core;
+
+import org.eclipse.core.runtime.preferences.PreferenceMetadata;
+
+/**
+ *
+ * The definition of {@link ILaunchConfiguration} attribute convenience to:
+ * <ul>
+ * <li>{@link ILaunchConfiguration#getAttribute(String, String)} and similar
+ * operations</li>
+ * <li>{@link ILaunchConfigurationWorkingCopy#setAttribute(String, String)} and
+ * similar operations</li>
+ * <li>Connecting {@link ILaunchConfiguration} attributes with preferences</li>
+ * <li>Representing {@link ILaunchConfiguration} attributes in UI</li>
+ * </ul>
+ *
+ * @see LaunchAttributeRead
+ * @see LaunchAttributeWrite
+ *
+ * @since 3.22
+ */
+public interface LaunchAttributeDefined<V> {
+
+	/**
+	 *
+	 * @return identity for defined attribute
+	 */
+	LaunchAttributeIdentity identity();
+
+	/**
+	 *
+	 * @return preference metadata for defined attribute
+	 */
+	PreferenceMetadata<V> metadata();
+
+}
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeRead.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeRead.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2024 ArSysOp.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.core;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.Status;
+
+/**
+ *
+ * Reads the value of launch attribute
+ *
+ * @since 3.22
+ */
+public final class LaunchAttributeRead<V> {
+
+	private final ILaunchConfiguration configuration;
+	private final String id;
+	private final Class<V> type;
+	private final Supplier<V> value;
+
+	public LaunchAttributeRead(ILaunchConfiguration configuration, LaunchAttributeDefined<V> defined) {
+		this(configuration, defined.identity().id(), defined.metadata().valueClass(), defined.metadata()::defaultValue);
+	}
+
+	public LaunchAttributeRead(ILaunchConfiguration configuration, String id, Class<V> type, Supplier<V> value) {
+		this.configuration = Objects.requireNonNull(configuration);
+		this.id = Objects.requireNonNull(id);
+		this.type = Objects.requireNonNull(type);
+		this.value = Objects.requireNonNull(value);
+	}
+
+	public V get() throws CoreException {
+		if (String.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, String.class.cast(value.get())));
+		}
+		if (Boolean.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Boolean.class.cast(value.get())));
+		}
+		if (Integer.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Integer.class.cast(value.get())));
+		}
+		if (List.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, List.class.cast(value.get())));
+		}
+		if (Map.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Map.class.cast(value.get())));
+		}
+		if (Set.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Set.class.cast(value.get())));
+		}
+		throw new CoreException(Status.error(id, new ClassCastException()));
+	}
+
+}
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeWrite.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeWrite.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2024 ArSysOp.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.debug.core;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+/**
+ *
+ * Writes the value of launch attribute. Useful in conjunction with UI.
+ *
+ * @since 3.22
+ */
+public final class LaunchAttributeWrite<V> implements Consumer<ILaunchConfigurationWorkingCopy> {
+
+	private final String id;
+	private final Class<V> type;
+	private final Supplier<V> value;
+
+	public LaunchAttributeWrite(LaunchAttributeDefined<V> defined) {
+		this(defined.identity().id(), defined.metadata().valueClass(), defined.metadata()::defaultValue);
+	}
+
+	public LaunchAttributeWrite(LaunchAttributeDefined<V> defined, Supplier<V> value) {
+		this(defined.identity().id(), defined.metadata().valueClass(), value);
+	}
+
+	public LaunchAttributeWrite(String id, Class<V> type, Supplier<V> value) {
+		this.id = Objects.requireNonNull(id);
+		this.type = Objects.requireNonNull(type);
+		this.value = Objects.requireNonNull(value);
+	}
+
+	@Override
+	public void accept(ILaunchConfigurationWorkingCopy working) {
+		if (String.class.equals(type)) {
+			working.setAttribute(id, String.class.cast(value.get()));
+		} else if (Integer.class.equals(type)) {
+			working.setAttribute(id, Integer.class.cast(value.get()).intValue());
+		} else if (Boolean.class.equals(type)) {
+			working.setAttribute(id, Boolean.class.cast(value.get()).booleanValue());
+		} else if (List.class.equals(type)) {
+			working.setAttribute(id, List.class.cast(value.get()));
+		} else if (Map.class.equals(type)) {
+			working.setAttribute(id, Map.class.cast(value.get()));
+		} else if (Set.class.equals(type)) {
+			working.setAttribute(id, Set.class.cast(value.get()));
+		} else {
+			working.setAttribute(id, value.get());
+		}
+	}
+
+}
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LauchAttributeIdentityRecord.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LauchAttributeIdentityRecord.java
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2024 ArSysOp.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.core;
+
+/**
+ * Default implementation for {@link LaunchAttributeIdentity}
+ *
+ * @since 3.22
+ */
+public record LauchAttributeIdentityRecord(String id) implements LaunchAttributeIdentity {
+
+	/**
+	 * Convenience way to compose full qualified name for launch attribute
+	 *
+	 * @param qualifier usually corresponds to Bundle-Symbolic-Name
+	 * @param key short key to name this very attribute in the scope of
+	 *            qualifier
+	 */
+	public LauchAttributeIdentityRecord(String qualifier, String key) {
+		this(qualifier + "." + key); //$NON-NLS-1$
+	}
+
+}
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeIdentity.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeIdentity.java
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2024 ArSysOp.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.core;
+
+/**
+ * Identifies an attribute in {@link ILaunchConfiguration}
+ *
+ * @since 3.22
+ */
+public interface LaunchAttributeIdentity {
+
+	/**
+	 * String id of {@link ILaunchConfiguration} attribute for "low-level"
+	 * operations
+	 *
+	 * @return id of attribute
+	 */
+	String id();
+
+}


mickaelistria:
As mentioned in the call, some example using this API proposal would be helpful. There is probably some content in org.eclipse.ui.externaltools/External Tools Base/org/eclipse/ui/externaltools/internal/launchConfigurations you can tweak to leverage this API.

ruspl-afed:
I'm not sure why it is is failing @HannesWell 
```
11:59:41.178 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.11-SNAPSHOT:verify (verify) on project org.eclipse.debug.core: 
Execute ApiApplication failed: InvocationTargetException: EmptyStackException -> [Help 1]
```

HannesWell:
> I'm not sure why it is is failing @HannesWell
> 
> ```
> 11:59:41.178 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.11-SNAPSHOT:verify (verify) on project org.eclipse.debug.core: 
> Execute ApiApplication failed: InvocationTargetException: EmptyStackException -> [Help 1]
> ```

It looks like your change is breaking API-tools^^
I have replayed the Jenkins build the `-e` set to get an more informative stack-trace:
https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-1669/7/

ruspl-afed:
Thank you @HannesWell for your assistance.

> It looks like your change is breaking API-tools^^

WOW! But that wasn't part of my plan. It looks like [this issue](https://github.com/eclipse-pde/eclipse.pde/issues/1575) was already reported. Interesting, may be I can have a deeper look to it soon.

And still interested in your feedback regarding PR content.

ruspl-afed:
After [supporting `record` for API Tools](https://github.com/eclipse-pde/eclipse.pde/pull/1598) everything is green @HannesWell

ruspl-afed:
ok, a few more days for discussion and I'm going to merge it

HannesWell:
> After [supporting `record` for API Tools](https://github.com/eclipse-pde/eclipse.pde/pull/1598) everything is green @HannesWell

Thank you for that fix/enhancement of API-tools! It's much appreciated.

> And still interested in your feedback regarding PR content.

Sorry for not responding. It has been some busy days recently. I have not forgotten this, just didn't have the time for it.
I try my best to have look at it in the next days.

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/launchConfigurations/LaunchAttributeArguments.java
```diff
@@ -0,0 +1,36 @@
...
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.externaltools.internal.launchConfigurations;
+
+import org.eclipse.core.externaltools.internal.IExternalToolConstants;
+import org.eclipse.core.runtime.preferences.PreferenceMetadata;
+import org.eclipse.debug.core.LaunchAttributeDefined;
+import org.eclipse.debug.core.LaunchAttributeIdentity;
+
+public final class LaunchAttributeArguments implements LaunchAttributeDefined<String> {
```
HannesWell, line 21: Couldn't we replace this entire class by a static field `LaunchAttributeDefined<String>LAUNCH_ATTRIBUTE_ARGUMENTS` e.g in the `IExternalToolConstants` class? Implemented either with an anonymous implementation for each field or a static factory method (that contains it).
I wonder the same for the companion `LaunchAttributeDefined` implementations.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeRead.java
```diff
@@ -0,0 +1,71 @@
...
+		this.configuration = Objects.requireNonNull(configuration);
+		this.id = Objects.requireNonNull(id);
+		this.type = Objects.requireNonNull(type);
+		this.value = Objects.requireNonNull(value);
+	}
+
+	public V get() throws CoreException {
+		if (String.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, String.class.cast(value.get())));
+		}
```
HannesWell, line 52: Couldn't we have this if-chain in the constructor and then encapsulate the action into a lambda and just execute it here?

HannesWell (changes requested):
I had a more detailed look at it and understand the goal and see the benefits from this: Better type-safety and more convenient read and writes.
Nevertheless I'm not sure that these benefits really carry the weight of the change, because currently the new API surface is very big.
Below I made some suggestion to make the API more compact, maybe we can then achieve a sufficiently got relation of advantages to API surface.

With these suggestions we would end up with just the `LaunchAttributeIdentity` and `LaunchAttributeDefined` types. Maybe we could even 'inline' the former into the latter and would then still get benefits mentioned above.
And eventually just having an `ILaunchAttribute` interface added with a few static factories to create them is something that could indeed be a in my opinion suitable addition.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeRead.java
```diff
@@ -0,0 +1,71 @@
...
+			return type.cast(configuration.getAttribute(id, List.class.cast(value.get())));
+		}
+		if (Map.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Map.class.cast(value.get())));
+		}
+		if (Set.class.equals(type)) {
+			return type.cast(configuration.getAttribute(id, Set.class.cast(value.get())));
+		}
+		throw new CoreException(Status.error(id, new ClassCastException()));
+	}
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeWrite.java
```diff
@@ -0,0 +1,69 @@
...
+		} else if (List.class.equals(type)) {
+			working.setAttribute(id, List.class.cast(value.get()));
+		} else if (Map.class.equals(type)) {
+			working.setAttribute(id, Map.class.cast(value.get()));
+		} else if (Set.class.equals(type)) {
+			working.setAttribute(id, Set.class.cast(value.get()));
+		} else {
+			working.setAttribute(id, value.get());
+		}
+	}
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/launchConfigurations/LaunchAttributeArguments.java
ruspl-afed: done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeRead.java
ruspl-afed: not so sure

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/LaunchAttributeDefined.java
```diff
@@ -0,0 +1,49 @@
...
+ * <li>Connecting {@link ILaunchConfiguration} attributes with preferences</li>
+ * <li>Representing {@link ILaunchConfiguration} attributes in UI</li>
+ * </ul>
+ *
+ * @see LaunchAttributeRead
+ * @see LaunchAttributeWrite
+ *
+ * @since 3.23
+ */
+public interface LaunchAttributeDefined<V> {
```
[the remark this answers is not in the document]
ruspl-afed: done

ruspl-afed:
Thank you for your feedback @HannesWell 

PR was reimplemented according to your comments.
I'm not sure why we are fighting for reducing number of API types with adding more API methods and more implementation code the the interface compilation unit, but the result looks much more closer to the old good Eclipse 2.0 style of Debug subsystem.

HannesWell:
> I'm not sure why we are fighting for reducing number of API types with adding more API methods and more implementation code the the interface compilation unit, but the result looks much more closer to the old good Eclipse 2.0 style of Debug subsystem.

I don't consider it as a fight, more as a search for (ideally) the best solution. Of course what is best is also a matter of opinion, but at least my personal opinion is that in general I like more compact APIs better. However can you please explain to me what the advantage is of defining in types what could be methods respectively why exactly this is Eclipse-2.0 style and why it's bad (I haven't been around back then)?
With the suggestions made we could still define typed launch-attributes using `LaunchAttribute(-Defined/Definition)` as static constant and read/write with it from/to a `LaunchConfiguration` in a more convenient way, couldn't we?

ruspl-afed:
You are right @HannesWell , it's definitely a matter of personal opinion. For me the old good Eclipse 2.0 style means a lot of static definitions, favoring of primitive types in API and a lot of setters to reuse one instance as long as one can. All this was actual 25 years ago for ancient JVMs. Nowadays we have modern JVM and very dynamic environment where non-replaceable static definition can cause problems and saving memory due to mutability of objects does not justify the cost of (much more complex) code support.

And even if I've stopped naming interfaces with "I" for new code and prefer simple immutable objects with very simple interfaces, I've done a lot following your comments to make this API improvement better suited to the surrounding code.

Now the question is: what else should be improved in this PR? I still can see your "change requested" veto but your latest comment doesn't content any direct change requests. Please clarify.

ruspl-afed:
Hello @HannesWell 
What else do you think should be changed in this PR?

HannesWell (commented):
> You are right @HannesWell , it's definitely a matter of personal opinion. For me the old good Eclipse 2.0 style means a lot of static definitions, favoring of primitive types in API and a lot of setters to reuse one instance as long as one can. All this was actual 25 years ago for ancient JVMs. Nowadays we have modern JVM and very dynamic environment where non-replaceable static definition can cause problems and saving memory due to mutability of objects does not justify the cost of (much more complex) code support.

My concern is not about a few objects more or less, I look at these new types from a consumer-perspective and how difficult a new API is to understand and learn. And there my experience/opinion is that 'distributed' types are more difficult and having it compact makes the capabilities easier to discover (as an IMO bad example: try to configure a JFace resource manager properly to be disposed with a Control and a child of the global registry).
As far as I know the problematic part of the Eclipse 2.0 style you referring to is that there tend to be only one large class that provided almost everything as static methods that don't allow dynamics, e.g. the `Platform` class in eclipse.runtime.
But at least from my perspective we are here talking about static factories on types (classes or interfaces), which I don't consider old-style. In fact I have the impression that's often the preferred way over constructors and many are added in Eclipse. From the top of my head I remember e.g. `ILog` or `IPath` or in the JDK there is `Path.of()` or the brand-new Classfile-API also makes heavy use of static factory methods AFAIK.

> And even if I've stopped naming interfaces with "I" for new code and prefer simple immutable objects with very simple interfaces

Personally I'm also not fully convinced that it's best to prefix all interfaces `I`. But in the end Eclipse is an existing code-base and when adding elements considering how new code fits into the existing code-base is IMO also an item to consider. Of course that doesn't make it simpler. Immutability is also something I think has many advantages and I try to achieve. For simple interfaces on the other hand, I would say it depends and that's IMO where the work starts and were the quality of an API is measured, if the 'right' size is found.

> 
> Now the question is: what else should be improved in this PR? I still can see your "change requested" veto but your latest comment doesn't content any direct change requests. Please clarify.

Sorry, I simply oversaw that you made changes. I was very busy of getting rid of the JXPath dependency in platform.ui.
I have now made another pass, please see the comments below.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,187 @@
...
+	 * @param working the launch configuration working copy to write value to,
+	 *            must not be <code>null</code>
+	 * @param value the value supplier, must not be <code>null</code>
+	 */
+	void write(ILaunchConfigurationWorkingCopy working, Supplier<V> value);
+
+	/**
+	 * Factory to create {@link ILaunchAttribute} objects
+	 */
+	interface Factory {
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,187 @@
...
+		<V> ILaunchAttribute<V> create(PreferenceMetadata<V> metadata);
+
+		/**
+		 * creates an instance of {@link ILaunchAttribute} using
+		 * primitives (useful for existing code)
+		 *
+		 * @param <V> the value type for the launch attribute
+		 * @param id the id of the attribute, must not be <code>null</code>
+		 * @param type the value type of the attribute, must not be
+		 *            <code>null</code>
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttributeIdentity.java
```diff
@@ -0,0 +1,61 @@
...
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.core;
+
+/**
+ * Identifies an attribute in {@link ILaunchConfiguration}
+ *
+ * @since 3.23
+ */
+public interface ILaunchAttributeIdentity {
```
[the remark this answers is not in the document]
ruspl-afed: This is funny. The introduction of "typed" identifier like `ILaunchAttributeIdentity` was the main goal of this change. The idea is to reduce programming errors when we have a lot of launch-related `String` constants without good way to distinguish them according to its type.

If you think that the ability to check at compile time is not worth introducing a new type, I'm not sure how to argue with that.

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/IExternalToolConstants.java
```diff
@@ -11,9 +11,16 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     dakshinamurthy.karra@gmail.com - bug 165371
+ *     Alexander Fedorov (ArSysOp) - https://github.com/eclipse-platform/eclipse.platform/issues/1668
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,187 @@
...
+		 * @param value the {@link Supplier} of default value of the attribute,
+		 *            must not be <code>null</code>
+		 * @return created instance of {@link ILaunchAttribute}
+		 */
+		<V> ILaunchAttribute<V> create(String id, Class<V> type, Supplier<V> value);
```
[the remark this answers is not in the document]
ruspl-afed: Since `PreferenceMetadata` instance is constructed internally, `Function` interface is hard to use because we have no argument for it.
I need PDE code example to understand this use case better @HannesWell

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,187 @@
...
+	Optional<V> read(ILaunchConfiguration configuration);
+
+	/**
+	 * Writes the supplied value to the launch configuration working copy
+	 *
+	 * @param working the launch configuration working copy to write value to,
+	 *            must not be <code>null</code>
+	 * @param value the value supplier, must not be <code>null</code>
+	 */
+	void write(ILaunchConfigurationWorkingCopy working, Supplier<V> value);
```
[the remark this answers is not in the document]
ruspl-afed: done

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/IExternalToolConstants.java
```diff
@@ -221,4 +241,37 @@ public interface IExternalToolConstants {
...
 	 */
 	String ATTR_INCLUDE_REFERENCED_PROJECTS = UI_PLUGIN_ID + ".ATTR_INCLUDE_REFERENCED_PROJECTS"; //$NON-NLS-1$
+
+	/**
+	 * {@link ILaunchAttribute<String>} identifying the location of an external.
+	 * Default value is <code>null</code>. Encoding is tool specific.
+	 */
+	ILaunchAttribute<String> LAUNCH_ATTRIBUTE_LOCATION = ILaunchAttribute.DEFAULT.create(//
+			new PreferenceMetadata<>(String.class, //
+					ILaunchAttributeIdentity.DEFAULT.create(ATTR_LOCATION).id(), //
```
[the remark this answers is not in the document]
ruspl-afed: reworked

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * Reads the value of launch configuration attribute. <br/>
+	 * Returns {@link Optional} in case of read failure of <code>null</code>
+	 * value to let caller decide how to proceed
+	 *
+	 * @param configuration the launch configuration to read attribute from,
+	 *            must not be <code>null</code>
+	 * @return the {@link Optional} with attribute value
+	 */
+	Optional<V> read(ILaunchConfiguration configuration);
```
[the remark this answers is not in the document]
ruspl-afed: yes, we have different use cases and they were covered by the pair `probe`/`read` in the initial PR version
now I'm returning this pair of methods back
1) `probe` - mostly to use from UI or other places where we are not interested to fail or receive `null`
2) `read` - for fail-fast control flows

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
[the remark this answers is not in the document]
ruspl-afed: ok, let's forget about `PreferenceMetadata`
done

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttributeIdentity.java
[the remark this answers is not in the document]
HannesWell: > This is funny. The introduction of "typed" identifier like `ILaunchAttributeIdentity` was the main goal of this change. 

It's not uncommon that a PR changes a lot over time. :)

> The idea is to reduce programming errors when we have a lot of launch-related `String` constants without good way to distinguish them according to its type.

Yes absolutely, and I think that this goal is still achieved. It's just that I'm not aware of any type that just wraps e.g. an Integer and does not have additional methods defined to work with it. E.g. `File`, `Path`, `URL`, `URI`, `Charset`, even `Integer` or `String` itself provide many methods to work with the wrapped element.

And since fixing API mistakes or changing API is usually very cumbersome and sometimes even a huge PITA, I think it's important to pay extra attention when designing new APIs and IMO it should be as small as possible and as large as necessary.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,170 @@
...
+		return new ConsiderStringVariables<>(identity, type, value, name, name);
+	}
+
+	/**
+	 * Default implementation for {@link ILaunchAttribute} that tries to
+	 * substitute variables for launch attributes with {@link String} type
+	 *
+	 * @see IStringVariableManager
+	 */
+	record ConsiderStringVariables<V>(ILaunchAttributeIdentity identity, Class<V> type, Function<ILaunchConfiguration, V> defaults, String name, String description) implements ILaunchAttribute<V> {
```
HannesWell, line 106: Records should really be used with great caution in APIs, because as far as I understand it, even just adding a component could be a binary incompatible change when considering newer lange-features like pattern matching.

So I would just make this a local class/record of the static `of` factory

HannesWell (commented):
Thanks for the update.
I didn't had the time for a detailed review. I still think we should handle the case where the default value depends on other attributes, because IIRC we have this case multiple times in PDE and I would like to have this covered. But I'll provide a specific example for that.

I'll be on vacation now, but when back I'll come back to this. Sorry for the further delay.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
ruspl-afed: extracted to package local type

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttributeIdentity.java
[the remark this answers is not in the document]
ruspl-afed: You won :) `ILaunchAttributeIdentity` was removed. 
I wonder why with all the repeating code around identifiers here and there you don't want to invest to a common notation to think less for routine tasks.

ruspl-afed:
@HannesWell I changed it to a state that we discussed on a meeting:
1) remove `ILaunchAttributeIdentity`
2) minimize API surface

ruspl-afed:
@HannesWell do you have more comments on this one?

HannesWell:
> @HannesWell do you have more comments on this one?

I didn't had the time yet to look into this, but definitely want to. 
Try to respond until the end of this week and will also look how to use this for PDE.

HannesWell:
> > @HannesWell do you have more comments on this one?
> 
> I didn't had the time yet to look into this, but definitely want to. Try to respond until the end of this week and will also look how to use this for PDE.

I only had the time for a short look at this change and in general I think this looks very good now. But I'd like to do an in depth review and try my best to do it tomorrow or Wednesday.

ant/org.eclipse.ant.ui/Ant Tools Support/org/eclipse/ant/internal/ui/launchConfigurations/AntMainTab.java
```diff
@@ -45,44 +47,27 @@
...
 	@Override
 	public void initializeFrom(ILaunchConfiguration configuration) {
 		super.initializeFrom(configuration);
-		try {
-			fCurrentLocation = configuration.getAttribute(IExternalToolConstants.ATTR_LOCATION, (String) null);
-		}
-		catch (CoreException e) {
-			// do nothing
-		}
+		fCurrentLocation = IExternalToolConstants.LAUNCH_ATTRIBUTE_LOCATION.probe(configuration);
```
HannesWell, line 57: Before `fCurrentLocation` was left unchanged, if `getAttribute()` threw. But as the code below also handles empty optional, I think this change is fine.

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/IExternalToolConstants.java
```diff
@@ -30,71 +35,77 @@ public interface IExternalToolConstants {
 	String EMPTY_STRING = ""; //$NON-NLS-1$
 
 	/**
-	 * Plug-in identifier for external tools UI (value <code>org.eclipse.ui.externaltools</code>).
+	 * Plug-in identifier for external tools UI (value
+	 * <code>org.eclipse.ui.externaltools</code>).
 	 */
-	String UI_PLUGIN_ID = "org.eclipse.ui.externaltools"; //$NON-NLS-1$;
+	String UI_PLUGIN_ID = "org.eclipse.ui.externaltools"; //$NON-NLS-1$ ;
```
HannesWell, line 41: Can you please remove the unnecessary trailing semicolon here and below?
```suggestion
	String UI_PLUGIN_ID = "org.eclipse.ui.externaltools"; //$NON-NLS-1$
```

debug/org.eclipse.core.externaltools/src/org/eclipse/core/externaltools/internal/launchConfigurations/ExternalToolsProgramMessages.java
```diff
@@ -29,6 +30,15 @@ public class ExternalToolsProgramMessages extends NLS {
...
 	public static String ExternalToolsUtil_invalidDirectory__0_;
 
+
+	public static String LaunchAttributeArguments_name;
+
+
+	public static String LaunchAttributeLocation_name;
+
+
+	public static String LaunchAttributeWorkingDirectory_name;
```
HannesWell, line 40: Please avoid excessive new lines.
```suggestion
	public static String LaunchAttributeArguments_name;
	public static String LaunchAttributeLocation_name;
	public static String LaunchAttributeWorkingDirectory_name;
```

debug/org.eclipse.ui.externaltools/External Tools Base/org/eclipse/ui/externaltools/internal/model/IExternalToolConstants.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2013 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
```
HannesWell, line 2: I cannot see any relevant change in this class, beside white-space changes.
If that's correct, please leave this class untouched.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,93 @@
...
+	/**
+	 * "Probes" the value of launch configuration attribute. <br/>
+	 * Returns {@link Optional} in case of read failure or <code>null</code>
+	 * value to let caller decide how to proceed
+	 *
+	 * @param configuration the launch configuration to read attribute from,
+	 *            must not be <code>null</code>
+	 * @return the {@link Optional} with attribute value
+	 */
+	Optional<V> probe(ILaunchConfiguration configuration);
```
HannesWell, line 64: I have to admit that I found the method name `probe` a bit unusual (and I'm not even sure it fit's perfectly here).
What about just naming this method to the ordinary name `get()`? With read one could somehow expect that it could fail.
```suggestion
	Optional<V> get(ILaunchConfiguration configuration);
```

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @param working the launch configuration working copy to write value to,
+	 *            must not be <code>null</code>
+	 * @param value the value, may be <code>null</code>
+	 */
+	void write(ILaunchConfigurationWorkingCopy working, V value);
+
+	/**
+	 * creates an instance of {@link ILaunchAttribute} using primitives (useful
+	 * for existing code)
```
HannesWell, line 77:
```suggestion
	 * Creates an instance of {@link ILaunchAttribute} using primitives (useful
	 * for existing code).
```

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @param <V> the value type for the launch attribute
+	 * @param id the id of the attribute, must not be <code>null</code>
+	 * @param type the value type of the attribute, must not be
+	 *            <code>null</code>
```
HannesWell, line 82: Instead of ``<code>null</code>` I generally would just use `{@code null}`.
```suggestion
	 * @param type the type of the attribute value, must not be {@code null}
```

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,93 @@
...
+	 * @param value the {@link Function} to calculate of default value of the
+	 *            attribute from {@link ILaunchConfiguration}, must not be
+	 *            <code>null</code>
```
HannesWell, line 85: Maybe name this parameter `defaultValue`?
```suggestion
	 * @param defaultValue the {@link Function} to calculate the default value of the
	 *            attribute for a given {@link ILaunchConfiguration}, must not be
	 *            {@code null}.
```
As it's mentioned very often that an attribute may not be `null`, what about stating that once in the interface description and only mention the cases where null is allowed (I assume nowhere?)

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ILaunchAttribute.java
```diff
@@ -0,0 +1,93 @@
...
+	 * @param name the name of the attribute, must not be <code>null</code>
+	 * @return created instance of {@link ILaunchAttribute}
+	 */
+	static <V> ILaunchAttribute<V> of(String id, Class<V> type, Function<ILaunchConfiguration, V> value, String name) {
+		return new ConsiderStringVariables<>(id, type, value, name, name);
+	}
```
HannesWell, line 91: As many callers just specify `lc -> null` as default value, what do you think about providing two more overloads for convenience for attributes that have a fixed default and no default:
- `of(String id, Class<V> type, V defaultValue, String name)`
- `of(String id, Class<V> type, String name)`

They can then just call the main `of` method with a corresponding lambda.

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ConsiderStringVariables.java
```diff
@@ -0,0 +1,95 @@
...
+			working.setAttribute(id, Boolean.class.cast(value).booleanValue());
+		} else if (List.class.equals(type)) {
+			working.setAttribute(id, List.class.cast(value));
+		} else if (Map.class.equals(type)) {
+			working.setAttribute(id, Map.class.cast(value));
+		} else if (Set.class.equals(type)) {
+			working.setAttribute(id, Set.class.cast(value));
+		} else {
+			working.setAttribute(id, value);
+		}
```
HannesWell, line 93: Is the distinction for the actual type actually necessary? AFAICT all special methods delegate to the object accepting variant any ways?
```suggestion
		working.setAttribute(id, value);
```
Similarly, although I haven't checked it in detail, the `readInternal` method maybe can be simplified too?

debug/org.eclipse.debug.core/core/org/eclipse/debug/core/ConsiderStringVariables.java
```diff
@@ -0,0 +1,95 @@
...
+import org.eclipse.core.variables.IStringVariableManager;
+import org.eclipse.core.variables.VariablesPlugin;
+
+/**
+ * Default implementation for {@link ILaunchAttribute} that tries to
+ * substitute variables for launch attributes with {@link String} type
+ *
+ * @see IStringVariableManager
+ */
+record ConsiderStringVariables<V>(String id, Class<V> type, Function<ILaunchConfiguration, V> defaults, String name, String description) implements ILaunchAttribute<V> {
```
HannesWell, line 34: Name and description seem unused atm.
```suggestion
record ConsiderStringVariables<V>(String id, Class<V> type, Function<ILaunchConfiguration, V> defaultValue) implements ILaunchAttribute<V> {
```

HannesWell (commented):
Thank you for the update.

Conceptually I'm really happy with it now and for the API I just have one suggestion about the name of the probe method. And we should consider to have more factories to avoid verbosity in simple cases.

Everything else are just minor internal implementation details.

PR #1670, opened 2024-12-20
runtime tests: throw dedicated TestException

jukzi:
To make it easier to filter test output for unintended exceptions.


jukzi:
windows verification failed with "java.lang.OutOfMemoryError: Java heap space"

PR #1673, opened 2025-01-07
Increase lower bound to re-exported Equinox Common

ptziegler:
New API has been added to the StringMatcher, which is going to be used in the Platform UI bundles.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/2672


ptziegler:
The micro version was updated with https://github.com/eclipse-platform/eclipse.platform/commit/3776c2f194dbb8b1132f2e7bb47f4c06a557a0ff as part of the 4.35 stream, but a change in the re-exported bundles requires a minor version increase.

iloveeclipse:
> New API has been added to the StringMatcher

See https://github.com/eclipse-equinox/equinox/pull/720

PR #1674, opened 2025-01-10
Don't use instance of for simple null checks

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739


laeubi:
Ignore infocenter OOM problem here (Github do not show any problems).

PR #1675, opened 2025-01-10
Move o.e.debug.tests to Java 21

akurtakov:
Needed by
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739 Fixed deprecated usage of Thread.getId.


akurtakov:
New warnings is due to URL constructor being deprecated but there is no clear replacement for it.

PR #1676, opened 2025-01-10
Move bundles with Java 21 in the dependency tree to that BREE

akurtakov:
Fixes warnings like:
```
Execution environment is lower than one of the plug-in's dependencies
(org.eclipse.core.tools) which has an execution environment of
JavaSE-21.	MANIFEST.MF	/org.eclipse.core.tools.resources/META-INF	line
19	Plug-in Problem
```


akurtakov:
New warnings are due to URL contstructors being deprecated but there is not 1:1 replacement so to be handled separately.

PR #1679, opened 2025-01-13
Use Java NIO in tests

akurtakov:


jukzi:
I generally like usage of Files.*. One just has to be careful about the default charset used.

PR #1680, opened 2025-01-13
Remove dispose call for Colors in AntPreviewerUpdater

vogella:
According to the Javadoc on Colors:

 * Colors do not need to be disposed, however to maintain compatibility
 * with older code, disposing a Color is not an error.


akurtakov:
There is a bit more to be cleaned up here.

Issue #894, opened 2023-11-29
Random failing org.eclipse.core.internal.expressions.tests.ExpressionTests.testSubTypeTiming 


HeikoKlare:
Failed once on MacOS: https://github.com/eclipse-platform/eclipse.platform/pull/892/checks?check_run_id=19130535605

```
cachedDelta: 90, instanceDelta: 73
java.lang.AssertionError: cachedDelta: 90, instanceDelta: 73
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.core.internal.expressions.tests.ExpressionTests.testSubTypeTiming(ExpressionTests.java:1135)
```

akurtakov:
Is this really timing measured ? If that's the case the test needs to be changed to check results and not timing.

HeikoKlare:
> Is this really timing measured ? If that's the case the test needs to be changed to check results and not timing.

It seems to be an attempt to provide a performance test for a caching mechanism. So you are right: it has to be rewritten to make a functional validation of the caching rather than a time-based one.

jukzi:
mac
```
cachedDelta: 148, instanceDelta: 74
java.lang.AssertionError: cachedDelta: 148, instanceDelta: 74
	at org.eclipse.core.internal.expressions.tests.ExpressionTests.testSubTypeTiming(ExpressionTests.java:1135)
```
https://github.com/eclipse-platform/eclipse.platform/pull/918/checks?check_run_id=19294549887

jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20250116-1800/testresults/html/org.eclipse.core.expressions.tests_ep435I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

```
java.lang.AssertionError: [assert cachedDelta is less than instanceDelta43]
Expecting actual:
55L
to be less than:
43L
at org.eclipse.core.internal.expressions.tests.ExpressionTests.testSubTypeTiming(ExpressionTests.java:1036)
```

PR #1692, opened 2025-01-17
disable flaky ExpressionTests.testSubTypeTiming() #894

jukzi:
https://github.com/eclipse-platform/eclipse.platform/issues/894


akurtakov:
It doesn't look like there are any asserts that make sense for anything but the time measurement. 
The test itself looks totally wrong as the "caching" version definitively can be slower on the first run as it populates the caches and even installs a BundleListener(!!!).
I would be in favor of totally removing the test after looking into it.

jukzi:
Locally running it works and performs a performancetest, so it could be used as such

akurtakov:
What I say is that even if fails locally that doesn't mean there is a problem. Please merge as is if you believe it's useful for smth.

jukzi:
i used it to verify the current implementation is faster then using Class::isAssignableFrom :-)

PR #1700, opened 2025-01-20
Stop maven-war-plugin from recompressing zip files

akurtakov:
There is no need for that and it should (hopefully) reduce the cases where build OOMs.


akurtakov:
Windows test failure totally unrelated to this changes.

PR #1702, opened 2025-01-21
Correct class name in extension point documentation

howlger:
In the extension point `org.eclipse.help.searchProcessor`, the supplied search processor class must extend the class `AbstractSearchProcessor` which is in the package `org.eclipse.help.search`, not in `org.eclipse.help`.


jukzi:
Hi @howlger, thanks for contribution

PR #1706, opened 2025-01-22
[Dark Theme]Fix for background color of bread crumbs in Debug view

mvm-sap:
The background color of bread-crumb items did not match the background color in the debug view, which has been fixed with this PR. Please take a look at the before and after screenshots below for the same.

Before:
Screenshot 2025-01-22 145235

After:
Screenshot 2025-01-22 150449


BeckerWdf:
Tested on macOS. Looks good. Thanks for fixing.

BeckerWdf (approved):

Issue #1684, opened 2025-01-15
[win][21] XmlDocumentFormatterTest fails since I20250113-1800


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20250113-1800/testresults/html/org.eclipse.ant.tests.ui_ep435I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

```
org.junit.ComparisonFailure: expected:<...<!-- = = = = = = -->[
<!-- formatTest -->
<!-- = = = = = = -->
<target name="formatTest" depends="init" description="--> test target one">
<!-- hello world -->
<greeting />
</target>

<!-- - - - - - - -->
<!-- init -->
<!-- - - - - - - -->
<target name="init">
<property name="greeting" value="hi" />
<presetdef name="greeting">]
<echo>${gr...> but was:<...<!-- = = = = = = -->[
<!-- formatTest -->
<!-- = = = = = = -->
<target name="formatTest" depends="init" description="--> test target one">
<!-- hello world -->
<greeting />
</target>

<!-- - - - - - - -->
<!-- init -->
<!-- - - - - - - -->
<target name="init">
<property name="greeting" value="hi" />
<presetdef name="greeting">
]
<echo>${gr...>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.ant.tests.ui.editor.formatter.XmlDocumentFormatterTest.simpleTest(XmlDocumentFormatterTest.java:100)
at org.eclipse.ant.tests.ui.editor.formatter.XmlDocumentFormatterTest.testGeneralFormat(XmlDocumentFormatterTest.java:44)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at org.eclipse.ant.tests.ui.testplugin.TestAgainExceptionRule$1.evaluate(TestAgainExceptionRule.java:39)
```

```
org.junit.ComparisonFailure: expected:<...<!-- = = = = = = -->[
<!-- formatTest -->
<!-- = = = = = = -->
<target name="formatTest" depends="init" description="--> test target one">
<!-- hello world -->
<greeting />
</target>
<!-- - - - - - - -->
<!-- init -->
<!-- - - - - - - -->
<target name="init">
<property name="greeting" value="hi" />
<presetdef name="greeting">]
<echo>${greeting...> but was:<...<!-- = = = = = = -->[
<!-- formatTest -->
[...]

PR #1712, opened 2025-01-23
Pay attention to system.lineseparator in XML formatter tests

akurtakov:
Use the helper method to unify on the system line separator.


akurtakov:
@jukzi Would you please that it fixes https://github.com/eclipse-platform/eclipse.platform/issues/1684 as I don't have access to windows?

jukzi:
As i am not able to reproduce the issue locally i can not verify this will fix it. i'd say just merge to try -  if verification builds succeeds. Meanwhile i wonder why the test did not fail in verification tests but only on i-build. The the i-build use another git autcrlf setting then the verification builds?

akurtakov:
That is a question which I can not answer.

akurtakov:
Merged, let's see next I-build results.

PR #1713, opened 2025-01-23
try core.autocrlf false in ci.yml

jukzi:


jukzi:
`reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps` don't know how to disable autocrlf

PR #1714, opened 2025-01-24
Fix NPE in switch case

laeubi:


team/bundles/org.eclipse.core.net/src/org/eclipse/core/internal/net/ProxyManager.java
```diff
@@ -66,11 +66,13 @@ public class ProxyManager implements IProxyService, IPreferenceChangeListener {
...
-			nativeProxyProvider = switch (os) {
-			case Constants.OS_LINUX -> new ProxyProviderLinux();
-			case Constants.OS_WIN32 -> new ProxyProviderWin32();
-			default -> null;
-			};
+			if (os != null) {
+				nativeProxyProvider = switch (os) {
+				case Constants.OS_LINUX -> new ProxyProviderLinux();
+				case Constants.OS_WIN32 -> new ProxyProviderWin32();
+				default -> null;
```
Bananeweizen, line 73: @laeubi you could also just write ...`case null, default -> null;`. The switch can deal with null, if it's declared in the cases.

laeubi: I'm not sure if its not only the case on Java 21,... but even `System.getProperty("osgi.os", "")` would be possible... so many options :-D

Bananeweizen: right, I sometimes forget that not everyone is on 21 yet. :)

PR #1719, opened 2025-01-31
Enhance the IContributionFactory to filter contributions

laeubi:
Currently there is no way for filtering a contribution to the model on a global level.

This adds a new method to IContributionFactory#isEnabled that allows extenders to mark a given contribution as being disabled.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2217

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.services/src/org/eclipse/e4/core/services/contributions/IContributionFactory.java
+++ b/runtime/bundles/org.eclipse.e4.core.services/src/org/eclipse/e4/core/services/contributions/IContributionFactory.java
@@ -16,8 +16,10 @@ package org.eclipse.e4.core.services.contributions;
 import org.eclipse.e4.core.contexts.IEclipseContext;
 import org.osgi.framework.Bundle;
 
-// TBD this became an utility method to create object from a bundle.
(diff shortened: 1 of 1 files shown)


runtime/bundles/org.eclipse.e4.core.services/src/org/eclipse/e4/core/services/contributions/IContributionFactory.java
```diff
@@ -26,4 +28,8 @@ public interface IContributionFactory {
 
 	Bundle getBundle(String uriString);
 
+	default boolean isEnabled(String uriString) {
```
jukzi, line 31: I wonder why this added API did not need a `@since` javadoc

laeubi: Because it is only preliminary (non public) API

jukzi: Thanks, i did not have in mind x-friends declares provisional APIs

Issue #1721, opened 2025-02-03
Random IndexOutOfBoundsException in PropertyBucket.readEntryValue


jukzi:
```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
at java.base/java.util.Objects.checkIndex(Objects.java:385)
at java.base/java.util.ArrayList.get(ArrayList.java:427)
at org.eclipse.core.internal.properties.PropertyBucket.readEntryValue(PropertyBucket.java:290)
at org.eclipse.core.internal.localstore.Bucket.loadEntries(Bucket.java:378)
at org.eclipse.core.internal.localstore.Bucket.load(Bucket.java:343)
at org.eclipse.core.internal.properties.PropertyBucket.load(PropertyBucket.java:274)
at org.eclipse.core.internal.localstore.Bucket.load(Bucket.java:310)
at org.eclipse.core.internal.localstore.BucketTree.loadBucketFor(BucketTree.java:128)
at org.eclipse.core.internal.properties.PropertyManager2.setProperty(PropertyManager2.java:183)
at org.eclipse.core.internal.resources.Resource.setPersistentProperty(Resource.java:1725)
at org.eclipse.core.internal.resources.ContentDescriptionManager.setCacheState(ContentDescriptionManager.java:556)
at org.eclipse.core.internal.resources.ContentDescriptionManager.invalidateCache(ContentDescriptionManager.java:453)
at org.eclipse.core.internal.resources.ContentDescriptionManager.handleEvent(ContentDescriptionManager.java:549)
at org.eclipse.core.internal.resources.Workspace.broadcastEvent(Workspace.java:460)
at org.eclipse.core.internal.resources.Resource.broadcastPreDeleteEvent(Resource.java:1893)
at org.eclipse.core.internal.resources.Resource.delete(Resource.java:788)
at org.eclipse.jdt.testplugin.JavaProjectHelper.delete(JavaProjectHelper.java:575)
at org.eclipse.jdt.ui.tests.refactoring.MoveRenameResourceChangeTests.testJavaProject01(MoveRenameResourceChangeTests.java:260)
[...]

PR #1722, opened 2025-02-03
Bucket: catch unchecked exception to continue with CoreException

jukzi:
was IndexOutOfBoundsException
https://github.com/eclipse-platform/eclipse.platform/issues/1721

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/Bucket.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/Bucket.java
@@ -380,6 +380,9 @@ public abstract class Bucket {
 				resultEntries.put(readEntryKey(source), readEntryValue(source));
 			}
 			return resultEntries;
+		} catch (RuntimeException e) {
+			String msg = NLS.bind(Messages.properties_readProperties, indexFile.toString());
+			throw new ResourceException(IResourceStatus.FAILED_READ_METADATA, null, msg, e);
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/Bucket.java
```diff
@@ -380,6 +380,9 @@ private Map<String, Object> loadEntries(File indexFile) throws CoreException, IO
 				resultEntries.put(readEntryKey(source), readEntryValue(source));
 			}
 			return resultEntries;
+		} catch (RuntimeException e) {
```
laeubi, line 383: I would either move this into `load` (line 349) or also catch the IO exception here and let the method only throw a core exception.

jukzi: From runtime it doesn't make a difference. I chose from possible callers that closest to the exception. It already throws such a CoreException from org.eclipse.core.internal.properties.PropertyBucket.readEntryValue(DataInputStream) Line 296, but could also fail with other RuntimeExceptions.
The function loadEntries() is documented to throw IOException and i rather keep the code changes as small as possible to avoid long git history.

jukzi:
ignoring "Unspecified error Waiting for Edge operation to terminate timed out" https://github.com/eclipse-platform/eclipse.platform.swt/issues/1801

PR #1727, opened 2025-02-05
Fix "The application has not been initialized." after shutdown

jukzi:
to "The application is already stopped."

As seen logged at the end of many junit tests.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
@@ -102,6 +102,7 @@ public final class InternalPlatform {
 
 	private boolean splashEnded = false;
 	private volatile boolean initialized;
+	private volatile boolean stopped;
 	private static final String KEYRING = "-keyring"; //$NON-NLS-1$
 	private String keyringFile;
 
@@ -182,8 +183,13 @@ public final class InternalPlatform {
 
 	private void assertInitialized() {
 		//avoid the Policy.bind if assertion is true
-		if (!initialized)
-			Assert.isTrue(false, Messages.meta_appNotInit);
+		if (!initialized) {
+			if (stopped) {
+				Assert.isTrue(false, Messages.meta_appStopped);
+			} else {
+				Assert.isTrue(false, Messages.meta_appNotInit);
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -182,8 +183,13 @@ public void addLogListener(ILogListener listener) {
 
 	private void assertInitialized() {
 		//avoid the Policy.bind if assertion is true
-		if (!initialized)
-			Assert.isTrue(false, Messages.meta_appNotInit);
+		if (!initialized) {
+			if (stopped) {
+				Assert.isTrue(false, "The application is already stopped."); //$NON-NLS-1$
```
laeubi, line 188: String needs to be externalized

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -391,7 +397,7 @@ public IPath getLocation() throws IllegalStateException {
 				if (explicitInitRequired) {
 					// See bug 514333: don't allow clients to initialize instance location if the
 					// instance area is not explicitly defined yet
-					throw new IllegalStateException(CommonMessages.meta_instanceDataUnspecified);
+					throw new IllegalStateException(""); //$NON-NLS-1$
```
laeubi, line 400: Why replace with empty (non externalized) string?

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -692,6 +699,7 @@ public void stop(BundleContext bundleContext) {
 		initialized = false;
 		closeOSGITrackers();
 		context = null;
+		stopped = true;
```
laeubi, line 702:
```suggestion
		stopped = true;
		context = null;
```

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
jukzi: that tooling...

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
jukzi: i don't get the difference, but done.

PR #1728, opened 2025-02-05
Test OOM build exception

laeubi:


laeubi:
Summary shows this after the helpcenter is build:

```
Total Memory: 1024.00mb
Free Memory:  98.00mb
Used Memory:  926.00mb
```

I now added to print the final flags, I don't see we set the memory here so maybe maven is only allowed to use 1GB what might not be enough.

PR #1735, opened 2025-02-10
Add option to disable all breakpoints except selected ones

SougandhS:
Reference : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/549#issuecomment-2610384125

PR introduces a new menu option that allows users to select multiple enabled breakpoints and disable all other breakpoints except the ones selected. This provides more flexibility and control over breakpoint management.

Screens - 

image
 <br><br>
After clicking - 
 <br><br>
image

<br><br> works with 
image

### DIFF
--- a/debug/org.eclipse.debug.ui/plugin.properties
+++ b/debug/org.eclipse.debug.ui/plugin.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2000, 2020 IBM Corporation and others.
+# Copyright (c) 2000, 2025 IBM Corporation and others.
 #
 # This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
@@ -83,6 +83,7 @@ DetailPaneFontDefinition.description=The detail pane text font is used in the de
 DetailPaneFactoriesExtension.name=Detail Pane Factories
 DisableBreakpointsAction.label=&Disable
 DisableAllBreakpointsAction.label=Disa&ble All
+DisableOtherBreakpointsAction.label=Disable Others
 EnableAllBreakpointsAction.label=Ena&ble All
 EnableBreakpointsAction.label=&Enable
 ExpandAll.label=Expand All
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
@@ -1,0 +1,99 @@
+/*******************************************************************************
+ * Copyright (c) 2025 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.internal.ui.actions.breakpoints;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.eclipse.core.resources.IWorkspaceRunnable;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.debug.core.DebugException;
+import org.eclipse.debug.core.DebugPlugin;
+import org.eclipse.debug.core.model.IBreakpoint;
+import org.eclipse.debug.internal.ui.DebugUIPlugin;
+import org.eclipse.debug.internal.ui.actions.ActionMessages;
+import org.eclipse.debug.internal.ui.breakpoints.provisional.IBreakpointContainer;
+import org.eclipse.jface.action.IAction;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.ui.IWorkbenchWindow;
+
+public class DisableAllOtherBreakpointsAction extends EnableBreakpointsAction {
+
+	/**
+	 * If this action can enable breakpoints
+	 *
+	 * @return always <code>false</code>
+	 */
+	@Override
+	protected boolean isEnableAction() {
+		return false;
+	}
+
+	@Override
+	public void run(IAction action) {
+		IStructuredSelection selection = getSelection();
+		if (selection.isEmpty()) {
+			return;
+		}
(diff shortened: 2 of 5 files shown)


SougandhS:
Hi @jukzi , could you please check this when you have time.

debug/org.eclipse.debug.ui/plugin.properties
```diff
@@ -83,6 +83,7 @@ DetailPaneFontDefinition.description=The detail pane text font is used in the de
 DetailPaneFactoriesExtension.name=Detail Pane Factories
 DisableBreakpointsAction.label=&Disable
 DisableAllBreakpointsAction.label=Disa&ble All
+DisableOtherBreakpointsAction.label=Disa&ble Others
```
jukzi, line 86: isn't the ampersand '&' used to denote a accelerator (key shortcut) that should be different from the other accelerators in that menu?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
```diff
@@ -0,0 +1,108 @@
...
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.ui.IWorkbenchWindow;
+
+public class DisableAllOtherBreakpointsAction extends EnableBreakpointsAction {
+
+	@Override
+	protected boolean isEnableAction() {
+		int enabledCount = 0;
+		for (IBreakpoint brkPoint : DebugPlugin.getDefault().getBreakpointManager().getBreakpoints()) {
+			if (enabledCount > 1) {
```
jukzi, line 41: i do not understand that enabledCount logic. shouldn't it be enabled for enabledCount>1?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
```diff
@@ -0,0 +1,108 @@
...
+
+			} catch (CoreException e) {
+				ms.merge(e.getStatus());
+			}
+		};
+
+		try {
+			ResourcesPlugin.getWorkspace().run(runnable, null, 0, new NullProgressMonitor());
+		} catch (CoreException e) {
+			// handled by runnable
```
jukzi, line 94: log or throw

debug/org.eclipse.debug.ui/plugin.properties
SougandhS: Looks like '&' came automatically for this.
Can this be valid - `DisableOtherBreakpointsAction.label=Disa&ble O&thers ` ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
SougandhS: Disable action returns false (`org.eclipse.debug.internal.ui.actions.breakpoints.DisableBreakpointsAction.isEnableAction()`) so implemented the same.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
SougandhS: okay 👍

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
SougandhS: will copy java doc too

debug/org.eclipse.debug.ui/plugin.properties
jukzi: No, it should be a single accelerator. Or none. Intuitive would be the key "O" like others, but "o" is already bound to "remove" (which would be better "r". but "r" is already bound to Properties - which better would be bound to "P". Well, either fix this mess up or just do not define any accelerator. 
You can test it by opening the context menu and typing the single key.
I can only see the accelerators when i open the context menu with a key:
image

image

SougandhS: Thanks for the info, looks like every characters are taken so I will remove the accelerator

jukzi: > My keyboard has a function key

wow, thanks for the hint, i just found it on my keyboard too. A hamburger key right to the space bar, i never used it yet.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/breakpoints/DisableAllOtherBreakpointsAction.java
jukzi: Ok, do i understand it right: DisableAllOtherBreakpointsAction will also **enable** the current one? That sounds like a usecase but is unexpected. May be it should not.

SougandhS: Actually it wont enable any disabled ones. if we select enabled ones out of n enables, then it will disable other enabled ones.

SougandhS: if the current one is already a disabled one, it wont enable it

SougandhS: true case is when user has only one enabled breakpoint, if there's only 1 enabled breakpoint then there wont be any use with Disable Others option

jukzi: but that does not fit the decription `If this action can enable breakpoints`

SougandhS: I think ideally it should be always false. 
I will make it false, (tested too)

SougandhS:
Hi @merks , I incorporated all the changes @jukzi suggested. if you have time could you please merge this PR ?

merks:
The Windows error looks unrelated and familiar:

>  Caused by: java.lang.OutOfMemoryError: Java heap space

SougandhS:
Thank you @merks

merks:
Thank you too.  This looks useful.  👍

PR #1743, opened 2025-02-18
Fix build block user action (#472)

haideryaqoob720:
User action is getting blocked ...
See: https://github.com/eclipse-platform/eclipse.platform.ui/issues/472

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/ThreadJob.java
@@ -208,7 +208,7 @@ class ThreadJob extends Job {
 		boolean interruptedDuringWaitForRun;
 		try {
 			// just return if lock listener decided to grant immediate access
-			if (manager.getLockManager().aboutToWait(blocker)) {
+			if (manager.getLockManager().aboutToWait(blocker) || identifyThreadAction()) {
 				return threadJob;
 			}
 			result = waitForRun(threadJob, monitor, blockingJob);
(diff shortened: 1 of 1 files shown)


iloveeclipse (changes requested):
1) Please explain the problem and solution strategy in the PR description & commit message.
2) Using job or thread names is a "NO GO", sorry. We can't make assumptions about job names. Either you use known job families or something else, but matching / guessing names is not acceptable.

akurtakov:
Would you please react to the comment here?

haideryaqoob720:
> Would you please react to the comment here?

I've been working on replacing string-based identifiers with job families

izubaire:
Hey @akurtakov @iloveeclipse, I’ve introduced `isUI()` methods in `LockManager` and `LockListener` to detect the UI thread and skip waiting when `isUI()` is true. This change prevents the UI from getting blocked by background build jobs, keeping interactions responsive for the user.

**Issue:** 
The user’s actions were getting blocked while a build job was running. In other words, the UI thread was waiting for the background build process to complete before allowing user interaction.

**What Changed:**
A new method, isUI(), has been introduced in both the LockManager and LockListener classes. This method checks if the current thread is the UI thread.
The joinRun method in ThreadJob has been updated to call isUI(). Now, if the thread is identified as the UI thread, it will not be forced to wait for the background build to finish.

**How It Solves the Issue:**
By detecting whether the calling thread is the UI thread, we ensure that UI interactions are no longer blocked by a long-running build job. This allows the user to continue interacting with the UI while the background build proceeds independently

iloveeclipse:
@haideryaqoob720 : 
1) there are two test failures caused by your patch: https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-1743/6/testReport/
2) A new test is needed to demonstrate the problem and the fix.

mickaelistria:
> The blocking behavior should be limited to actions that explicitly rely on the build process, such as Run and Debug.

I don't think this assertion is true. The strategy to decide whether to block an action or not is to rely on SchedulingRules. If a build job has a rule holding a file, then it's expected that the file being hold cannot be modified.
In general, instead of trying to bend the Jobs framework, a saner and better strategy involves finding the Job that holds the rule causing the blocked operation and then trying to minimize the impact of its scheduling rule (ie not block a whole project, but only block the few files in needs to read, only for the shortest possible time).

mickaelistria:
The point is that whether an operation is UI or not doesn't make a difference: if the operation uses a scheduling rule that is lock, then it *must* wait for the scheduling rule to be free; otherwise a lot of bugs can happen. Please see https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fruntime_jobs_rules.htm or https://www.eclipse.org/articles/Article-Concurrency/jobs-api.html ; anything that tries to just ignore that is basically introducing big risks of bugs.
The "Build All" operation could be refactored if some reliable optimizations can be found. Usually, not using rules or using more "narrow" rules helps a lot. I don't see a particular reason why "Build All" would require to lock the whole workspace for the duration of the build; as it mostly reads the projects to establish a build order and then each individual builder on the project sets a scheduling rule to lock the right resources. So it could be that just not setting a rule on Build All would be a good solution.

haideryaqoob720:
@iloveeclipse @akurtakov @mickaelistria
To enhance UI responsiveness during builds while preserving SchedulingRules integrity:

- **Fine-Grained Locking:**
    - Build All: Lock and build projects individually instead of the entire workspace lock.
    - Build Project: Lock only the specific project or finer where possible.
    - Auto-Build: Maintain minimal locking (file/resource level) wherever feasible.
- **Resource Tracking:** Detect file edits during builds and trigger incremental rebuilds.
- **SchedulingRules Compliance:** Fully honor concurrency guarantees without bypassing locks.
This approach improves responsiveness safely and aligns Eclipse behavior closer to modern IDEs. I’d appreciate your feedbacks.
Thanks!

PR #1745, opened 2025-02-18
Remove unused code

howlger:
In `org.eclipse.help.internal.search.DefaultAnalyzer`, the `locale` is no longer used, so it does not need to be calculated.


akurtakov:
Jenkins built fine. Merging.

Issue #1758, opened 2025-03-05
beginTask was called on SaveManager more than once


ptziegler:
## Steps to reproduce

Call `IWorkspace.save(true, monitor)`

## Here is some relevant log output

```
!MESSAGE beginTask was called on this instance more than once
!STACK 0
java.lang.Throwable
	at org.eclipse.core.runtime.SubMonitor.logProblem(SubMonitor.java:1187)
	at org.eclipse.core.runtime.SubMonitor.beginTask(SubMonitor.java:713)
	at org.eclipse.core.runtime.ProgressMonitorWrapper.beginTask(ProgressMonitorWrapper.java:53)
	at org.eclipse.core.internal.resources.SaveManager.save(SaveManager.java:1214)
	at org.eclipse.core.internal.resources.Workspace.save(Workspace.java:2494)
	at org.eclipse.core.internal.resources.Workspace.save(Workspace.java:2483)
```

The `beginTask` method is called implicitly in the InternalMonitorWrapper:

https://github.com/eclipse-platform/eclipse.platform/blob/8186345249a0b4ba54edca6771661a4181cd2df8/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java#L1190-L1192

And later explicitly in the `save()` method:

https://github.com/eclipse-platform/eclipse.platform/blob/8186345249a0b4ba54edca6771661a4181cd2df8/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java#L1208-L1217

I.e. `beginTask` is called on the internal SubMonitor of the InternalMonitorWrapper instance in line 1209 and 1214.

## Tested under this environment:
* Eclipse IDE/Platform version (as shown in *Help > About*): Eclipse 2024-12

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

ptziegler:
PR will hopefully follow

PR #1759, opened 2025-03-05
Don't use SubMonitor when saving project in SaveManager

ptziegler:
The migration from the IProgressMonitor to the SubMonitor done as part of 40114d2ce9a41cbec7f01d8342d4371a6d970af0 also converts the progress monitor used within the InternalMonitorWrapper.

If the parent monitor is a SubMonitor, this calls beginTask() both when the InternalMonitorWrapper is created, as well as explicitly later on. To avoid this, switch back to create a null-safe progress monitor via Policy.monitorFor(...).

Closes https://github.com/eclipse-platform/eclipse.platform/issues/1758

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


ptziegler:
The other usages of the SubMonitor look fine, so that seems to be the only place that needs to be adapted.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/SaveManager.java
```diff
@@ -1188,7 +1188,7 @@ class InternalMonitorWrapper extends ProgressMonitorWrapper {
 		private boolean ignoreCancel;
 
 		public InternalMonitorWrapper(IProgressMonitor monitor) {
-			super(SubMonitor.convert(monitor));
+			super(Policy.monitorFor(monitor));
```
ptziegler, line 1191: I noticed that `Policy.monitorFor`

https://github.com/eclipse-platform/eclipse.platform/blob/8186345249a0b4ba54edca6771661a4181cd2df8/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/utils/Policy.java#L177-L179

and `IProgressMonitor.nullSafe(...)`

```java
static IProgressMonitor nullSafe(IProgressMonitor monitor) {
    if (monitor == null) {
        return new NullProgressMonitor();
    }
    return monitor;
}
```

are functionally identical. It probably makes sense to get rid of the former altogether...

ptziegler:
The failure on the Windows node also happens on the master, so it's likely unrelated.

PR #1766, opened 2025-03-10
Branding images for 4.36 (2025-06)

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2856


merks:
There's nothing in the build that verifies this.

PR #1773, opened 2025-03-13
[Debug] Automatic cleanup - instanceof pattern matching in Debug.core

SougandhS:
This commit implements automatic cleanup using instanceof pattern matching in the debug.core package of the platform debug module.


SougandhS:
Thank you @mickaelistria

Issue #1776, opened 2025-03-17
Improper use of progress monitor when deleting a folder


ptziegler:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

Enable the progress monitor tracing and call `IFolder.delete()` with a SubMonitor. e.g.

```java
folder.delete(true, true, subMonitor.newChild(1));
```

## Here is some relevant log output

```
!ENTRY org.eclipse.core.runtime 2 0 2025-03-17 14:47:31.884
!MESSAGE You must allocate ticks using beginTask or setWorkRemaining before trying to consume them
!STACK 0
java.lang.Throwable
	at org.eclipse.core.runtime.SubMonitor.logProblem(SubMonitor.java:1187)
	at org.eclipse.core.runtime.SubMonitor.consume(SubMonitor.java:635)
	at org.eclipse.core.runtime.SubMonitor.internalWorked(SubMonitor.java:742)
	at org.eclipse.core.runtime.SubMonitor.worked(SubMonitor.java:759)
	at org.eclipse.core.internal.localstore.DeleteVisitor.recursiveKeepHistory(DeleteVisitor.java:125)
	at org.eclipse.core.internal.localstore.DeleteVisitor.recursiveKeepHistory(DeleteVisitor.java:116)
	at org.eclipse.core.internal.localstore.DeleteVisitor.delete(DeleteVisitor.java:60)
	at org.eclipse.core.internal.localstore.DeleteVisitor.visit(DeleteVisitor.java:156)
	at org.eclipse.core.internal.localstore.UnifiedTree.accept(UnifiedTree.java:126)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.delete(FileSystemResourceManager.java:413)
	at org.eclipse.core.internal.resources.ResourceTree.internalDeleteFolder(ResourceTree.java:363)
	at org.eclipse.core.internal.resources.ResourceTree.standardDeleteFolder(ResourceTree.java:815)
	at org.eclipse.core.internal.resources.Resource.unprotectedDelete(Resource.java:1830)
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:803)
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:767)
```

[...]

PR #1778, opened 2025-03-18
Cleanup progress monitor handling in resource plugin

ptziegler:
This resolves several trace errors thrown by an improper use of the
progress monitor while moving, deleting or copying files or folders.

- ResourceTree

The deprecated SubProgressMonitor has been replaced by a SubMonitor, due
to an improved handling of rounding errors when reporting fractional
work back to its parent. In addition, the standardMoveFile() method now
explicitly calls beginTask() on its given progress monitor, to remain
consistent with the standardMoveFolder() and standardMoveProject()
methods and to avoid an error being reported when moving files.

- DeleteVisitor

beginTask() is called on the given progress monitor by converting it to
a SubMonitor with the given number of ticks. This number is based on the
number of to-be-deleted files. When deleting a file, a SubMonitor is
used instead of the deprecated SubProgressMonitor, consuming a number of
ticks based on the total number of ticks.

- CopyVisitor

This method has been adapted similarly to the DeleteVisitor as
beginTask() also needs to be called on the given progress monitor here.
Additionally, the progress handling has been adapted to avoid creating
"zero ticks" SubMonitors, as this also causes an error.

Closes https://github.com/eclipse-platform/eclipse.platform/issues/1776

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/CopyVisitor.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/CopyVisitor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


ptziegler:
CC @vogella due to changes to 8539653d6fb34e6955f678c43fac91c83484ae1d and b7534d00886d1047427f61bfec1bf35cb7753f0c.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/CopyVisitor.java
```diff
@@ -188,41 +195,37 @@ protected void synchronize(UnifiedTreeNode node) throws CoreException {
...
+			String message = NLS.bind(Messages.localstore_locationUndefined, path);
+			status.add(new ResourceStatus(IResourceStatus.FAILED_READ_LOCAL, path, message, null));
+			return false;
+		}
+		boolean wasSynchronized = isSynchronized(node);
+		if (force && !wasSynchronized) {
+			synchronize(node);
+			// If not synchronized, the monitor did not take this resource into account.
+			ticks++;
+			rootMonitor.setWorkRemaining(ticks);
```
ptziegler, line 213: I'm not overly happy with this solution, as it might happen that the last file is out-of-sync and wherefore all ticks have already been consumed. The original implementation doesn't really work either, as the progress monitor complains if you work for zero ticks.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/CopyVisitor.java
```diff
@@ -113,8 +118,8 @@ protected boolean copyContents(UnifiedTreeNode node, Resource source, Resource d
 			IFileStore destinationStore = destination.getStore();
 			//ensure the parent of the root destination exists (bug 126104)
 			if (destination == rootDestination)
-				destinationStore.getParent().mkdir(EFS.NONE, monitor.newChild(0));
-			sourceStore.copy(destinationStore, EFS.SHALLOW, monitor.newChild(0));
+				destinationStore.getParent().mkdir(EFS.NONE, subMonitor.newChild(1));
+			sourceStore.copy(destinationStore, EFS.SHALLOW, subMonitor.newChild(1));
 			//create the destination in the workspace
```
ptziegler, line 123: Similarly to the visit method, the progress monitor complains if you create children with zero ticks...

ptziegler:
The OOM error on Windows seems sadly unrelated. Unless there are any objections, I'd like to merge this at the end of the week.

ptziegler:
The other builds are looking fine. Merging.

PR #1781, opened 2025-03-19
Modernized help UI prototype: Fix deleting and applying a search scope

howlger:
This fixes the following issues in the modernized help UI prototype:
- Deleting a set search scope causes JavaScript error.
- Applying via search scope drop-down, choosing _Scopes..._, going to the scope and hitting _Apply_ does not work; as workaround a search scope can directly selected via search scope drop-down.


hhermann:
Thank you Holger for this fix. I have tested it on Windows 11 and Linux RHEL 9.2 and it works fine.

PR #1789, opened 2025-03-24
Jenkinsfile: Use Tycho 4.0.13 way to display api tools warnings

sratz:
Also make sure the issues are published even in case of a failed build, as otherwise they are not visualized.


sratz:
Same as for https://github.com/eclipse-platform/eclipse.platform.swt/pull/1932.
Build failure is unrelated.

PR #1797, opened 2025-03-31
Add SVGs to org.eclipse.debug.ui

Michael5601:
This PR adds all SVGs of `org.eclipse.debug.ui` and changes related paths except for the following as these are not available as SVG yet:

obj16:
launchConfiguration.svg

ov15:
quickaccess_debug
quickaccess_profile
quickaccess_run

wizban:
adddir_wiz
addsrcloc_wiz
debug_wiz
editdir_wiz
edtsrclkup_wiz
export_brkpts_wizban
export_config_wizban
import_brkpts_wizban
import_config_wizban
profile_wiz
run_wiz

The changes can be seen in the toolbar of the debug perspective:

Default (PNG):
image

Updated (SVG):
image

We should add this change in M1 so we can properly test the SVG feature that was introduced in this [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638) with a bundle that contains SVGs.


HeikoKlare:
You can ignore my previous comment. I found the issue for the wrongly scaled SVGs: the Image implementation was erroneous: if there the data for an image at the zoom level of the file (usually 100) was already loaded, this data is scaled instead of loading the file in proper zoom anew (as required for SVGs). That's easy to fix with a PR I provided today, so I added a fix that PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1987

Then with the same scenario as above it looks like this:
image

HannesWell:
When we merge this, we should also have https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936 available, shouldn't we? Otherwise I expect the disabled SVGs to not look good.

HannesWell:
> When we merge this, we should also have [eclipse-platform/eclipse.platform.swt#1936](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936) available, shouldn't we? Otherwise I expect the disabled SVGs to not look good.

On the other hand I recall a (private) discussion with Heiko where he said, that the current state of that SWT PR is not yet fully ready and needs further refinement.

Michael5601:
> When we merge this, we should also have [eclipse-platform/eclipse.platform.swt#1936](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936) available, shouldn't we? Otherwise I expect the disabled SVGs to not look good.

The disabled icons are not affected by this. I only changed the paths of the normal icons and ignored all disabled icons.

HannesWell (approved):
Thank you for this work! The change looks good and it's great to finally have the first SVG icons in real use!

Altough this contributon is much larger than 1000 LOC (which usually requires an IP-review), this is fine because all the SVG files originate from https://github.com/eclipse-platform/eclipse.platform.images.
Eventually I think all SVG files that have been moved into the using plugin should be removed from that repo to avoid maintaining it at two different places.

As soon as https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593 is submitted, this can be submitted too.

BeckerWdf:
maybe a stupid question: Do I need to activate the rendering of the SVG icons somehow on macOS?

BeckerWdf:
> I did not test it yet on MacOs but can you try adding the following line in your `eclipse.ini`:
> `-Dswt.autoScale=quarter`

I did this (added to the launch config not the ini file) but I could not see a difference on my macOS retina display. Is the scaling on macOS so good?

Michael5601:
> > I did not test it yet on MacOs but can you try adding the following line in your `eclipse.ini`:
> > `-Dswt.autoScale=quarter`
> 
> I did this (added to the launch config not the ini file) but I could not see a difference on my macOS retina display. Is the scaling on macOS so good?

I think so. In the discussions I read, before I implemented this feature, it seemed like people using macOS had less problems with scaling icons.

HeikoKlare:
MacOS does only support 100% and 200% scaling. Since we had already provided the rasterized versions for those two scale values, you will not see any difference on MacOS. They will be equally sharp or blurry (like everything on MacOS) as they were before.
Using `quarter` or any other auto-scale setting will make no difference on MacOS since are not such "quarter" values for the scaling.

PR #1810, opened 2025-04-11
Streamline ImageDescriptor implementations and avoid getImageData()

HannesWell:
ImageDescriptor.getImageData() is deprecated and sub-classes are advised to stop re-implementing it and instead implement getImageData(int). The still existing implementations just replicate the default implementation and therefore can just be removed.

Additionally the custom and internal 'ImageImageDescriptor' and '(Abstract)OverlayIcon' classes are replaced with implementations provided by standard JFace.


HannesWell:
Both places where custom `ImageDescriptor` implementations have been replaced look as before: The breakpoint icon and the icons with the pen in the book in the tree:
grafik

HannesWell:
Jenkins failure is due to: https://github.com/eclipse-platform/eclipse.platform/pull/1809#issuecomment-2796726688

PR #1815, opened 2025-04-14
Add SVGs to org.eclipse.ui.intro bundles

Michael5601:
This PR adds SVGs for all icons in the bundles `org.eclipse.ui.intro` and `org.eclipse.ui.intro.universal` except for the following as it is not available as SVG yet:

org.eclipse.ui.intro/etool16/restore_welcome.svg

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.


Michael5601:
> There are code changes which are unrelated to svgs. Please drop them from this PR.

Doing it right now :D

Michael5601:
> There are code changes which are unrelated to svgs. Please drop them from this PR.

Should be clean now. I did not notice the changes from the autoformatter thus the unwanted changes.
I will do this today for all other PRs too.

Michael5601:
> Seeing all these changes from `png` > `svg` (and previously from `gif` > `png`), maybe we should have a `ImageDescriptor` with autodetection e.g. if no extension is given it first tries `svg`, then `png`, then `gif`?
> 
> Then one can already remove all extensions and if ther will ever be a svg provided it is picked up automatically.

Sounds like a good idea but wouldn't the performance be affected by this? I think we would need to try loading the SVG and if it fails we load the PNG.

Michael5601:
> Of course it could require some more lookups, but the same is true for looking for higher res images and so on, and given we use it not on a cloud-image it should be not that costly given that we have svg for most of the things already.
> 
> But that was just an idea that came into my mind seeing all these extension changes :-)

I would handle this as a follow up for now but keep it in mind. Thanks :)

laeubi:
> I would handle this as a follow up for now but keep it in mind. Thanks :)

What just came into my mind: If we remove the png now and people reference them in their plugin (what of course is bad practice) this might fail now.

So what might be useful as well is that if one looks for `myImage.png` but it is not found, we give another try for `myImage.svg`... I think this would be acceptable to have it "slower" here instead of fail completely.

Michael5601:
> > I would handle this as a follow up for now but keep it in mind. Thanks :)
> 
> What just came into my mind: If we remove the png now and people reference them in their plugin (what of course is bad practice) this might fail now.
> 
> So what might be useful as well is that if one looks for `myImage.png` but it is not found, we give another try for `myImage.svg`... I think this would be acceptable to have it "slower" here instead of fail completely.

The PNGs are not removed yet for the exact reason you mentioned but of course this is a goal in the future. We already discussed the approach of redirecting to the SVG if the PNG is not available and referenced somewhere. We don't have a final decision here but thanks for the note.

HeikoKlare (approved):

PR #1821, opened 2025-04-14
Different Xmx opts for Jenkins / GitHub actions

sratz:
See #1809.


HannesWell:
Thanks. That's a good solution.

Issue #1823, opened 2025-04-16
Help webapp UI colors wrong on Linux


akurtakov:
On Fedora 42 Help webapp many colors are wrong leading to unreadable buttons like the "GO" button .
Standalone:

![Image](https://github.com/user-attachments/assets/a6489fde-6cea-41d0-9212-5ca96e904b05)
Totally invisible

Firefox:

![Image](https://github.com/user-attachments/assets/5d1e545a-fe8e-4709-904d-eb0270176936)
Barely readable

This is coming from usage of deprecated css colors (https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#deprecated_system_color_keywords ) which Webkit seems to no longer support at all and Firefox has some very basic and no longer good support.

akurtakov:
Firefox has the suboptimal colors for Go button on Windows too.

akurtakov:
As per previous comment this seems to be the correct fix for now so I'll push the PR once verifications are done.

PR #1824, opened 2025-04-16
Fix invisible/barely visible Go button in help webapp

akurtakov:
This is coming from usage of deprecated css colors (https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#deprecated_system_color_keywords ) which Webkit seems to no longer support at all and Firefox has some very basic and no longer good support.
Replaced the deprecated colors with their "analogs" as per that page. Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1823


akurtakov:
It looks like 
image
 with this PR for me.

PR #1828, opened 2025-04-17
Build jsps with jetty-ee8-jspc-maven-plugin

akurtakov:
Use same jetty version that we ship with instead of old Jetty 10 jspc maven plugin.
This version finally allows to set the source/target compiler version to Java 21 to match the BREE.


akurtakov:
Test failure unrelated to jsp compiler change in different bundle.

PR #1834, opened 2025-04-23
Touch bundles affected by compiler changes, take two

iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3951 
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2995


iloveeclipse:
Should be merged only when the fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3951 is available & merged in JDT.

PR #1837, opened 2025-04-25
Remove hover images and replace faulty icon in org.eclipse.ui.console

HeikoKlare:
The hover images only replicate the ordinary icons without giving any benefit but just duplicate resources. One of the just added SVGs icons (for wordwrap) is even incorrect (it's different version of the `wordwrap` icon, so that original and hover icon are slightly different). This change completely removes the hover images in org.eclipse.ui.console, including their code usages and the just provisioned SVGs.

Before:
wordwrap_hover_before

After:
wordwrap_hover_after


HeikoKlare:
Failing version increment check is for infrastructure reasons and necessary version bump has already been made via https://github.com/eclipse-platform/eclipse.platform/pull/1813

PR #1840, opened 2025-04-25
Remove usages of deprecated fastViewWidthRatio

HeikoKlare:
Some plugins contained definitions of fastViewWidthRatio even though that value is not read for a while and was just deprecated. This change removes those definitions.


ptziegler (approved):
Whoops. I didn't notice that the platform also contains UI contributions. Otherwise I would've updated them myself when I deprecated the attribute.

vogella:
LGTM, please merge

PR #1848, opened 2025-05-01
Force qualifier update to fix build failure

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3006


akurtakov:
Yet another comparator issue caused by png to svg switch and no builds for extend period. Merging.

akurtakov:
image

HannesWell:
Looks like the constant is inlined by the compiler. So yet another example, why the baseline check is important.

PR #1851, opened 2025-05-03
Merge terminal control

laeubi:
This migrates the terminal control currently located at CDT to platform see:

- https://github.com/eclipse-platform/eclipse.platform/pull/1762
- https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.control

From the readme:

> The Terminal is a UI-less model of a grid of characters,
> plus an SWT widget that's updated asynchronously for 
> maximum performance. The widget can be hooked up to various
> ITerminalConnectors providing an InputStream, OutputStream,
> and a method for setting the Terminal Size.
> 
> The widget processes ANSI control characters, including NUL,
> backspace, carriage return, linefeed, and a subset of ANSI
> escape sequences sufficient to allow use of screen-oriented
> applications, such as vi, Emacs, and any GNU readline-enabled
> application (Bash, bc, ncftp, etc.).
> 
> This is not yet a fully compliant vt100 / vt102 terminal 
> emulator!

FYI @akurtakov @jonahgraham @stbischof


merks (approved):
I have not reviewed in detail.  But I've seen the discussion and I think folks are comfortable with the plan to migrate this to the platform so that it can be more widely reused.

iloveeclipse:
@laeubi : Do you plan to migrate tests or to provide new tests for that bundle?

laeubi:
There is a test-suite here:

- https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.test

but it covers more aspects that the control itself. If things work out well we should migrate more parts (e.g. terminal view and connectors) so probably also the tests.

Issue #1852, opened 2025-05-04
Missing *.svg in org.eclipse.team.ui/icons/full/dlcl16


merks:
The entire dlcl16 folder is missing all the *.svg

https://github.com/eclipse-platform/eclipse.platform/tree/master/team/bundles/org.eclipse.team.ui/icons/full/dlcl16

unlike the elcl16 folder which has them

https://github.com/eclipse-platform/eclipse.platform/tree/master/team/bundles/org.eclipse.team.ui/icons/full/elcl16

I noticed because the team perspective has red view toolbar icons:

![Image](https://github.com/user-attachments/assets/2bc5d0c6-0606-4dba-8fa1-58dff3429d9f)

HannesWell:
The folder actually exists in the images repo:
https://github.com/eclipse-platform/eclipse.platform.images/tree/478ce88bf7465cb04af36c9819a4bf833f27847b/org.eclipse.images/eclipse-svg/org.eclipse.team.ui/icons/full/elcl16

I assume this was just forgotten in
- https://github.com/eclipse-platform/eclipse.platform/pull/1818

@Michael5601, @HeikoKlare could you please take care of it ASAP as we are close before our final M2 contribution (Monday should be sufficient)?

merks:
It's the dlcl16  folder that's missing content, but is also missing here:

https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images/eclipse-svg/org.eclipse.team.ui/icons/full

As I understand it, all the disabled images are to be generated automatically...

HannesWell:
> As I understand it, all the disabled images are to be generated automatically...

Yes that's right. Sorry I didn't read the problem precisely. You fix also looks good. Thank you for that.
It's really great how all the work about SVGs and disabled icons reduces our requirements to provide and manage resources!

PR #1853, opened 2025-05-04
Don't populate disabled images from dcl16

merks:
- There are no svg images in that folder because the disabled image is synthesized from the regular image.

https://github.com/eclipse-platform/eclipse.platform/issues/1852


merks:
This fixes the problem and we see proper disabled images:

image

merks:
The failures are known:

https://github.com/eclipse-equinox/equinox/issues/937

HeikoKlare:
Thank you for the fix, @merks! This is the right thing to do. The pre-generated disabled images became obsolete with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936, as they are now generated on-the-fly with the same result/quality. This was a preparation for the SVG support, as it made the provision of disabled icon SVGs obsolete. I tried to adapt all references to disabled icons together with the SVG-introducing PRs, but seems like a missed at least this place.

PR #1854, opened 2025-05-04
Simplify NLS.bind() invocations in jsch integration

HannesWell:
This also avoid compilation errors due to some _source_ incompatible cases of:
- https://github.com/eclipse-equinox/equinox/pull/936


HannesWell:
Submitting this early to be able to restart the I-build that just failed due to the initially mentioned source-compatibility problems.
The changed projects were compiled successfully and I don't expect an effect on tests from this change.

PR #1858, opened 2025-05-05
Migrate org.eclipse.e4.core.tests to JUnit 5 (javax + jakarta variant)

ptziegler:


runtime/tests/org.eclipse.e4.core.javax.tests/src/org/eclipse/e4/core/internal/tests/contexts/performance/ContextPerformanceTest.java
```diff
@@ -26,31 +26,20 @@
...
-		// return suite;
-	}
-
-	public ContextPerformanceTest(String name) {
-		super(name);
-	}
-
-	@Override
-	protected void setUp() throws Exception {
-		super.setUp();
```
ptziegler, line 51: I don't think there is anything gained by keeping this test as a test suite. The test was never executed before and if it's only a single class, it makes more sense adding it to the parent suite.

ptziegler:
Both test suites have 418 test in total, but the test suites now have different names. Partially because of Junit 5 and partially because I gave the javax variant a dedicated name, to avoid any further confusion.

ptziegler:
Everything looks good now. Unless there are any objections, I plan on merging later today.

runtime/tests/org.eclipse.e4.core.javax.tests/src/org/eclipse/e4/core/internal/tests/contexts/EclipseContextTest.java
```diff
@@ -53,18 +53,18 @@ public void setUp() throws Exception {
 
 	@Test
 	public void testContainsKey() {
-		assertFalse("1.0", context.containsKey("function"));
-		assertFalse("1.1", context.containsKey("separator"));
+		assertFalse(context.containsKey("function"), "1.0");
```
HannesWell, line 56: In my opinion, such trivial location marks are not necessary nowadays when we have good stack traces available. Therefore I would just remove them. Explaining comments are still relevant for me and worth migrating.

runtime/tests/org.eclipse.e4.core.javax.tests/src/org/eclipse/e4/core/internal/tests/contexts/performance/ContextPerformanceTest.java
```diff
@@ -26,31 +26,20 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 
-import junit.framework.Test;
-import junit.framework.TestCase;
-import junit.framework.TestSuite;
-
-public class ContextPerformanceTest extends TestCase {
+public class ContextPerformanceTest {
+	TestInfo testInfo;
```
HannesWell, line 34: Wouldn't it be sufficient to just save the display name here?

HannesWell (commented):
Thank you for working on this. 
In general this looks good to me, I just have two minor remarks.

runtime/tests/org.eclipse.e4.core.javax.tests/src/org/eclipse/e4/core/internal/tests/contexts/EclipseContextTest.java
ptziegler: Good point. They don't really add anything of value and will just be a pain, if JUnit decides to change the order of operations again. I've updated the PR.

runtime/tests/org.eclipse.e4.core.javax.tests/src/org/eclipse/e4/core/internal/tests/contexts/performance/ContextPerformanceTest.java
ptziegler: Yes, we only need the name. I've also included that in the updated PR.

vogella:
LGTM Thanks @ptziegler

Issue #1689, opened 2025-01-16
NullPointerException in DefaultHelpUi.displayContext


2wendex2:
In ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/DefaultHelpUI.java, lines 330-336:

```IContextProvider adapter = activePart.getAdapter(IContextProvider.class);
if (adapter != null)
	context = adapter.getContext(c);
String contextText = context.getText();
```

But javadoc of `IContextProvider.getContext(Object)` says:

> return context help for the provided target or `null` if none is defined.

I.e. `adapter.getContext(c)` may return `null` then we call `context.getText()` and give `NullPointerException`.

jukzi:
If you have an actual NPE please quote the whole stacktrace and give a reproducer. Or are you just speculating that there is a possible NPE?

2wendex2:
@jukzi I have an actual NPE. Stack trace:
```!ENTRY org.eclipse.ui 4 0 2025-01-16 21:22:02.836
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.help.IContext.getText()" because "context" is null
	at org.eclipse.help.ui.internal.DefaultHelpUI.displayContext(DefaultHelpUI.java:336)
	at org.eclipse.help.ui.internal.DefaultHelpUI.displayContext(DefaultHelpUI.java:305)
	at org.eclipse.ui.internal.help.WorkbenchHelpSystem.displayContext(WorkbenchHelpSystem.java:750)
	at org.eclipse.ui.internal.help.WorkbenchHelpSystem$WorkbenchHelpListener.helpRequested(WorkbenchHelpSystem.java:133)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:157)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4273)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1066)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4071)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3659)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
[...]

PR #1863, opened 2025-05-08
Null check for org.eclipse.help.IContext

SougandhS:
Check whether context is null or not before invoking getText()

Fixes : https://github.com/eclipse-platform/eclipse.platform/issues/1689

### DIFF
--- a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/DefaultHelpUI.java
+++ b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/DefaultHelpUI.java
@@ -1,5 +1,5 @@
 /***************************************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


SougandhS:
> The referenced issue was introduced by exactly the code that now may lead to a null value. 

Thank you for the detailed review and context. I agree that introducing a null check here may mask the underlying issue rather than addressing its root cause, especially considering the regression mentioned in bug 533828.

Given that, I'll go ahead and close this PR. It makes more sense to investigate and resolve the underlying cause of the null value instead of working around.

laeubi:
If the value must not be null at this place, a null check can still be sufficient, e.g using `Objects.requireNonNull(context, "describe why it must not be null here");`
This is much better than having the code run into a generic NPE...

SougandhS:
> If the value must not be null at this place, a null check can still be sufficient, e.g using `Objects.requireNonNull(context, "describe why it must not be null here");` This is much better than having the code run into a generic NPE...

thanks I'll update with this

HeikoKlare:
> If the value must not be null at this place, a null check can still be sufficient, e.g using `Objects.requireNonNull(context, "describe why it must not be null here");`
> This is much better than having the code run into a generic NPE...

Makes sense. The updated change is fine for me and it fits to your proposal, @laeubi, doesn't it?

SougandhS:
Thank you @HeikoKlare

PR #1864, opened 2025-05-09
Marked long-time deprecated API in Plugin for removal

vogella:
Plugin contains methods which are marked since a long time as deprecated. Marking these for removal to give us the flexibility to remove these after API freeze period if needed.

Triggered by an investigation of ILegacyPreferences from Equinox which calls some of these methods in PreferenceServiceRegistryHelper during startup.

### DIFF
--- a/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Plugin.java
+++ b/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Plugin.java
@@ -213,7 +213,7 @@ public abstract class Plugin implements BundleActivator {
 	 * @return a URL for the given path or <code>null</code>
 	 * @deprecated use {@link FileLocator#find(Bundle, IPath, Map)}
 	 */
-	@Deprecated
+	@Deprecated(forRemoval = true, since = "2025-06 (removal in 2027-06 or later)")
 	public final URL find(IPath path) {
 		return FileLocator.find(getBundle(), path, null);
 	}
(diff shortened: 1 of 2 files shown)


vogella:
the preference API is still used by Ant.... So this should be updated to not mark this API for deletion as I assume no one plans to work on updating Ant to the non-deprecated preferences.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Plugin.java
```diff
@@ -180,7 +180,7 @@ public abstract class Plugin implements BundleActivator {
 	 * @since 2.0
 	 * @deprecated
 	 */
-	@Deprecated
+	@Deprecated(forRemoval = true, since = "2025-06 (removal in 2027-06 or later)")
 	private Preferences preferences = null;
```
HannesWell, line 184: Deprecating private fields doesn't seem to make sense to me as none outside of this class can access it.

HannesWell (commented):
> the preference API is still used by Ant.... So this should be updated to not mark this API for deletion as I assume no one plans to work on updating Ant to the non-deprecated preferences.

Would it be difficult to migrate away from the old API in Eclipse-ANT?

vogella:
> > the preference API is still used by Ant.... So this should be updated to not mark this API for deletion as I assume no one plans to work on updating Ant to the non-deprecated preferences.
> 
> Would it be difficult to migrate away from the old API in Eclipse-ANT?

I don't know, never looked into this. In my more active time Ant was heavily "guarded" by others so it never felt like worth trying to invest into this code.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Plugin.java
vogella: Yes, thanks. Changed in latest version

vogella:
The test failure 1 out of 3 runs failed: testBug546710_ConsoleCreationRaceCondition (org.eclipse.debug.tests.console.ProcessConsoleManagerTests  in Windows seems unrelated.

merks:
> the preference API is still used by Ant.... So this should be updated to not mark this API for deletion as I assume no one plans to work on updating Ant to the non-deprecated preferences.

It's really quite widely used beyond just Ant, including by other methods that return a deprecated Preferences instance.

image

PR #1867, opened 2025-05-09
Mark for removal  i platform configuration factory

vogella:


vogella:
@fedejeanne do you know how I can remove the Javadoc errors? This is similar to what you might have seen in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2966/

fedejeanne:
> @fedejeanne do you know how I can remove the Javadoc errors? This is similar to what you might have seen in [eclipse-platform/eclipse.platform.ui#2966](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2966)

I don't I had warnings in the _imports_, I only had warnings in the usage of the methods (they were not imported statically so it was a different scenario). 

I also don't see a warning in my IDE and can't find any configuration to activate such warning. The only way I know to get rid of such warnings is to not have the imports _i.e._ use fully qualified names. I took the liberty of do that for `BundleGroupComponent` in b07a32f6709df7187b779996419d80410fb84660, I hope that's OK. If that works and you like the final state of the class, you could try the same in the other classes. Otherwise just revert to your last version, no harm no foul :-)

vogella:
Thanks @fedejeanne the removal of the import is a great idea.

vogella:
Thanks @fedejeanne for the help

fedejeanne:
My pleasure :-)

Sorry for the late response, I finally managed to stay offline for the rest of the weekend after [my last comment](https://github.com/eclipse-platform/eclipse.platform/pull/1867#issuecomment-2868815097) (it took some convincing though 😅 )

Have a nice day!

PR #1871, opened 2025-05-12
Whitespace clean-up in resources and runtime

vogella:


vogella:
cc @HannesWell

vogella:
.project should be removed from this change

PR #1872, opened 2025-05-12
Clean-up whitespace in team

vogella:
Using the JDT clean-up action to run the whitespace clean-up in team.


vogella:
cc @HannesWell I think this one is for the part of the repo which did not get a lot of clean-up hence the larger number of files. In my opinion most changes will not affect the Git history as they hit empty lines or Javadoc lines and only a few affect code lines.

vogella:
Planning to merge today, if I do not here concerns

PR #1873, opened 2025-05-12
Add Missing SVGs to Platform Bundles

Michael5601:
This PR adds all missing SVGs to Platform Bundles except for OS-specific icons in the bundle `org.eclipse.platform/images` for which a deletion is proposed in this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1875).

All added icons were either created by me or copied if I found a duplicate with another name.

All `wizban` icons were created by using an Inkscape function to redraw the bitmap (PNG to SVG conversion). This returns pretty good results with only minimal manual work but also creates many paths resulting in SVG that are about 10x larger than other wizban SVGs. Rasterizing the SVGs "manually" by using JSVG did not yield significantly longer rasterization time in my testing. As I don't know a feasible way to create the wizban SVGs otherwise I propose to use this method.

I also added available SVGs for `org.eclipse.ui.intro.universal/themes` which were forgotten before but already existed in the images repository.
 
All missing SVGs are documented in this [issue](https://github.com/eclipse-platform/eclipse.platform.images/issues/129).


HeikoKlare (commented):
Thank you for creating the SVGs! The 16x16 icons and their references look fine to me. Some debug.ui icons are already referenced as SVGs in code even though they were not provided yet. So it definitely makes sense to have them in soon/now.

Like mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2970#pullrequestreview-2839383590, the wizard banner icons are no proper vector graphics, so we should exclude them form the change for now.

HeikoKlare (approved):
I've removed the wizard banner icons for the reasons mentioned in https://github.com/eclipse-platform/eclipse.platform.ui/pull/2970#pullrequestreview-2839383590 again to make this PR ready to merge. We should have a separate look at them.

PR #1875, opened 2025-05-13
Delete unused PNGs in org.eclipse.platform/images

Michael5601:
This PR deletes unused PNGs in folder `org.eclipse.platform/images`. These PNGs were directly referenced via url-function within CSS-files 13 years ago which can be seen in the History.

image
image

These CSS files were moved or deleted, so the PNGs remain unused.

This change is related to the documentation of [missing SVGs](https://github.com/eclipse-platform/eclipse.platform.images/issues/129) as for the mentioned PNGs no corresponding SVGs exist.


HeikoKlare (approved):
Since those images are unused for a long time (used to be referenced by the UI themes several years ago), removing them looks fine to me.

Issue #1877, opened 2025-05-16
Enable Tycho clean-ups for platform


vogella:
If I run the clean-ups we also use in our Tycho build for platform, I get lots of clean-ups. But I see no PR for these clean-ups created by Tycho.

I suggest we also activate them here.  Or are they paused for a reason? How can I check that? In Equinox I see  some ignore configuration but PDE does not  seem to have a configuration for cleanup. 

Equinox:
````
     <groupId>org.eclipse.tycho</groupId>
          <artifactId>tycho-cleancode-plugin</artifactId>
          <configuration>
              <ignores>
                  <ignore>felix/src/.*</ignore>
                  <ignore>osgi/src/.*</ignore>
                  <ignore>felix/src_test/.*</ignore>
              </ignores>
          </configuration>
      </plugin>
````

How are these clean-ups activated for a repository?

laeubi:
This is planned for the next release stream, to not create too much rumor now we are in RC phase.

The cleanup is activated by a timed github action for example here:
https://github.com/eclipse-pde/eclipse.pde/blob/master/.github/workflows/doCleanCode.yml

PR #1878, opened 2025-05-16
Enable code cleanups for platform repository

laeubi:
FYI @vogella 

My plan would be to merge this one master opens for the next release.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1877


laeubi:
> Nice. Do you have something similar for platform.ui?

If there is enough interest and enough people that are dedicated to reviewing/merging resulting PRs and possibly reporting issues to JDT then it is quite similar like this.

PR #1927, opened 2025-06-09
Add Orbit key to o.e.equinox.p2.engine.pgp:trustedKey

merks:
- Organize all the keys in a keys folder with a meaningful name for each key.


merks:
These are then available:

image

akurtakov (approved):
Looks good to me!

PR #1945, opened 2025-06-11
Remove unnecessary casts

HeikoKlare:
Otherwise cleanup https://github.com/eclipse-platform/eclipse.platform/pull/1943 will assign useless variable names to casted values

### DIFF
--- a/ua/org.eclipse.help/META-INF/MANIFEST.MF
+++ b/ua/org.eclipse.help/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %help_plugin_name
 Bundle-SymbolicName: org.eclipse.help; singleton:=true
-Bundle-Version: 3.10.500.qualifier
+Bundle-Version: 3.10.600.qualifier
 Bundle-Activator: org.eclipse.help.internal.HelpPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/ua/org.eclipse.help/pom.xml
+++ b/ua/org.eclipse.help/pom.xml
@@ -18,6 +18,6 @@
   </parent>
   <groupId>org.eclipse.platform</groupId>
   <artifactId>org.eclipse.help</artifactId>
-  <version>3.10.500-SNAPSHOT</version>
+  <version>3.10.600-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 </project>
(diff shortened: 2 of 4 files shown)


laeubi:
> Otherwise cleanup eclipse-jdt/eclipse.jdt.ui#1943 will assign useless variable names to casted values

@HeikoKlare is there an enhancement request for JDT already?

HeikoKlare:
> is there an enhancement request for JDT already?

I don't think it makes sense to improve that, which is why I did not create a request/proposal yet. Consider that in this case, the code is rather weird: you have an object of "proper type" (in this case, the parameter type of a method it is passed to), but you check it for being some subtype and only if it's not that specific subtype, you create and pass a different object. In my opinion, this is more an OO smell than a necessity for a proper cleanup. I just came across it because I wanted to improve the generated variable names and then found that it can be removed completely.

FWIW: the following is a generic example. Applying the "remove unnecessary cast" cleanup removes the cast from the first statement but not from the second:
```java
public class CastInTernary {
	private static class Super {
	}
	
	private static class Sub extends Super {
	}
	
	public void method (Super type) {
		Super result1 = (Sub) type;
		Super result2 = type instanceof Sub ? (Sub) type : new Super(); 
	}
}
```

laeubi:
> FWIW: the following is a generic example. Applying the "remove unnecessary cast" cleanup removes the cast from the first statement but not from the second:

That's what I have in mind, even if the code is weird the Cleanup should properly clean it up, @jjohnstn WDYT?

jjohnstn:
@laeubi I think it should have worked.   I'll look at it today.

akurtakov:
There is the example for Jeff so I'm merging this one. If/when JDT improves - even better.

jjohnstn:
@laeubi The cleanup is simple and uses the compiler to determine all the locations of unnecessary casts.  In this case, the compiler does not recognize it.

laeubi:
So something to report to jdt.core?

jjohnstn:
> So something to report to jdt.core?

Yes.

HeikoKlare:
> So something to report to jdt.core?

If I am not mistaken, the cleanup is implemented in jdt.ui ([UnusedCodeFixCore](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.core.manipulation/core%20extension/org/eclipse/jdt/internal/corext/fix/UnusedCodeFixCore.java) class), isn't it? I thus created this issue:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4103

If I am mistaken, please let me know or just move the issue to jdt.core.

Issue #1951, opened 2025-06-12
Proxy messages written to log that cannot be satisfied


rubenporras:
Similar to what is described in https://github.com/eclipse-platform/eclipse.platform/issues/258 I have the messages

> !MESSAGE System property https.proxyHost is not set but should be ...
> !MESSAGE System property https.proxyPort is not set but should be ...
> !MESSAGE System property https.proxyHost is not set but should be ...
> !MESSAGE System property https.proxyPort is not set but should be ...

I understand that this is advisable to do, as otherwise there is a window between the start of the java application and the time that the `org.eclipse.core.net` is activated, where the system properties are not set to the desired value.

I have set the system property in my `.ini` file to match the settings set for the `org.eclipse.core.net` plug-in, but then I get the messages

> "System property {0} has been set to {1} by an external source. This value will be overwritten using the values from the preferences"

which is quite confusing because it does not realize that the `external source` and the one from the preferences define the same values. Since it looks like `ProxyType` already checks for the properties being equal, I suggest we remove the check that outputs the later message.

fedejeanne:
Thank you for reporting this issue and suggesting the enahancement, @rubenporras !

Are you planning on providing a PR for this?

rubenporras:
Hi @fedejeanne , I did the PR in https://github.com/eclipse-platform/eclipse.platform/pull/1952

fedejeanne:
Thank you, I missed that. Can you please adapt the commit message in your PR so it references this issue as described in the [recommendations](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations) (you need to add 2 references to this issue)? 

Also, a simple way to reproduce this behavior (just the steps) and also your current preferences (you may of course anonymize the actual IPs) would help understanding the issue more in detail. I have to confess that I am not experienced in this particular area but it seems to me that your PR disables the verification for all OSes which may also introduce a security issue. You mentioned that the check for equal is done later but it is difficult to validate this without knowing the exact steps you used to reproduce this behavior.

Just for informational purpose, I am linking the original approach that you took, which was originally added to fix an issue for Mac:
- Issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=564923
- Fix: https://github.com/eclipse-platform/eclipse.platform/commit/7f7dbae6777773c1b54c3c9864728b4215e19219

rubenporras:
Thanks for the comment. I have adapted the commit message.

[...]

Issue #1370, opened 2024-05-15
[mac] LinkTest failed on I20240514-1800: help/topic/org.eclipse.jdt.doc.user/nav/1_3 doens't exist.


jukzi:
only on mac:
https://download.eclipse.org/eclipse/downloads/drops4/I20240514-1800/testresults/html/org.eclipse.ua.tests.doc_ep432I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
```
Expecting empty but was: ["Link from http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/reference/views/breakpoints/ref-breakpoints_view.htm to ../../../nav/1 is broken: target URI http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/nav/1 doens't exist.", "Link from http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/reference/views/breakpoints/ref-breakpoints_view.htm to ../../../nav/1_3 is broken: target URI http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/nav/1_3 doens't exist."]

java.lang.AssertionError:
Expecting empty but was: ["Link from http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/reference/views/breakpoints/ref-breakpoints_view.htm to ../../../nav/1 is broken: target URI http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/nav/1 doens't exist.",
"Link from http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/reference/views/breakpoints/ref-breakpoints_view.htm to ../../../nav/1_3 is broken: target URI http://127.0.0.1:50948/help/topic/org.eclipse.jdt.doc.user/nav/1_3 doens't exist."]
at org.eclipse.ua.tests.doc.internal.linkchecker.LinkTest.testAllLinks(LinkTest.java:103)
```

jukzi:
got even worse on I20240515-1800:


```
[...]

PR #1952, opened 2025-06-12
fix: do not verify proxy data during initialization

rubenporras:
as it produces `System property {0} has been set to {1} by an external source. This value will be overwritten using the values from the preferences` when the user sets the system properties to avoid the info message `System property {0} is not set but should be {1}`

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/1951


rubenporras:
for #1951

fedejeanne (approved):

fedejeanne:
Failing tests are unrelated:
- https://github.com/eclipse-platform/eclipse.platform/issues/1370
- #1956 

Thank you @rubenporras for the fix!

PR #1959, opened 2025-06-14
Migrate org.eclipse.tm.terminal.view.core

laeubi:
This migrate another part of the terminal code currently in CDT to platform:

https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.view.core

FYI @jonahgraham @akurtakov once this is merged and M1 is available CDT would want to use the platform M1 repo and remove the plugin from CDT. I increased the version of the bundle to 4.11 to account for the location change.


laeubi:
We need to enable merge-commits for this (FYI @HannesWell )

laeubi:
To make the migration smooth, CDT should merge this before the next release:

- https://github.com/eclipse-cdt/cdt/pull/1203

laeubi:
I now recovered as much of the history as possible (dating back 11 years) and adjusted it to the codebase so this should be ready to be merged.

akurtakov:
I'm totally deferring this one to @jonahgraham as it needs changes in cdt too.

laeubi:
Looking at the history there seem to be not much changes, so one option would be to start clean and not migrate the history at all WDYT?

vogella:
Nice to see the terminal moved to platform. Thanks @laeubi for this work. Is their an issue describing the background of this change?

merks:
There's this N&N entry now:

https://eclipse.dev/eclipse/markdown/?f=news/4.37/platform.md

jonahgraham:
> Looking at the history there seem to be not much changes, so one option would be to start clean and not migrate the history at all WDYT?

This code has been moved around so much in the past that key parts of its history are missing from the history. I would recommend copying the current CDT HEAD commit for these bundles and make a reference to CDT's commit sha1 to keep tracking the history farther back. 

FTR this code originated in TCF, was moved to TM and then CDT before ending up here.

laeubi:
> > Looking at the history there seem to be not much changes, so one option would be to start clean and not migrate the history at all WDYT?
> 
> This code has been moved around so much in the past that key parts of its history are missing from the history. I would recommend copying the current CDT HEAD commit for these bundles and make a reference to CDT's commit sha1 to keep tracking the history farther back.
> 
> FTR this code originated in TCF, was moved to TM and then CDT before ending up here.

Done in
- https://github.com/eclipse-platform/eclipse.platform/pull/1965

PR #1965, opened 2025-06-17
Migrate org.eclipse.tm.terminal.view.core

laeubi:
Migrated from
https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.view.core Commit 9b6fded373b3518ebfa6abe307d5ee6673f30c33


laeubi:
@jonahgraham I plan to merge this once the build is green if there are no objections.

merks:
This sounds like a good plan. 

Have you applied all the formatting and cleanups?

laeubi:
> Have you applied all the formatting and cleanups?

I have added minimal fixups to make the build happy, just in case someone wants to compare this with the CDT state. Once the migration is done I plan to add the `terminal` folder to the usual cleanup actions and then let the cleanup be performed on top of it.

PR #1966, opened 2025-06-17
Migrate org.eclipse.tm.terminal.view.ui

laeubi:
Migrated from
https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.view.ui Commit a1cd8a948f6d788e0c59fb81f11f1ff4f96facaf

FYI @jonahgraham now [the core bundle is in](https://github.com/eclipse-platform/eclipse.platform/pull/1965), this is the next one that can be migrated as it now only depends on other platform things.

As before I raised the version to 4.12.0 to account for the migration.


laeubi:
The build is now green, if there are no concerns I would like to merge this, the change to not include a fixed version in CDT is here:

- https://github.com/eclipse-cdt/cdt/pull/1209

jonahgraham:
LGTM.

Although I guess this initiates another job which is to migrate all the new icons added here. They are all .gif files, so two technologies (gif -> png -> svg) beyond where Platform is now leading.

laeubi:
@jonahgraham yes I expect some kind of cleanups happen here, I currently just migrate with as less changes as possible (just in case some one wants to compare with previous CDT state), but more has to come on top of the migration after time :-)

PR #1968, opened 2025-06-18
Create a feature of the terminal bundles for target/product consumption

laeubi:
Now we have more and more bundles migrated from the cdt-terminal it seems useful to group them into a feature, for example to be consumed in a target or product.


laeubi:
With this in place I'm able to start a Child Eclipse from the SDK with the Terminal View available (but complaining no connectors are there), so the next step would be to migrate the local connector at least.

PR #1970, opened 2025-06-18
Migrate `org.eclipse.tm.terminal.connector.process`

laeubi:
Migrated from

https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.connector.process

at commit `2ba5b3d68a373e3aa301fa50ec48a3a77d92a2ad`

FYI @jonahgraham 

This is required for the local connector what itself should be migrated to make this feature standalone usable. Currently using the Terminal View (and also the new https://github.com/eclipse-platform/eclipse.platform/pull/1969 ) will result in a message that no connector is available. While one can install them from the CDT site it is quite inconvenient, so it is the next logical step to bring them over here.


laeubi:
Removal form CDT feature is here:
- https://github.com/eclipse-cdt/cdt/pull/1212

jonahgraham:
This LGTM - but I assume there is some cleanup like you had to do on the last one because of Javadoc/etc errors.

PR #1975, opened 2025-06-19
migrate org.eclipse.tm.terminal.connector.ssh

laeubi:
Migrated from

https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.connector.ssh

at commit

2ba5b3d68a373e3aa301fa50ec48a3a77d92a2ad

FYI @jonahgraham 

This is the last connector to be migrated, we now then have:

- local terminal
- shh
- telnet

available.


laeubi:
Only flapping compiler difference

PR #1977, opened 2025-06-20
Migrate org.eclipse.tm.terminal.test

laeubi:
Migrated from

https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.test

at commit

2ba5b3d68a373e3aa301fa50ec48a3a77d92a2ad

FYI @jonahgraham 

As we have now migrated all relevant bits here it seems a good time to also add the tests.


laeubi:
@jonahgraham Even though the project has gone renaming maybe far to often in the future I think it hopefully will now settle at platform.

Therefore I would suggest to rename `org.eclipse.tm.terminal.x` -> `org.eclipse.terminal.x`  like we have already for other parts in platform (e.g. team and debug).

This would mean adjusting the remaining two connectors in cdt ([remote](https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.connector.remote) and [cdtserial](https://github.com/eclipse-cdt/cdt/tree/main/terminal/plugins/org.eclipse.tm.terminal.connector.cdtserial)) to the new namespace, and either _delete_ or _update_ the [corresponding features](https://github.com/eclipse-cdt/cdt/tree/main/terminal/features) to the new name(s). If you like I can propose a PR once M1 is out.

jonahgraham:
> If you like I can propose a PR once M1 is out.

Yes please.

jonahgraham:
Or earlier - we can update CDT to use a pre-M1 I-build of platform and synchronize all the change from M1. Whatever works best for you. I am almost completely unavailable though from Jun 25 - Jul 2 inclusive.

PR #1979, opened 2025-06-24
Rename eclipse.tm.terminal > eclipse.terminal

laeubi:
FYI @merks @akurtakov @jonahgraham


terminal/plugins/org.eclipse.tm.terminal.control/META-INF/MANIFEST.MF
```diff
@@ -1,25 +1,25 @@
...
- org.eclipse.tm.terminal.model
-Automatic-Module-Name: org.eclipse.tm.terminal.control
+Export-Package: org.eclipse.terminal.internal.connector;x-internal:=true,
+ org.eclipse.terminal.internal.control,
+ org.eclipse.terminal.internal.control.actions,
+ org.eclipse.terminal.internal.control.impl;x-internal:=true,
+ org.eclipse.terminal.internal.emulator;x-internal:=true,
+ org.eclipse.terminal.internal.model;x-internal:=true,
+ org.eclipse.terminal.internal.preferences,
+ org.eclipse.terminal.internal.provisional.api,
```
jonahgraham, line 21: Note that the before naming of these packages isn't great - but many of these packages with "internal" in their name are API (no x-internal tag). This means that consumers of this API now have to adapt.

For example in other parts of CDT we [use these packages](https://github.com/eclipse-cdt/cdt/blob/1e42e5f0e1e6591b5b609db0b864affc746ed5bf/dsf-gdb/org.eclipse.cdt.dsf.gdb.ui/src/org/eclipse/cdt/dsf/gdb/internal/ui/console/GdbFullCliConsolePage.java#L49-L51).

These packages were explicitly API, see for example https://github.com/eclipse-cdt/cdt/commit/3d2cdcc9860c0841b2f7237ffaa7233a2a2cdcf0

Therefore, I don't think we can rename these packages without even more coordination with CDT (and possibly other clients - but honestly I don't know if there are any)

jonahgraham: PS If I said something earlier that implied approving of this change then I apologize for my confusion over what was being renamed. I am ok to rename the features, and I would be ok to rename the bundles/packages in alignment with other projects (mostly CDT)

laeubi: @jonahgraham yes consumers need to adapt when migrating to the new namespace, I also think we should make things that are not really "internal" should be made proper API now. I Plan to do that for the `provisional.api` packages as a next step.

PR #1992, opened 2025-06-25
Stop using deprecated Locale constructors

akurtakov:


laeubi:
It would be great if one could have cleanups for such things but it seems it is not detected, @jjohnstn can such things only be discovered in user-libraries or also in JDK code?

Issue #2003, opened 2025-07-01
Error when changing the name of a new Debug Configuration


pdeville2:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against Eclipse 2025-06 (https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

* I open the debug configuration
* Create a new configuration (for example Eclipse Application) and save clicking in Apply button
* Modify the name of the same debug configuration
* Click on "new launch configuration"
* A Pop up ask for saving the first config -> click on save

I expected: the first config is saved and the new config can be editing

But got: the first config is saved but I got this error message and the new configuration has not been created 

![Image](https://github.com/user-attachments/assets/0e7884b7-65b1-4f7a-92d5-7114f4dee7c3)

## Here is some relevant log output

There is no exception in the console

## Tested under this environment:
* OS & version: Windows 10
* Eclipse IDE/Platform version (as shown in *Help > About*): ~Version: 2025-06 (4.36.0) - Build id: 20250605-1316~ **2021-09 (4.21)** and maybe even earlier, I didn't look further (@fedejeanne, see https://github.com/eclipse-platform/eclipse.platform/issues/2003#issuecomment-3022494651)

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

pdeville2:
I now that the exception is throw here `org.eclipse.debug.internal.core.LaunchManager.getInfo(LaunchConfiguration)` beacause the `config` does not exist anymore

fedejeanne:
Thank you for reporting this, @pdeville2 !

I tested it in a very old installation I had around (**2021-09 (4.21)**) and it was reproducible too. I didn't went further back but I assume this must have been there for ages, glad that you caught it!

Would you be open to proposing a PR to fix it? Or even a smaller PR to even properly show the error in the _Error Log_ view?

I only peeked at it and found the error message in the code, so those would be the places to start digging :-)

![Image](https://github.com/user-attachments/assets/985c3a3f-10f6-4a9b-b5a4-44313fa4d2be)

pdeville2:
[...]

PR #2004, opened 2025-07-01
Show errors in ErrorLog when exception is throw in launch configuration

pdeville2:
Fixes: #2003


fedejeanne (changes requested):
Thank you for the contribution! The code looks fine but when I reproduce the error, I see some extra errors that I can't explain:

image

Do you see them too? 

It would also be nice to have the stack trace in the details too. 

image

The good thing is that the stack trace appears in the (text) log file itself, so it's not a big problem.

Also, please correct your commit text: this PR does not **fix** #2003 (the bug still exists), it **contributes** to it by helping to find it. I would write it like this:

```
Log exceptions thrown when creating launch configurations #2003

Both in the log file and in the Error Log view.

Contributes to #2003

Signed-off-by: Pauline DEVILLE <pauline.deville@cea.fr>
```

The thing is that if you write `fixes #xyz` then the issue will be closed by GitHub, which is not what we want :-)

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/AbstractLaunchConfigurationAction.java
```diff
@@ -128,6 +130,7 @@ public void dispose() {
 	 */
 	protected void errorDialog(CoreException exception) {
 		ErrorDialog.openError(getShell(), null, null, exception.getStatus());
+		DebugUIPlugin.log(exception);
```
iloveeclipse, line 133: Just curious, would it work by replacing the code with
```
StatusManager.getManager().handle(exception.getStatus(), StatusManager.SHOW | StatusManager.LOG);
```

fedejeanne: Not really. I tried it and the result is that one looses the stack trace in the error log (text file). The details also don't show the stack trace:
image

Why do you ask?

laeubi: BTW it all heavily depends on **if the status is actually carry an exception** (what it seems not do) as the `Core Exception` is just a carrier object (what even can contain an OK status...).

As it seems not an _exception_ occurred, but a _condition_ is false that triggers a plain message, one will never see any exception trace in the error details.

fedejeanne: Thank you for the clarification! 

I see what you mean: I wanted exception 1 but got 2 (null) instead:

image

Then I'm fine with the code as it is right now, no need to do a big change here since one can always get the stack trace from the log file.

fedejeanne (approved):
I changed the commit text myself, LGTM now.

@iloveeclipse , @laeubi , should I merge?

fedejeanne:
BTW the odd messages about "Blurf" and so on are unrelated. I just tested this PR in my real IDE by hacking the corresponding JAR file in my Bundle Pool and I don't see those errors, just the one I want:

image

iloveeclipse:
> I changed the commit text myself, LGTM now.
> may I merge?

What is the point to merge an error log that says exactly the same as the dialog shown to user, without providing any additional information (not even a stack trace)?
Reading #2003 I don't see how this can help with further work or improve user experience.

fedejeanne:
I agree that it may not seem like much (in the UI) but the current state (`master`) is also missing the stack trace in the log file (text). This PR not only adds the entry in the _Error Log_ view but also the stack trace in the log files.

fedejeanne:
Here's the relevant point in #2003 
image

pdeville2:
Thank you for the review and modifications, sorry I was not active yesterday. It is not a big contribution for the user but at least we get a stack trace now.

fedejeanne:
I still think this is an improvement, however small, so I plan to merge this later today if there are no objections.

fedejeanne:
Thank you again @pdeville2 for reporting the issue and for this contribution! 👏 

I hope this is the first...
image

... of many to come :-)

PR #1862, opened 2025-05-08
Compare selection or editor with clipboard

SougandhS:
This feature will allow users to compare either the selected text or the entire editor content against the current clipboard contents.
<br>
image
<br>
If users want to just compare just a selection in an editor, select the section and choose _Compare with-> Clipboard_
(I had changes copied..)

https://github.com/user-attachments/assets/38d38751-1cd2-4c88-80f2-6d2f329744b7

<br>

https://github.com/user-attachments/assets/fc0d271b-fa47-4585-83cc-d19daabe7443

<br>
If User want to compare entire Editor, either select all or select None and choose _Compare with-> Clipboard_

https://github.com/user-attachments/assets/1d6d9f67-2e00-48c7-8f45-a5fbba2ba5c9

PR #2007, opened 2025-07-03
Support for modifying source in compare clipboard editor

SougandhS:
For https://github.com/eclipse-platform/eclipse.platform/pull/1862#issuecomment-3028138890

follow-up to https://github.com/eclipse-platform/eclipse.platform/pull/1862

Supports modifying the source from compare editor itself

https://github.com/user-attachments/assets/2f1b60b2-8786-491c-adb1-e30e751009ee

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ClipboardCompare.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ClipboardCompare.java
@@ -13,46 +13,13 @@
  *******************************************************************************/
 package org.eclipse.compare.internal;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.lang.reflect.InvocationTargetException;
-import java.nio.charset.StandardCharsets;
-
-import org.eclipse.compare.CompareConfiguration;
-import org.eclipse.compare.CompareEditorInput;
-import org.eclipse.compare.CompareUI;
-import org.eclipse.compare.IEncodedStreamContentAccessor;
-import org.eclipse.compare.ITypedElement;
-import org.eclipse.compare.structuremergeviewer.DiffNode;
 import org.eclipse.core.resources.IFile;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.jface.action.IAction;
 import org.eclipse.jface.dialogs.MessageDialog;
-import org.eclipse.jface.text.ITextSelection;
 import org.eclipse.jface.viewers.ISelection;
-import org.eclipse.osgi.util.NLS;
-import org.eclipse.swt.dnd.Clipboard;
-import org.eclipse.swt.dnd.TextTransfer;
-import org.eclipse.swt.graphics.Image;
-import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
-import org.eclipse.ui.IEditorPart;
-import org.eclipse.ui.IObjectActionDelegate;
-import org.eclipse.ui.IViewPart;
-import org.eclipse.ui.IWorkbenchPage;
-import org.eclipse.ui.IWorkbenchPart;
-import org.eclipse.ui.PlatformUI;
-import org.eclipse.ui.forms.editor.FormEditor;
-import org.eclipse.ui.texteditor.ITextEditor;
 
-public class ClipboardCompare extends BaseCompareAction implements IObjectActionDelegate {
 
-	private final String clipboard = "Clipboard"; //$NON-NLS-1$
-	private final String compareFailed = "Comparision Failed"; //$NON-NLS-1$
-
-	private IWorkbenchPart activePart;
+public class ClipboardCompare extends BaseCompareAction {
 
 	@Override
 	protected void run(ISelection selection) {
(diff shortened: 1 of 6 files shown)


SougandhS:
One more thing I have to handle

SougandhS:
https://github.com/user-attachments/assets/21b185b3-140b-4ceb-9699-72bfaaadcbc0

SougandhS:
Hi @iloveeclipse, I have added the missing **modify** functionality 👍

vogella:
@iloveeclipse I think you suggest that @SougandhS adds this. Can you review?

team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ClipboardCompare.java
```diff
@@ -74,32 +89,38 @@ protected boolean isEnabled(ISelection selection) {
...
+			String fileContents = new String(currentResouce.getContents().readAllBytes(), currentResouce.getCharset());
 			showComparison(fileContents, fileName, cb, parentShell);
 			return;
 		}
 		final String selectionContents;
-		if (editor instanceof FormEditor fromEditor) {
-			editor = fromEditor.getActiveEditor();
+		if (editor instanceof MultiPageEditorPart mpe) {
+			Object page2 = mpe.getSelectedPage();
+			if (page2 instanceof EditorPart e) {
```
tomaswolf: Use `IEditorPart` instead of `EditorPart`.

And in #2100 the same FormEditor code was also in `ClipboardReplace`. Should probably be done there, too (in some other commit), if that "Replace by clipboard" command is kept at all.

SougandhS: Done

vogella:
@SougandhS can you solve the merge conflict?

SougandhS:
> @SougandhS can you solve the merge conflict?

Done 👍

vogella (commented):
A few review notes (haven't pulled the branch, just read the diff):

1. Likely bug in `setContent` (around line 218): the branch `if (selectionLength <= 1)` overwrites the entire file with `newContent`. With a 1‑character partial selection, this silently replaces the whole file with just the edited char. The condition should be `if (!partialSelection)`, mirroring how `getContents` decides.

2. Mutable per-action state on a singleton delegate: `currentResouce`, `offSet`, `len`, `partialSelection` are now instance fields mutated inside the per-file loop and later read by `EditableFileNode#setContent` after the compare editor is open. A later invocation can flip behavior of an already-open compare editor. Pass these as parameters / capture them in the node instead.

3. Sentinel mixed with `Integer.MAX_VALUE`: `offSet`/`len` use `-1` as "no selection", and the else branch then constructs `new EditableFileNode(currentResouce, 0, Integer.MAX_VALUE)`. It only works because `getContents` clamps via `Math.min`. A single boolean inside the node would be cleaner.

4. Dialogs from model code: `EditableFileNode#getContents` and `setContent` call `MessageDialog.openError(...)` directly; these can be invoked off the UI thread by the compare framework. Prefer `ILog.of(getClass()).error(...)` and rethrow / return, so the framework surfaces the error.

5. Cosmetics: typo `currentResouce` (should be `currentResource`), field `offSet` (should be `offset`), variable name `page2`, and `ClipboardCompare.java` is missing a trailing newline. Pre-existing typo `"Comparision Failed"` could be fixed in passing.

6. No tests added; given the offset/length math, a test that opens a clipboard compare on a partial selection and writes back would be valuable.

7. Edge case: captured `selectionOffset`/`selectionLength` become stale if the underlying file is modified between opening the compare and saving.

SougandhS:
>A few review notes (haven't pulled the branch, just read the diff):

Thanks, will check this comments 👍

SougandhS:
> A few review notes (haven't pulled the branch, just read the diff):
> 
> 1. Likely bug in `setContent` (around line 218): the branch `if (selectionLength <= 1)` overwrites the entire file with `newContent`. With a 1‑character partial selection, this silently replaces the whole file with just the edited char. The condition should be `if (!partialSelection)`, mirroring how `getContents` decides.

Fixed 👍 

> 2. Mutable per-action state on a singleton delegate: `currentResouce`, `offSet`, `len`, `partialSelection` are now instance fields mutated inside the per-file loop and later read by `EditableFileNode#setContent` after the compare editor is open. A later invocation can flip behavior of an already-open compare editor. Pass these as parameters / capture them in the node instead.

Fixed 👍 

> 3. Sentinel mixed with `Integer.MAX_VALUE`: `offSet`/`len` use `-1` as "no selection", and the else branch then constructs `new EditableFileNode(currentResouce, 0, Integer.MAX_VALUE)`. It only works because `getContents` clamps via `Math.min`. A single boolean inside the node would be cleaner.

Refactored 👍 

> 4. Dialogs from model code: `EditableFileNode#getContents` and `setContent` call `MessageDialog.openError(...)` directly; these can be invoked off the UI thread by the compare framework. Prefer `ILog.of(getClass()).error(...)` and rethrow / return, so the framework surfaces the error.

Used `ILog.` 👍 

> 5. Cosmetics: typo `currentResouce` (should be `currentResource`), field `offSet` (should be `offset`), variable name `page2`, and `ClipboardCompare.java` is missing a trailing newline. Pre-existing typo `"Comparision Failed"` could be fixed in passing.

Refactored 👍 

> 6. No tests added; given the offset/length math, a test that opens a clipboard compare on a partial selection and writes back would be valuable.

handled this now in a simpler way now 👍 

> 7. Edge case: captured `selectionOffset`/`selectionLength` become stale if the underlying file is modified between opening the compare and saving.

Fixed this huge bug 👍

SougandhS:
Hi @vogella, could you please re-check this when you get some time ?

PR #2013, opened 2025-07-08
Make delays depending on available processors

laeubi:
Currently we use a fixed delay for a given job priority, depending on the number of jobs running and other factors this can considerably delay some actions and make eclipse feel laggy.

This now makes these values dependent on the number of CPUs installed in the machine, as we can assume more CPUs can do the work better in parallel without additional sleep delays.

FYI @iloveeclipse @merks @vogella 

I'll keep this PR open for a few days in case someone likes to review this, but the change seem save enough as I tried to even reduce the delay always to zero and didn't see any negative effects here. In many cases (e.g. no builds or other jobs running) this is even zero in some cases as the value is multiplied with the number of running jobs. In other cases it seems even unused or only used for rescheduling purpose.

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
+++ b/runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
@@ -78,6 +78,19 @@ import org.eclipse.osgi.util.NLS;
  */
 public class JobManager implements IJobManager, DebugOptionsListener {
 
+
+	private static final int CORES = Runtime.getRuntime().availableProcessors();
+
+	private static final long DELAY_PRIORITY_INTERACTIVE = 0;
+
+	private static final long DELAY_PRIORITY_DECORATE = 1000 / CORES;
+
+	private static final long DELAY_PRIORITY_BUILD = 500 / CORES;
+
+	private static final long DELAY_PRIORITY_LONG = 100 / CORES;
+
+	private static final long DELAY_PRIORITY_SHORT = 50 / CORES;
+
 	private static final int NANOS_IN_MS = 1_000_000;
 
 	/**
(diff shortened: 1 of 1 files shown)


iloveeclipse:
I've run JDT and JDT debug tests with this change, looks fine. In general this change makes sense, with high number of cores there is no need to wait a second for decoration thread to start, there is almost always a core free. Original code was written more than 20 years ago, [highly likely on a single or dual core PC](https://www.google.com/search?q=what+is+a+typical+PC+from+2005), therefore it was valid to delay "unimportant" jobs. Now on my workstation with 32 cores it is a completely different picture, delaying only wastes my time.

iloveeclipse:
I tend to merge this PR now and trigger an IBuild to see whether any unexpected side effects will appear due the timing change. Any objections?

laeubi:
Alright then go a haed and trigger a build.

iloveeclipse:
https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/58/ started

merks:
@iloveeclipse  @akurtakov 

I'm considering whether it maybe makes sense to update the Platform's SimRel M1 contribution before tomorrow's deadline?  What's your opinion?  How serious/noticeable is this problem generally?

iloveeclipse:
> @iloveeclipse @akurtakov
> 
> I'm considering whether it maybe makes sense to update the Platform's SimRel M1 contribution before tomorrow's deadline? What's your opinion? How serious/noticeable is this problem generally?

For me it was a deal breaker. So if the tests would not show any issues, why not?

merks:
Yes, that was my sense/guess.

I think the only potential drawback is that RAP's contribution might introduce duplicates, but I don't think those duplicates are harmful.

@mknauer

Would it be problematic to contribute a new version of the Platform to SimRel later today?  When do you actually finalize the RAP contribution?

vogella:
I todays I-Build I see a lot of errors during startup of an RCP if I do not clean-up the project before running.

Stuff like:

java.lang.Error: Unresolved compilation problems: 
	The import org.eclipse cannot be resolved
	The import org.eclipse cannot be resolved

If I run "Project -> Clean" the the project is fine again, auto-build is active.

What I do to see the errors:

1.) Start RCP app
2.) Stop it
3.) Modify some code
4.) Start again -> errors during startup, no error markers in the IDE
5.) Clean-all and start works again

Not sure if this is related to this change but it seems to be a good candidate, so if someone else is seeing this....

merks:
It works for me:

image

PR #2014, opened 2025-07-09
Use HttpWhiteboard instead of extension point for NavServlet

laeubi:
Currently the help center uses equinox specific extension points to register its servlets but there is actually a standard OSGi way to do so with R7 the Http Whiteboard Service.

This migrates the NavServlet as a very first step to investigate if any issues are to be discovered. If that works all other servlets should be migrated as well.


laeubi:
Has to be finally validated in tomorrows I-Build...

akurtakov:
I would recommend starting an I-build now and monitoring/testing it earlier.

laeubi:
Whatever seems suitable, I just don't want to waste to much build time resources here :-)

merks:
I did test that the help system still works.  What part might stop working?

laeubi:
The `NavServlet` handles these links on the left:

grafik

merks:
That and all other buttons and links worked fine when I tried them.  So we can be hopeful there are no problems introduced.

laeubi:
I updated my SDK and HelpCenter is still working...

PR #2017, opened 2025-07-09
Add missing package versions to terminal view

laeubi:


merks:
I know there isn't necessarily any relation between the package version and the bundle version `4.11.0` but it seems a little strange that even the major version number is entirely different.  

---

Do we expect that we will start properly maintaining package version going forward?  This was last changed 5 years ago:

image

But API was added 9 months ago:

image

laeubi:
> I know there isn't necessarily any relation between the package version and the bundle version 4.11.0 but it seems a little strange that even the major version number is entirely different.

The Bundle version was taken from CDT (+1 minor version) so if we like we can possibly restart the bundle version with 1.0 now as well if we want as its another namespace. Or we can use 4.37 as per the Eclipse release... I have no strong opinion here.

> Do we expect that we will start properly maintaining package version going forward? 

API tools will complain if you add new API and do not raise the package version (unless there is no version at all), clearly bundle version and package version will not align after some time if we do only change some of the APIs. Sadly there won't be a chance to properly do this for previous releases, but since last release it should be checked, if not please let me know!

merks:
A bundle version 1.0 would have made sense but now this has escaped into the wild so back leveling it seems a bad idea. 

@HannesWell 

Do you have an opinion?

laeubi:
Its only a milestone so not officially released, I think its not too late we still can use `1.0` ... also the goal with adding package versions is that people use import package from now on.

merks:
Anyone regularly updating their ide might never see another update.  It’s just seems like inviting problems to happen.  I’m interested in a 3rd opinion.  Maybe no ide installs it by default and I worry about nothing.

HannesWell:
> A bundle version 1.0 would have made sense but now this has escaped into the wild so back leveling it seems a bad idea. 
>

Since this was only in a milestone (as Christoph pointed out) and is currently not in the SDK feature (maybe some EPPs have already changed to the migrated terminal?) I think it's not installed that often yet and it's acceptable to set back the bundle version.

laeubi:
Yes it should not be installed anywhere beside from someone downloading the Ibuild SDK and then install it manually. But in the end its just a random number and no one probably care... We can even say we go to Version `5` ...

> maybe some EPPs have already changed to the migrated terminal?

I'm currently struggle to integrate this into CDT and linux tools so that's higly unlikely...

merks:
Yes I’m probably being over conservative.   You guys can decide.

merks:
We're in luck, I forgot that I was supposed to add the new feature to SimRel so it's not actually in SimRel M1:

image

SO version 1.0 for bundle and package (for all these new things), would be good...

PR #2021, opened 2025-07-10
Set terminal bundle versions to 1.0 and import package API

laeubi:
FYI @merks @HannesWell this now makes use of exported package to ensure no internal API is required to implement extensions. There is only one exception where the NLS of another bundle is reused but this should be cleared out later.

The package structure sometimes look a bit more nested / overengineered in some places (e.g. `org.eclipse.terminal.view.core` / `org.eclipse.terminal.view.core.interfaces` / `org.eclipse.terminal.view.core.interfaces.constants` ... ) but this is also something for another PR.

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/META-INF/MANIFEST.MF
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/META-INF/MANIFEST.MF
@@ -2,24 +2,30 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.terminal.connector.local;singleton:=true
-Bundle-Version: 4.9.0.qualifier
+Bundle-Version: 1.0.0.qualifier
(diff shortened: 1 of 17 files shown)


stephan-herrmann:
:warning:  It looks like this change is breaking builds as witnessed by [today's Y-build](https://ci.eclipse.org/releng/job/YPBuilds/job/Y-build-4.37/19/console), because https://github.com/eclipse-platform/eclipse.platform/blob/master/debug/org.eclipse.debug.terminal/META-INF/MANIFEST.MF requires bundle `org.eclipse.terminal.control;bundle-version="5.6.0"` whereas this bundle has just been downgraded here from 5.6.100 to 1.0.0.

(despite being a Y-build, all platform artifacts where built from master).

merks:
While fixing this problem I noticed another error in the IDE as well.  The following should address them both:

https://github.com/eclipse-platform/eclipse.platform/pull/2026

laeubi:
@stephan-herrmann thanks for reporting and thanks @merks for fixing this. The problem in this particular case is that the build fetches the old artifact from the I-Builds repository in this case so it can build here without a problem but in the i-build it will fail as we do not have the repo enabled there.

PR #2025, opened 2025-07-10
Reduce the number of subpackages for public API

laeubi:
Currently there are a quite complex distinction between "the package", "the interfaces" and "the constants" so consumers need to properly import most of the time three packages.

This now reduces this by moving the interfaces and constant holders into the main package, so less package versions to maintain and less to import.


laeubi:
Now I was able to make a lot of stuff internal / non API and I checked with linux tools and cdt that they also won't need to access any internal anymore with this change.

PR #2026, opened 2025-07-10
Fix version problems

merks:


merks:
@laeubi 

FYI these seem to be two errors that went undetected by verification builds and unnoticed in the IDE.

HannesWell:
Thanks for this fix. The latest rebuild of my PR in the `eclipse.platform.releng.aggregator` confirms that everything is fine again.

laeubi:
Thanks for fixing!

Issue #2028, opened 2025-07-11
API error in org.eclipse.core.resources since I20250710-1800


iloveeclipse:
There is an API error reported since I20250710-1800 on `org.eclipse.core.resources` project due the super class changed for `org.eclipse.core.resources.ProjectScope` which is public API. 

The super class is coming from `org.eclipse.equinox.preferences`, the related change is https://github.com/eclipse-equinox/equinox/pull/1066.

Additionally to the API error reported:
- `org.eclipse.core.runtime` should reexport **latest** `org.eclipse.equinox.preferences` version.
- minor segment of `org.eclipse.core.runtime` bundle version should be bumped because it re-exports new API from `org.eclipse.equinox.preferences` that bumped minor segment
- `org.eclipse.core.resources` need a bump on `org.eclipse.core.runtime` dependency and need a version bump due changed supertype of `org.eclipse.core.resources.ProjectScope`.

I will push a PR fixing issues above.

PR #2029, opened 2025-07-11
Updated bundle dependencies and versions after API change in equinox

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2028

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
+++ b/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.22.300.qualifier
+Bundle-Version: 3.23.0.qualifier
 Bundle-Activator: org.eclipse.core.resources.ResourcesPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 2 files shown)


merks:
I tried to fix this yesterday evening, but then the build complained that the version increment wasn't needed; but I didn't change the import ranges so maybe that will work better.

merks:
@iloveeclipse 

Yes, this is what failed last night.  😭 

image

@laeubi 

Somehow the IDE and the build don't agree on whether there are API changes or not.  Certainly the class hierarchy has changed...

laeubi:
> Somehow the IDE and the build don't agree on whether there are API changes or not. Certainly the class hierarchy has changed...

The problem is that we have cyclic dependencies between repositories and unless things are published to the Ibuilds its not visible to the build. But when running inside an IDE everything is there (but might be shaded by something from the target platform)... then we have these nasty reexports and Api tools do not always handle them right.

So sometimes one need to change things "incremental" for it to be fully passing the validation.

resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.22.300.qualifier
+Bundle-Version: 3.23.0.qualifier
```
laeubi, line 5: I tend to agree with API tools here, according to the changed manifest you do not change API but only required bundles (that are not reexported).

iloveeclipse: API is changed because supertype is changed. API tools in IDE says that.

iloveeclipse:
@merks, @laeubi : I believe there was a way to "mute" new *jenkins* warnings for the quality gate, I can't remember how.
 
We should marge this PR & "fix" that on jenkins side.

laeubi:
Well you can always press the merge button :-)

merks:
@iloveeclipse 

I _guess_ if we just merge this that there will be a new build that will become the reference build and then outstanding will become 55:

image

@laeubi 

Is that correct?

PR #2027, opened 2025-07-11
Make connector API provisional and prevent static singleton access

laeubi:
Currently Connector API is public API (what seems unnecessary as one can
query connectors by ID), also there are some static singleton services
that binds consumers to that explicit class and might hinder future
evolution.

This now makes all connector API provisional except for one shared with
another connector but reducing that part as much as possible. Also some
interfaces are extracted and registered as services to allow loose
coupling. In one case the static access can be even avoided completely
now.

PR #2030, opened 2025-07-11
Introduce `TerminalViewId` instead of nullable `id` and `secondaryId`

ruspl-afed:
Currently there is a pair of arguments in a number of signatures `String id, String secondaryId` each of which may be `null`.

Instead of this pair a new type introduced that also handles the logic of calculation for the next secondary id.

As a result the client code may be simplified to look like this: `UIPlugin.getConsoleManager().showConsoleView(new
TerminalViewId().next());`

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2025 Wind River Systems, Inc. and others. All rights reserved.
+ * This program and the accompanying materials are made available under the terms
+ * of the Eclipse Public License 2.0 which accompanies this distribution, and is
+ * available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Wind River Systems - initial API and implementation
+ * Max Weninger (Wind River) - [361363] [TERMINALS] Implement "Pin&Clone" for the "Terminals" view
+ * Alexander Fedorov (ArSysOp) - further evolution
+ *******************************************************************************/
+package org.eclipse.terminal.view.ui;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import org.eclipse.ui.IViewReference;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+
+/**
+ * Encapsulates the pair of mandatory primary terminal view id and optional secondary id
+ *
+ * @see IUIConstants#ID
+ * @see IViewReference#getSecondaryId()
+ */
+public record TerminalViewId(String primary, Optional<String> secondary) {
+
+	public TerminalViewId() {
+		this(IUIConstants.ID);
+	}
+
+	public TerminalViewId(String primary) {
+		this(primary, Optional.empty());
+	}
+
+	public TerminalViewId(String primary, String secondary) {
+		this(primary, Optional.ofNullable(secondary));
+	}
+
+	public TerminalViewId next() {
+		return new TerminalViewId(primary, nextSecondaryId(primary));
+	}
+
+	private static String nextSecondaryId(String primary) {
+		Map<String, IViewReference> terminalViews = new HashMap<>();
+		int maxNumber = 0;
+		for (IViewReference ref : viewReferences()) {
+			if (ref.getId().equals(primary)) {
+				if (ref.getSecondaryId() != null) {
+					terminalViews.put(ref.getSecondaryId(), ref);
+					int scondaryIdInt = Integer.parseInt(ref.getSecondaryId());
+					if (scondaryIdInt > maxNumber) {
+						maxNumber = scondaryIdInt;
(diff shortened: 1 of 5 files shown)


ruspl-afed:
@laeubi this is what I would like to suggest as further step for [our discussion](https://github.com/eclipse-platform/eclipse.platform/pull/2027)

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
```diff
@@ -0,0 +1,135 @@
...
+		return primary;
+	}
+
+	public Optional<String> secondary() {
+		return secondary;
+	}
+
+	public TerminalViewId next() {
+		return new TerminalViewId(primary, nextSecondaryId());
+	}
```
laeubi: All call-sites of next() seem to perform `new TerminalId().next()` so it this really suitable here as an instance method and should not be more a static method like


```suggestion
	public static TerminalViewId create() {
		return new TerminalViewId(IUIConstants.ID, nextSecondaryId(IUConstants.ID));
	}
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
```diff
@@ -0,0 +1,135 @@
...
+					return null;
+				}
+				return Integer.toString(i);
+			}
+		}
+		// add a new one
+		return Integer.toString(i + 1);
+	}
+
+	private List<IViewReference> viewReferences() {
```
laeubi: Can be static as it does not access any instance fields.

```suggestion
	private static List<IViewReference> viewReferences() {
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
```diff
@@ -0,0 +1,135 @@
...
+
+	public Optional<String> secondary() {
+		return secondary;
+	}
+
+	public TerminalViewId next() {
+		return new TerminalViewId(primary, nextSecondaryId());
+	}
+
+	private String nextSecondaryId() {
```
laeubi: Could be static by passing in `primary` that would safe us creating an instance just to call this method, see above.

```suggestion
	private static String nextSecondaryId(String primary) {
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
```diff
@@ -0,0 +1,135 @@
...
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+
+/**
+ * Encapsulates the pair of mandatory primary terminal view id and optional secondary id
+ *
+ * @see IUIConstants#ID
+ * @see IViewReference#getSecondaryId()
+ */
+public final class TerminalViewId {
```
laeubi: Should we probably make this a record class?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
ruspl-afed: I'm not so sure regarding this. 
We allow to specify primary id for some reason, perhaps we expect terminal view with another view id to use this API (may be similar with what we have for "Problem" view family).

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
ruspl-afed: I thought we don't want to use record for API types. But ok, I did.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
ruspl-afed: Personally I prefer to avoid static utilities, however I can adopt a preferred codebase style. Done.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
laeubi: @HannesWell @akurtakov do we have any such agreement for platform?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/TerminalViewId.java
ruspl-afed: Personally I prefer to avoid static utilities, however I can adopt a preferred codebase style. Done

ruspl-afed:
Thank you for your review @laeubi
All your comments were considered, the only point to discuss is "what is the purpose of using different primary terminal view id"

laeubi:
Sadly infrastructure is still down :-\
But Github looks good already.

@jonahgraham any thoughts about having different primary view IDs?

laeubi:
@ruspl-afed there are three javadoc issues to be fixed:

https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-2030/4/javadoc-warnings/new/

ruspl-afed:
> @ruspl-afed there are three javadoc issues to be fixed:

@laeubi everything is green now, I suggest to merge this PR and then create others if needed

PR #2033, opened 2025-07-13
Restore natures and buildspec for project descriptions

laeubi:
Currently a private project description does not store natures and builder, but these are mostly important these days.

This now also stores natures and builder specs.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocalMetaArea.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocalMetaArea.java
@@ -24,13 +24,17 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import org.eclipse.core.filesystem.URIUtil;
+import org.eclipse.core.internal.events.BuildCommand;
 import org.eclipse.core.internal.localstore.SafeChunkyInputStream;
 import org.eclipse.core.internal.localstore.SafeChunkyOutputStream;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
 import org.eclipse.core.resources.IBuildConfiguration;
+import org.eclipse.core.resources.ICommand;
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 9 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocalMetaArea.java
```diff
@@ -408,6 +428,33 @@ public void readPrivateDescription(IProject target, ProjectDescription descripti
...
 			description.setBuildConfigReferences(m);
+			// read parts since 3.23
+			String[] natureIds = new String[dataIn.readInt()];
+			for (int i = 0; i < natureIds.length; i++) {
+				natureIds[i] = dataIn.readUTF();
+			}
+			description.setNatureIds(natureIds);
+			int buildspecs = dataIn.readInt();
+			ICommand[] buildSpecData = new ICommand[buildspecs];
+			for (int i = 0; i < buildSpecData.length; i++) {
```
merks:
```suggestion
			for (int i = 0; i < buildspecs; i++) {
```

merks (approved):
Just a nit comment.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocalMetaArea.java
laeubi: I would assume that the compiler is able to do this equally well (as `buildSpecData.length` is final), but find it a bit more readable as it is the usual loop style over an array, while without that one always would thing a little about it if loops look uncommon.

merks: If you look at the byte code you won't need to assume. 😝 Anyway, it a nit and I don't actually care...

mickaelistria:
Is this easily compatible with a pre-3.23 workspace? I see
```
// read parts since 3.23
int natures = dataIn.readInt();
```
but when opening an older workspace, one wouldn't get the expected int from `dataIn`. Would an exception be thrown and be silenced without breaking anything later?

laeubi:
> Is this easily compatible with a pre-3.23 workspace

The code works to read as much as possible and stops on the first error (that was the case before). So an old Eclipse will simply ignore additional bytes, and old project specs will be read without the missing information.

mickaelistria:
Ack, thanks.

PR #2035, opened 2025-07-13
Continue #2027: use List instead of array for `ILaunchDelegateManager`

ruspl-afed:
* see https://github.com/eclipse-platform/eclipse.platform/pull/2027#discussion_r2200920811
* see https://github.com/eclipse-platform/eclipse.platform/pull/2027#discussion_r2200926031

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/ILaunchDelegateManager.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/ILaunchDelegateManager.java
@@ -9,8 +9,11 @@
  * Contributors:
  * Wind River Systems - initial API and implementation
  * Christoph Läubrich - extract to interface
+ * Alexander Fedorov (ArSysOp) - further evolution
  *******************************************************************************/
 package org.eclipse.terminal.view.ui.launcher;
+
+import java.util.List;
 
 import org.eclipse.jface.viewers.ISelection;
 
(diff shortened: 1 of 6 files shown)


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/ILaunchDelegateManager.java
```diff
@@ -22,9 +25,9 @@ public interface ILaunchDelegateManager {
 	 * @param unique If <code>true</code>, the method returns new instances for each
 	 *               contributed terminal launcher delegate.
 	 *
-	 * @return The list of contributed terminal launcher delegates, or an empty array.
+	 * @return The list of contributed terminal launcher delegates, or an empty list.
 	 */
-	ILauncherDelegate[] getLauncherDelegates(boolean unique);
+	List<ILauncherDelegate> getLauncherDelegates(boolean unique);
```
laeubi, line 30: I would rather use

```suggestion
	Stream<ILauncherDelegate> getLauncherDelegates(boolean unique);
```

instead of List what would require to document otherwise if the returned list is a copy and can be modified by the caller.

ruspl-afed: I was scared by sibling method, that is asked for size of list and for element at specific index.
Currently implementation doesn't provide access to internals and has no constraints for the caller
If you insist that we either need to use `Stream` here or document an ability to modify the returned list please let me know.

laeubi: > I was scared by sibling method, that is asked for size of list and for element at specific index.

I think this is more because of the previous array nature of return type than really required.

If we look for example at `LocalLauncherHandler` there is actually a filtering happen and then the first match is chosen, something that can easily be expressed with `Stream.filter(...).findFirst()`

Then in `PropertyTester` it is checked for empty what can become `Stream.findFirst().isPresent()`

`LaunchTerminalSettingsDialog` also performs a filtering and then a mapping ...

`LaunchTerminalCommandHandler`  is the only place where we actually only want to know if we can omit the dialog, here one can easily collect a List (or even array as before).

I therefore think `Stream<..>` instead of list should be used here.

ruspl-afed: Created https://github.com/eclipse-platform/eclipse.platform/pull/2044

PR #2041, opened 2025-07-14
Make service trackers instance scoped in terminal.view.ui

laeubi:
Currently service trackers are using a static field, what could lead to wrong behavior when the bundle is restarted.

This now converts the tracker to instance fields to account for bundle restarts and clears up an unnecessary static reference in `TerminalService`.

FYI @ruspl-afed

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
@@ -53,7 +53,7 @@ import org.osgi.util.tracker.ServiceTracker;
  */
 public class UIPlugin extends AbstractUIPlugin {
 	// The shared instance
-	private static UIPlugin plugin;
+	private static volatile UIPlugin plugin;
 	// The scoped preferences instance
 	private static volatile ScopedEclipsePreferences scopedPreferences;
 	// The trace handler instance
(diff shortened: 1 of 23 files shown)


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
```diff
@@ -201,12 +203,12 @@ void activateContexts() {
 
 	@Override
 	public void stop(BundleContext context) throws Exception {
+		plugin = null;
```
ruspl-afed: I would make this one the very last statement, because statements below may try to access it via hooks (in theory)

laeubi:
> Shouldn't we close service trackers in `stop` method? I know that today this is not the use case, but just to show the right example.

After the stop method exit all trackers will be closed implicitly, but it would of course not harm.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
laeubi: All dependents should already be shut down at this point and we do only remove listeners, so it should not be an issue, but it also don't harm...

ruspl-afed: The recipe of a potential trouble is the following:
1) `plugin = null;`
2) continue to shutdown bundle with hooks that call `getDefault()`
3) NPE

PR #2044, opened 2025-07-14
Continue #2027 (and #2035): use `Stream` for ILaunchDelegateManager

ruspl-afed:
* see https://github.com/eclipse-platform/eclipse.platform/pull/2035#discussion_r2203837671

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/LauncherDelegateManager.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/LauncherDelegateManager.java
@@ -19,7 +19,9 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.eclipse.core.expressions.EvaluationContext;
 import org.eclipse.core.expressions.EvaluationResult;
@@ -31,6 +33,7 @@ import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExtension;
 import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.IExtensionRegistry;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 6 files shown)


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/dialogs/LaunchTerminalSettingsDialog.java
```diff
@@ -421,7 +421,8 @@ protected List<String> getTerminals() {
 						ITraceIds.TRACE_LAUNCH_TERMINAL_COMMAND_HANDLER, LaunchTerminalSettingsDialog.this);
 			}
 
-			List<ILauncherDelegate> delegates = UIPlugin.getLaunchDelegateManager().getLauncherDelegates(false);
+			List<ILauncherDelegate> delegates = UIPlugin.getLaunchDelegateManager().getLauncherDelegates(false)
+					.toList();
```
ruspl-afed, line 425: we use `toList()` here and below to keep tracing meaningful @laeubi

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/LauncherDelegateManager.java
```diff
@@ -224,23 +226,23 @@ public int compare(IExtension o1, IExtension o2) {
...
 	@Override
-	public List<ILauncherDelegate> getLauncherDelegates(boolean unique) {
+	public Stream<ILauncherDelegate> getLauncherDelegates(boolean unique) {
 		List<ILauncherDelegate> contributions = new ArrayList<>();
 		for (Proxy launcherDelegate : getExtensions().values()) {
 			ILauncherDelegate instance = unique ? launcherDelegate.newInstance() : launcherDelegate.getInstance();
 			if (instance != null && !contributions.contains(instance)) {
 				contributions.add(instance);
 			}
 		}
```
laeubi: Instead of adding everything to a list first and then return a stream of it this whole method can be rewritten into a stream like

`return getExtensions().values().stream().map(proxy->unique ? launcherDelegate.newInstance() : launcherDelegate.getInstance()).filter(Objects::nonNull).distinct();`

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/dialogs/LaunchTerminalSettingsDialog.java
laeubi: I really wonder if the tracing is useful at all but we can keep it for now I think.

PR #2045, opened 2025-07-14
Rename `ILaunchDelegateManager` to `ILauncherDelegateManager`

ruspl-afed:
Since this type is all about `ILauncherDelegate` and not about `ILaunchDelegate` from Debug subsystem, I believe it is better to rename it and avoid confusion.


ruspl-afed:
@laeubi I think it was a kind of misprint at some stage, as I also typed `ILaunchDelegate` instead of `ILauncherDelegate` several times

laeubi (approved):
Seems reasonable.

PR #2047, opened 2025-07-15
Terminal Connector creation: throw instead of returning `null`

ruspl-afed:
Returning `null` from the `create` method is very confusing. If terminal connector cannot be created the `CoreException` should be thrown


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/ILauncherDelegate.java
```diff
@@ -93,6 +94,7 @@ public interface ILauncherDelegate extends IExecutableExtension, IAdaptable {
 	 *
 	 * @param properties The terminal properties. Must not be <code>null</code>.
 	 * @return The terminal connector or <code>null</code>.
```
laeubi: The method no longer return `null`

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/connector/Messages.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.terminal.internal.connector;
+
+import org.eclipse.osgi.util.NLS;
+
+public class Messages extends NLS {
```
laeubi: Should we probably reuse `TerminalMessages` here?

ruspl-afed: Do you mean that plan `Messages` may later conflict with another class with the same name and this is a kind of good practice to use compound class name for messages?
If so, why `TerminalMessages` and not `ConnectorMessages`? (I'm looking for recipe to compose such a name for general case)

ruspl-afed: Yes, sure, I tried to find an existing bundle but failed

PR #2049, opened 2025-07-15
Move util `Env` class closer to its single usage

ruspl-afed:
`Env` class is partially copied from CDT and has very limited usage. Therefore, it should not be a part of API and it is better to move it to internal package, closer to its single usage.


laeubi:
Good catch!

PR #2050, opened 2025-07-15
Make `org.eclipse.terminal.view.core.utils` subsystem-internal

ruspl-afed:
The API package `org.eclipse.terminal.view.core.utils` may be converted to subsystem-internal, since types it provides has questionable value outside the scope of terminal subsystem implementation.

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
@@ -108,7 +108,7 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 		// Initialize the local terminal working directory.
 		if (!properties.containsKey(ITerminalsConnectorConstants.PROP_PROCESS_WORKING_DIR)) {
 			// By default, start the local terminal in the users home directory
-			String initialCwd = IPreferenceKeys.getPreferences()
+			String initialCwd = UIPlugin.getScopedPreferences()
 					.getString(IPreferenceKeys.PREF_LOCAL_TERMINAL_INITIAL_CWD);
 			String cwd = null;
 			if (initialCwd == null || IPreferenceKeys.PREF_INITIAL_CWD_USER_HOME.equals(initialCwd)
@@ -277,7 +277,7 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 			}
 		}
 		if (shell == null) {
-			shell = IPreferenceKeys.getPreferences().getString(IPreferenceKeys.PREF_LOCAL_TERMINAL_DEFAULT_SHELL_UNIX);
+			shell = UIPlugin.getScopedPreferences().getString(IPreferenceKeys.PREF_LOCAL_TERMINAL_DEFAULT_SHELL_UNIX);
 			if (shell == null || "".equals(shell)) { //$NON-NLS-1$
 				if (System.getenv("SHELL") != null && !"".equals(System.getenv("SHELL").trim())) { //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 					shell = System.getenv("SHELL").trim(); //$NON-NLS-1$
@@ -312,7 +312,7 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 
 		String arguments = (String) properties.get(ITerminalsConnectorConstants.PROP_PROCESS_ARGS);
 		if (arguments == null && !Platform.OS_WIN32.equals(Platform.getOS())) {
-			arguments = IPreferenceKeys.getPreferences()
+			arguments = UIPlugin.getScopedPreferences()
 					.getString(IPreferenceKeys.PREF_LOCAL_TERMINAL_DEFAULT_SHELL_UNIX_ARGS);
 		}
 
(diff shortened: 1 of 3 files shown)


terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
```diff
@@ -108,7 +108,7 @@ public void execute(Map<String, Object> properties, ITerminalService.Done done)
 		// Initialize the local terminal working directory.
 		if (!properties.containsKey(ITerminalsConnectorConstants.PROP_PROCESS_WORKING_DIR)) {
 			// By default, start the local terminal in the users home directory
-			String initialCwd = IPreferenceKeys.getPreferences()
+			String initialCwd = UIPlugin.getScopedPreferences()
```
laeubi, line 111: This is not the same! Previously the preferences of `org.eclipse.terminal.view.ui` where used, now `org.eclipse.terminal.connector.local` is used, what likely will not work!

The problem is that the terminal.view has the preference page to configure some aspects here that should be reused.

terminal/bundles/org.eclipse.terminal.view.core/META-INF/MANIFEST.MF
```diff
@@ -12,5 +12,10 @@ Bundle-ActivationPolicy: lazy
 Bundle-Localization: plugin
 Export-Package: org.eclipse.terminal.view.core;version="1.0.0",
  org.eclipse.terminal.view.core.internal;x-internal:=true,
- org.eclipse.terminal.view.core.utils;version="1.0.0"
+ org.eclipse.terminal.view.core.utils;version="1.0.0";
```
laeubi, line 15: A version on an internal package is not really useful (or you need to import it with provider version range) and I'm also not a fan of "friends" concept. Either others need to access the API, then it should be a proper API available for everyone or it should not accessed by others (except tests for example).

Also note that for example CDT still has two connector implementations as well, one would then check if they need some classes here as well.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/IPreferenceKeys.java
```diff
@@ -71,8 +69,4 @@ public interface IPreferenceKeys {
 	public final String PREF_LOCAL_TERMINAL_DEFAULT_SHELL_UNIX_ARGS = PREF_TERMINAL
 			+ ".localTerminalDefaultShellUnixArgs"; //$NON-NLS-1$
 
-	static ScopedEclipsePreferences getPreferences() {
```
laeubi, line 74: If you remove this, the whole `IPreferenceKeys` becomes quite useless as there is no simple way without using internal API to ever access any of these keys.

laeubi (changes requested):

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
ruspl-afed: Good catch, thanks!

terminal/bundles/org.eclipse.terminal.view.core/META-INF/MANIFEST.MF
ruspl-afed: > and I'm also not a fan of "friends" concept

This is important to know, I also have mixed feeling when I see dependent bundles mentioned.

ruspl-afed:
My main goal here was to remove `util` types from API package, but it seems that we should discuss them individually
1) `ScopedEclipsePreferences` - it is not yet clear for me why `ScopedPreferenceStore` was not used
2) `TraceHandler` - I'm in doubts that we need it so hard, may be we need employ `org.eclipse.osgi.service.debug.DebugTrace` for critical occurrences.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/IPreferenceKeys.java
ruspl-afed: What if we try to switch to `ScopedPreferenceStore`? And what would be the right way to provide its instance to other bundles? I don't like the idea of making Activator an API type and would love to learn more dynamic approach of doing so.

laeubi:
> My main goal here was to remove util types from API package, but it seems that we should discuss them individually

Yes probably replace one item at a time so we can get rid of the package then.

> ScopedEclipsePreferences - it is not yet clear for me why ScopedPreferenceStore was not used

I assume it is as often, that it is for historic reasons. I think best is to try replacing it and see if something is missing / not working anymore.

> TraceHandler - I'm in doubts that we need it so hard, may be we need employ org.eclipse.osgi.service.debug.DebugTrace for critical occurrences.

Yes using the build in tracing facility seems better as it is standard and configurable.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/IPreferenceKeys.java
laeubi: Activator is not really an API the method just redirects there (where one of course can unwind the call to the native ones to not reference it).

akurtakov:
It would be nice to get this one in for M2 this Friday.

ruspl-afed:
Obsolete by
* https://github.com/eclipse-platform/eclipse.platform/pull/2083
* https://github.com/eclipse-platform/eclipse.platform/pull/2082

ruspl-afed:
> It would be nice to get this one in for M2 this Friday.

Merged part by part @akurtakov

PR #2051, opened 2025-07-16
Replace `Done` callback with `CompletableFuture` result

ruspl-afed:
Currently `null` is tolerated as a callback for a number of terminal-related operations. The introduced empty implementation `NOOP` allows to prohibit `null` as a callback argument.

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.cdt.utils.pty.PTY;
 import org.eclipse.core.runtime.Assert;
@@ -44,7 +45,6 @@ import org.eclipse.terminal.connector.local.activator.UIPlugin;
 import org.eclipse.terminal.connector.local.controls.LocalWizardConfigurationPanel;
 import org.eclipse.terminal.connector.process.ProcessSettings;
 import org.eclipse.terminal.view.core.ILineSeparatorConstants;
-import org.eclipse.terminal.view.core.ITerminalService;
 import org.eclipse.terminal.view.core.ITerminalServiceOutputStreamMonitorListener;
 import org.eclipse.terminal.view.core.ITerminalsConnectorConstants;
 import org.eclipse.terminal.view.ui.IMementoHandler;
@@ -75,7 +75,7 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 	}
 
 	@Override
-	public void execute(Map<String, Object> properties, ITerminalService.Done done) {
+	public CompletableFuture<?> execute(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
 
 		// Set the terminal tab title
@@ -221,12 +221,10 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 				}
 			}
 		}
-
-		// Get the terminal service
-		ITerminalService terminal = getTerminalService();
-		// If not available, we cannot fulfill this request
-		if (terminal != null) {
-			terminal.openConsole(properties, done);
+		try {
+			return getTerminalService().openConsole(properties);
+		} catch (RuntimeException e) {
+			return CompletableFuture.failedFuture(e);
 		}
 	}
 
--- a/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
@@ -14,6 +14,7 @@ package org.eclipse.terminal.connector.local.launcher;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
@@ -37,9 +38,12 @@ public class LocalLauncherHandler extends AbstractHandler {
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		ISelection selection = selection(event);
-		UIPlugin.getLaunchDelegateManager().getApplicableLauncherDelegates(selection)
-				.filter(d -> "org.eclipse.terminal.connector.local.launcher.local".equals(d.getId())).findFirst() //$NON-NLS-1$
-				.ifPresent(d -> executeDelegate(selection, d));
+		Optional<ILauncherDelegate> delegate = UIPlugin.getLaunchDelegateManager()
+				.getApplicableLauncherDelegates(selection)
+				.filter(d -> "org.eclipse.terminal.connector.local.launcher.local".equals(d.getId())).findFirst(); //$NON-NLS-1$
+		if (delegate.isPresent()) {
+			executeDelegate(selection, delegate.get());
+		}
 		return null;
 	}
 
@@ -64,11 +68,15 @@ public class LocalLauncherHandler extends AbstractHandler {
 		return selection;
 	}
 
-	private void executeDelegate(ISelection selection, ILauncherDelegate delegate) {
+	private void executeDelegate(ISelection selection, ILauncherDelegate delegate) throws ExecutionException {
 		Map<String, Object> properties = new HashMap<>();
 		properties.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegate.getId());
 		properties.put(ITerminalsConnectorConstants.PROP_SELECTION, selection);
-		delegate.execute(properties, null);
+		try {
+			delegate.execute(properties);
+		} catch (Exception e) {
+			throw new ExecutionException(e.getMessage(), e);
+		}
 	}
 
 }
--- a/terminal/bundles/org.eclipse.terminal.connector.process/src/org/eclipse/terminal/connector/process/internal/ProcessLauncherDelegate.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.process/src/org/eclipse/terminal/connector/process/internal/ProcessLauncherDelegate.java
@@ -13,6 +13,7 @@
 package org.eclipse.terminal.connector.process.internal;
 
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.cdt.utils.pty.PTY;
 import org.eclipse.core.runtime.Assert;
@@ -22,7 +23,6 @@ import org.eclipse.terminal.connector.ITerminalConnector;
 import org.eclipse.terminal.connector.InMemorySettingsStore;
 import org.eclipse.terminal.connector.TerminalConnectorExtension;
 import org.eclipse.terminal.connector.process.ProcessSettings;
-import org.eclipse.terminal.view.core.ITerminalService;
 import org.eclipse.terminal.view.core.ITerminalServiceOutputStreamMonitorListener;
 import org.eclipse.terminal.view.core.ITerminalsConnectorConstants;
 import org.eclipse.terminal.view.ui.launcher.AbstractLauncherDelegate;
@@ -45,14 +45,12 @@ public class ProcessLauncherDelegate extends AbstractLauncherDelegate {
 	}
 
 	@Override
-	public void execute(Map<String, Object> properties, ITerminalService.Done done) {
+	public CompletableFuture<?> execute(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
-
-		// Get the terminal service
-		ITerminalService terminal = getTerminalService();
-		// If not available, we cannot fulfill this request
-		if (terminal != null) {
-			terminal.openConsole(properties, done);
+		try {
+			return getTerminalService().openConsole(properties);
+		} catch (RuntimeException e) {
+			return CompletableFuture.failedFuture(e);
 		}
 	}
 
--- a/terminal/bundles/org.eclipse.terminal.view.core/src/org/eclipse/terminal/view/core/ITerminalService.java
+++ b/terminal/bundles/org.eclipse.terminal.view.core/src/org/eclipse/terminal/view/core/ITerminalService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 - 2018 Wind River Systems, Inc. and others. All rights reserved.
+ * Copyright (c) 2011 - 2025 Wind River Systems, Inc. and others. All rights reserved.
  * This program and the accompanying materials are made available under the terms
  * of the Eclipse Public License 2.0 which accompanies this distribution, and is
  * available at https://www.eclipse.org/legal/epl-2.0/
@@ -8,12 +8,12 @@
  *
  * Contributors:
  * Wind River Systems - initial API and implementation
+ * Alexander Fedorov (ArSysOp) - further evolution
  *******************************************************************************/
 package org.eclipse.terminal.view.core;
 
 import java.util.Map;
-
-import org.eclipse.core.runtime.IStatus;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Terminal service.
@@ -21,40 +21,28 @@ import org.eclipse.core.runtime.IStatus;
 public interface ITerminalService {
 
 	/**
-	 * Client call back interface.
-	 */
-	public interface Done {
-		/**
-		 * Called when the terminal service operation is done.
-		 *
-		 * @param status The status of the terminal service operation.
-		 */
-		public void done(IStatus status);
-	}
-
-	/**
 	 * Opens a terminal asynchronously and invokes the given callback if done.
 	 *
 	 * @param properties The terminal properties. Must not be <code>null</code>.
-	 * @param done The callback to invoke if finished or <code>null</code>.
+	 * @return the {@link CompletableFuture}
 	 */
-	public void openConsole(Map<String, Object> properties, Done done);
+	public CompletableFuture<?> openConsole(Map<String, Object> properties);
 
 	/**
-	 * Close the terminal asynchronously and invokes the given callback if done.
+	 * Close the terminal asynchronously.
 	 *
 	 * @param properties The terminal properties. Must not be <code>null</code>.
-	 * @param done The callback to invoke if finished or <code>null</code>.
+	 * @return the {@link CompletableFuture}
 	 */
-	public void closeConsole(Map<String, Object> properties, Done done);
+	public CompletableFuture<?> closeConsole(Map<String, Object> properties);
 
 	/**
 	 * Terminate (disconnect) the terminal asynchronously and invokes the given callback if done.
 	 *
 	 * @param properties The terminal properties. Must not be <code>null</code>.
-	 * @param done The callback to invoke if finished or <code>null</code>.
+	 * @return the {@link CompletableFuture}
 	 */
-	public void terminateConsole(Map<String, Object> properties, Done done);
+	public CompletableFuture<?> terminateConsole(Map<String, Object> properties);
 
 	/**
 	 * Register the given listener to receive notifications about terminal events.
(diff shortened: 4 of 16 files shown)


terminal/bundles/org.eclipse.terminal.view.core/src/org/eclipse/terminal/view/core/ITerminalService.java
```diff
@@ -32,34 +33,48 @@ public interface Done {
 		public void done(IStatus status);
 	}
 
+	/**
+	 * Empty callback if no action is needed
+	 */
+	Done NOOP = new Done() {
```
laeubi: I think the whole `Done` interface can be replaced by a `Consumer<IStatus>` instead.

I also don't thing we need something like a noop one can simply calls 

`openConsole(Map<String, Object> properties, s->{});`

or we even better add default implemented methods that do not use a "done" parameter at all.

terminal/bundles/org.eclipse.terminal.view.core/src/org/eclipse/terminal/view/core/ITerminalService.java
```diff
@@ -32,34 +33,48 @@ public interface Done {
...
+
 	/**
 	 * Opens a terminal asynchronously and invokes the given callback if done.
 	 *
 	 * @param properties The terminal properties. Must not be <code>null</code>.
-	 * @param done The callback to invoke if finished or <code>null</code>.
+	 * @param done The callback to invoke if finished. Must not be <code>null</code>.
+	 * @see ITerminalService#NOOP
 	 */
 	public void openConsole(Map<String, Object> properties, Done done);
```
laeubi:
```suggestion
	public void openConsole(Map<String, Object> properties, Consumer<IStatus> statusConsumer);
	
	public default void openConsole(Map<String,Object> properties) {
	     openConsole(properties, s->{});
	}
```

and so on ...

laeubi: Thinking further about this returning a `CompletableFuture<IStatus>` might even be much better here ...

```suggestion
	public CompletableFuture<IStatus> openConsole(Map<String, Object> properties);
```

terminal/bundles/org.eclipse.terminal.view.core/src/org/eclipse/terminal/view/core/ITerminalService.java
ruspl-afed: Thank you for the comment that is very helpful to understand the expected style of code @laeubi

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/ILauncherDelegate.java
```diff
@@ -84,9 +86,9 @@ public interface ILauncherDelegate extends IExecutableExtension, IAdaptable {
 	 * Execute the terminal launch.
 	 *
 	 * @param properties The properties. Must not be <code>null</code>.
-	 * @param done The callback or <code>null</code>.
+	 * @return the {@link CompletableFuture} with result {@link IStatus}
 	 */
-	public void execute(Map<String, Object> properties, ITerminalService.Done done);
+	public CompletableFuture<IStatus> execute(Map<String, Object> properties);
```
ruspl-afed: @laeubi  may be we should return `IStatus` here to simplify caller's life?

I mean that caller statistically not interested in result at all, and want thing to `just happen`.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
```diff
@@ -221,12 +222,10 @@ public void execute(Map<String, Object> properties, ITerminalService.Done done)
...
-
-		// Get the terminal service
-		ITerminalService terminal = getTerminalService();
-		// If not available, we cannot fulfill this request
-		if (terminal != null) {
-			terminal.openConsole(properties, done);
+		try {
+			return getTerminalService().openConsole(properties);
+		} catch (CoreException e) {
+			return CompletableFuture.completedFuture(e.getStatus());
```
laeubi: Do you think it makes sense here to use 
```suggestion
			return CompletableFuture.failedFuture(e);
```

Maybe in general we should return CompletableFuture<?> and drop the status all togehter, in that case if a failure is happening (of any kind) we have a proper exception with a stacktrace.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
```diff
@@ -221,12 +222,10 @@ public void execute(Map<String, Object> properties, ITerminalService.Done done)
...
 		}
-
-		// Get the terminal service
-		ITerminalService terminal = getTerminalService();
-		// If not available, we cannot fulfill this request
-		if (terminal != null) {
-			terminal.openConsole(properties, done);
+		try {
+			return getTerminalService().openConsole(properties);
+		} catch (CoreException e) {
```
laeubi: Maybe even

```suggestion
		} catch (CoreException | RuntimeException e) {
```

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
```diff
@@ -37,9 +38,12 @@ public class LocalLauncherHandler extends AbstractHandler {
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
```
laeubi, line 39: Might be a candidate for CF as well?
```suggestion
	public CompletableFuture<?> execute(ExecutionEvent event) {
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -77,10 +79,9 @@ protected static abstract class TerminalServiceRunnable {
 		 * @param title The terminal tab title. Must not be <code>null</code>.
 		 * @param connector The terminal connector. Must not be <code>null</code>.
 		 * @param data The custom terminal data node or <code>null</code>.
-		 * @param done The callback to invoke if the operation finished or <code>null</code>.
+		 * @return the {@link CompletableFuture} with result {@link IStatus}
 		 */
-		public abstract void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data,
-				Done done);
+		public abstract IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data);
```
laeubi, line 83: this dies not match the javadoc, should be CF as well?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -186,26 +187,19 @@ protected final void executeServiceOperation(final Map<String, Object> propertie
...
 		} else {
-			try {
-				Display display = PlatformUI.getWorkbench().getDisplay();
-				display.asyncExec(() -> runnable.run(tvid, title, connector, data, done));
-			} catch (Exception e) {
-				// if display is disposed, silently ignore.
-			}
+			return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
+					Display.getDefault());
 		}
```
laeubi:
```suggestion
if (runnable.isExecuteAsync()) {
	return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
					Display.getDefault());
} else {
	return CompletableFuture.completedFuture(runnable.run(tvid, title, connector, data));
}
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/handler/LaunchTerminalCommandHandler.java
```diff
@@ -165,4 +163,20 @@ private boolean isValidSelection(ISelection selection) {
...
+	private void executeDelegate(ISelection selection, ILauncherDelegate delegate) throws ExecutionException {
+		Map<String, Object> properties = new HashMap<>();
+		properties.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegate.getId());
+		properties.put(ITerminalsConnectorConstants.PROP_SELECTION, selection);
+		executeDelegate(properties, delegate);
+	}
+
+	private void executeDelegate(Map<String, Object> properties, ILauncherDelegate delegate) throws ExecutionException {
+		try {
+			delegate.execute(properties).get();
```
laeubi: is waiting for the execution really necessary? We should better hook into the future and execute any actions that must happens after in a `thenAccept` or similar.

But it looks like the code before has also not waited for the completion at all.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/local/showin/DynamicContributionItems.java
```diff
@@ -128,7 +129,11 @@ public void run() {
 				properties.put(ITerminalsConnectorConstants.PROP_TRANSLATE_BACKSLASHES_ON_PASTE,
 						Boolean.valueOf(translate));
 
-				delegate.execute(properties, null);
+				try {
+					delegate.execute(properties).get();
```
laeubi: Same as above, we maybe want to pass the future even to the caller of the method(s)

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/AbstractLauncherDelegate.java
```diff
@@ -138,7 +140,15 @@ protected String getDefaultTerminalTitle(Map<String, Object> properties) {
...
-	protected ITerminalService getTerminalService() {
-		return UIPlugin.getTerminalService();
+	/**
+	 * Get the terminal service
+	 *
+	 * @return an instance of {@link ITerminalService}
+	 * @throws CoreException if service is not available
+	 *
+	 */
+	protected final ITerminalService getTerminalService() throws CoreException {
```
laeubi: Maybe return an Optional and let the caller decide if an exception should be thrown?

laeubi (commented):
> @laeubi may be we should return IStatus here to simplify caller's life?

I think a CF is better to reflect the async nature of this, but using `IStatus` at all seems questionable to me. an anonymous `CompletableFuture<?>` that simply is failed on any error seems more appropriate.. if we feel to pass a status a `CoreException` can then be used.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
ruspl-afed: But this return value is a part of `org.eclipse.core.commands.IHandler` signature and not used in practice, isn't it?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/AbstractLauncherDelegate.java
ruspl-afed: Here I would say that in this hierarchy, the absence of an `ITerminalService` is an abnormal condition.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
laeubi: One can call a Command (that is then handled by a handler)  programmatically and then retrieve the result, here it is most likely always ignored, so maybe we can just return null always.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/launcher/AbstractLauncherDelegate.java
laeubi: The better throw `IllegalStateEception` than a checked exception (or even `AssertionError`).

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
ruspl-afed: I think it should return `IStatus` here otherwise it will be the second level of Futures

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
ruspl-afed: I'll better keep both the signature and returned `null` to conform Platform essentials

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
laeubi: I can't follow...? As we do things async here (previously with done) it seems legal to return CF and then simply use `compose` ?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/handler/LaunchTerminalCommandHandler.java
ruspl-afed: I'm not sure what kind of action we can execute here, this is the end of the main flow for a handler

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/local/showin/DynamicContributionItems.java
ruspl-afed: I'm not sure what kind of action we can execute here, this is the end of the main flow for an `IAction`

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/handler/LaunchTerminalCommandHandler.java
laeubi: Then simply let it go and do not get (wait) for it... It would still be nice to maybe log any errors if they happen.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
laeubi: Okay strike my previous comment this is the runnable already executed inside an async operation so it feels safe here to block.

ruspl-afed: > this is the runnable already executed inside an async operation

yes, this is what I was trying to express, sorry for the vague wording.

ruspl-afed:
Now the snippet is @laeubi 
```java
		try {
			delegate.execute(properties);
		} catch (Exception e) {
			throw new ExecutionException(e.getMessage(), e);
                        //or log
		}
```

terminal/bundles/org.eclipse.terminal.connector.process/src/org/eclipse/terminal/connector/process/internal/ProcessLauncherDelegate.java
```diff
@@ -45,14 +45,12 @@ public IConfigurationPanel getPanel(IConfigurationPanelContainer container) {
...
 		Assert.isNotNull(properties);
-
-		// Get the terminal service
-		ITerminalService terminal = getTerminalService();
-		// If not available, we cannot fulfill this request
-		if (terminal != null) {
-			terminal.openConsole(properties, done);
+		try {
+			return getTerminalService().openConsole(properties);
+		} catch (Exception e) {
```
laeubi:
```suggestion
		} catch (RuntimeException e) {
```

While I'm not sure we should actually catch these, do we really expect  getTerminalService() or openConsole to ever throw?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -277,18 +269,19 @@ public void run(TerminalViewId tvid, final String title, final ITerminalConnecto
...
 					// After that schedule opening the requested console
+					IStatus[] result = new IStatus[1];
 					try {
 						Display display = PlatformUI.getWorkbench().getDisplay();
-						display.asyncExec(() -> doRun(tvid, title, connector, data, done));
+						display.asyncExec(() -> result[0] = doRun(tvid, title, connector, data));
 					} catch (Exception e) {
 						// if display is disposed, silently ignore.
 					}
+					return result[0];
```
laeubi: we would need to wait for the display thread here otherwise result is probably not set yet. What makes me feel that instead of returning `IStatus` we need to pass the future to the run method and let the implementation complete it (either successfully or exceptionally).

terminal/bundles/org.eclipse.terminal.connector.process/src/org/eclipse/terminal/connector/process/internal/ProcessLauncherDelegate.java
ruspl-afed: We have a number of `Assert` calls there, so I changed to `RuntimeException`

laeubi: Given that we have an assert in the line 49 already might it be valid to let them just bubble up?

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
ruspl-afed: I tried to remove this additional nesting and works for me

terminal/bundles/org.eclipse.terminal.connector.process/src/org/eclipse/terminal/connector/process/internal/ProcessLauncherDelegate.java
ruspl-afed: > might it be valid to let them just bubble up?

TBH I don't know. I would like to agree on policy regarding assert and then apply it through the terminal code base with dedicated PRs.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -186,26 +186,19 @@ protected final void executeServiceOperation(final Map<String, Object> propertie
...
+			return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
+					Display.getDefault());
 		} else {
-			try {
-				Display display = PlatformUI.getWorkbench().getDisplay();
-				display.asyncExec(() -> runnable.run(tvid, title, connector, data, done));
-			} catch (Exception e) {
-				// if display is disposed, silently ignore.
-			}
+			return CompletableFuture.completedFuture(runnable.run(tvid, title, connector, data));
```
jonahgraham, line 201: Note that effectively both of the branches of this if statement run the runnable immediately instead of queuing it on the display queue.

See https://github.com/eclipse-platform/eclipse.platform/issues/2155#issuecomment-3299092302

and Display.execute(Runnable) which is used by asyncSupplyStage. When Display.execute is called on the UI thread, then the runnable is run immediately.

laeubi: > When Display.execute is called on the UI thread, then the runnable is run immediately.

That's indeed the case... I think one can simply pass in `display::asyncExec` as the executor to get the desired effect as before.

PR #2052, opened 2025-07-16
Removes log methods from Terminal activator

jnerlich:
Using ILog.of is the more modern approach


laeubi:
Looks like a good approach, why closing?

jnerlich:
Replaced by: https://github.com/eclipse-platform/eclipse.platform/pull/2053

PR #2057, opened 2025-07-16
Add constant for "Open Terminal on Selection" command id

ruspl-afed:
The "Open Terminal on Selection" command is referenced from another bundle by its string identifier. The new constant is introduced to put this String value under API control.

### DIFF
--- a/debug/org.eclipse.debug.terminal/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.terminal/META-INF/MANIFEST.MF
@@ -5,7 +5,8 @@ Bundle-SymbolicName: org.eclipse.debug.terminal;singleton:=true
 Bundle-Vendor: %providerName
 Bundle-Version: 1.0.0.qualifier
 Import-Package: org.eclipse.terminal.connector;version="[1.0.0,2.0.0)",
- org.eclipse.terminal.control;version="[1.0.0,2.0.0)"
+ org.eclipse.terminal.control;version="[1.0.0,2.0.0)",
+ org.eclipse.terminal.view.ui;version="[1.0.0,2.0.0)"
 Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.debug.core;bundle-version="3.23.0",
  org.eclipse.cdt.core.native;bundle-version="[6.4.0,7.0.0)",
--- a/debug/org.eclipse.debug.terminal/src/org/eclipse/debug/terminal/ui/TerminalConsoleFactory.java
+++ b/debug/org.eclipse.debug.terminal/src/org/eclipse/debug/terminal/ui/TerminalConsoleFactory.java
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Christoph Läubrich - initial API and implementation
+ *     Alexander Fedorov (ArSysOp) - further evolution
  *******************************************************************************/
 package org.eclipse.debug.terminal.ui;
 
@@ -19,6 +20,7 @@ import org.eclipse.core.commands.Command;
 import org.eclipse.core.commands.ParameterizedCommand;
 import org.eclipse.core.runtime.ILog;
 import org.eclipse.terminal.connector.ITerminalConnector;
+import org.eclipse.terminal.view.ui.CommandIds;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.commands.ICommandService;
 import org.eclipse.ui.console.ConsolePlugin;
@@ -38,7 +40,7 @@ public class TerminalConsoleFactory implements IConsoleFactory {
 		if (commandService == null || handlerService == null) {
 			return;
 		}
-		Command command = commandService.getCommand("org.eclipse.terminal.view.ui.command.launchConsole");
+		Command command = commandService.getCommand(CommandIds.COMMAND_LAUNCHCONSOLE);
 		if (command == null) {
 			return;
 		}
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/CommandIds.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/CommandIds.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2025 ArSysOp and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Alexander Fedorov (ArSysOp) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.terminal.view.ui;
+
+/**
+ * Commands declared with "org.eclipse.ui.commands" extension point
+ */
+public interface CommandIds {
+
+	/**
+	 * "Open Terminal on Selection" command id
+	 */
+	String COMMAND_LAUNCHCONSOLE = "org.eclipse.terminal.view.ui.command.launchConsole"; //$NON-NLS-1$
+
+}


ruspl-afed:
Personally, I would prefer to have the dedicated type for Command Id instead of the usual `String`, but this may lead discussion too far. What do you think in general regarding exposing identifiers from plugin.xml as a regular API symbols? @laeubi

laeubi:
If it is needed externally why not in E4 Editor we even have a wizard to generate ID classes.

ruspl-afed:
> in E4 Editor we even have a wizard to generate ID classes.

In this case I would like to conform with it, do we have an example somewhere in Platform already?

laeubi:
It is nothing magical it looks for the example like this:

```
public class AppModelId {
	public static final String COMMANDPARAMETER_ORG_ECLIPSE_PDE_SPY_CORE_COMMAND_PARTID = "org.eclipse.pde.spy.core.command.partID";
	public static final String COMMAND_ORG_ECLIPSE_PDE_SPY_CORE_COMMAND = "org.eclipse.pde.spy.core.command";
	public static final String DYNAMICMENUCONTRIBUTION_ORG_ECLIPSE_PDE_SPY_CORE_DYNAMICMENUCONTRIBUTION_0 = "org.eclipse.pde.spy.core.dynamicmenucontribution.0";
	public static final String HANDLER_ORG_ECLIPSE_PDE_SPY_CORE_HANDLER_0 = "org.eclipse.pde.spy.core.handler.0";
	public static final String MENUCONTRIBUTION_ORG_ECLIPSE_PDE_SPY_CORE_MENUCONTRIBUTION = "org.eclipse.pde.spy.core.menucontribution";
	public static final String MENUCONTRIBUTION_ORG_ECLIPSE_PDE_SPY_CORE_MENUCONTRIBUTION_0 = "org.eclipse.pde.spy.core.menucontribution.0";
	public static final String MENUSEPARATOR_ORG_ECLIPSE_PDE_SPY_CORE_MENUSEPARATOR_0 = "org.eclipse.pde.spy.core.menuseparator.0";
	public static final String MENU_ORG_ECLIPSE_PDE_SPY_CORE_MENU_E4SPIES = "org.eclipse.pde.spy.core.menu.e4spies";
	public static final String MENU_WINDOW = "window";
	public static final String PARTSTACK_ORG_ECLIPSE_PDE_SPY_CORE_PARTSTACK = "org.eclipse.pde.spy.core.partStack";
	public static final String TOOLBAR_ORG_ECLIPSE_PDE_SPY_CORE_TOOLBAR = "org.eclipse.pde.spy.core.toolbar";
	public static final String TRIMBAR_ORG_ECLIPSE_PDE_SPY_CORE_TRIMBAR = "org.eclipse.pde.spy.core.trimbar";
	public static final String TRIMMEDWINDOW_ORG_ECLIPSE_PDE_SPY_CORE_WINDOW = "org.eclipse.pde.spy.core.window";
}
```

ruspl-afed:
Hmm, I'm  in doubts
I may change this to
1)
```java
interface IUIConstants {
    String COMMAND_ORG_ECLIPSE_TERMINAL_VIEW_UI_COMMAND_LAUNCHCONSOLE = "org.eclipse.terminal.view.ui.command.launchConsole"; //$NON-NLS-1$
}
```
but may be it would look better (and still machine and e4 friendly) as
2)
```java
interface IUIConstants { //well, I'm not sure, may be another one with better name
    interface CommandIds { //E4elementIds
        String COMMAND_LAUNCHCONSOLE //everything after bundle id?
            = "org.eclipse.terminal.view.ui.command.launchConsole"; //$NON-NLS-1$
    }
}

```

laeubi:
If you fell more comfortable with an `CommandIds` I see no problem but do not make it an inner interface of an interface :-P

ruspl-afed:
I'm looking for some more a less standard way to specify such things, 
having `CommandIds` in the "main" bundle package is fine for me, so for both human and machine it should be relatively simple to find it.

laeubi:
I think static final fields in interfaces are "standard" enough for everyone, also likely this is nothing used very often but only for integrator that usually closely inspect the code they are integrate with.

ruspl-afed:
Yes, true, but this is all about manual work that I would like to minimize. 
I'm looking for support from a Tool:
if I have any e3 extension or e4 model with referenceable id , this is a kind of API, that others may want to use from another bundle.
So I want two things to be addressed:
1) a standard and easily discoverable location of a corresponding Java symbol
2) an automated control for consistency for such non-Java contributions and Java symbols

Don't you see any benefit from this approach?

laeubi:
These are all good intentions, but I doubt we can fix that within the given time frame and scope here. So maybe just add the constant to `IUIConstants` now and we deprecate/replace with something better later... its just a string the compiler will inline anyways :-)

ruspl-afed:
This particular occurrence of cause is not critical and more a show case of suggested approach.

If you have strong arguments why introducing such a notation will make things even worse and more expensive to support than old good string constants inside some public type - I would love to learn it.

I suggest to discuss it with a wider audience on today's `Eclipse IDE - Developers community call` @laeubi @merks

laeubi:
I have a meeting there before but try to join. In any case it would be good to have a written concept on how such thing should look like. 

The main problems I see:

1. It should be easy to use and string constants are a concept that is easy and know to developers so we need to really proof the added value
2. While having tools is great, someone needs to write and maintain such tools. It already becomes hard to maintain existing things so adding more is likely not happen without someone dedicating to this tools.

ruspl-afed:
Very good points.
I doubt that I'm ready to invest time to a written spec in case that it is already doomed to serious skepticism. On the over hand, why others should waste time to discuss something not yet well shaped?

What if E4 Tools will (magically, i.e. via PRs) provide such an output? That will lead us to more reasonable starting position: when E4 Tool can do so, than we can continue much more practical exchange of opinions regarding details.

laeubi:
The problem with automation often is as with the externalize string that then there are sometimes cases where one do not want it... others might not be well understood (e.g. placeholders or wildcards) by the tool and so on.

So I dont think we need a real "spec" but at least a little thing that gives a rough overview about the operation. I think we have something similar for resources somewhere already (icons) that also has not getting much progress yet. Mostly because the current things work "good enough" what is the enemy of "better" :-)

From a tooling perspective some annotations would of course be nice, so lets assume you have a Commandhandler and it would be annotated with `@Exported`, a tool might then look into the `plugin.xml` find it there by the classname and update a given file (e.g. an ENUM implementing an Interface with getId()). Then also this can be validated of course... again we would likely want tools in the IDE and in the build (like API tools).

ruspl-afed:
> a little thing that gives a rough overview about the operation

The situation highlighted in this PR
1) we have an identified contribution in bundle A.
2) we want to use this identified contribution from the Java code of bundle B
3) we are lucky to introduce a string literal in bundle B that is equal to an attribute value from bundle A

We of cause can introduce a string constant to bundle A and reuse it from bundle B, however we cannot:
1) automate control of correspondence between string constant and attribute value
2) simplify a detection of the FQN for such a constant (if it ever exists) for another identified contribution

The suggested approach to improve the general flow:
1) identified contributions from `plugin.xml` and `.e4xmi` should have corresponding Java symbols
2) these Java symbols have predictable and calculable FQN (and, preferably, dedicated types)
3) these Java symbols are promoted to an API eligible for control with existing API Tools
4) PDE contributes ignorable builder checks for correspondence between Java symbols and E3/E4 manifests
5) PDE contributes Quick Fixes /generators to restore/create the correspondence between Java symbols and E3/E4 manifests

laeubi:
> we are lucky to introduce a string literal in bundle B that is equal to an attribute value from bundle A

Just as a side not, I'm not really "happy" about this because it actually couple Bundle A + Bundle B directly in code while the whole purpose of _Commands_ (and _Handlers_) is to decouple code. Decoupling of course comes at a price of maybe loosing the connection but that's why I think copy the ID is not always bad depending on how close your code is related.

ruspl-afed:
> Just as a side not, I'm not really "happy" about this because it actually couple Bundle A + Bundle B directly in code while the whole purpose of _Commands_ (and _Handlers_) is to decouple code.

Good point. However, I think that here we don't violate the purpose of _Commands_&_Handlers_ since we ask for a _Command_ (interface) and not for a _Handler_ (implementation)

ruspl-afed:
We had a discussion on the meeting: 
1) String constants may be inlined by compiler and this won't save us from the "contribution id changed" issues
2) Typed identity may solve this problem but this idea needs to be polished more before the introduction
3) More comprehensive demo of approach is welcome

Since my intention here is to introduce notation and area for tools, that will simplify our life during support of big code bases, and I don't want to just introduce another "ad hoc" constant to the API.

May be it is better to close this one.

PR #1998, opened 2025-06-26
All Eclipse Platform Icons Upgraded to SVG

Michael5601:
This PR changes the usage of all remaining PNGs and GIFs in Eclipse Platform to SVG. Currently this is a draft as I also want to include the Eclipse-Logo SVG to Eclipse Platform as soon as this [issue](https://github.com/eclipse-platform/eclipse.platform.images/issues/153) is done.

In detail the following changes will be performed:

1. Include missing SVGs for existing GIFs in bundle `org.eclipse.core.tools.resources`. All SVGs were either copied from other bundles or created from scratch by myself.
2. Enabling the image comparison of mutliple SVGs with each other.
3. Changing the reference of two JFace PNGs in bundle `org.eclipse.team.ui` to SVG.
4. Introduce a SVG for the Eclipse Logo in different sizes e.g. `eclipse16.svg`, `eclipse32.svg` etc. and changing the references to SVG. This allows better scaling for the window icon in the top left, the toolbar icon and the icons in the About dialog.

PR #2058, opened 2025-07-16
Add SVGs for Eclipse Logo

Michael5601:
This PR introduces a SVG for the Eclipse Logo in different sizes e.g. `eclipse16.svg`, `eclipse32.svg` etc. and changing the references to SVG. This allows better scaling for the window icon in the top left and the icons in the About dialog.


Michael5601:
Please see this [comment](https://github.com/eclipse-platform/eclipse.platform/pull/1998#issuecomment-3079285483) for clarification why the Eclipse logo needs further refinement.

The stripes in the middle of the Eclipse logo as proposed in this draft are not exactly in the middle of the icon (y=5.5px) for 16x16px size.

image

 To allign pixel perfect they either must be at y=6.0px which is too low or at y=5.0px which is too high. 



A solution to this would be to change the placement of the logo in the grid. Right now the logo is alligned in the middle of the view area. I do not know if there is any restriction for the placement of the logo in the style guidelines.

BeckerWdf:
> A solution to this would be to change the placement of the logo in the grid. Right now the logo is alligned in the middle of the view area. I do not know if there is any restriction for the placement of the logo in the style guidelines.

I think this is a good solution. We have to make compromises in this case and I think this is a good one.

HeikoKlare:
> > A solution to this would be to change the placement of the logo in the grid. Right now the logo is alligned in the middle of the view area. I do not know if there is any restriction for the placement of the logo in the style guidelines.
> 
> I think this is a good solution. We have to make compromises in this case and I think this is a good one.

Agreed. Do you already have a draft of that version? Would be nice to test around with it in Eclipse, as I am just concerned that the usages as a program icon (in the top left of the window on Windows, in the taskbar etc.) may look a bit strange if the icon is not centered, so that's something we should test.

Michael5601:
> > > A solution to this would be to change the placement of the logo in the grid. Right now the logo is alligned in the middle of the view area. I do not know if there is any restriction for the placement of the logo in the style guidelines.
> > 
> > 
> > I think this is a good solution. We have to make compromises in this case and I think this is a good one.
> 
> Agreed. Do you already have a draft of that version? Would be nice to test around with it in Eclipse, as I am just concerned that the usages as a program icon (in the top left of the window on Windows, in the taskbar etc.) may look a bit strange if the icon is not centered, so that's something we should test.

Not yet. I try to provide it today, so you can test tomorrow. :)

Michael5601:
I added the non-centered pixel perfect Eclipse Logos to this draft so one can try them out in the IDE. 
I made sure the displacement is consistent for all three sizes:

image
image
image

I also have an alternative suggestion:
The sizes 32x32px and 48x48px can be improved by moving the logo one pixel above, so it looks more centered. This is not possible for the 16x16px variant as the space is limited. Thus we can improve the looks of the sizes 32x32px and 48x48px by sacrificing the consistency with the 16x16px variant.

BeckerWdf:
> Thus we can improve the looks of the sizes 32x32px and 48x48px by sacrificing the consistency with the 16x16px variant.

That's a good idea. The 16x16 one is so small that this sacrifice is even not visible in real life...

HeikoKlare:
I have taken some screenshots at different zooms. It looks really good to me. The only concern (as I expected) is with respect to the task bar: if your are sensitive, you will notice that the icon is off there (compared to the others around). You can see it in the 100% taskbar screenshot: the middle icon is the new one, the ones around are the old ones. The old ones fit slightly better to the icons around. But maybe that's something that is fixed when using different alignment for the 32x32 and 48x48 versions, as the task bar icons are obviously larger. So I am in favor of following that approach:

> Thus we can improve the looks of the sizes 32x32px and 48x48px by sacrificing the consistency with the 16x16px variant.

### Title bar
100%
image

150%
image

175%
image

#### Task bar
100%
image

150%
image

175%
image

Michael5601:
As suggested I improved the placement of the 32x32px and 48x48px in the current state of this draft.

image
image
image

I changed this draft to a PR.

HeikoKlare (approved):
This looks really great to me now. Let's bring it into M2 to have more people test it. @BeckerWdf do you have any objections on merging this as is?

BeckerWdf:
> This looks really great to me now. Let's bring it into M2 to have more people test it. @BeckerWdf do you have any objections on merging this as is?

Go for it 👍

PR #2064, opened 2025-07-19
Remove explicit hover and disabled icons from terminal bundles

HeikoKlare:
The recently added terminal bundles contain explicit icons for the hover and disabled state of UI elements. The hover icons are equal to the ordinary enabled icons and the disabled icons can be generated by SWT on-the-fly.

This change removes the just added hover and disabled icons. The hover icons are removed from the image registry and the registry's disabled icons are generated on-the-fly instead.

This change also removes some public API, but since that API has just been moved (and is only about constants for icons) I don't see any risk in this.

This is a preparatory step for replacing the remaining PNG/GIF icons with SVGs.


laeubi:
As these seem to be only used internally, we might move that interface to the internal package as well. I don't really see any benefit for making them public.

ruspl-afed:
I'm also for minimizing API surface, we can always declare API symbols later if we really need them.
I also don't like `null` in signatures
But I think the change itself is consistent and we may create other PRs regarding API reduction

HeikoKlare:
Sounds good, thank you. I wanted to ask you about that change anyway. I will update the PR soon.

laeubi:
@HeikoKlare @ruspl-afed its great to see getting all the dust-off this rather old code so I'm quite happy that I'm not the only one interested to get this into platform and being modernized! Thanks for helping with that :+1:

ruspl-afed (approved):
Looks reasonable for me.
I plan to take care about `org.eclipse.terminal.control.actions` package after this one will be merged with a goal to remove it from the API at least partially.

ruspl-afed:
> I will update the PR soon

@HeikoKlare  Please let me know if you still plan to do so, otherwise let's merge

HeikoKlare:
I've now just moved the image constants interfaces to the internal packages containing the plugin initialization classes that also fill the registry with the icons. I think that's a reasonable place for them. So this should be ready to go now.

> I'm quite happy that I'm not the only one interested to get this into platform and being modernized!

I really like to see the terminal being integrated into platform. Thank you all for working on that! I also look forward to see how it will evolve and will try my best to find some time to contribute to it.

PR #2066, opened 2025-07-20
Replace terminal icons with SVGs

HeikoKlare:
This replaces the existing GIF and PNG icons used by the terminal bundles with SVGs. It covers all icons except for the terminal view icon and the derived icon for creating a new terminal view.

Since the bundles were just introduced/renamed, there might be no one referencing the GIF/PNG icons, which is why I propose to remove them instead of keeping them for two more years like we do for the other platform bundles.

### Terminal View Icons
__Note:__ There is a now a dedicated PR for this:
- https://github.com/eclipse-platform/eclipse.platform/pull/2067

Replacements for the `terminal_view.gif` and `new_terminal_view.gif` are not part of this PR as there seem to be no existing SVGs that can easily be reused but according SVGs have to be created first.

With respect to that, I would like to know what you think about taking the chance of exchanging the terminal view icon. I don't want to strongly argue for it, but instead of putting effort into converting the existing GIF into an SVG I would first like to ask if we want to keep that at all. Nowadays, there seem to be rather common icons for a terminal (based on the string `> _` in a window), so what about adopting something more common today? If we want to preserve the connecting aspect in the icon, the connector shown in the existing icon seems to be inspired by a physical connector that many people probably don't even know anymore.

Additionally, the visual distinction between opening a terminal (inside a terminal view, using icon for the console view) and opening a new terminal view (using the terminal view icon with the connector) does not seem to be that easy to understand for me. To me, it would even make more sense to swap those icons, as currently the terminal view icon contains the connector (even though it is just a container for actual terminal connections) and opening a terminal (which actually connects to something) does not contain the connector.

So would be great to have your opinion on this.

This is the icon we have:
terminal_view

We could for example reuse and adapt the existing icon for the consolve view:
terminal_in_console_view

Or even simplify it:
terminal_view
image

Just as some preliminary thoughts.


laeubi:
> Or even simplify it:
> terminal_view

I like this icon what is much better than the current one.

HeikoKlare:
> I like this icon what is much better than the current one.

Thank you for the feedback! Then I will make a proper SVG out of that and draft a PR with it.

HeikoKlare:
I have created a proposal for terminal view icon:
- https://github.com/eclipse-platform/eclipse.platform/pull/2067

This PR replaces all other terminal icons and should be ready to go.

BeckerWdf:
> Or even simplify it:
> terminal_view

I like this idea of simpliciation.

PR #2067, opened 2025-07-20
Replace view icon for terminal view

HeikoKlare:
This is a proposal to exchange the terminal view icons with a new version. The existing icons are only present as raster graphics and are based on elements that do nowadays not fit that well anymore. At the same time, other icons for terminals became common in other applications. The proposed icons (for the view itself and the toolbar item to create a new view) align with such common terminal icons.

These are the existing and the proposed icons for comparison:
Existing:
terminal_view
new_terminal_view

Proposal:
terminal_view
new_terminal_view

### Screenshots

Taken on Windows at different zooms.

100%
image

150%
image

175%
image


BeckerWdf:
My only question:
Do we really want to have these tiny stars:

image

In the "File > New" main menu icons we also don't have them.

HeikoKlare:
> My only question:
> Do we really want to have these tiny stars:

I think we should remove them. I just took the "add overlay" from the icon for creating a new project, which contains those stars. I didn't check other icons with an "add overlay", which all do not contain such stars. And I would remove the tail underneath the stars as well, so only keeping the "+" symbol.

BeckerWdf:
> I just took the "add overlay" from the icon for creating a new project, which contains those stars.

If you tell me which file this is I can have a look at it and provide a PR that removes them also there.

HeikoKlare:
> If you tell me which file this is I can have a look at it and provide a PR that removes them also there.

It only seems to be present in the rather generic "new" icon, but that one is spread across multiple bundles. Just a few I found via the image browser:

image
image
image
image

I first thought it was the general "new project" icon, but then I found that the "new project" icon is inconsistent in the "New" toolbar drop-down menu and in the "New" wizard that is opened by that tool item:
image
image

Not sure if that has always been like that or if it has been mixed up when replacing PNGs with SVGs.

BeckerWdf:
> I first thought it was the general "new project" icon, but then I found that the "new project" icon is inconsistent in the "New" toolbar drop-down menu and in the "New" wizard that is opened by that tool item

So the "New > Project", "New >Example" and the "New > Other" all use this icon with the stars and the wizard that opens also has these stars in the wizard banner. So I would say this is consistent.
Or what do you exactly mean with inconsistent?

We anyway can decide that the stars in both icons (toolbar and wizard banner) should be removed from both. But this is more a question of "style" and "taste" and not one of consistency.

> Not sure if that has always been like that or if it has been mixed up when replacing PNGs with SVGs.
I don't think this got mixed up.

HeikoKlare:
> So the "New > Project", "New >Example" and the "New > Other" all use this icon with the stars and the wizard that opens also has these stars in the wizard banner. So I would say this is consistent.
> Or what do you exactly mean with inconsistent?

I referred to the `Project` entry in the "New" drop-down menu and the `Project` entry inside the `New` wizard having different icons:
image
image
But I just noticed that the formed does not start the wizard to create a new plain project but starts the wizard for different kinds of projects (as indicated by the `...` in the drop-down menu entry). I have to admit that I, as a user, am not able to easily distinguish all these different usages of "Project" and have to try on my own, but that's nothing to blame the icons for and also nothing I consider anything severe and necessary to address.

> We anyway can decide that the stars in both icons (toolbar and wizard banner) should be removed from both. But this is more a question of "style" and "taste" and not one of consistency.

Yes, we could do that, but it's also just fine for me as is. This all boils down to that I just made a bad choice for the icon to take the "add" symbol from and that everything can be considered just fine as is. So from my side, there is no need to change anything and we could close the discussion.

BeckerWdf:
> Yes, we could do that, but it's also just fine for me as is. This all boils down to that I just made a bad choice for the icon to take the "add" symbol from and that everything can be considered just fine as is. So from my side, there is no need to change anything and we could close the discussion.

Same for me.

PR #2069, opened 2025-07-21
Add Eclipse logo SVGs to build.properties

HeikoKlare:
The recently added Eclispe logo SVGs have not been added to the build.properties to make the deployed binary bundle also contain them. This change adds them accordingly.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2327

### DIFF
--- a/platform/org.eclipse.platform/build.properties
+++ b/platform/org.eclipse.platform/build.properties
@@ -17,6 +17,9 @@ bin.includes = about.html,\
                about.properties,\
                eclipse_lg.png,\
                eclipse_lg@2x.png,\
+               eclipse16.svg,\
+               eclipse32.svg,\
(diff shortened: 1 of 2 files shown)


fedejeanne (approved):

akurtakov:
Why do we need 3 svgs ?

HeikoKlare:
> Why do we need 3 svgs ?

The same reason why we had 3 PNGs: the icon is required at different sizes (independent from HiDPI support with zoomed versions). Since there is no API to load an image (including an SVG) at a specific size yet (we are working on extending on that), images need to be provided in the size in which they are required, which for the Eclipse icon seem to be 16x16, 32x32 and 48x48 versions.

PR #2072, opened 2025-07-22
Improve Eclipse SVGs

BeckerWdf:
- Add black outline
- Fix length of white strips (that where standing out of the purple circle). This is only visible if the background of the icon is not white.

See: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2327


BeckerWdf:
New versions (first line) vs. Before versions (second line):

eclipse16
eclipse16

eclipse32
eclipse32

eclipse48
eclipse48

Just zoom in your browser and you will clearly see the difference.

laeubi:
@HeikoKlare can you explain why 3 images are needed here should we not be able to derive the size from the filename? e.g assume we would have `eclipse.svg` can we have something like using it as 

/my/path/eclipse.svg?size=16x16
/my/path/eclipse.svg?size=32x32
/my/path/eclipse.svg?size=48x48

it really feels odd to have a scalable file 3 times just because of **different size** it would of course be good to have it with **different details levels**.

Alternatives:

/my/path/eclipse.svg#16x16
/my/path/eclipse.svg#32x32
/my/path/eclipse.svg#48x48

or 

/my/path/eclipse.svg?16x16
/my/path/eclipse.svg?32x32
/my/path/eclipse.svg?48x48

merks:
With respect to the 3 images, on another issue/pr heiko explained that it's not current possible to scale an image independent of the zoom of the image.   When that's supported somehow, then we could use just one image.  I think the same applies for the image in the empty editor area, i.e., not possible to just scale the image relative to the area where the image is displayed.

HeikoKlare:
Let's stick to actual issue. Since questions about the three images have already been raised yesterday (https://github.com/eclipse-platform/eclipse.platform/pull/2069#issuecomment-3096721739), maybe we should have a separate issue for that.

iloveeclipse:
> New versions:
> 
> eclipse16 eclipse32 eclipse48

Could you please post how it was looking before (the current state)?

laeubi:
>  i.e., not possible to just scale the image relative to the area where the image is displayed

I don't think that this is true it just requires a bit more of math magic, if I know the current dimension of the image and the desired one I can compute a scale factor and asked for an image with that factor (if we assume it is fine to scale by 1pt steps) e.g. `org.eclipse.swt.graphics.Image.getImageData(int)`

BeckerWdf:
> * Fix length of white strips (that where standing out of the purple circle). This is only visible if the background of the icon is not white.

You can see that here:
image

iloveeclipse (approved):
Thanks.

HeikoKlare (approved):
Much better, thank you! Interesting that we did not notice the missing border before. And with the white background in Inkscape I also did not notice that the white bars outreaching the circle to the left and right have a negative impact on the result when being placed on a non-white background.

One might still improve how the SVG itself is composed, as the different circle fragments (for the purple circle) are not perfectly aligned, but with the black circle around it does not matter for the result anymore as that misalignment is hidden by that black circle. So I think we can merge this as proposed.

Here are some screenshots with the current icon always to the left and the proposed one to the right.

100%
image
image

150%
image
image

PR #2073, opened 2025-07-23
Import com.jcraft.jsch packages #958

ptziegler:
This removes the hard dependency to the 'com.jcraft.jsch' bundle and allows clients to switch to the 'com.github.mwiede.jsch' fork.

The original JSch bundle is no longer maintained and suffers from several problems, such as an incompatibility with modern OpenSSH versions.

Note:
- Both plugins share the same package name and can be used interchangeably.
- The current version of the mwiede fork is 2.27.2, which is why the upper bound needs to be extended.

### DIFF
--- a/team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
+++ b/team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
@@ -8,12 +8,12 @@ Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.ui;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.jsch.core;bundle-version="[1.2.0,2.0.0)",
- com.jcraft.jsch;bundle-version="[0.1.50,1.0.0)"
+ org.eclipse.jsch.core;bundle-version="[1.2.0,2.0.0)"
 Bundle-ActivationPolicy: lazy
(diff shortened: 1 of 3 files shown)


team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
```diff
@@ -8,12 +8,12 @@ Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.ui;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.jsch.core;bundle-version="[1.2.0,2.0.0)",
- com.jcraft.jsch;bundle-version="[0.1.50,1.0.0)"
+ org.eclipse.jsch.core;bundle-version="[1.2.0,2.0.0)"
```
laeubi, line 11: Should require bundle not be removed?

laeubi (commented):
Thanks I think we should merge this before M2 so there is enough time for testing!

team/bundles/org.eclipse.jsch.ui/META-INF/MANIFEST.MF
ptziegler: What else should be removed? `com.jcraft.jsch` is removed, but `org.eclipse.jsch.core` obviously needs to stay. The additional diff is only because of the comma at the end.

laeubi: > The additional diff is only because of the comma at the end.

You are right... it seems still to early in the morning :-\

PR #2074, opened 2025-07-23
Move terminal actions to internal package

ruspl-afed:
Currently actions are available as API types from "o.e.t.control" bundle, but used only from internals of "o.e.t.v.ui" bundle. 
So, actions can be moved to internal package of of "o.e.t.v.ui" bundle to reduce API surface.


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
```diff
@@ -248,6 +249,12 @@ protected void initializeImageRegistry(ImageRegistry registry) {
 		ImageDescriptor disabledTerminalViewIcon = ImageDescriptor.createWithFlags(enabledInputFieldIcon,
 				SWT.IMAGE_DISABLE);
 		registry.put(ImageConsts.ACTION_NewTerminalView_Disabled, disabledTerminalViewIcon);
+		URL enabledClearAllIconUrl = bundle
```
laeubi, line 252: You might want to use

`org.eclipse.jface.resource.ImageDescriptor.createFromFile(Class<?>, String)`

ruspl-afed: Thank you for your feedback @laeubi !
Here I use literally the same style of code that we already have above this line for other images, I plan to work more on it after API change.

ruspl-afed:
If there is no objections I would like to merge this today

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
ruspl-afed: Now I'm not sure that exactly here it will the best choice, since we will solve the same `Bundle` 5 times, I'll keep this advice for the case of single URL resolution.

PR #2075, opened 2025-07-23
[#3124] CME in `DebugElementHelper.getImageDescriptor`

ruspl-afed:
Use `ConcurrentSkipListMap` to avoid `ConcurrentModificationException`

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/launch/DebugElementHelper.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/launch/DebugElementHelper.java
@@ -13,8 +13,8 @@
  *******************************************************************************/
 package org.eclipse.debug.internal.ui.views.launch;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.debug.core.model.Breakpoint;
 import org.eclipse.debug.internal.ui.DelegatingModelPresentation;
@@ -44,7 +44,7 @@ public class DebugElementHelper {
 	private static DelegatingModelPresentation fgPresenetation;
 
 	// map of images to image descriptors
-	private static Map<Image, ImageDescriptor> fgImages = new HashMap<>();
+	private static Map<Image, ImageDescriptor> fgImages = new ConcurrentHashMap<>();
 
 	/**
 	 * Disposes this adapter


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/launch/DebugElementHelper.java
```diff
@@ -44,7 +44,7 @@ public class DebugElementHelper {
 	private static DelegatingModelPresentation fgPresenetation;
 
 	// map of images to image descriptors
-	private static Map<Image, ImageDescriptor> fgImages = new HashMap<>();
+	private static Map<Image, ImageDescriptor> fgImages = new ConcurrentSkipListMap<>();
```
iloveeclipse: Why not `ConcurrentHashMap`? Any specific order needed here? It wasn't sorted before.

ruspl-afed: Yes, sorting will not work for `Image`, my fault, reworked to old fashion before looking to this comment.
But we may try `ConcurrentHashMap` as well.

ruspl-afed: Now it is green. 
Sorry for wasting your time on red PR. Please let me know if you prefer `ConcurrentHashMap`

iloveeclipse: I believe for the functionality here it is irrelevant, but I would prefer `ConcurrentHashMap` simply because the use of `ConcurrentSkipListMap` would later always lead to question "why this specific one is used?" and `ConcurrentHashMap`  is generic enough to avoid such questions.

ruspl-afed: `ConcurrentSkipListMap` fails with ClassCastException here, because `Image` is not `Comparable`. And this is good, because `Comparator` for `Image` class may consume a lot of resources for almost nothing.

I'll try to use `ConcurrentHashMap` as you suggested.

ruspl-afed: now it is green with `ConcurrentHashMap`

iloveeclipse: Thanks.

PR #2078, opened 2025-07-23
Rework image initialization for `org.eclipse.terminal.view.ui`

ruspl-afed:
* reduce code duplication
* fix copy-n-paste error for disabled Terminal View Icon

see https://github.com/eclipse-platform/eclipse.platform/pull/2074#discussion_r2224922478


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
```diff
@@ -226,35 +226,21 @@ protected void initializeImageRegistry(ImageRegistry registry) {
...
-		registry.put(ImageConsts.ACTION_ToggleCommandField_Enabled, enabledInputFieldIcon);
-		ImageDescriptor disabledInputFieldIcon = ImageDescriptor.createWithFlags(enabledInputFieldIcon,
-				SWT.IMAGE_DISABLE);
-		registry.put(ImageConsts.ACTION_ToggleCommandField_Disabled, disabledInputFieldIcon);
-
-		URL enabledTerminalViewIconUrl = bundle
-				.getEntry(ImageConsts.IMAGE_DIR_ROOT + ImageConsts.IMAGE_DIR_ELCL + "new_terminal_view.svg"); //$NON-NLS-1$
-		ImageDescriptor enabledTerminalViewIcon = ImageDescriptor.createFromURL(enabledTerminalViewIconUrl);
-		registry.put(ImageConsts.ACTION_NewTerminalView_Enabled, enabledTerminalViewIcon);
-		ImageDescriptor disabledTerminalViewIcon = ImageDescriptor.createWithFlags(enabledInputFieldIcon,
```
ruspl-afed, line 249: > ImageDescriptor disabled**TerminalViewIcon** = ImageDescriptor.createWithFlags(enabled**InputFieldIcon**

It seems that we overlooked a copy-n-paste issue here @HeikoKlare

HeikoKlare: Good catch! We probably also didn't notice when testing because the according tool item is never or seldom disabled.
The refactoring significantly reduces the risk for this kind of mistake. We have applied a similar pattern of having a method combining the initialization for one icon (including its enabled and disabled version) in several other bundles, so it definitely makes sense to do something similar here as well.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/UIPlugin.java
```diff
@@ -226,35 +226,21 @@ protected void initializeImageRegistry(ImageRegistry registry) {
...
-				SWT.IMAGE_DISABLE);
-		registry.put(ImageConsts.ACTION_NewTerminalView_Disabled, disabledTerminalViewIcon);
-		URL enabledClearAllIconUrl = bundle
-				.getEntry(ImageConsts.IMAGE_DIR_ROOT + ImageConsts.IMAGE_DIR_ELCL + "clear_co.svg"); //$NON-NLS-1$
-		ImageDescriptor enabledClearAllIcon = ImageDescriptor.createFromURL(enabledClearAllIconUrl);
-		registry.put(ImageConsts.ACTION_ClearAll_enabled, enabledClearAllIcon);
-		ImageDescriptor disabledClearAllIcon = ImageDescriptor.createWithFlags(enabledClearAllIcon, SWT.IMAGE_DISABLE);
-		registry.put(ImageConsts.ACTION_ClearAll_disabled, disabledClearAllIcon);
+	private void putActionImages(ImageRegistry registry, Bundle bundle, String file, String ekey, String dkey) {
```
HeikoKlare, line 239: Nit: you might call this `enabledKey` and `disabledKey` to improve comprehensibility.
```suggestion
	private void putActionImages(ImageRegistry registry, Bundle bundle, String file, String enabledKey, String disabledKey) {
```

ruspl-afed: Only if you will insist. I follow a different religion: avoid compound names since they cries that further decomposition is needed. 
However you are right, `ekey` and `dkey` looks like an ugly compromise. How about `enabled` and `disabled`?

HeikoKlare: That's also fine for me. As said, I am also fine with the PR as is. I think in any case we should try to merge it today, so that it likely goes into M2.

ruspl-afed: ok, let's merge

PR #2083, opened 2025-07-24
Move `TraceHandler` to internals of `org.eclipse.terminal.view.ui`

ruspl-afed:
This type is not in fact used in other bundles therefore it should not be a part of API

FYI @laeubi this one looks even better to merge first


laeubi (commented):
Seems like a good catch!

PR #2082, opened 2025-07-24
Replace `ScopedEclipsePreferences` with standard `IPreferenceStore`

ruspl-afed:
Ideologically it looks very similar and we definitely don't want to have custom API type to work with preferences

This will allow to fully remove `org.eclipse.terminal.view.core.utils` package (that is currently an API package)

FYI @laeubi

PR #2085, opened 2025-07-24
Remove dependencies to removed terminal utils package

HeikoKlare:
The package `org.eclipse.terminal.view.core.utils` has been removed but some bundles still import it, leading to errors. This change removes those unnecessary dependencies.

image

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/META-INF/MANIFEST.MF
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/META-INF/MANIFEST.MF
@@ -9,7 +9,6 @@ Import-Package: org.eclipse.cdt.utils.pty;mandatory:=native,
  org.eclipse.terminal.connector;version="[1.0.0,2.0.0)",
  org.eclipse.terminal.connector.process;version="[1.0.0,2.0.0)",
  org.eclipse.terminal.view.core;version="[1.0.0,2.0.0)",
- org.eclipse.terminal.view.core.utils;version="[1.0.0,2.0.0)",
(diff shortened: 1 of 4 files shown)


HannesWell:
Thanks for this, this saved tonight's I-build (and already broke the aggregator build).
I wonder why, this didn't show up in:
- https://github.com/eclipse-platform/eclipse.platform/pull/2082#issuecomment-3113993451

laeubi:
> I wonder why, this didn't show up in

I also recently broke it because of a forgotten dependency removal. The problem is that we are building against I-Build repo as a repository, and of course there is a version that exports the package and Tycho happily downloads it from there if nothing else is providing it and there are no constrains forbidding it.

In OSGi that would mean some use-constraint for example, but P2 does not support that and Tycho disable it for performance reasons during compilation (apart from that large parts of Eclipse are not have proper use-clauses anyways)..

HeikoKlare:
> Thanks for this, this saved tonight's I-build (and already broke the aggregator build).

Yes, I expected it would probably break, which is why just merged. We need some good final build for M2.

> The problem is that we are building against I-Build repo as a repository, and of course there is a version that exports the package and Tycho happily downloads it from there if nothing else is providing it and there are no constrains forbidding it.

In my understanding, such a broken dependency will show up in the next I-Build at latest. Given how seldom this happens, I am not sure whether there is a need for some action item here. In this case, the issue even showed up in the IDE already (at least for me). So I think there are sufficient mechanisms in place to deal with the seldom case of some API/package being removed.

PR #2093, opened 2025-07-30
[Debug] Update breakpoint icon

SougandhS:
This PR changes the dark breakpoint icon to lighter one to provide more visibility

Before 
image

After
image

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/731

### DIFF
--- a/debug/org.eclipse.debug.ui/icons/full/obj16/brkp_obj.svg
+++ b/debug/org.eclipse.debug.ui/icons/full/obj16/brkp_obj.svg
@@ -1,100 +1,11 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-
-<svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:xlink="http://www.w3.org/1999/xlink"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="16"
-   height="16"
-   id="svg2"
-   version="1.1"
-   inkscape:version="0.48.0 r9654"
-   sodipodi:docname="brkp_obj.svg">
-  <defs
-     id="defs4">
-    <linearGradient
-       id="linearGradient4894">
-      <stop
-         id="stop4896"
-         offset="0"
-         style="stop-color:#b4d1e5;stop-opacity:1" />
-      <stop
-         style="stop-color:#5395be;stop-opacity:1"
-         offset="0.5"
-         id="stop4898" />
-      <stop
-         id="stop4900"
-         offset="1"
-         style="stop-color:#91bcd7;stop-opacity:1" />
-    </linearGradient>
-    <linearGradient
-       y2="6.8716693"
-       x2="8.7378912"
-       y1="2.1550355"
-       x1="8.7378912"
-       gradientUnits="userSpaceOnUse"
-       id="linearGradient4871"
-       xlink:href="#linearGradient4894"
-       inkscape:collect="always" />
-  </defs>
-  <sodipodi:namedview
-     id="base"
-     pagecolor="#ffffff"
-     bordercolor="#666666"
-     borderopacity="1.0"
-     inkscape:pageopacity="0.0"
-     inkscape:pageshadow="2"
-     inkscape:zoom="45.254831"
-     inkscape:cx="6.3291708"
-     inkscape:cy="5.9636676"
-     inkscape:document-units="px"
-     inkscape:current-layer="layer1"
-     showgrid="true"
-     inkscape:window-width="1241"
-     inkscape:window-height="814"
-     inkscape:window-x="88"
-     inkscape:window-y="106"
-     inkscape:window-maximized="0"
-     showguides="true"
-     inkscape:guide-bbox="true"
-     inkscape:snap-global="true">
-    <inkscape:grid
-       type="xygrid"
-       id="grid3999" />
-  </sodipodi:namedview>
-  <metadata
-     id="metadata7">
-    <rdf:RDF>
-      <cc:Work
-         rdf:about="">
-        <dc:format>image/svg+xml</dc:format>
-        <dc:type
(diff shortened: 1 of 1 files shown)


SougandhS:
Hi @iloveeclipse 
Could you please check this too..

vogella:
Look good to me, could you also upload before / after for the light theme?

SougandhS:
> Look good to me, could you also upload before / after for the light theme?

Actually on mac, child eclipse in light theme is not rendered properly (On every Eclipse versions)😕

image

vogella:
> > Look good to me, could you also upload before / after for the light theme?
> 
> Actually on mac child eclipse in light theme is not rendered properly (On every Eclipse versions)😕
> 
> image

I think you need to restart child window after the theme switch. Did you try that?

SougandhS:
> I think you need to restart child window after the theme switch. Did you try that?

Yes, I did tried that but no use.

vogella:
> > I think you need to restart child window after the theme switch. Did you try that?
> 
> Yes, I did tried that but no use.

@BeckerWdf is this the same to you? Runtime light Eclipse theme  is broken on Mac?

vogella:
Old icon in light theme

image


New  icon in light theme

image


Looks fresher for me. WDYT? @Michael5601 @HeikoKlare @BeckerWdf ?

HeikoKlare:
I have to admit that I am not in favor of the proposed change yet. Here are some opinions/thoughts on this:
- I (personally) don't like the new color, but that might be because I am used to the existing color and maybe it's also because of how that new color looks like on my monitors (note that a color looks different depending on the color representation/profile of your monitor).
- I am not sure whether the proposal is the right solution for the actual issue; the reported issue https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/731 is about the breakpoint icon not being well visible on top of the blue selection gutter. That selection gutter is code-rendered, so it would also be possible to adapt that gutter to the theme color. Maybe that could also improve visibility of that gutter (and of the breakpoint icons everyone is used to at the same time).
- When changing the color of such a central icon, we should thoroughly think about the proper color to choose and not just select an arbitrary other color. How was the proposed color selected? Is it just a rather arbitrary lighter blue color? Does it fit to anything else in the IDE / in existing icons? Taking a look at other IDEs (IntelliJ, VS Code), using red dots for breakpoints seems to be common there. Maybe this would be a chance to adapt. I don't think the breakpoint icon color is a USP or distinguishing element from other IDEs, so that might be a chance to align to common colors.
- There is a bunch of other icons related to breakpoints, which include the color of the breakpoint icon (several of them in the same bundle, maybe also others in other bundles). Shouldn't they be adapted at the same time so that they fit together? E.g., 
brkp_grp 
brkp_type

merks:
I agree with @HeikoKlare.  My first reaction is kind of like "that's weird, the color that is familiar for the last 25 years has changed to some strange new color".    It just doesn't seem to "fit" with other colors and not clear how this will "fit in" overall.

image

SougandhS:
>Taking a look at other IDEs (IntelliJ, VS Code), using red dots for breakpoints seems to be common there. Maybe this would be a chance to adapt.

Bp with red icon would look like this (can be modified with different red saturations)

image

image

image



<br><br>

>There is a bunch of other icons related to breakpoints, which include the color of the breakpoint icon (several of them in the same bundle, maybe also others in other bundles). Shouldn't they be adapted at the same time so that they fit together? 

Absolutely, if the change is approved, I’m ready to put in the effort for that.

BeckerWdf:
I agree with Ed and Heiko. We should not change that icon. But fix the issue via other ways.
There are also lot's of downstream solutions that have own icons based on the Eclipse break point icon. These would also need to adapt - and that probably will not happen.

laeubi (changes requested):
This completely mess up the SVG, it replaces the vector data with an embedded (png?!) image.

That is a (technical) blocker apart from the other concerns already raised.

Beside that I wanted to mention, that now we have SVGs it could be independent of this change be evaluated if we can use CSS to style icons in a theme dependent way.

In this example one can see that the SVG in the style of the path references a `linearGradient4871` where one might:

1. Chose a more meaningful name e.g. `linearGradient_defaultTheme`
2. Define another one named e.g. `linearGradient_darkTheme`
3. Allow the SVG Render to collect e.g. a `SVGTransformerService`
4. Such service can then implemented by platform and look for specific things in an SVG and for example modify the `style` attribute in a way if it finds a reference that uses `_defaultTheme` as a suffix and there is defined one with the name `_<currentThemeId>Theme` it replace it (here linearGradient_defaultTheme` with `linearGradient_darkTheme`)

Then one would be able to have a somehow dynamic changing icon depending on the theme.

Similar might be for stroke colors or other things as well (the engine maybe even can read such styles from the themes css, e.g. if we add an ID to the icons or use the icon name)

SougandhS:
>Then one would be able to have a somehow dynamic changing icon depending on the theme.

This would be great!

SougandhS:
Thank you everyone for your feedback. As the majority disagrees with this icon change, I'm closing this PR.

Michael5601:
> This completely mess up the SVG, it replaces the vector data with an embedded (png?!) image.
> 
> That is a (technical) blocker apart from the other concerns already raised.
> 
> Beside that I wanted to mention, that now we have SVGs it could be independent of this change be evaluated if we can use CSS to style icons in a theme dependent way.
> 
> In this example one can see that the SVG in the style of the path references a `linearGradient4871` where one might:
> 
> 1. Chose a more meaningful name e.g. `linearGradient_defaultTheme`
> 2. Define another one named e.g. `linearGradient_darkTheme`
> 3. Allow the SVG Render to collect e.g. a `SVGTransformerService`
> 4. Such service can then implemented by platform and look for specific things in an SVG and for example modify the `style` attribute in a way if it finds a reference that uses `_defaultTheme` as a suffix and there is defined one with the name `_<currentThemeId>Theme` it replace it (here linearGradient_defaultTheme`with`linearGradient_darkTheme`)
> 
> Then one would be able to have a somehow dynamic changing icon depending on the theme.
> 
> Similar might be for stroke colors or other things as well (the engine maybe even can read such styles from the themes css, e.g. if we add an ID to the icons or use the icon name)

This is a great feature and should be implemented in the future but I doubt it will work with the current icons as they consist of many gradients. Thus monochromatic icons should be better for this.

laeubi:
> This is a great feature and should be implemented in the future but I doubt it will work with the current icons as they consist of many gradients

The mentions thing would work with as many gradients as you want :-D

Michael5601:
> > This is a great feature and should be implemented in the future but I doubt it will work with the current icons as they consist of many gradients
> 
> The mentions thing would work with as many gradients as you want :-D

I don't doubt it will work technically but will they be styled consistently? Will overlay icons still be visible? Do some icons need a special treatment? ...

For monochromatic icons these difficulties do not exist as we only change the base color which will probably always be a gray tone even before recoloring.

PR #2096, opened 2025-07-30
Exclude docs folder and README.md from Github build trigger

### LINKED ISSUE #2095: Exclude docs folder and README.md from Github build trigger
Based on a discussion with @merks I researched the exclusion pattern for GH actions and Jenkins files.

For GH actions I think we could use exclude patterns as described here: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-excluding-paths

For Jenkins pipeline it looks like it is also supported https://www.jenkins.io/doc/book/pipeline/syntax/#when

I suggest we add an exclusion filter for "docs" in this repo for our GH ci.ymal to avoid unnecessary builds (and to have an example for others to copy from).

vogella:
Every change to the existing repository triggers a build that is pretty much pointless and expensive for a doc change that literally has no impact on the actual content being built.
With this change documentation changes are excluded.

Using documentation from
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-excluding-paths

Fixes #2095

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,14 @@ concurrency:
 on:
   push:
     branches: [ master ]
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
   pull_request:
     branches: [ master ]
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
 
 jobs:
   build:


fedejeanne (approved):
LGTM, just a small question: couldn't we exclude all md files? 

... or do we keep them in case we decide to check for broken links in them on every PR (like Ed mentioned in https://github.com/eclipse-oomph/oomph/issues/164#issuecomment-3135802306)

merks:
Maybe we can ask @HannesWell to take a look in general how we might best exploit exclusions for both GHA and Jenkins.  This is applicable for a number of repositories and avoiding wasted build cycles is good for everything and everyone:

image

I'm not show how one would specify "all *.md" for root files, just *.md I guess?

Oh, and to answer @fedejeanne's question, I do think it's literally everything in the docs folder, including images and maybe even zip samples.

fedejeanne:
> I'm not show how one would specify "all *.md" for root files, just *.md I guess?

According to the documentation for [including paths](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths), I'd say it's `**.md` (two stars `*`)

merks:
Yes, but the question was how to specify just the ones in the root.  😁  

I think probably generally all *.md could be included as you suggest.  But I think it's important also to exclude all content under the docs folder too...

fedejeanne:
> Yes, but the question was how to specify just the ones in the root. 😁

Ah, you're right, I missed that part 😅 . According to [this documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths), `*.md` should match the files in the root folder (as you proposed initially).

> I think probably generally all *.md could be included as you suggest. But I think it's important also to exclude all content under the docs folder too...

👍 
... and whatever we don't do now, we can do in the next PR. So just FTR, no need to hold this PR on my account ✔️ 🙂

Sorry for the noise.

vogella:
Root md files are now excluded. I plan to merge this after the verification builds are done, finer adjustments can be done in follow-up commits.

fedejeanne:
> **I** plan to merge this after...

Zing! too slow :-P

HannesWell:
> Maybe we can ask @HannesWell to take a look in general how we might best exploit exclusions for both GHA and Jenkins.  This is applicable for a number of repositories and avoiding wasted build cycles is good for everything and everyone:

I can have a look into implementing a check in the called workflow to early exit if the set of changed files only contains markdown files.
For GH actions this could be available from the context, for Jenkins I have to check where to get this information from. 
But on the other hand, projects might want to customize the set of exclusions and then it can be done individually just like it was done here. 
So I'm not sure if a unified solution pays off.
Especially as I wonder how common doc-only changes are, compared to changes that require a build?

PR #2107, opened 2025-08-06
[Debug] minor spelling fix in Import/Export Attribute

SougandhS:


laeubi:
As it might be that the wrong attribute name is used in existing data, should at least reading the wrong name be used to recover such values?

merks:
Yes, of course this is used in a memento to serialize the value so all existing serializations will lose that value and that can't be a good thing.

SougandhS:
I can keep attribute value as misspelled and refactor variables name only

SougandhS:
Hi @merks 
If there are no issues can this be merged ?

laeubi:
> I can keep attribute value as misspelled and refactor variables name only

I think this was misunderstanding here, I think we should fix it, but when we _read_ the value, then first try the old (misspelled) so one gradually can upgrade to the right one.

SougandhS:
> I think this was misunderstanding here, I think we should fix it, but when we _read_ the value, then first try the old (misspelled) so one gradually can upgrade to the right one.

My apologies, I have updated it now

SougandhS:
Hi @laeubi 
build looks good now :)

SougandhS:
Thank you @laeubi & @merks

PR #2114, opened 2025-08-08
[Debug] Prompt when launching debug session with 'Skip All Breakpoints' enabled

SougandhS:
This pull request introduces a warning dialog that appears when a user attempts to start a debug session while the "Skip All Breakpoints" option is enabled. The goal is to prevent confusion caused by breakpoints being silently ignored, which often leads users to believe the debugger is malfunctioning. 
eg https://www.reddit.com/r/eclipse/comments/13zt4ll/debugger_not_working/
eg https://bugs.eclipse.org/bugs/show_bug.cgi?id=578000

The popup will provide option to disable the warning forever (which can be enabled later on from Run/Debug settings) and option disable skip
image

image
<br>
This will be helpful for new users who may not understand the debug settings and could get confused when breakpoints don’t work as expected.

<br>

https://github.com/user-attachments/assets/c13ff93f-d543-44c0-b4c6-ce97a0abba94

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -92,6 +92,7 @@ import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.custom.BusyIndicator;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorPart;
@@ -974,6 +975,14 @@ public class DebugUITools {
 		} catch (CoreException e) {
 			DebugUIPlugin.log(e);
 		}
+		boolean breakpointsDisabled = DebugPlugin.getDefault().getBreakpointManager().isEnabled();
+		boolean showWarningPrompt = DebugUIPlugin.getDefault().getPreferenceStore()
+				.getBoolean(IInternalDebugUIConstants.PREF_SKIP_ALL_BREAKPOINTS_PROMPT);
+		if (!breakpointsDisabled && showWarningPrompt && mode.contains("debug")) { //$NON-NLS-1$
+			SkipBreakpointsWarning showWarning = new SkipBreakpointsWarning(DebugUIPlugin.getShell());
+			Display display = DebugUIPlugin.getStandardDisplay();
+			display.syncExec(() -> showWarning.open());
+		}
 		if (launchInBackground) {
 			DebugUIPlugin.launchInBackground(configuration, mode);
 		} else {
@@ -1372,5 +1381,5 @@ public class DebugUITools {
 		}
 		return null;
 	}
-
 }
+
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/SkipBreakpointsWarning.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/SkipBreakpointsWarning.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2025 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.debug.ui;
+
+import org.eclipse.debug.core.DebugPlugin;
+import org.eclipse.debug.internal.ui.DebugUIMessages;
+import org.eclipse.debug.internal.ui.DebugUIPlugin;
+import org.eclipse.debug.internal.ui.IInternalDebugUIConstants;
+import org.eclipse.jface.dialogs.Dialog;
+import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Shell;
+
+/**
+ * Provides warning prompt dialog
+ *
+ * @since 3.19
+ */
+public class SkipBreakpointsWarning extends Dialog {
+
+	protected SkipBreakpointsWarning(Shell parentShell) {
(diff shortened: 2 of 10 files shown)


SougandhS:
Hi @iloveeclipse 
Could you please check this when you are available ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/SkipBreakpointsWarning.java
```diff
@@ -0,0 +1,116 @@
...
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Shell;
+
+/**
+ * Provides warning prompt dialog
+ *
+ * @since 3.19
+ */
+public class SkipBreakpointsWarning extends Dialog {
```
iloveeclipse: I believe you have just implemented `MessageDialogWithToggle` again?

iloveeclipse:
> The popup will provide option to disable the warning forever (which can be enabled later on from Run/Debug settings) and option disable skip
> image

This looks like a yet another implementation of `MessageDialogWithToggle`. Please use standard implementation if possible.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/SkipBreakpointsWarning.java
SougandhS: But `MessageDialogWithToggle` supports only 1 toggle option, so created a custom one

iloveeclipse: Why do we need two options? Just warn and user can enable or disable "skip breakpoints" after that if needed.

SougandhS: I thought this approach would help the user quickly disable the option if needed. But I can go with your suggestion

merks: Also you could provide a push button for the disable disabled action so that there is no expectation that the check state might be persisted like the don’t ask again check box.

SougandhS: >Also you could provide a push button for the disable disabled action so that there is no expectation that the check state might be persisted like the don’t ask again check box

You mean remove the checkbox and add a "disable" button right ?

merks: Yes. Just keep in mind that the double negative on disable the disable all is also a potential point of confusion. Enable Breakpoints might be better.  The opinions of others would be welcome.

SougandhS: @merks 
How about this ?
image

merks: Yes that seems fine (to me).

SougandhS: > https://github.com/eclipse-platform/eclipse.platform/pull/2114#discussion_r2265759455

@iloveeclipse is this design fine ?

SougandhS:
Hi @iloveeclipse 
I have updated the PR to show warn only as you suggested
image

SougandhS:
Hi @iloveeclipse, If there are no more changes can this be merged ?

iloveeclipse:
@SougandhS : I'm sorry for letting you wait, I'm mostly focused of regressions and have no free time for features. It looks better now, I could try to review next week.

SougandhS:
>I could try to review next week.

Thanks a lot!

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
```diff
@@ -974,6 +978,31 @@ public static void launch(final ILaunchConfiguration configuration, final String
...
+			final boolean[] msgValue = new boolean[] { true };
+			Display display = DebugUIPlugin.getStandardDisplay();
+
+			if (!display.isDisposed()) {
+				display.syncExec(() -> {
+					MessageDialogWithToggle dialog = MessageDialogWithToggle.openOkCancelConfirm(
+							display.getActiveShell(), DebugUIMessages.skipBreakpointWarningTitle,
+							DebugUIMessages.skipBreakpointWarningLabel, DebugUIMessages.skipBreakpointWarningToggle1,
+							false, null, null);
+					if (dialog.getToggleState()) {
```
iloveeclipse: This if block would be non needed if the call above would properly use requested values instead of passing `null` to two last arguments.
```
					MessageDialogWithToggle dialog = MessageDialogWithToggle.openOkCancelConfirm(
							display.getActiveShell(), DebugUIMessages.skipBreakpointWarningTitle,
							DebugUIMessages.skipBreakpointWarningLabel, DebugUIMessages.skipBreakpointWarningToggle1,
							false, DebugUIPlugin.getDefault().getPreferenceStore(),
							IInternalDebugUIConstants.PREF_SKIP_ALL_BREAKPOINTS_PROMPT);
					msgValue[0] = dialog.getReturnCode() == IDialogConstants.OK_ID;
```

Also if one now selects the checkbox but cancels the dialog, preference will be set (wrong). Using API as shown above would avoid this bug.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
```diff
@@ -974,6 +978,31 @@ public static void launch(final ILaunchConfiguration configuration, final String
 		} catch (CoreException e) {
 			DebugUIPlugin.log(e);
 		}
+		boolean breakpointsDisabled = DebugPlugin.getDefault().getBreakpointManager().isEnabled();
+		boolean showWarningPrompt = DebugUIPlugin.getDefault().getPreferenceStore()
```
iloveeclipse: `DebugUIPlugin.getDefault().getPreferenceStore()` is used three times in same method, even if there is a `getPreferenceStore()` available in same class and doing same.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
```diff
@@ -974,6 +978,31 @@ public static void launch(final ILaunchConfiguration configuration, final String
...
+				.getBoolean(IInternalDebugUIConstants.PREF_SKIP_ALL_BREAKPOINTS_PROMPT);
+		if (!breakpointsDisabled && showWarningPrompt && mode.contains("debug")) { //$NON-NLS-1$
```
iloveeclipse: `mode.contains("debug")` is not OK : hard coding and matching more as expected.
`ILaunchManager.DEBUG_MODE.equals(mode)` should be used

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/DebugUITools.java
```diff
@@ -974,6 +978,31 @@ public static void launch(final ILaunchConfiguration configuration, final String
...
+			final boolean[] msgValue = new boolean[] { true };
+			Display display = DebugUIPlugin.getStandardDisplay();
+
+			if (!display.isDisposed()) {
+				display.syncExec(() -> {
```
iloveeclipse: `syncExec()` is not needed, the precondition of this method is to be executed on UI thread.
The shell should be taken from `DebugUIPlugin.getShellForModalDialog()`

iloveeclipse (changes requested):
I will push requested changes shortly.

iloveeclipse (approved):

iloveeclipse:
@SougandhS : please provide N&N for this change.

SougandhS:
Thank you @iloveeclipse for the changes.
>please provide N&N for this change.

Sure 👍

PR #2122, opened 2025-08-18
Mention the Eclipse-IDE GitHub page in README and Oomph configuration

HannesWell:
Migrate off the `Eclipse_Platform_SDK_Provisioning` guide hosted at the deprecated wiki.

This is the first repo of the all Eclipse-SDK repos, where this change is applied, so please add your remarks here. Subsequently I'll apply it to all other repos in platform, equinox, JDT and PDE.

Once https://github.com/eclipse-ide is mentioned in all READMEs, we can remove general information from https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md that's already provided in https://github.com/eclipse-ide (and add there general information that's missing).

CC @merks

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ Update provides Java interfaces like `IPlatformConfiguration`. `IPlatformConfigu
 
 Contributions are most welcome. There are many ways to contribute, from entering high quality bug reports, to contributing code or documentation changes.
 
-For a complete guide, see the [CONTRIBUTING](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md) page.
+For a general contributing information and a complete guide see the [Eclipse-IDE](https://github.com/eclipse-ide) developer page.<br>
+For detailed information about the contributing process of this project, see [Eclipse-Platform CONTRIBUTING](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md).
 
 [![Create Eclipse Development Environment for Eclipse Platform](https://download.eclipse.org/oomph/www/setups/svg/Eclipse_Platform.svg)](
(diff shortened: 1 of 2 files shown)


HeikoKlare:
Thank you for taking care of this, Hannes! I really like to see this information cleaned up and centralized. We just had discussions few days ago about how to properly set up an IDE just for Platform development (as the current setup description will result in a quite large setup with the complete SDK of Platform, PDE, Equinox, JDT). That's much better with the information at eclipse-ide now!
I really like how the eclipse-ide contributing information explains how to set up an environment, how it links the different setups, and describes the contribution process. So streamlining and centralizing information there is much appreciated.

> Once https://github.com/eclipse-ide is mentioned in all READMEs, we can remove general information from https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md that's already provided in https://github.com/eclipse-ide (and add there general information that's missing).

I was wondering whether that should be a first step: extract the general information from Platform's contributing page (if necessary at all) to eclipse-ide (or simply remove it if obsolete) and add a reference from Platform's contributing page to the eclipse-ide contributing page. Otherwise, with the proposed change, we add another link in Platform's readme first and the page contains even another set of information on how to contribute, which may lead to more confusion during the transition time.

HannesWell:
> > Once https://github.com/eclipse-ide is mentioned in all READMEs, we can remove general information from https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md that's already provided in https://github.com/eclipse-ide (and add there general information that's missing).
> 
> I was wondering whether that should be a first step: extract the general information from Platform's contributing page (if necessary at all) to eclipse-ide (or simply remove it if obsolete) and add a reference from Platform's contributing page to the eclipse-ide contributing page. Otherwise, with the proposed change, we add another link in Platform's readme first and the page contains even another set of information on how to contribute, which may lead to more confusion during the transition time.

Yes that's absolutely the cleaner way. When writing this PR's message I thought the same.
Will do that before continuing with this one.

Issue #2121, opened 2025-08-18
NPE in PreferencePage.performOk() caused by null values.


merks:
@ruspl-afed 

The changes for https://github.com/eclipse-platform/eclipse.platform/pull/2082 continue to cause problems for M3. 

I noticed the following when I visited all the preferences pages installing all the staging train contents:
```
java.lang.NullPointerException
	at org.eclipse.core.internal.preferences.EclipsePreferences.put(EclipsePreferences.java:840)
	at org.eclipse.ui.preferences.ScopedPreferenceStore.setValue(ScopedPreferenceStore.java:568)
	at org.eclipse.terminal.view.ui.internal.preferences.PreferencePage.performOk(PreferencePage.java:661)
	at org.eclipse.jface.preference.PreferenceDialog$7.run(PreferenceDialog.java:903)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:167)
	at org.eclipse.jface.preference.PreferenceDialog.okPressed(PreferenceDialog.java:887)
	at org.eclipse.ui.internal.dialogs.FilteredPreferenceDialog.okPressed(FilteredPreferenceDialog.java:449)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.oomph.util.ReflectUtil.invokeMethod(ReflectUtil.java:126)
	at org.eclipse.oomph.util.ReflectUtil.invokeMethod(ReflectUtil.java:156)
	at org.eclipse.oomph.setup.ui.recorder.PreferenceInitializationDialog$Initializer$2$1.run(PreferenceInitializationDialog.java:505)
	at org.eclipse.oomph.ui.UIUtil$5.run(UIUtil.java:604)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4135)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3751)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.block(ModalContext.java:167)
[...]

PR #2123, opened 2025-08-19
[#2121] NPE in PreferencePage.performOk() caused by null values.

ruspl-afed:
* use empty `String` instead of `null`


ruspl-afed:
@merks I tried to find an explanation for the Platform merge process in freeze period, but failed. Do we have it?

Issue #2126, opened 2025-08-21
Improve Text Selection in Terminal


Philipp0205:
I noticed recent improvements to the terminal and I’d like to suggest a feature that could further enhance usability.

In most terminals, when a user double-clicks on text, the first double-click selects a word and a triple-click selects the entire line. It would be great if Eclipse’s terminal could support similar progressive selection behavior.

laeubi:
@Philipp0205 this sounds useful do you like to propose a PR?

Philipp0205:
@laeubi I could try to implement it myself, but my eclipse development experience is limited. Would you have any place/tips to start?

laeubi:
As a first step you should setup an IDE with all the code:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment

then I would start with

- https://github.com/eclipse-platform/eclipse.platform/blob/master/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/view/TerminalsView.java
- https://github.com/eclipse-platform/eclipse.platform/blob/master/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/tabs/TabFolderManager.java

and poke a bit around, I can see that `TabFolderManager` already registers some listener regarding mouse handling in the `TerminalControlSelectionListener`

PR #2131, opened 2025-08-26
Implement word and line selection modes in TextCanvas

Philipp0205:
Implemented word/line selection for the terminal view similar to the behavior eclipse already has in the console / editors.
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2126

What it does:

- Double‑click selects a word, dragging after that extends by whole words.
- Triple‑click selects a line, dragging after that extends by whole lines

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
+++ b/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
@@ -294,7 +294,7 @@ public class TextCanvas extends GridCanvas {
 			}
 			redraw();
 		});
-		serVerticalBarVisible(true);
+		setVerticalBarVisible(true);
 		setHorizontalBarVisible(false);
 	}
 
--- a/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/VirtualCanvas.java
+++ b/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/VirtualCanvas.java
@@ -335,7 +335,7 @@ public abstract class VirtualCanvas extends Canvas {
 		return getVerticalBar().isVisible();
 	}
 
-	protected void serVerticalBarVisible(boolean showVScrollBar) {
+	protected void setVerticalBarVisible(boolean showVScrollBar) {
 		ScrollBar vertical = getVerticalBar();
 		vertical.setVisible(showVScrollBar);
 		vertical.setSelection(0);


terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/VirtualCanvas.java
```diff
@@ -335,7 +335,7 @@ protected boolean isVertialBarVisible() {
 		return getVerticalBar().isVisible();
 	}
 
-	protected void serVerticalBarVisible(boolean showVScrollBar) {
```
ruspl-afed: This change is not related and should go to a separate PR

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
```diff
@@ -215,10 +288,80 @@ public void mouseUp(MouseEvent e) {
...
+		}
+
+		boolean isValid() {
+			return start != null && end != null;
+		}
+	}
+
+	private Range wordRangeAt(Point pt) {
+		if (pt == null)
+			return new Range(null, null);
```
ruspl-afed: Let's avoid `null` values, we can declare special value for `Range` instead

ruspl-afed:
@Philipp0205 thank you for your contribution

Please make sure that your PR corresponds to the coding style of the changed bundle. For example I can see missed brackets in a few places.
Also I would avoid using `null` as a special value where possible.

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
```diff
@@ -215,10 +288,80 @@ public void mouseUp(MouseEvent e) {
...
+			this.start = start;
+			this.end = end;
+		}
+
+		boolean isValid() {
+			return start != null && end != null;
+		}
+	}
+
+	private Range wordRangeAt(Point pt) {
```
laeubi, line 315: Is null really a valid input here? Are there other cases where a Range can not be supplied? If yes you should use


```suggestion
	private Optional<Range> wordRangeAt(Point pt) {
```

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
Philipp0205: After some testing, I realized that null is not a valid input here, as @laeubi also pointed out. Therefore, I can just remove this check.

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextCanvas.java
Philipp0205: I tested different scenarios, and you have is not valid input here. From my perspective, we can remove the null check and there is no need to make the method Optional.

laeubi:
@ruspl-afed any further comments or can we merge this (maybe it requires a rebase @Philipp0205 )

Philipp0205:
@laeubi I am currently on vacation, will rebase once I am back next week :)

Philipp0205:
@laeubi I see some failing checks since the build seems to be failing. I took a quick look, and it seems to be unrelated to my changes. What is the process here, should I update my branch to rerun the checks?

laeubi:
You need to rebase your commit on current master, for some reason I can't do that automatically here.

ruspl-afed:
> @ruspl-afed any further comments or can we merge this (maybe it requires a rebase @Philipp0205 )

Now it looks much better, thank you @Philipp0205

merks:
This wasn't just rebased and force pushing but rather did some type of merge.

Philipp0205:
@merks yes something went wrong. Let me try to redo it

Philipp0205:
@laeubi I again see two failing checks that seem unrelated to my changes. Can I somehow re-run them?

laeubi:
@Philipp0205 that seems a flapping test and not related to your change as it only fails on one platform.

I would suggest you squash and rebase your change and if @ruspl-afed has no further request merge this.

Philipp0205:
@laeubi done :)

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/VirtualCanvas.java
```diff
@@ -335,7 +335,7 @@ protected boolean isVertialBarVisible() {
 		return getVerticalBar().isVisible();
 	}
 
-	protected void setVerticalBarVisible(boolean showVScrollBar) {
+	protected void serVerticalBarVisible(boolean showVScrollBar) {
```
ruspl-afed: what was the reason to rename this method? looks like a misprint

Philipp0205: Ah, I fixed this typo in https://github.com/eclipse-platform/eclipse.platform/pull/2135 and somehow forgot it here. Will fix, and thanks for pointing it out.

PR #2135, opened 2025-08-31
Fix method name typo for vertical bar visibility.

Philipp0205:
This is just a typo fix that came up while implementing #2131


HeikoKlare (approved):
Thank you! That's obviously a typo and since the class is internal, it's safe to change the method name.

PR #2136, opened 2025-08-31
Remove unused reference to PlatformAdmin / do not use PackageAdmin

laeubi:
PlatformAdmin + PackageAdmin are part of the old legacy Resolver API and should not be used anymore.

### DIFF
--- a/update/org.eclipse.update.configurator/src/org/eclipse/update/internal/configurator/Utils.java
+++ b/update/org.eclipse.update.configurator/src/org/eclipse/update/internal/configurator/Utils.java
@@ -26,17 +26,16 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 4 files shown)


update/org.eclipse.update.configurator/src/org/eclipse/update/internal/configurator/Utils.java
```diff
@@ -200,44 +198,21 @@ public static String getNL() {
...
 		}
+		Bundle[] bundles = context.getBundles();
 		//Return the first bundle that is not installed or uninstalled
 		for (Bundle bundle : bundles) {
-			if ((bundle.getState() & (Bundle.INSTALLED | Bundle.UNINSTALLED)) == 0) {
+			if (bundle.getSymbolicName().equals(symbolicName)
+					&& (bundle.getState() & (Bundle.INSTALLED | Bundle.UNINSTALLED)) == 0) {
 				return bundle;
 			}
 		}
```
HannesWell: Can't this entire method be just replaced with `org.eclipse.core.runtime.Platform.getBundle(String)`?

laeubi: Maybe yes, I haven checked why it was not used here.

laeubi: I replaced all calls with a call to platform now.

PR #2137, opened 2025-09-01
Depend on Lucene using the Import-Package header

ptziegler:
Using the Import-Header entries allows consumers to contribute their own Lucene bundles, without having to match the symbolic name defined by Orbit.

### DIFF
--- a/ua/org.eclipse.help.base/META-INF/MANIFEST.MF
+++ b/ua/org.eclipse.help.base/META-INF/MANIFEST.MF
@@ -42,11 +42,29 @@ Export-Package: org.apache.lucene.demo.html;x-internal:=true,
 Require-Bundle: org.eclipse.ant.core;bundle-version="[3.2.200,4.0.0)";resolution:=optional,
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
  org.eclipse.help;bundle-version="[3.5.0,4.0.0)";visibility:=reexport,
- org.eclipse.core.net;bundle-version="[1.2.200,2.0.0)",
- org.apache.lucene.analysis-common;bundle-version="[10.0.0,11.0.0)",
- org.apache.lucene.core;bundle-version="[10.0.0,11.0.0)",
- org.apache.lucene.analysis-smartcn;bundle-version="[10.0.0,11.0.0)"
-Import-Package: org.eclipse.equinox.http.jetty;resolution:=optional
+ org.eclipse.core.net;bundle-version="[1.2.200,2.0.0)"
+Import-Package: org.apache.lucene.analysis;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.br;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.cjk;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.cn.smart;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.core;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.cz;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.de;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.el;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.en;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.fr;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.miscellaneous;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.nl;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.ru;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.standard;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.tokenattributes;version="[10.0.0,11.0.0)",
+ org.apache.lucene.analysis.util;version="[10.0.0,11.0.0)",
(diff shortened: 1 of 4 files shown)


ptziegler:
For context: We maintain three releases of our application, the oldest one based on the Eclipse 2023-12. Because of [CVE-2024-45772](https://nvd.nist.gov/vuln/detail/CVE-2024-45772), we need to update to a newer Lucene version. But because the Help plugin expects a specific bundle name, this became slightly more frustrating.

merks (approved):
It's a better design.

laeubi (approved):
Looks resonable

ptziegler:
I've udated the PR to handle the failing test cases. Following import needed to be added as well: `org.apache.lucene.analysis.cn.smart`.

Which raises a potential problem I'm not sure how to handle. Users can contribute their own analyzers using the `org.eclipse.help.base.luceneAnalyzer` extension point. And given that Lucene has somehow decided to define a package for each language, all of those packages need to be imported, so that they are visible to the classloader.

Or to rephrase: Using `Import-Package` seems really scary, now that I have a closer look...

laeubi:
@ptziegler 

> Users can contribute their own analyzers using the `org.eclipse.help.base.luceneAnalyzer` extension point

That seems reasonable and I don't see an immediate problem here?

> And given that Lucene has somehow decided to define a package for each language, all of those packages need to be imported, so that they are visible to the classloader

Can you explain this more especially how require-bundle makes the situation better? We then maybe will find a solution out here. Maybe `Dynamic-ImportPackage org.apache.lucene.analysis.*;version="[10.0.0,11.0.0)"` would be a solution? 

But I'm not completely sure why the imports are needed for a custom Analyzer, should not the custom one then make sure all packages are properly imported?

ptziegler:
@laeubi 

Assume someone contributes the following analyzer:

```
  <analyzer
        class="org.apache.lucene.analysis.ta.TamilAnalyzer"
        locale="ta">
  </analyzer>
```

Because the `org.apache.lucene.analysis.ta` package is not imported, this leads to a ClassNotFoundException:

```
java.lang.ClassNotFoundException: org.apache.lucene.analysis.ta.TamilAnalyzer cannot be found by org.eclipse.help.base_4.5.300.qualifier
	at org.eclipse.osgi.internal.loader.BundleLoader.generateException(BundleLoader.java:567)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass0(BundleLoader.java:562)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:438)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:195)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.loadClass(EquinoxBundle.java:652)
	at org.eclipse.core.internal.registry.osgi.RegistryStrategyOSGI.createExecutableExtension(RegistryStrategyOSGI.java:226)
	at org.eclipse.core.internal.registry.ExtensionRegistry.createExecutableExtension(ExtensionRegistry.java:1034)
	at org.eclipse.core.internal.registry.ConfigurationElement.createExecutableExtension(ConfigurationElement.java:286)
	at org.eclipse.core.internal.registry.ConfigurationElementHandle.createExecutableExtension(ConfigurationElementHandle.java:65)
	at org.eclipse.help.internal.search.AnalyzerDescriptor.createAnalyzer(AnalyzerDescriptor.java:113)
	at org.eclipse.help.internal.search.AnalyzerDescriptor.<init>(AnalyzerDescriptor.java:40)
```

You wouldn't have this issue when using `Require-Bundle`, because all packages are visible.

> But I'm not completely sure why the imports are needed for a custom Analyzer, should not the custom one then make sure all packages are properly imported?

The problem is that the classloader for `org.eclipse.help.base` is used, not the one from the contributing bundle. Otherwise you would be correct.

laeubi:
> The problem is that the classloader for org.eclipse.help.base is used, not the one from the contributing bundle. 

This sounds strange as it would mean that the user of the extension point would need to import all requirements what should not be the case, so it it ensured that the provider has a proper import package or was it just omitted because it "worked anyways"?

I'm asking because the code in question explicitly says it loads the class from the contributing bundle:


RegistryStrategyOSGI.java
```
// load the requested class from this bundle
Class<?> classInstance = null;
try {
	classInstance = contributingBundle.loadClass(className);
} catch (Exception | LinkageError e1) {
	throwException(
			NLS.bind(RegistryMessages.plugin_loadClassError, contributingBundle.getSymbolicName(), className),
			e1);
}
```

laeubi:
@ptziegler I now found that the extensions are directly provided by the `org.eclipse.help.base` bundle, so `contributor` == `consumer`.

In this case I would recommend to use the `@Referenced(... list of implicitly required class names ...)` for example at the `AnalyzerDescriptor` (alternatively a `package-info.class` can be used) to make this explicit and visible. Otherwise the imported packages will be assumed to be unused sooner or later!

ptziegler:
> I'm asking because the code in question explicitly says it loads the class from the contributing bundle:

Yes, you are correct. I got misled by the stacktrace.

I've extended the UA test with the Tamil analyzer, which is also where I contributed to the extension point. After I added the additional package to the test bundle, the stack trace went away.

So that is indeed a non-issue.

ptziegler:
> In this case I would recommend to use the @Referenced(... list of implicitly required class names ...) for example at the AnalyzerDescriptor (alternatively a package-info.class can be used) to make this explicit and visible. Otherwise the imported packages will be assumed to be unused sooner or later!

Done

ptziegler:
Merging, as the only issue seems to be my misunderstanding of which classes are loaded by which classloader.

PR #2139, opened 2025-09-02
[RelEng] Add release preparation script to update the Splash-screen

HannesWell:
The script will be called by the pipeline to prepare a new development cycle and automates the update of the splash-screen for the new cycle. This automates manual changes in the release preparation process like
- https://github.com/eclipse-platform/eclipse.platform/pull/2133

and therefore simplifies and speeds up the preparation itself.

At the same time it adds the requirement to make all the updated images available in advance of the preparation. This means that if a new splash-screen style/template is introduced (which typically happens once a year) the image files for the splash-screen and the about-dialog must be added before the preparation is done. In general I don't think that isn't a problem, we just have to remember that it should happen at least in the very end of a development cycle.
Furthermore the files in the `futureSplashScreens` 'staging' directory now have to follow a strict naming schema.

The script is written in a way that it moves the splash-screen images into the target location, replacing the previous version. This also means that the 'staging' directory will be empty of previous splash-screen images when the new ones are added.
I could also change to just copy the corresponding files into the target location. Then the old files should be deleted when new ones are added. This might maybe help those that add new files a little bit, but could also be looked up in the history.

@MohananRahul, @subyssurendran666 is this fine and is the documentation sufficient?
@HeikoKlare while preparing this I was wondering why the About-Dialog files are not yet SVGs? Is it because they are simply not available as such and we should ask for them in the next round? Or are these images too complex?


MohananRahul (approved):
Looks fine, ie we can store all future Splash screen in this folder.

HannesWell:
> Looks fine, ie we can store all future Splash screen in this folder.

Great, thanks. The verification build doesn't test this, so this can be submitted immediately.
I'll update the release documentation soon.

Btw. we need a new splash-screen for 2025-06. So I assume the discussion about it will start in the next months?

PR #2140, opened 2025-09-05
Version bump for new stream

akurtakov:
Manual bump needed as it's caused by change in other repo. Needed for verification of
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3286


akurtakov:
Merging despite failing test as it's unrelated to version bump and we have to unbreak builds.

PR #2144, opened 2025-09-08
Add configurable elapsed time display with format customization

### LINKED ISSUE #2112: console elapsed time format
console title is updated every second with elapsed time from program start; into console preferences there should an option for elapsed time format used in [org.eclipse.debug.internal.ui.views.console.ProcessConsole ](https://github.com/eclipse-platform/eclipse.platform/blob/401d48340e4b43d76bb306acc01e306c3c93cb4e/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/console/ProcessConsole.java#L352)or to hide it (usually not useful in development phase)

SougandhS:
This PR introduces a toggle to enable or disable the display of elapsed time in the console output in Console settings. 

Added support for customizing the format of the elapsed time, allowing users to choose from multiple display styles based on their preferences or requirements.

image

https://github.com/user-attachments/assets/a6f8e2dd-8bb0-4ec1-9345-d76443b5b283

fixes : https://github.com/eclipse-platform/eclipse.platform/issues/2112

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -15,9 +15,14 @@ package org.eclipse.debug.internal.ui.preferences;
 
 
 import java.text.MessageFormat;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.regex.Pattern;
 
 import org.eclipse.debug.internal.ui.DebugUIPlugin;
 import org.eclipse.debug.internal.ui.IDebugHelpContextIds;
+import org.eclipse.debug.internal.ui.SWTFactory;
+import org.eclipse.debug.internal.ui.views.console.ProcessConsole;
 import org.eclipse.jface.preference.BooleanFieldEditor;
 import org.eclipse.jface.preference.ColorFieldEditor;
 import org.eclipse.jface.preference.FieldEditor;
@@ -25,12 +30,17 @@ import org.eclipse.jface.preference.FieldEditorPreferencePage;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.preference.IntegerFieldEditor;
 import org.eclipse.jface.util.PropertyChangeEvent;
+import org.eclipse.jface.viewers.ArrayContentProvider;
+import org.eclipse.jface.viewers.ComboViewer;
+import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Combo;
 import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Label;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPreferencePage;
 import org.eclipse.ui.PlatformUI;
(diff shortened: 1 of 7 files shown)


SougandhS:
Hi @merks 
Could you please check this PR when you are available ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
```diff
@@ -159,6 +165,27 @@ public void widgetSelected(SelectionEvent e) {
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_OUT, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_output_3, SWT.NONE, getFieldEditorParent()));
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_ERR, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_error_3, SWT.NONE, getFieldEditorParent()));
 
+		fshowElapsedTime = new BooleanFieldEditor2(IDebugPreferenceConstants.CONSOLE_ELAPSED_TIME,
+				DebugPreferencesMessages.ConsoleElapsedTime, SWT.NONE, getFieldEditorParent());
+		SWTFactory.createLabel(getFieldEditorParent(), "Format ", 1); //$NON-NLS-1$
+		fElapsedFormat = new Combo(getFieldEditorParent(), SWT.BORDER|SWT.READ_ONLY);
+		fElapsedFormat
+				.setItems(new String[] { "H:MM:SS", "HH:MM:SS", "HH:MM:SS.mmm", "MM:SS.mmm",
```
merks: I see later on there is a relation between the index of each of these strings, and the magic numbers used in `getElapsedFormat`.  

```
pref.setValue(IDebugPreferenceConstants.CONSOLE_ELAPSED_FORMAT, fElapsedFormat.getSelectionIndex());
```

 That knowledge is likely to be lost in the future. 

@laeubi 

Do you have a good idea how one might improve this?

laeubi:
> Also added support for customizing the format of the elapsed time, allowing users to choose from multiple display styles based on their preferences or requirements.

I would suggest to make this a text field where the user can enter any formt wanted, and simply let empty mean its disabled.

SougandhS:
> I would suggest to make this a text field where the user can enter any formt wanted, and simply let empty mean its disabled.

Sure 👍

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
laeubi: If one really wants to use an index, it is best to define an enum and then use `ordinal()`,b but this will require a strong hint that order  in the enum matters!

As suggested I would better store the "string format" itself and let it being parsed (what of course is a bit more effort but much more flexible on the long run).

SougandhS: Changed to String preference 👍

SougandhS:
>the user can enter any formt wanted, and simply let empty mean its disabled.

Updated as suggested

https://github.com/user-attachments/assets/a132cb55-238b-4a16-97e4-81215306319b



image

laeubi:
@SougandhS thanks as the format is a bit ugly and maybe not everyone is familiar with it I would suggest that you add a litte description (e.g. a tooltip to tell what are valid I assume it has to be a java format string, but what about positional parameters?) and a preview filed that shows the result of the user input so one do not needs to start a process to see how it will look like.

As a bonus, you maybe want to have using `AutoCompleteField` to give the user some suggestions, or a `ComboViewer` with edit capability where one can choose some predefined values from the dropdown.

SougandhS:
> @SougandhS thanks as the format is a bit ugly and maybe not everyone is familiar with it I would suggest that you add a litte description (e.g. a tooltip to tell what are valid I assume it has to be a java format string, but what about positional parameters?) and a preview filed that shows the result of the user input so one do not needs to start a process to see how it will look like.
> 
> As a bonus, you maybe want to have using `AutoCompleteField` to give the user some suggestions, or a `ComboViewer` with edit capability where one can choose some predefined values from the dropdown.

Sure, I'll work on this 👍

SougandhS:
I have updated as suggested 

https://github.com/user-attachments/assets/50dcd0c3-09d0-4602-a441-ed91f210e4ff



image

laeubi:
@SougandhS that looks good already, but in the drowdown I would use a more human readable label for the entries like in your initial PR

SougandhS:
> @SougandhS that looks good already, but in the drowdown I would use a more human readable label for the entries like in your initial PR

Will update it 👍

SougandhS:
Hi @laeubi 
Updated to human readable form just like the initial PR


https://github.com/user-attachments/assets/a6f8e2dd-8bb0-4ec1-9345-d76443b5b283


image

laeubi:
Wow that's much more better now (and actually more than I suggested here), but seems much more user friendly now!

SougandhS:
> @laeubi

Thank you for your suggestions :)

SougandhS:
Hi @laeubi, If there are no more changes can this be merged ? Test failures are unrelated.

SougandhS:
Thanks a lot!

jukzi:
Hi, just tried to use this option and found "MM:SS.mmm" gets it wrong (actually prints HH:MM:SSS), which is even visible in the preview:
image
image

SougandhS:
> Hi, just tried to use this option and found "MM:SS.mmm" gets it wrong (actually prints HH:MM:SSS), which is even visible in the preview:

hi Jörg, long time no see :D
Will check this 👍

Issue #2148, opened 2025-09-08
Few WorkManager fields are not MT safe


iloveeclipse:
While debugging https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4370 I've noticed that in the `org.eclipse.core.internal.resources.WorkManager` few fields are not volatile/atomic but could be accessed from different threads, so even if the modification happens while holding workspace locks, other threads might see stale values.

Fields in question are 
- `WorkManager.operationCanceled`
- `WorkManager.nestedOperations`
- `WorkManager.preparedOperations`

I don't have reproducer, but I think making the fields either volatile or atomic shouldn't harm.

laeubi:
> I don't have reproducer, but I think making the fields either volatile or atomic shouldn't harm.

I must confess that I disagree here.

 It harms at least in the following ways:

1. It will now make the code more complex 
2. It make it look like it is handling concurrent access, so the next time one will maybe want to add other "missing" synchronization elsewhere
3. `volatile` does not allow certain caching and forces to read/write values immediately so can impact performance due to pipeline stalls
4. atomic is most useful when we really make use of any of its concurrent methods (e.g. compareAndSet) and act accordingly

PR #2149, opened 2025-09-08
Few WorkManager fields are not MT safe

iloveeclipse:
Few fields are not volatile/atomic but could be accessed from different threads, so even if the modification happens while holding workspace locks, other threads might see stale values.

Fields in question are
- `WorkManager.operationCanceled`
- `WorkManager.nestedOperations`
- `WorkManager.preparedOperations`

Additionally changed Workspace.endOperation() to read workManager.getPreparedOperationDepth() only once to avoid inconsistent reads.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2148

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/WorkManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/WorkManager.java
@@ -15,13 +15,17 @@
  *******************************************************************************/
 package org.eclipse.core.internal.resources;
 
+import java.util.concurrent.atomic.AtomicInteger;
 import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.internal.utils.Policy;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceStatus;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.core.runtime.jobs.*;
+import org.eclipse.core.runtime.jobs.IJobManager;
+import org.eclipse.core.runtime.jobs.ILock;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
 
 /**
  * The work manager governs concurrent access to the workspace tree.  The {@link #lock}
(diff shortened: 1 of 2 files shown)


laeubi:
> so even if the modification happens while holding workspace locks, other threads might see stale values

Wo other threads are *reading* the values outside the workspace lock? Looking at the code changes it seem not to really benefit / improve from the changes otherwise, e.g. the atomic operation is never used to compareAndSet and otherwise only unconditional set values.

iloveeclipse:
> Wo other threads are _reading_ the values outside the workspace lock?

I'm not saying that they are reading outside the lock, I'm saying they are probably reading stale values.
`WorkManager.operationCanceled()` and `.WorkManager.shouldBuild()` are called from methods that are not synchronized, the filed is not volatile, so IMO other threads might see stale values.

laeubi:
> I'm not saying that they are reading outside the lock, I'm saying they are probably reading stale values.

If they are reading them under the workspace look, how can they ever see stale values?

> WorkManager.operationCanceled() and .WorkManager.shouldBuild() are called from methods that are not synchronized, the filed is not volatile, so IMO other threads might see stale values.

I always wondered where the impression comes from that only synchronized methods are barriers and anyone should see "stale" data, if that would be the case we would need synchronized everywhere and java would be completely unusable as a programming language.

iloveeclipse:
> I always wondered where the impression comes from that only synchronized methods

I never said that, please carefully quote what I've said.

> If they are reading them under the workspace look

The workspace lock I mean is `WorkManager.lock`, and it is for humans very hard to proof which `WorkManager` operation is inside a sequence of calls that acquired this lock and how many times the lock was already acquired/released.

Given that, quote myself, note the "might" verb:

> Few fields are not volatile/atomic but could be accessed from different threads, so even if the modification happens while holding workspace locks, other threads might see stale values.

Therefore IMO if we know the code is runing in multiple threads, the fields are not volatile, and we can't easily proof all paths that read/write of the fields are properly guarded by either synchronized blocks or the lock above, it is safer to make the code protected from possible MT issues.

laeubi:
> Therefore IMO if we know the code is runing in multiple threads, the fields are not volatile, and we can't easily proof all paths that read/write of the fields are properly guarded by either synchronized blocks or the lock above, it is safer to make the code protected from possible MT issues.

But that's the point, if we can't prove that then the change does not gives us anything better regards MT issues, it would only solve e.g. if we can see that one code hangs forever (e.g. boolean value is not visible due to missing volatile as its cached) but as you said you can't give a proof here.

Second is the update of the filed, now made atomic but still do not prevent from a race condition (that is one Thread is writing a value where another is seeing a stale value and behave wrong).

So I'm really a bit reserved on changing code based on "feelings" ( = might help but might not and we hope it does not harm) instead of at least facts (what would be for example a callstack that shows problematic access or at least a description of the case to fix). You are referencing explicitly an issue in JDT, so how/why would one even expect to make the situation **better** there with this change?

PR #2161, opened 2025-09-15
Resources / Runtime tests: Do not depend on higher-level plug-ins.

sratz:
* PreferencesServiceTest is not deterministic but its behavior changes depending on the presence of downstream plug-in org.eclipse.core.resources. Take this into account.

* IResourceTest should not depend on downstream plug-in org.eclipse.ui.ide (even inside a downstream repository). Use a different (upstream) filter provider in the unit test instead.

These downstream plug-ins are currently only visible because Tycho adds the full-blown UI test harness to the target platform unconditionally, even if there is no such actual dependency [1]. This is a cleanup in preparation for [2].

[1] https://github.com/eclipse-tycho/tycho/issues/5349
[2] https://github.com/eclipse-tycho/tycho/pull/5362

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IResourceTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IResourceTest.java
@@ -1101,9 +1101,9 @@ public class IResourceTest {
 		sourceProj.open(createTestMonitor());
 		// create a new filter.
 		if (applyResFilter) {
-			String MULTI_FILT_ID = "org.eclipse.ui.ide.multiFilter";
-			String FILT_ARG = "1.0-length-equals-false-false-10485760";
-			FileInfoMatcherDescription filterDesc = new FileInfoMatcherDescription(MULTI_FILT_ID, FILT_ARG);
+			String filterProviderId = "org.eclipse.core.resources.regexFilterMatcher";
+			String filterArguments = "foo.*";
+			FileInfoMatcherDescription filterDesc = new FileInfoMatcherDescription(filterProviderId, filterArguments);
 			int EXCL_FILE_GT = IResourceFilterDescription.EXCLUDE_ALL + IResourceFilterDescription.FILES
 					+ IResourceFilterDescription.INHERITABLE;
 			sourceProj.createFilter(EXCL_FILE_GT, filterDesc, IResource.BACKGROUND_REFRESH, createTestMonitor());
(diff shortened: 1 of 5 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IResourceTest.java
```diff
@@ -1101,8 +1101,8 @@ private IProject createProject(boolean applyResFilter) throws CoreException {
 		sourceProj.open(createTestMonitor());
 		// create a new filter.
 		if (applyResFilter) {
-			String MULTI_FILT_ID = "org.eclipse.ui.ide.multiFilter";
-			String FILT_ARG = "1.0-length-equals-false-false-10485760";
+			String MULTI_FILT_ID = "org.eclipse.core.resources.regexFilterMatcher";
+			String FILT_ARG = "foo.*";
```
laeubi: I must confess I don't understand what this actually does and where it is needed in the test but would trust your analysis here and it looks much more understandable than the previous version.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
```diff
@@ -222,11 +223,25 @@ public void testImportExportBasic() throws Exception {
 	@Test
 	public void testLookupOrder() {
 		IPreferencesService service = Platform.getPreferencesService();
-		String[] defaultOrder = new String[] {"project", //$NON-NLS-1$
+		// "project" scope is added via
+		// PreferencesService.prependScopeToDefaultDefaultLookupOrder(String)
+		// by org.eclipse.core.internal.resources.Workspace, when the the
```
laeubi:
```suggestion
		// by org.eclipse.core.internal.resources.Workspace, when the
```

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/preferences/PreferencesServiceTest.java
```diff
@@ -222,11 +223,25 @@ public void testImportExportBasic() throws Exception {
...
 				InstanceScope.SCOPE, //
 				ConfigurationScope.SCOPE, //
 				UserScope.SCOPE, //
-				DefaultScope.SCOPE};
+				DefaultScope.SCOPE) //
+		);
+		Bundle resourcesBundle = Platform.getBundle("org.eclipse.core.resources");
+		if (resourcesBundle != null) {
+			if (resourcesBundle.getState() == Bundle.ACTIVE) {
+				defaultOrderList.add(0, "project");
```
laeubi, line 241: Should the comment about project scope not better be here?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/IResourceTest.java
sratz: Yeah, the test basically just wants any `<filteredResources><filter>...`  in the `.project` metadata. And we should use one here that is actually known here (or upstream) :)

Issue #2155, opened 2025-09-12
Restoring Terminal View on start broken


ppalaga:
- [ ] I have not verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) because I am experiencing this with a fresh release that was published yesterday

## Steps to reproduce

From a fresh installation and clean workspace:

* Invoke "Show in local terminal" from two different nodes (folder-1 and folder-2) in Project Explorer
* This opens two tabs inside the Terminal View
* Restart Eclipse
* Expected: the two terminals set to the respective folders are restored within a single Terminal view upon restart 
* Actual: There are three Terminal views opened: 
  - "Terminal" containing no terminal tabs
  - "Terminal 1" containing folder-1
  - "Terminal 2" containing folder-2

* Restart Eclipse again
* Expected: the terminal views are restored in the same state as they were before the restart
* Actual: There are five Terminal views opened: 
  - "Terminal" containing no terminal tabs
  - "Terminal 1" containing no terminal tabs
  - "Terminal 2" containing no terminal tabs
  - "Terminal 3" containing folder-1
  - "Terminal 4" containing folder-2

## Here is some relevant log output

I do not think it is relevant.

## Tested under this environment:
* OS & version: Linux, Fedora 42 with KDE Plasma
* Eclipse IDE/Platform version (as shown in *Help > About*): ...
```
Eclipse IDE for Java Developers (includes Incubating components)

Version: 2025-09 (4.37.0)
Build id: 20250905-1456

org.eclipse.terminal.connector.local (1.0.0.v20250724-2146) "Terminal Local Connector" [Active]
org.eclipse.terminal.connector.process (1.0.0.v20250724-2146) "Terminal Process Connector" [Active]
org.eclipse.terminal.connector.ssh (1.0.0.v20250724-2146) "Terminal SSH Connector" [Active]
org.eclipse.terminal.connector.telnet (1.0.0.v20250724-2146) "Terminal Telnet Connector" [Active]
org.eclipse.terminal.control (1.0.0.v20250723-1539) "Terminal Control (Embeddable Widget)" [Active]
org.eclipse.terminal.view.core (1.0.0.v20250724-1302) "Terminal View Core" [Active]
org.eclipse.terminal.view.ui (1.0.0.v20250819-0800) "Terminal View" [Active]
```

## Community

[...]

PR #2163, opened 2025-09-16
Ensure terminal async runnables are run asynchronously

jonahgraham:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2155

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
@@ -19,6 +19,7 @@ import java.util.Optional;
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
@@ -72,11 +73,11 @@ public class LocalLauncherHandler extends AbstractHandler {
 		Map<String, Object> properties = new HashMap<>();
 		properties.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegate.getId());
 		properties.put(ITerminalsConnectorConstants.PROP_SELECTION, selection);
-		try {
-			delegate.execute(properties);
-		} catch (Exception e) {
-			throw new ExecutionException(e.getMessage(), e);
-		}
+		delegate.execute(properties).whenComplete((r, e) -> {
+			if (e != null) {
+				ILog.get().error("Error occurred while running delegate to open console", e); //$NON-NLS-1$
+			}
+		});
 	}
 
 }
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
@@ -16,12 +16,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.ISafeRunnable;
-import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.ListenerList;
 import org.eclipse.core.runtime.SafeRunner;
 import org.eclipse.core.runtime.Status;
@@ -68,8 +68,11 @@ public class TerminalService implements ITerminalService {
 
 	/**
 	 * Common terminal service runnable implementation.
+	 *
+	 * These runnables are run asynchronously on the UI thread.
 	 */
-	protected static abstract class TerminalServiceRunnable {
+	@FunctionalInterface
+	private static interface TerminalServiceRunnable {
 
 		/**
 		 * Invoked to execute the terminal service runnable.
@@ -78,21 +81,10 @@ public class TerminalService implements ITerminalService {
 		 * @param title The terminal tab title. Must not be <code>null</code>.
 		 * @param connector The terminal connector. Must not be <code>null</code>.
 		 * @param data The custom terminal data node or <code>null</code>.
-		 * @return the result {@link IStatus}
+		 * @throws CoreException if operation does not complete successfully
+		 * @throws NullPointerException if any required arguments are null.
 		 */
-		public abstract IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data);
-
-		/**
-		 * Returns if or if not to execute the runnable asynchronously.
-		 * <p>
-		 * The method returns per default <code>true</code>. Overwrite to
-		 * modify the behavior.
-		 *
-		 * @return <code>True</code> to execute the runnable asynchronously, <code>false</code> otherwise.
-		 */
-		public boolean isExecuteAsync() {
-			return true;
-		}
+		void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) throws CoreException;
 	}
 
 	@Activate
@@ -194,18 +186,14 @@ public class TerminalService implements ITerminalService {
 	private CompletableFuture<?> executeServiceOperation(final TerminalServiceRunnable runnable, TerminalViewId tvid,
 			final String title, final ITerminalConnector connector, final Object data) {
 		try { // Execute the operation
-			if (runnable.isExecuteAsync()) {
-				return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
-						/*
-						 * A special executor is used here because the default Display executor will
-						 * do a syncExec here if called from the UI thread, but isExecuteAsync is true so we
-						 * must run async on the UI thread.
-						 */
-						Display.getDefault()::asyncExec);
+			return CompletableFuture.runAsync(() -> {
+				try {
+					runnable.run(tvid, title, connector, data);
+				} catch (CoreException e) {
+					throw new CompletionException(e);
+				}
+			}, Display.getDefault()::asyncExec);
 
-			} else {
-				return CompletableFuture.completedFuture(runnable.run(tvid, title, connector, data));
-			}
 		} catch (RuntimeException e) {
 			return CompletableFuture.failedFuture(e);
 		}
@@ -265,9 +253,10 @@ public class TerminalService implements ITerminalService {
 		final boolean restoringView = fRestoringView;
 		return executeServiceOperation(properties, new TerminalServiceRunnable() {
 			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
+			public void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data)
+					throws CoreException {
 				if (restoringView) {
-					return doRun(tvid, title, connector, data);
+					doRun(tvid, title, connector, data);
 				} else {
 					// First, restore the view. This opens consoles from the memento
 					fRestoringView = true;
@@ -277,11 +266,12 @@ public class TerminalService implements ITerminalService {
 						ILog.get().log(e.getStatus());
 					}
 					fRestoringView = false;
-					return doRun(tvid, title, connector, data);
+					doRun(tvid, title, connector, data);
 				}
 			}
 
-			private IStatus doRun(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
+			private void doRun(TerminalViewId tvid, String title, ITerminalConnector connector, Object data)
+					throws CoreException {
 				// Determine the terminal encoding
 				String encoding = (String) properties.get(ITerminalsConnectorConstants.PROP_ENCODING);
 				// Create the flags to pass on to openConsole
@@ -302,16 +292,11 @@ public class TerminalService implements ITerminalService {
 					flags.put(ITerminalsConnectorConstants.PROP_TITLE_DISABLE_ANSI_TITLE, false);
 				}
 				// Open the new console
-				try {
-					Widget console = consoleViewManager.openConsole(tvid, title, encoding, connector, data, flags);
-					// Associate the original terminal properties with the tab item.
-					// This makes it easier to persist the connection data within the memento handler
-					if (!console.isDisposed()) {
-						console.setData("properties", properties); //$NON-NLS-1$
-					}
-					return Status.OK_STATUS;
-				} catch (CoreException e) {
-					return e.getStatus();
+				Widget console = consoleViewManager.openConsole(tvid, title, encoding, connector, data, flags);
+				// Associate the original terminal properties with the tab item.
+				// This makes it easier to persist the connection data within the memento handler
+				if (!console.isDisposed()) {
+					console.setData("properties", properties); //$NON-NLS-1$
 				}
 			}
 		});
@@ -320,25 +305,12 @@ public class TerminalService implements ITerminalService {
 	@Override
 	public CompletableFuture<?> closeConsole(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
-		return executeServiceOperation(properties, new TerminalServiceRunnable() {
-			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
-				consoleViewManager.closeConsole(tvid, title, connector, data);
-				return Status.OK_STATUS;
-			}
-		});
+		return executeServiceOperation(properties, consoleViewManager::closeConsole);
 	}
 
 	@Override
 	public CompletableFuture<?> terminateConsole(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
-
-		return executeServiceOperation(properties, new TerminalServiceRunnable() {
-			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
-				consoleViewManager.terminateConsole(tvid, title, connector, data);
-				return Status.OK_STATUS;
-			}
-		});
+		return executeServiceOperation(properties, consoleViewManager::terminateConsole);
 	}
 }
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
@@ -193,12 +193,21 @@ public class TerminalService implements ITerminalService {
 
 	private CompletableFuture<?> executeServiceOperation(final TerminalServiceRunnable runnable, TerminalViewId tvid,
 			final String title, final ITerminalConnector connector, final Object data) {
-		// Execute the operation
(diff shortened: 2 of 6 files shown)


terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -196,7 +196,19 @@ private CompletableFuture<?> executeServiceOperation(final TerminalServiceRunnab
...
+					 * do a syncExec here, but isExecuteAsync is true, so we need to return from
+					 * executeServiceOperation *before* we execute the runnable.
+					 */
+					(command) -> {
+						try {
+							Display display = Display.getDefault();
+							display.asyncExec(command);
+						} catch (Exception e) {
+							// if display is disposed, silently ignore.
+						}
```
laeubi:
```suggestion
					Display.getDefault()::asyncExec
```

and then surround this whole block with 

```
try {
     ....
} catch(RuntimeException e) {
   return CompletableFuture.failedFuture(e);
}
```

jonahgraham:
A bunch of effort was put in during #2051 to convert the terminal's own Async API (ITerminalService.Done and friends) with the standard Java concurrent API. However I don't see any code using the `CompletableFuture` that `executeServiceOperation` is returning. So I have done a quick version of this, but this code isn't correct because the returned CompletableFuture will never complete. ~I think what should be happening here is a throw RejectedExecutionException instead - new patch coming soon.~ <-- instead of this solution see https://github.com/eclipse-platform/eclipse.platform/pull/2163#discussion_r2352815118

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: I was writing https://github.com/eclipse-platform/eclipse.platform/pull/2163#issuecomment-3299187206 while you commented this - I will try what you said here.

jonahgraham: @laeubi I tried what you suggested - please let me know if that is what you intended.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -195,8 +195,18 @@ private CompletableFuture<?> executeServiceOperation(final TerminalServiceRunnab
...
+				return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
+						/*
+						 * A special executor is used here because the default Display executor will
+						 * do a syncExec here, but isExecuteAsync is true, so we need to return from
+						 * executeServiceOperation *before* we execute the runnable.
+						 */
+						Display.getDefault()::asyncExec);
+			} catch (RuntimeException e) {
+				return CompletableFuture.failedFuture(e);
+			}
```
laeubi: This must be moved a bit further so it covers the whole method.

Assume Line 211 throws an exception then it will bubble up here, while it is more semantic to be reported through the `CompletableFuture`, so callers will never get a (direct) exception here.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -193,12 +193,21 @@ protected final CompletableFuture<?> executeServiceOperation(final Map<String, O
...
-		} else {
-			return CompletableFuture.completedFuture(runnable.run(tvid, title, connector, data));
+		try { // Execute the operation
+			if (runnable.isExecuteAsync()) {
+				return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
+						/*
+						 * A special executor is used here because the default Display executor will
+						 * do a syncExec here, but isExecuteAsync is true, so we need to return from
+						 * executeServiceOperation *before* we execute the runnable.
+						 */
```
laeubi: I think this comment is here a bit misleading because it depends a bit on what thread calls this.

e.g. assume I call it from a "normal" thread, there is no guarantee that it executes *after* (or _before_) `executeServiceOperation` returns.

So I would simply explain here that `isExecuteAsync` means it should always be executed async in the UI thread.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: Done.

FWIW in practice this code is still overly complicated - nothing ever examines the completedfuture returned and `isExecuteAsync` always returns true in practice (in at least the code base I can see)

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: Done.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
laeubi: Its perfectly possible that now some code path become obsolete here, especially I see some call sites calling runtime exceptions (that never should happen with a `CompletableFuture` API anyways.

Also `TerminalServiceRunnable` now is internal API (and an implementation detail) so we can probably remove the method completely and always execute asynchron.

I would even argue the `IStatus` return value is usless (as most of the time we return OK anyways) and we better allow to throw a `CoreException` (that then can be passed to the caller as a failed future state).

The question is more how eager we are to further cleanup that area :-)

jonahgraham: > The question is more how eager we are to further cleanup that area :-)

I will do a quick second commit that removes `isExecuteAsync` from TerminalServiceRunnable and removes `run` return type in favour of throwing CoreException

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -68,6 +68,8 @@ public class TerminalService implements ITerminalService {
 
 	/**
 	 * Common terminal service runnable implementation.
+	 *
+	 * These runnables are run asynchronously on the UI thread.
 	 */
 	protected static abstract class TerminalServiceRunnable {
```
laeubi:
```suggestion
	private static interface TerminalServiceRunnable {
```

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -193,12 +185,17 @@ protected final CompletableFuture<?> executeServiceOperation(final Map<String, O
...
-			return CompletableFuture.supplyAsync(() -> runnable.run(tvid, title, connector, data),
-					Display.getDefault());
-		} else {
-			return CompletableFuture.completedFuture(runnable.run(tvid, title, connector, data));
+		try { // Execute the operation
+			return CompletableFuture.runAsync(() -> {
+				try {
+					runnable.run(tvid, title, connector, data);
+				} catch (CoreException e) {
+					throw new CompletionException(e);
```
laeubi, line 192: Interesting I didn't know that!

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -268,11 +266,12 @@ public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connect
 						ILog.get().log(e.getStatus());
```
laeubi: Should this exception really be suppressed here or should we retrow it?

If not I would suggest adding a message to the log e.g.

ILog.get().error("Can't restore view", e);

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -313,9 +307,8 @@ public CompletableFuture<?> closeConsole(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
 		return executeServiceOperation(properties, new TerminalServiceRunnable() {
 			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
+			public void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
```
laeubi: could become a lamda here

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -326,9 +319,8 @@ public CompletableFuture<?> terminateConsole(Map<String, Object> properties) {
 
 		return executeServiceOperation(properties, new TerminalServiceRunnable() {
 			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
+			public void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
```
laeubi: same here, can be a lamda now

laeubi (approved):
It all looks good to me, just some smaller suggestions.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: Done in next push.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: Yes, once I turned TerminalServiceRunnable into an interface.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: I'm going to do the logging of the error for now - I may do an additional commit to address all the lost exceptions by handling the completed futures

jonahgraham: I added another new commit that logs all errors instead of silently discarding them.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
```diff
@@ -73,9 +74,14 @@ private void executeDelegate(ISelection selection, ILauncherDelegate delegate) t
...
-			delegate.execute(properties);
+			delegate.execute(properties).whenComplete((r, e) -> {
+				if (e != null) {
+					ILog.get().error("Error occurred while running delegate to open console", e); //$NON-NLS-1$
+				}
+			});
 		} catch (Exception e) {
-			throw new ExecutionException(e.getMessage(), e);
+			// This wraps any synchronous exceptions - asynchronous exceptions are logged in the whenComplete above
+			throw new ExecutionException("Error occurred while running delegate to open console", e); //$NON-NLS-1$
```
laeubi: I think this is no longer needed, there should be no exceptions anymore, same below.

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
```diff
@@ -73,9 +74,14 @@ private void executeDelegate(ISelection selection, ILauncherDelegate delegate) t
 		properties.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegate.getId());
 		properties.put(ITerminalsConnectorConstants.PROP_SELECTION, selection);
 		try {
-			delegate.execute(properties);
+			delegate.execute(properties).whenComplete((r, e) -> {
+				if (e != null) {
+					ILog.get().error("Error occurred while running delegate to open console", e); //$NON-NLS-1$
+				}
+			});
```
laeubi: if you are only interested in errors you can use `exceptionally`

```suggestion
			delegate.execute(properties).exceptionally(e -> 
			    ILog.get().error("Error occurred while running delegate to open console", e); //$NON-NLS-1$
			    );
```

same blow at the other places.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -311,11 +302,11 @@ private IStatus doRun(TerminalViewId tvid, String title, ITerminalConnector conn
 	@Override
 	public CompletableFuture<?> closeConsole(Map<String, Object> properties) {
 		Assert.isNotNull(properties);
-		return executeServiceOperation(properties, new TerminalServiceRunnable() {
-			@Override
-			public IStatus run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data) {
-				consoleViewManager.closeConsole(tvid, title, connector, data);
-				return Status.OK_STATUS;
+		var future = executeServiceOperation(properties,
```
laeubi: I find var a bit hard to read often because you need to now derive the type of the variable and as we do not have a complex or varing eeturn type I would better just use 
```suggestion
		CompletableFuture<?> future = executeServiceOperation(properties,
```

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
jonahgraham: I don't think you can (cleanly) use exceptionally because exceptionally takes a Function, where as whenComplete takes a BiConsumer

Concretely, with your code change (after removing the extra `;`) you get error `Cannot return a void result`.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
jonahgraham: I just realized I made a mistake and should not have added the whenComplete in this location at all because we should leave the caller to handle the exceptions since we are returning future here.

BTW TerminalService's `closeConsole` and `terminateConsole` are not actually called from any code I know of

terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherHandler.java
jonahgraham: I agree - I left it there because I thought it was there for some specific reasons. I added better message as `ExecutionException` are shown to users, but looking through the history I agree with your assessment. I think the catch shouldn't be there.

@ruspl-afed as you recently wrote this code, if @laeubi or I have overlooked something please let me know.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/handler/LaunchTerminalCommandHandler.java
```diff
@@ -172,11 +173,11 @@ private void executeDelegate(ISelection selection, ILauncherDelegate delegate) t
 	}
 
 	private void executeDelegate(Map<String, Object> properties, ILauncherDelegate delegate) throws ExecutionException {
-		try {
-			delegate.execute(properties);
-		} catch (Exception e) {
-			throw new ExecutionException(e.getMessage(), e);
-		}
+		delegate.execute(properties).whenComplete((r, e) -> {
```
laeubi, line 176:
```suggestion
		delegate.execute(properties).exceptionally(e -> {
```

then you can be sure it only triggers on error.

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/TerminalService.java
```diff
@@ -78,21 +80,11 @@ protected static abstract class TerminalServiceRunnable {
...
-		 * modify the behavior.
-		 *
-		 * @return <code>True</code> to execute the runnable asynchronously, <code>false</code> otherwise.
+		 * @throws CoreException if operation does not complete successfully
+		 * @throws NullPointerException if any required arguments are null.
 		 */
-		public boolean isExecuteAsync() {
-			return true;
-		}
+		public abstract void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data)
```
laeubi:
```suggestion
		void run(TerminalViewId tvid, String title, ITerminalConnector connector, Object data)
```

laeubi (commented):
It looks sane to me!

terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/handler/LaunchTerminalCommandHandler.java
jonahgraham: I think I can't make this change for the same reason as https://github.com/eclipse-platform/eclipse.platform/pull/2163#discussion_r2355500278 - Let me know if I am misunderstanding something

PS. notifications from GitHub are behaving badly on this PR as some notifications are arriving out of order in my inbox.

laeubi: > I don't think you can (cleanly) use exceptionally because exceptionally takes a Function, where as whenComplete takes a BiConsumer
> Concretely, with your code change (after removing the extra ;) you get error Cannot return a void result.

I somehow missed that between the lines, just return `null` as a result from the function (it does not really matters).

This is only important when it comes to chaining, because `exceptionally` "handle" the exception and returns a result.

jonahgraham: I don't think it is appropriate to use exceptionally - whenComplete is specifically designed to observe without changing results and is generally recommended to be used when logging (references [1](https://mincong.io/2020/05/30/exception-handling-in-completable-future/#logging), [2](https://medium.com/@prawin609/error-handling-in-completablefuture-understanding-exceptionally-handle-and-whencomplete-5cd9d8bcc8a8#:~:text=Differences%20Between%20exceptionally()%2C%20handle()%2C%20and%20whenComplete())).

However I don't care that much, and if you want it to use exceptionally with a dummy return (that is what my "cleanly" was about) then let me know and I will make the change.

ruspl-afed:
Now it is all green. Thank you @jonahgraham for working on this

jonahgraham:
Thanks @ruspl-afed for rebasing and getting me all green ticks on this.

I plan to rebase and merge this soon if there is no further comment.

PR #2162, opened 2025-09-15
org.eclipse.core.contenttype: osgi.extender=osgi.component requirement

sratz:
org.eclipse.core.contenttype declares a service component, so it should also declare the requirement to the service extender.

Otherwise, this can lead to exceptions such as the one below, in case no extender is added.

This also adjusts org.eclipse.core.tests.harness accordingly, which builds a custom OSGI config.ini in CustomSessionConfigurationImpl and must now also include a SCR bundle (and its dependencies). We use org.apache.felix.scr here.

This serves also as a preparation for
https://github.com/eclipse-tycho/tycho/pull/5362

```
org.osgi.framework.BundleException: Exception in org.eclipse.core.resources.ResourcesPlugin.start() of bundle org.eclipse.core.resources.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:847)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:768)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1066)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:394)
	at org.eclipse.osgi.container.Module.doStart(Module.java:643)
	at org.eclipse.osgi.container.Module.start(Module.java:500)
	at org.eclipse.osgi.framework.util.SecureAction.start(SecureAction.java:528)
	at org.eclipse.osgi.internal.hooks.EclipseLazyStarter.postFindLocalClass(EclipseLazyStarter.java:122)
	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.findLocalClass(ClasspathManager.java:622)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.findLocalClass(ModuleClassLoader.java:353)
	at org.eclipse.osgi.internal.loader.BundleLoader.findLocalClass(BundleLoader.java:417)
	at org.eclipse.osgi.internal.loader.sources.SingleSourcePackage.loadClass(SingleSourcePackage.java:41)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass0(BundleLoader.java:521)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:438)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:195)
[...]

PR #2165, opened 2025-09-17
Test Harness: Session Tests: List felix.scr dependencies explicitly

sratz:
* We don't need the optional ones

* We should not install org.eclipse.osgi again, as this is already set as the 'osgi.framework' in the config.ini

It's easier to just list the few dependencies we really want explicitly.

Another follow-up on #2162.
Addresses https://github.com/eclipse-equinox/equinox/issues/1146.


HeikoKlare:
Please note that `ConfigurationSessionTestSuite` is actually outdated (I should probably mark it `deprecated`) and should be removed as soon as possible to avoid the duplicate maintenance efforts. The only remaining consumer is Equinox, for which I have implemented the according migration started with https://github.com/eclipse-equinox/equinox/pull/654, which is still pending.

sratz:
> Please note that `ConfigurationSessionTestSuite` is actually outdated (I should probably mark it `deprecated`) and should be removed as soon as possible to avoid the duplicate maintenance efforts. The only remaining consumer is Equinox, for which I have implemented the according migration started with [eclipse-equinox/equinox#654](https://github.com/eclipse-equinox/equinox/pull/654), which is still pending.

Ah, thanks @HeikoKlare! That explains why I missed adjusting one of them (besides not looking thoroughly enough) in my initial change to include SCR.

Issue #2167, opened 2025-09-23
Launch Group never terminates when the last configuration is disabled


pdeville2:
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

* create a new basic java main
* create a run configuration "java application"
* copy many times this configuration
* create a group with all config as "wait until terminated"
* disable the last configuration

<img width="795" height="439" alt="Image" src="https://github.com/user-attachments/assets/35cfeb3b-c156-43d7-95f5-1fa6743a0667" />

When running the group, the groupe never terminate even when all configuration are terminated : 

<img width="473" height="237" alt="Image" src="https://github.com/user-attachments/assets/e024c176-790c-40b5-9660-c7a1235e0b4d" />

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

pdeville2:
Hello, 

Thanks you for merging my patch before. 

However I make a mistake in it :
if the number of configuration Is bigger than the number of enabled configuration, then `i` is not the good reference point, since `i` iterate over the entire list of configuration. I propose to iterate over only enabled configurations.

I will submit a new patch for it.

I am sorry for this mistake.

iloveeclipse:
Fixed via https://github.com/eclipse-platform/eclipse.platform/pull/2169

PR #2168, opened 2025-09-23
GH-2167 Launch Group never terminates when the last configuration is disabled

pdeville2:
Fix issue
https://github.com/eclipse-platform/eclipse.platform/issues/2167


mickaelistria:
Thank you!

PR #2169, opened 2025-09-23
GH-2167 Launch Group never terminates when many configurations are disabled

pdeville2:
Fix issue
https://github.com/eclipse-platform/eclipse.platform/issues/2167

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
@@ -98,7 +98,7 @@ public class GroupLaunchConfigurationDelegate extends LaunchConfigurationDelegat
 			List<GroupLaunchElement> launches = createLaunchElements(groupConfig);
 			List<GroupLaunchElement> enabledLaunches = launches.stream().filter(l -> l.enabled).toList();
 			for (int i = 0; i < enabledLaunches.size(); ++i) {
-				GroupLaunchElement le = launches.get(i);
+				GroupLaunchElement le = enabledLaunches.get(i);
 
 				// find launch; if not found, skip (error?)
 				final ILaunchConfiguration conf = findLaunchConfiguration(le.name);


iloveeclipse:
PR still in draft state?

pdeville2:
Sorry for that I did not pay attention to it. It is now mark as ready.

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
```diff
@@ -96,14 +96,10 @@ public void launch(ILaunchConfiguration groupConfig, String mode, final ILaunch
 			SubMonitor progress = SubMonitor.convert(monitor, NLS.bind(DebugCoreMessages.GroupLaunchConfigurationDelegate_Launching, groupConfig.getName()), 1000);
 
 			List<GroupLaunchElement> launches = createLaunchElements(groupConfig);
-			int nbEnabledLaunches = launches.stream().filter(l -> l.enabled).toList().size();
-			for (int i = 0; i < launches.size(); ++i) {
+			List<GroupLaunchElement> enabledLaunches = launches.stream().filter(l -> l.enabled).toList();
+			for (int i = 0; i < enabledLaunches.size(); ++i) {
```
mickaelistria, line 100: This patch is good, however it seems to me that just replacing `launches.sized()` by `nbEnabledLaunch` would be equivalent, wouldn't it? And if so, is any actually better?

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
```diff
@@ -96,14 +96,10 @@ public void launch(ILaunchConfiguration groupConfig, String mode, final ILaunch
...
 				GroupLaunchElement le = launches.get(i);
 
-				if (!le.enabled) {
-					continue;
-				}
-
```
mickaelistria, line 106: Why is this now discareded?

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
```diff
@@ -96,14 +96,10 @@ public void launch(ILaunchConfiguration groupConfig, String mode, final ILaunch
 			SubMonitor progress = SubMonitor.convert(monitor, NLS.bind(DebugCoreMessages.GroupLaunchConfigurationDelegate_Launching, groupConfig.getName()), 1000);
 
 			List<GroupLaunchElement> launches = createLaunchElements(groupConfig);
-			int nbEnabledLaunches = launches.stream().filter(l -> l.enabled).toList().size();
-			for (int i = 0; i < launches.size(); ++i) {
+			List<GroupLaunchElement> enabledLaunches = launches.stream().filter(l -> l.enabled).toList();
+			for (int i = 0; i < enabledLaunches.size(); ++i) {
 				GroupLaunchElement le = launches.get(i);
```
mickaelistria: Should it be `enabledLaunches.get(i)` instead?

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
pdeville2: Because if the rest of the patch is well done then we loop into only enabled launches.

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
pdeville2: yes it should of course,

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
pdeville2: yes this patch is in fact not complete as you mention later so it does nothing sorry for the pour quality.

pdeville2:
thank you for the review and sorry for the quality of the patch,  indeed we were not iterate over the good list.

PR #2171, opened 2025-09-25
Update Skip Breakpoints text values

SougandhS:


SougandhS:
cc @merks

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DebugUIMessages.properties
```diff
@@ -289,6 +289,6 @@ ExpressionManagerContentProvider_1=Add new expression
 
 LaunchShortcutAction_combineLaunchShortcutName={0} ({1})
 
-skipBreakpointWarningTitle = Breakpoints set to skip
-skipBreakpointWarningLabel = Breakpoints are disabled due to 'Skip All Breakpoints' being active. Debugger will continue without stopping.
+skipBreakpointWarningTitle = Breakpoints Set To Skip
```
merks: Title case is a little bit picky, so it's not simply make every word upper case.

Even here there are different rules!

https://titlecaseconverter.com/


```suggestion
skipBreakpointWarningTitle = Breakpoints Set to Skip
```

SougandhS: Updated 👍

SougandhS:
If there are no more changes can this be merged ? I need an IBuild to grab a screenshot in light theme
because of https://github.com/eclipse-platform/eclipse.platform/pull/2093#issuecomment-3135119781

merks (approved):

SougandhS:
Thanks 👍

PR #2174, opened 2025-09-30
Eliminate the nested platform.jar from org.eclipse.platform

merks:
- I believe this remains like this for historical purposes when there was also a startup.jar in that bundle.


merks:
From generating SBOMs I noticed that two bundle provide nested jars:

image


From this file I am assuming there was previous a separate launcher jar in this bundle:

https://github.com/eclipse-platform/eclipse.platform/tree/master/platform/org.eclipse.platform/src#readme

But now there are no other classes and I can think of no reason why the two classes should be bundled in a separate nested jar.

merks:
FYI, I tested the build this morning and this appears to have caused no problems.

PR #2176, opened 2025-10-01
Add option to show import configuration in launch tool bar

SougandhS:
Previously, the toolbar only provided an option to export launch configurations. New users often found it difficult to locate the import functionality quickly.  This PR adds an "Import Launch Configurations" option to the toolbar, 
making it consistent with the existing export option.

image


SougandhS:
Hi @laeubi, Could you please check this PR when you are available ?

laeubi:
The icons are a bit hard to distinguish but this could be done in a separate PR.

SougandhS:
Thank you @laeubi 
>The icons are a bit hard to distinguish but this could be done in a separate PR.

Sure I'll redesign it 👍

PR #2184, opened 2025-10-03
Fix IllegalArgumentException with duplicate natures

laeubi:
Changed ProjectDescription.hasPrivateChanges() to use HashSet instead of Set.of() to handle duplicate natures gracefully. Set.of() throws IllegalArgumentException on duplicates, but HashSet tolerates them and correctly compares nature sets.

Also added test case testDuplicateNatures() to ensure this regression doesn't occur again.

Derived from here:
- https://github.com/eclipse-platform/eclipse.platform/pull/2183

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectDescription.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ProjectDescription.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


laeubi:
ECA check failure is due:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6562
 
and currently unavoidable, I plan to merge anyways if build succeeds here as it is the same solution that was suggested by @merks and its more a technical detail. If there will be ever concerns regarding AI generated content, the only new thing is the test what then can be deleted without any problem.

merks:
The content is simple, follows the original suggestion of mine in the issue, and the test case is mechanical in nature.

PR #2187, opened 2025-10-05
Convert runtime performance test suite from JUnit3 to JUnit5

vogella:
This commit converts AllPerfTests.java from JUnit3 to JUnit5:
- Converted from JUnit3 TestCase with suite() method to JUnit5 @Suite/@SelectClasses
- Replaced static Test suite() method with @TestFactory for dynamic performance tests
- Preserved the original behavior for warming up startup tests
- Preserved headless startup test configuration with eclipse.activateRuntimePlugins=false
- Preserved UI performance test integration
- Maintained complex ContentTypePerformanceTest suite behavior via dynamic test execution

The conversion maintains all original test functionality while modernizing to JUnit5:
- BenchPath.class and PreferencePerformanceTest.class are included via @SelectClasses
- Complex performance session tests are handled via @TestFactory and DynamicTest
- Error handling for setup failures is preserved using RuntimeException wrapping

All tests (395) continue to pass after the conversion with 4 skipped tests unchanged from the original behavior.

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/perf/AllPerfTests.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/perf/AllPerfTests.java
@@ -13,14 +13,29 @@
  *******************************************************************************/
 package org.eclipse.core.tests.runtime.perf;
 
-import junit.framework.*;
 import org.eclipse.core.tests.runtime.RuntimeTestsPlugin;
-import org.eclipse.core.tests.session.*;
+import org.eclipse.core.tests.session.PerformanceSessionTestSuite;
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
I don't think this kind of migration makes sense. `SessionTestSuite` is outdated (as it relies on JUnit 3/4 concepts). After this change, the test class still relies on `SessionTestSuite`, which is not a good way forward. All existing `SessionTestSuite` usages (except for usages of the specialized `ConfigurationSessionTestSuite` in Equinox https://github.com/eclipse-equinox/equinox/pull/654) have been replaced with a JUnit 5 concept based on the added `SessionTestExtension` and related functionality. Instead of introducing another way of migrating session tests to JUnit 5 and preserving dependencies to JUnit 3/4 classes, the same pattern should be adopted here as well (or a different approach should be applied consistently everywhere).

PR #2197, opened 2025-10-14
Experiment with the new JUnit-Platform Mojo

laeubi:
This is an experiment in how much the new tycho junit-platform mojo can replace "traditional" tycho-surefire test and what are possibly missing pieces.

For this is:
- it creates a pom.xml so we have full control and do not need to modify the parent
- there we explicitly disable maven surefire
- configure the new mojo as an additional execution

### DIFF
--- a/team/tests/org.eclipse.compare.tests/META-INF/MANIFEST.MF
+++ b/team/tests/org.eclipse.compare.tests/META-INF/MANIFEST.MF
@@ -13,7 +13,9 @@ Require-Bundle: org.eclipse.compare,
  org.eclipse.core.tests.resources,
  org.eclipse.core.tests.harness,
  org.eclipse.core.filesystem,
- org.eclipse.team.ui
+ org.eclipse.team.ui,
+ org.eclipse.platform.testing,
+ org.eclipse.ui.ide.application;bundle-version="1.5.900"
 Import-Package: org.assertj.core.api,
  org.assertj.core.api.iterable,
  org.junit.jupiter.api;version="[5.14.0,6.0.0)",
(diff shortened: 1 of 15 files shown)


laeubi:
Executing this locally results in

```
Test run finished after 1315 ms
[        40 containers found      ]
[         0 containers skipped    ]
[        40 containers started    ]
[         0 containers aborted    ]
[        40 containers successful ]
[         0 containers failed     ]
[       259 tests found           ]
[         0 tests skipped         ]
[       259 tests started         ]
[         0 tests aborted         ]
[       222 tests successful      ]
[        37 tests failed          ]
```

because currently the "use UI harness" feature is not implemented / available. But as we can already see a lot of tests simply don't need that at all!

So next would be to evaluate if we can add the "missing" support with an extension here.

team/tests/org.eclipse.compare.tests/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,9 @@ Require-Bundle: org.eclipse.compare,
  org.eclipse.core.tests.resources,
  org.eclipse.core.tests.harness,
  org.eclipse.core.filesystem,
- org.eclipse.team.ui
+ org.eclipse.team.ui,
+ org.eclipse.platform.testing,
+ org.eclipse.ui.ide.application;bundle-version="1.5.900"
```
mickaelistria, line 18: Are those really necessary? Couldn't they be added as requirements in build.properties or pom.xml instead?

laeubi: This is currently a bit work in progress and I just wanted it to run, the first is likely not required and the second would better be an additional requirement in the target platform.

PR #2198, opened 2025-10-14
Improve Copying Watch Expressions 

SougandhS:
This PR improves the behavior when copying a watch expression, preventing unwanted UI label text from being included in the clipboard. It ensures that only the intended expression is copied, enhancing usability and providing a cleaner, more consistent debugging experience.

(disabled)

Currently copying the expressions will generate this result in Clipboard - Making users to manually remove the labels
```
stringVal + " a "	sadsd a	
stringVal + " a "	(disabled)	
invalidS + "e"	<error(s)_during_the_evaluation> (disabled)	
spd2	(pending)	
"(disabled)"	(disabled)	
Add new expression		

```

With this change - 
```
stringVal + " a "	sadsd a	
stringVal + " a "		
invalidS + "e"		
spd2		
"(disabled)"		
Add new expression	
```

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2017 Bachmann electronic GmbH and others.
+ *  Copyright (c) 2017, 2025 Bachmann electronic GmbH and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -10,15 +10,17 @@
  *
  *  Contributors:
  *     Bachmann electronic GmbH - initial API and implementation
+ *     IBM Corporation - Exclude Expression UI values while copying
  *******************************************************************************/
 package org.eclipse.debug.internal.ui.actions.expressions;
 
+import org.eclipse.debug.internal.core.WatchExpression;
+import org.eclipse.debug.internal.ui.DebugUIMessages;
 import org.eclipse.debug.internal.ui.viewers.model.VirtualCopyToClipboardActionDelegate;
 
 public class CopyExpressionsToClipboardActionDelegate extends VirtualCopyToClipboardActionDelegate {
 
 	private static final String QUOTE = "\""; //$NON-NLS-1$
-
 	@Override
 	protected String trimLabel(String rawLabel) {
 		String label = super.trimLabel(rawLabel);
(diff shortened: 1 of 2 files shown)


SougandhS:
Hi @laeubi,
Could you please check this PR when you are available ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
```diff
@@ -34,5 +36,31 @@ protected String trimLabel(String rawLabel) {
...
+				if (label.equals(DebugUIMessages.DefaultLabelProvider_12)
+						|| label.contains(DebugUIMessages.DefaultLabelProvider_13)) {
+					return null;
+				}
+			}
+			if (!watchExp.isEnabled()) {
+				if (label.equals(DebugUIMessages.DefaultLabelProvider_15)) {
+					return null;
+				}
+				if (label.contains(DebugUIMessages.DefaultLabelProvider_15)) {
```
merks, line 52: This guard isn't actually needed because index will become -1 anyway which you test at line 54.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
```diff
@@ -34,5 +36,31 @@ protected String trimLabel(String rawLabel) {
...
+					int index = label.lastIndexOf(DebugUIMessages.DefaultLabelProvider_15);
+					if (index != -1) {
+						label = label.substring(0, index)
+								+ label.substring(index + DebugUIMessages.DefaultLabelProvider_15.length());
+					}
+				}
+			}
+			return label;
```
merks, line 60: This is the same return as below so it's redundant.

SougandhS:
Thanks a lot !

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
```diff
@@ -34,5 +36,31 @@ protected String trimLabel(String rawLabel) {
...
+			}
+			if (!watchExp.isEnabled()) {
+				if (label.equals(DebugUIMessages.DefaultLabelProvider_15)) {
+					return null;
+				}
+				if (label.contains(DebugUIMessages.DefaultLabelProvider_15)) {
+					int index = label.lastIndexOf(DebugUIMessages.DefaultLabelProvider_15);
+					if (index != -1) {
+						label = label.substring(0, index)
+								+ label.substring(index + DebugUIMessages.DefaultLabelProvider_15.length());
```
merks, line 56: If I understand correctly, this label will be at the end of the string and the value at line 56 is always the empty string, right?

image

laeubi:
@SougandhS please investigate the remarks from @merks it seems I was to eagerly merging this :-\

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
SougandhS: the if at line 49 is for this particular case 
image


the if at 52 is for cases of 

image

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
SougandhS: yes, label will always be in the end and 56 will be empty string

SougandhS:
I have addressed the changes in https://github.com/eclipse-platform/eclipse.platform/pull/2202

PR #2202, opened 2025-10-19
Remove redundant return and additional empty string calculation

SougandhS:


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/CopyExpressionsToClipboardActionDelegate.java
```diff
@@ -52,12 +52,10 @@ protected String exludeLabels(Object itemData, String label) {
 				if (label.contains(DebugUIMessages.DefaultLabelProvider_15)) {
```
merks: I think this test can be removed.

SougandhS: The idea was to simply `return null` if the expression started as disabled, I guess what you proposed is better

merks: Yes, but `label.lastIndexOf(DebugUIMessages.DefaultLabelProvider_15)` also does a "contains" test that will be -1 so it seems to me there are two guards.  So the early null seems okay but the contains test is redundant:

image

SougandhS: Thanks, Updated it now 👍

merks (approved):
Very good.  Thanks.

SougandhS:
Thank you @merks for checking this changes again and helping in optimising further

PR #2203, opened 2025-10-19
Add Support for Expression Context Pinning

SougandhS:
This PR introduces a new feature in the Expressions view that allows expressions to be pinned to a specific stack frame, ensuring their values remain visible and consistent even when switching between threads or stack frames or even launches. This enhancement improves the debugging experience by preventing frequent `<error on evaluation>` issues (if one wishes to see the variables/expressions again) and enables developers to track and compare expressions or variables (by watch) across different contexts more effectively.

image
<br><br>
Pinned expressions will be highlighted with Bold and italics 
image
<br><br>
This can be removed/unpinned manually or will be removed/unpinned if context not available
image

https://github.com/user-attachments/assets/63e68e0a-3a8a-4ab2-8bd3-951098a9cd33

(apologies for the poor video quality - had to compress the size below 10 MB)

### DIFF
--- a/debug/org.eclipse.debug.core/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.core; singleton:=true
-Bundle-Version: 3.23.200.qualifier
+Bundle-Version: 3.24.0.qualifier
 Bundle-Activator: org.eclipse.debug.core.DebugPlugin
 Bundle-Vendor: %providerName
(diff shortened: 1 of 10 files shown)


mickaelistria (changes requested):
I don''t think the workflow demonstrated in the video is satisfying. It has many usability quirks:
* the feedback that the value is pinned to a particular context (only bold) is not explicit; The context on which the evaluation is applied should be made more explicit
* When an expression is pinned to a context, it's easily possible to get the value from another context
* The fact that we can't see the values side-by-side makes it complex to compare different complex values

SougandhS:
> * the feedback that the value is pinned to a particular context (only bold) is not explicit; The context on which the evaluation is applied should be made more explicit

Additional label can be provided 👍 

> * When an expression is pinned to a context, it's easily possible to get the value from another context

But in different launches it will be difficult.

PR #2209, opened 2025-10-23
Fix Job.schedule() not rescheduling after cancel()

laeubi:
The aboutToRunCanceled flag was not being cleared when reschedule=true, causing jobs to be canceled instead of scheduled. This fix moves the flag clearing to after the early returns, ensuring it's always cleared when actually scheduling the job.

Test verifies that Job.schedule() works correctly after cancel() when called on a running job. This reproduces the issue where the aboutToRunCanceled flag was not being cleared during reschedule.

FYI @basilevs 

Maybe we can build on this?

### DIFF
--- a/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.core.jobs/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.jobs; singleton:=true
-Bundle-Version: 3.15.700.qualifier
+Bundle-Version: 3.15.800.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.core.internal.jobs;x-internal:=true,


basilevs:
@iloveeclipse has pointed out, that job seems to start despite `getState()` we will have to reevaluate, what are we fixing here and why.

I had to create #2210 formulated in terms of `Job.join()` instead. `Job.join()` turns out to have no guarantees because it is documented in terms of `Job.getState()` internal API, which provides no guarantees or contracts.

Therefore the solution might be to fix Javadoc for `Job.join()` to indicate, that it has no contract, or invent one and adapt `Job.join()` to follow it.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/Bug_550738.java
```diff
@@ -134,4 +134,41 @@ public void testReportDoneOncePerSchedule() throws InterruptedException {
...
+						startedLatch.await(5, TimeUnit.SECONDS));
+
+				// While job is running, cancel it and then immediately schedule it again
+				Thread.sleep(i);
+				job.cancel();
+				job.schedule();
+
+				// The job should reschedule and not remain in NONE state
+				// Wait a bit to ensure the rescheduling happens
+				Thread.sleep(50);
```
basilevs, line 164: This is not enough to reproduce a problem with `getState()`  returning spurious `Job.NONE`. Why has this test failed in the first place?

runtime/bundles/org.eclipse.core.jobs/src/org/eclipse/core/internal/jobs/JobManager.java
```diff
@@ -1498,6 +1495,10 @@ protected boolean scheduleInternal(InternalJob job, long delay, boolean reschedu
 		if (job.internalGetState() != Job.NONE) {
 			return false;
 		}
+		// Clear the about to run canceled flag when actually scheduling.
+		// A new explicit schedule() call should override any previous cancel() call.
+		// See https://github.com/eclipse-platform/eclipse.platform/issues/160
+		job.setAboutToRunCanceled(false);
```
basilevs, line 1501: I've managed to apply this delta and `org.eclipse.core.tests.runtime.jobs.Bug_550738.testCancelSchedule()` has started to fail. Does Copilot run any tests?

PR #2216, opened 2025-10-28
Do not enable Link Prototypes if there are no prototypes

SougandhS:
This PR disables the `Link Prototypes` context menu in launch configurations when no prototypes are defined for the selected launch type.

Previously, selecting this option displayed an empty list when no prototypes were available.
<br>
image

image

<br><br>

With this change, the context menu option will only be enabled if at least one prototype is defined for the selected launch type, preventing users from seeing an empty list.

image
Screenshot 2025-10-28 at 5 46 23 PM
image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
@@ -13,9 +13,6 @@
  *******************************************************************************/
 package org.eclipse.debug.internal.ui.launchConfigurations;
 
-import java.util.Collection;
-import java.util.HashSet;
-
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationType;
@@ -104,34 +101,37 @@ public class LinkPrototypeAction extends AbstractLaunchConfigurationAction {
 	 */
 	@Override
 	protected boolean updateSelection(IStructuredSelection selection) {
-		// Enable action only if launch configuration(s) of the same type
-		// is(are) selected and the launch configuration type allows prototypes
-		Collection<ILaunchConfigurationType> launchConfigurationTypes = new HashSet<>();
+		if (selection.isEmpty()) {
+			return false;
+		}
+		ILaunchConfigurationType type = null;
 		for (Object object : selection.toList()) {
-			if (object instanceof ILaunchConfiguration) {
-				if (((ILaunchConfiguration) object).isPrototype()) {
+			if (!(object instanceof ILaunchConfiguration launchConfig)) {
+				return false;
+			}
+
+			if (launchConfig.isPrototype()) {
+				return false;
+			}
+
+			ILaunchConfigurationType currentType;
+			try {
+				currentType = launchConfig.getType();
+				if (currentType.getPrototypes().length == 0) {
 					return false;
-				} else {
-					ILaunchConfigurationType type = null;
-					try {
-						type = ((ILaunchConfiguration) object).getType();
-					} catch (CoreException e) {
-						DebugUIPlugin.log(e.getStatus());
-					}
-					if (type != null) {
-						launchConfigurationTypes.add(type);
-					} else {
-						return false;
-					}
 				}
-			} else {
+			} catch (CoreException e) {
+				DebugUIPlugin.log(e.getStatus());
+				return false;
+			}
+
+			if (type == null) {
+				type = currentType;
+			} else if (!type.equals(currentType)) {
 				return false;
 			}
 		}
-		if (launchConfigurationTypes.size() == 1) {
-			return launchConfigurationTypes.iterator().next().supportsPrototypes();
-		}
-		return false;
+		return true;
 	}
 
 	@Override


SougandhS:
Hi @vogella, could you please check this PR when you get some time ?

SougandhS:
Hi @merks, could you please merge this small change when you get a moment ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
```diff
@@ -128,8 +128,13 @@ protected boolean updateSelection(IStructuredSelection selection) {
 				return false;
 			}
 		}
-		if (launchConfigurationTypes.size() == 1) {
-			return launchConfigurationTypes.iterator().next().supportsPrototypes();
+
+		if (launchConfigurationTypes.size() == 1 && type != null) {
```
merks: I think it's not possible for `type == null` `if launchConfigurationTypes.size()== 1` because you only add non-null types this list.

merks (changes requested):

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
SougandhS: Yes, but compiler was throwing warnings there, so only added the null check
image

SougandhS: To avoid unnecessary iterations for getting the `type`, I went with this approach

merks: Let's get another opinion. @akurtakov  @laeubi

laeubi: I think the compiler is confused by the hard to understand logic here.

From the description I think you actually not need the set here at all and I would not premature optimize this, we can expect this is not containing millions of elements so splitting up the logic into different steps would help e.g. you can early exit if the selection is empty.

Then the if/else is unnecessarily nested (as you immediately return from the method in the case).

Next your check for `null` can become

```
if (type == null || type == launchConfig.getType()) {
  type = launchConfig.getType();
} else {
 return false;
}
```

(and a core exception should probably be rethrown or also return false..)

iloveeclipse: What about this patch (applied on master):

```
diff --git a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
index e8d2a66..7f1a9d2 100644
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
@@ -15,5 +15,2 @@
 
-import java.util.Collection;
-import java.util.HashSet;
-
 import org.eclipse.core.runtime.CoreException;
@@ -108,11 +105,12 @@
 		// is(are) selected and the launch configuration type allows prototypes
-		Collection<ILaunchConfigurationType> launchConfigurationTypes = new HashSet<>();
-		for (Object object : selection.toList()) {
-			if (object instanceof ILaunchConfiguration) {
-				if (((ILaunchConfiguration) object).isPrototype()) {
+		for (Object object : selection) {
+			if (object instanceof ILaunchConfiguration config) {
+				if (config.isPrototype()) {
 					return false;
 				} else {
-					ILaunchConfigurationType type = null;
 					try {
-						type = ((ILaunchConfiguration) object).getType();
+						ILaunchConfigurationType type = config.getType();
+						if (type != null) {
+							return type.supportsPrototypes() && type.getPrototypes().length > 0;
+						}
 					} catch (CoreException e) {
@@ -120,7 +118,2 @@
 					}
-					if (type != null) {
-						launchConfigurationTypes.add(type);
-					} else {
-						return false;
-					}
 				}
@@ -130,5 +123,2 @@
 		}
-		if (launchConfigurationTypes.size() == 1) {
-			return launchConfigurationTypes.iterator().next().supportsPrototypes();
-		}
 		return false;
```

merks: There is a comment about all the selection being of the same type so I think this maybe misses that point by returning early.  Or?

SougandhS: @iloveeclipse with your patch, one of the launch config will be removed if different launch configs are selected accidently by the user - and do a prototype selection

https://github.com/user-attachments/assets/aea318fc-05e5-4b45-88e2-634cd9c99a75

SougandhS: Option should show only if its a same launch config type

SougandhS: I have optimised the code bit more

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LinkPrototypeAction.java
```diff
@@ -104,34 +101,37 @@ protected void performAction() {
...
+				type = currentType;
+			} else if (!type.equals(currentType)) {
 				return false;
 			}
 		}
-		if (launchConfigurationTypes.size() == 1) {
-			return launchConfigurationTypes.iterator().next().supportsPrototypes();
+
+		try {
+			return type != null && type.getPrototypes().length > 0;
```
merks: I hate to pick things to death, but if this test were done after line 116, you could return early at that point, there would be one less try/catch block, and if you then get to the end of the loop, you can return true (because line 125 will return false when all the types are not the same).  Or?

SougandhS: >I hate to pick things to death, but if this test were done after line 116, you could return early at that point, there would be one less try/catch block

```
currentType = launchConfig.getType();
return currentType != null && currentType.getPrototypes().length > 0;
```
I tried this but got the same issue as  https://github.com/eclipse-platform/eclipse.platform/pull/2216#discussion_r2671882315 :(

merks: The point is to test `if currentType.getPrototypes().length == 0` and return false early in that case, not to return true early for any other reason.  We of course must get through the whole loop before we can know to return true ever.

SougandhS: Thanks, Updated 👍

merks:
I will assume that all reviewers are happy and will merge shortly.

SougandhS:
Thanks for the review!

PR #2218, opened 2025-10-28
Use workspace root modification rule when removing all breakpoints

### LINKED ISSUE #2217: ResourceException Marker id not found when calling BreakpointManager.removeBreakpoints()
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=564872

During one of our tests we observed the following exception:

```
org.eclipse.core.internal.resources.ResourceException: Marker id 1332 not found.
	at org.eclipse.core.internal.resources.Marker.checkInfo(Marker.java:63)
	at org.eclipse.core.internal.resources.Marker.setAttribute(Marker.java:260)
	at org.eclipse.jdt.internal.debug.ui.JavaDebugOptionsManager$2.run(JavaDebugOptionsManager.java:894)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2292)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2317)
	at org.eclipse.jdt.internal.debug.ui.JavaDebugOptionsManager.updateBreakpointMessages(JavaDebugOptionsManager.java:906)
	at org.eclipse.jdt.internal.debug.ui.JavaDebugOptionsManager.breakpointsChanged(JavaDebugOptionsManager.java:965)
	at org.eclipse.debug.internal.core.BreakpointManager$BreakpointsNotifier.run(BreakpointManager.java:1104)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.debug.internal.core.BreakpointManager$BreakpointsNotifier.notify(BreakpointManager.java:1124)
	at org.eclipse.debug.internal.core.BreakpointManager.fireUpdate(BreakpointManager.java:975)
	at org.eclipse.debug.internal.core.BreakpointManager.access$4(BreakpointManager.java:962)
	at org.eclipse.debug.internal.core.BreakpointManager$BreakpointManagerVisitor.update(BreakpointManager.java:788)
	at org.eclipse.debug.internal.core.BreakpointManager.resourceChanged(BreakpointManager.java:699)
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:305)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.events.NotificationManager.notify(NotificationManager.java:295)
	at org.eclipse.core.internal.events.NotificationManager.broadcastChanges(NotificationManager.java:158)
	at org.eclipse.core.internal.resources.Workspace.broadcastBuildEvent(Workspace.java:366)
	at org.eclipse.core.internal.events.AutoBuildJob.doBuild(AutoBuildJob.java:157)
	at org.eclipse.core.internal.events.AutoBuildJob.run(AutoBuildJob.java:244)
!SUBENTRY 1 org.eclipse.core.resources 4 376 2020-06-30 21:21:07.013
```

Upon investigation, the exception coincides with a call to `org.eclipse.debug.internal.core.BreakpointManager.removeBreakpoints(IBreakpoint[], boolean)` from our test. The code in `BreakpointManager.removeBreakpoints()` deletes a breakpoint marker without specifying a rule, and so can cause race condition (and e.g. cause the error stack trace above).

As in [bug 540132](https://bugs.eclipse.org/bugs/show_bug.cgi?id=540132) (https://git.eclipse.org/c/jdt/eclipse.jdt.debug.git/commit/?id=897972b60a60c5bd9d0b8138d5be9b07addfcaf5), the breakpoints operation should specify a workspace modification rule if there is no matching outer rule.

To reproduce:

1. Debug Eclipse.
2. In the debuggee Eclipse, set a breakpoint at the `System.out.println()` in this snippet:
```
package test;

public class TestClass {

	public static void main1(String[] args) {
		System.out.println("hello world");
	}

}
```
3. Set a breakpoint at `org.eclipse.jdt.internal.debug.ui.JavaDebugOptionsManager.updateBreakpointMessages(IBreakpoint[])`, at:
```
marker.setAttribute(IMarker.MESSAGE, type + info);
```
4. Set a breakpoint at the start of: `org.eclipse.debug.internal.core.BreakpointManager.removeBreakpoints(IBreakpoint[], boolean)`
5. Change the `main` method name to `main1` (simply by typing, no need for refactor rename) and save. This hits the breakpoint at `JavaDebugOptionsManager.updateBreakpointMessages`.
6. Remove all breakpoints with the respective Breakpoints view action, this hits the breakpoint at `BreakpointManager.removeBreakpoint()`.
7. Resume first the thread in `BreakpointManager` (twice), then the thread in `JavaDebugOptionsManager`.

trancexpress:
When removing breakpoints, its possible that another job is accessing the breakpoint marker. This can result in an exception, since removing a breakpoint will also remove the marker.

This change moves breakpoint removal code in a job that holds a workspace root modification rule, if the current thread does not hold a matching rule.

Fixes: #2217

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/BreakpointManager.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/BreakpointManager.java
@@ -21,6 +21,7 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.Vector;
 
@@ -45,7 +46,9 @@ import org.eclipse.core.runtime.ListenerList;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.SafeRunner;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.MultiRule;
 import org.eclipse.debug.core.DebugException;
 import org.eclipse.debug.core.DebugPlugin;
 import org.eclipse.debug.core.IBreakpointListener;
@@ -387,7 +390,7 @@ public class BreakpointManager implements IBreakpointManager, IResourceChangeLis
 	 * Convenience method to get the workspace
 	 * @return the default {@link IWorkspace}
 	 */
-	private IWorkspace getWorkspace() {
+	private static IWorkspace getWorkspace() {
 		return ResourcesPlugin.getWorkspace();
 	}
 
@@ -527,7 +530,12 @@ public class BreakpointManager implements IBreakpointManager, IResourceChangeLis
 					}
 				}
 			};
-			getWorkspace().run(r, null, 0, null);
+			ISchedulingRule rule = MultiRule.combine(remove.stream()
+					.map(IBreakpoint::getMarker).filter(Objects::nonNull)
+					.map(IMarker::getResource).filter(Objects::nonNull)
+					.map(BreakpointManager::getMarkerRule)
+					.toArray(ISchedulingRule[]::new));
+			getWorkspace().run(r, rule, 0, null);
 		}
 	}
 
@@ -1404,6 +1412,10 @@ public class BreakpointManager implements IBreakpointManager, IResourceChangeLis
 		} catch (CoreException e) {
 			DebugPlugin.log(e);
 		}
+	}
+
+	private static ISchedulingRule getMarkerRule(IResource resource) {
+		return getWorkspace().getRuleFactory().modifyRule(resource);
 	}
 }
 
--- a/debug/org.eclipse.debug.core/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.core; singleton:=true
-Bundle-Version: 3.23.200.qualifier
+Bundle-Version: 3.23.300.qualifier
 Bundle-Activator: org.eclipse.debug.core.DebugPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin


trancexpress:
This revives the now gone: https://git.eclipse.org/r/c/platform/eclipse.platform.debug/+/165753

mickaelistria:
Would just passing the rule `MultiRule.combine(update.stream().map(getWorkspace()::markerRule).toArray(ISchedulingRule[]::new))` prevent from comcurrent access without blocking too much?

trancexpress:
> Would just passing the rule `MultiRule.combine(update.stream().map(getWorkspace()::markerRule).toArray(ISchedulingRule[]::new))` prevent from comcurrent access without blocking too much?

I'll check when I can, I'd have more confidence in the change that way.

If you mean `org.eclipse.core.resources.IResourceRuleFactory.markerRule(IResource)`, it returns `null` always?

trancexpress:
The `MultiRule` suggestion works, I wonder if we should use it also in `JavaDebugOptionsManager`... https://github.com/eclipse-jdt/eclipse.jdt.debug/commit/897972b60a60c5bd9d0b8138d5be9b07addfcaf5

When testing the change, I noticed its not enough to prevent the logged error. This code is removing the info:

```
"Worker-18: Remove Breakpoints" #131 [77058] prio=5 os_prio=0 cpu=418.64ms elapsed=1291.43s tid=0x00007f5d68000de0 nid=77058 at breakpoint [0x00007f5da91ee000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.core.internal.resources.MarkerManager.removeMarker(MarkerManager.java:513)
        at org.eclipse.core.internal.resources.Marker.delete(Marker.java:84)
        at org.eclipse.ui.ide.undo.AbstractMarkersOperation.deleteMarkers(AbstractMarkersOperation.java:117)
        at org.eclipse.ui.ide.undo.DeleteMarkersOperation.doExecute(DeleteMarkersOperation.java:61)
        at org.eclipse.debug.internal.ui.views.breakpoints.DeleteBreakpointMarkersOperation.doExecute(DeleteBreakpointMarkersOperation.java:48)
        at org.eclipse.ui.ide.undo.AbstractWorkspaceOperation.lambda$0(AbstractWorkspaceOperation.java:199)
        at org.eclipse.ui.ide.undo.AbstractWorkspaceOperation$$Lambda/0x00007f5e94bdc8c8.run(Unknown Source)
        at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2505)
        at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2533)
        at org.eclipse.ui.ide.undo.AbstractWorkspaceOperation.execute(AbstractWorkspaceOperation.java:199)
        at org.eclipse.core.commands.operations.DefaultOperationHistory.execute(DefaultOperationHistory.java:488)
        at org.eclipse.debug.ui.DebugUITools.deleteBreakpoints(DebugUITools.java:353)
        at org.eclipse.debug.internal.ui.actions.breakpoints.RemoveAllBreakpointsAction$1.run(RemoveAllBreakpointsAction.java:126)
```

So apparently there is more to change...

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/BreakpointManager.java
```diff
@@ -1405,5 +1413,9 @@ private void touchAllBreakpoints() {
 			DebugPlugin.log(e);
 		}
 	}
+
+	private static ISchedulingRule getMarkerRule(IResource resource) {
+		return getWorkspace().getRuleFactory().modifyRule(resource);
```
trancexpress, line 1418: I'm not sure which rule we should use here, as mentioned before in the PR `org.eclipse.core.internal.resources.Rules.markerRule(IResource)` returns `null`:

```
	/**
	 * Obtains the scheduling rule from the appropriate factory for a marker change operation.
	 */
	@Override
	public ISchedulingRule markerRule(IResource resource) {
		//team hook currently cannot change this rule
		return null;
	}
```

trancexpress:
> When testing the change, I noticed its not enough to prevent the logged error. This code is removing the info:

When testing again, I'm not running into this error. Maybe something changed...

Anyway, if we can, lets proceed here. If there are more sporadic marker errors, we can always open another issue.

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/BreakpointManager.java
iloveeclipse: Can you please add this to the method javadoc?

iloveeclipse:
As commented on [other PR](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/873), see this discussion: https://bugs.eclipse.org/bugs/show_bug.cgi?id=66538.

I wonder if the `org.eclipse.debug.core.model.Breakpoint.getMarkerRule()` should be used instead and the underlined logic fixed where needed in one place for all breakpoint clients.

trancexpress:
> As commented on [other PR](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/873), see this discussion: https://bugs.eclipse.org/bugs/show_bug.cgi?id=66538.
> 
> I wonder if the `org.eclipse.debug.core.model.Breakpoint.getMarkerRule()` should be used instead and the underlined logic fixed where needed in one place for all breakpoint clients.

The marker rule seems to be a placeholder for a change that was never done, or am I seeing it wrong?

I don't think we can fix this even for `JavaDebugOptionsManager`, since it does multiple operations on the marker... check if the marker still exists, then do something with it. Maybe if `Marker.setAttribute()` held a lock on the respective rule, and we catch the thrown exception if it no longer exists...

Issue #2220, opened 2025-10-29
UnifiedTree.isRecursiveLink() links check may skip some links


iloveeclipse:
Users & tools on Linux are creative in creating links. Both link parent and link itself might be not normalized or being part of other link structures. I've per occasion found a case where recursive link check skips some links (it ignores `IOException`s on purpose) because the code didn't considered level of creativity from our users.

Anyway, the `UnifiedTree.isRecursiveLink()` code might (silently) skip valid links tests while testing for the "real path" constructed by using "wrong" parts (normalized and not normalized) of the valid link.
In such case exception below will be swallowed.
```
java.nio.file.NoSuchFileException: /some/path/with/link
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixPath.toRealPath(UnixPath.java:834)
```
I will provide a test & a fix tomorrow.

PR #2221, opened 2025-10-29
UnifiedTree.isRecursiveLink() links check may silently ignore some links

iloveeclipse:
UnifiedTree.isRecursiveLink() links check may silently ignore some links

- Main fix: the child path created from the link should use `realParentPath` to avoid silent `NoSuchFileException`'s if the constructed link does not match actual (fully resolved) file system state of the local file we are checking. Original code run into (silent) exceptions and because of that was not able to detect some of possible recursive links.
- Removed pattern checking for "../X" because backwards links can also be created with absolute paths, as seen in
`Bug_185247_recursiveLinks.test5_linkParentDirectoyTwiceWithAbsolutePath(boolean)`
- Added regression test `SymlinkResourceTest.testGithubBug2220`.
- Added `UnifiedTree` methods to set/read `disable_advanced_recursive_link_checks` flag in tests so we can test
both variants of link checking code.
- Updated existing tests to test both advanced / simple link checks (where possible): `SymlinkResourceTest`, `Bug_185247_recursiveLinks` & `Bug_185247_LinuxTests`
- Improved `WorkspaceResetExtension` to properly report test names (it only reported `false` or `true` for parametrized tests).

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2220

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/UnifiedTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/UnifiedTree.java
@@ -438,9 +438,6 @@ public class UnifiedTree {
 		//Pattern: A UNIX or Windows relative path that just points backward
 		static final Pattern TRIVIAL_SYMLINK_PATTERN = Pattern.compile( //
 				Platform.OS.isWindows() ? "\\.[.\\\\]*" : "\\.[./]*"); //$NON-NLS-1$//$NON-NLS-2$
-
-		static final Pattern REPEATING_BACKWARDS_PATTERN = Pattern.compile( //
-				Platform.OS.isWindows() ? "(\\.\\.\\\\)+.*" : "(\\.\\./)+.*"); //$NON-NLS-1$//$NON-NLS-2$
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


iloveeclipse:
TODO: test

iloveeclipse:
Two problems:
1) JUnit 4 assumptions are ignored by JUnit 5 tests (fixed now)
2) [Releng data storage troubles](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6744), so we run into 
```
14:15:02.676 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.1-SNAPSHOT:validate-classpath (default-validate-classpath) on project org.eclipse.debug.terminal: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:5.0.1-SNAPSHOT:validate-classpath failed: Could not mirror artifact osgi.bundle,org.eclipse.cdt.core.linux.source,6.1.100.202402230238 into the local Maven repository.See log output for details. Server returned HTTP code: 400 for URL https://download.eclipse.org/eclipse/updates/4.38-I-builds/I20251102-1800/plugins/org.eclipse.cdt.core.linux.source_6.1.100.202402230238.jar
```

PR #2223, opened 2025-10-30
Fix Terminate & Remove showing every time

SougandhS:
This PR will enable Terminate & Remove option in debug view context only when there's a valid selection

before : 

image

after : 
image


iloveeclipse (approved):

SougandhS:
Thank you for approving, build looks good now 👍

PR #2230, opened 2025-10-30
Handle long executable paths in Windows process launching

laeubi:
Add shortenWindowsCommandLine() to shorten all command line arguments that exceed Windows MAX_PATH limit, not just the working directory. This fixes issues where Eclipse is unable to launch processes with long executable paths on Windows.

The fix uses Windows GetShortPathName API via Win32Handler to convert long paths to their 8.3 short form before passing them to ProcessBuilder or Runtime.exec().

This was reported in one of my projects, for example when passing a params file (to overcome limitations of too long arguments) it can happen that the path of the parameter file (e.g. java, javadoc and alike tools) itself exceed the path limit.
The same might be if I try to launch `c:\java\inaverylongpathaht... exxeedcsthelimit\java.exe`

As there was already similar workaround for the working directory it feels good to have similar for the arguments itself but as I'm not a regular windows user it would be great to get some feedback from others, the fix is quite trying to avoid unnecessary checks so in almost all cases would only be a no-op


laeubi:
As no one has objected and to get possible feedback from M3 merging now...

HannesWell:
> Add shortenWindowsCommandLine() to shorten all command line arguments that exceed Windows MAX_PATH limit, not just the working directory. This fixes issues where Eclipse is unable to launch processes with long executable paths on Windows.

In general there is a Registry setting in Windows to permit (almost) arbitrary long paths on Windows. But for some reason this setting is not effective when launching new processes via the Windows API.
I linked the corresponding JDK bug in the PR where I introduced the called method:
- https://github.com/eclipse-platform/eclipse.platform/pull/1477

I cannot tell if this is also relevant for arguments itself, but AFAIK there are limitations for the overall length of a CLI invocation. In that case it's often a stragegy to pass argument files instead, but I don't know if this just a performance/handling optimization or if there is a hard limit in any case.
Nevertheless I wonder if the registry setting permitting long paths would also help in your case?

PR #2240, opened 2025-11-02
Auto pin behavior for opening new console

mehmet-karaman:
The console view has a "pin console" action, which pins the current console. If the user wants a new console, the original one should be pinned, otherwise a console switch will switch all unpinned console views.

- Added auto pin behavior for the console view. If user opens a new
console, it will raise the question if auto pin should be done.
- Added a new preference to the ConsolePreferencePage

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
@@ -177,6 +177,10 @@ public class ConsolePreferencePage extends FieldEditorPreferencePage implements 
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_OUT, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_output_3, SWT.NONE, getFieldEditorParent()));
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_ERR, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_error_3, SWT.NONE, getFieldEditorParent()));
 
+		BooleanFieldEditor autoPinEditor = new BooleanFieldEditor(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,
+				DebugPreferencesMessages.ConsolePreferencePage_ConsoleAutoPinEnable, SWT.NONE, getFieldEditorParent());
+		autoPinEditor.setPreferenceStore(ConsolePlugin.getDefault().getPreferenceStore());
+		addField(autoPinEditor);
 		Label comboLabel = new Label(getFieldEditorParent(), SWT.NONE);
 		comboLabel.setText(DebugPreferencesMessages.ConsoleElapsedTimeLabel);
 		fElapsedFormat = new ComboViewer(getFieldEditorParent(), SWT.DROP_DOWN | SWT.BORDER);
@@ -324,6 +328,9 @@ public class ConsolePreferencePage extends FieldEditorPreferencePage implements 
 		updateBufferSizeEditor();
 		updateInterpretCrAsControlCharacterEditor();
 		updateElapsedTimePreferences();
+
+		IPreferenceStore prefStore = ConsolePlugin.getDefault().getPreferenceStore();
+		prefStore.setValue(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, false);
 	}
 
 	protected boolean canClearErrorMessage() {
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/DebugPreferencesMessages.properties
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/DebugPreferencesMessages.properties
@@ -23,6 +23,7 @@ ConsolePreferencePage_Wrap_text_1=Fixed &width console
 ConsolePreferencePage_Console_width=&Maximum character width:
 ConsolePreferencePage_Limit_console_output_1=&Limit console output
 ConsolePreferencePage_Console_buffer_size__characters___2=Console &buffer size (characters):
+ConsolePreferencePage_ConsoleAutoPinEnable=Pin current and the new Console views when a new Console view is opened
 ConsolePreferencePage_The_console_buffer_size_must_be_at_least_1000_characters__1=Buffer size must be between 1000 and {0} inclusive.
 ConsolePreferencePage_console_width=Character width must be between 80 and 1000 inclusive.
 ConsolePreferencePage_12=Displayed &tab width:
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
@@ -164,6 +164,22 @@ public interface IConsoleConstants {
 	String P_CONSOLE_WORD_WRAP = ConsolePlugin.getUniqueIdentifier() + ".P_CONSOLE_WORD_WRAP"; //$NON-NLS-1$
 
 	/**
+	 * The preference name for the auto pin question to avoid that opening for every
+	 * new console opening.
+	 *
+	 * @since 3.15
+	 */
+	String REMEMBER_AUTO_PIN_DECISION_PREF_NAME = ConsolePlugin.getUniqueIdentifier() + ".AUTO_PIN_ASKED"; //$NON-NLS-1$
+
+	/**
+	 * The preference name for automatically pinning current Console view when
+	 * opening a new Console view.
+	 *
+	 * @since 3.15
+	 */
+	String AUTO_PIN_ENABLED_PREF_NAME = ConsolePlugin.getUniqueIdentifier() + ".AUTO_PIN_ENABLED"; //$NON-NLS-1$
+
+	/**
 	 * The default tab size for text consoles.
 	 *
 	 * @since 3.1
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
@@ -59,3 +59,6 @@ PatternMatchListenerExtension_3=Console Pattern Match Listener
 PatternMatchListenerExtension_4=contributed by
 PatternMatchListenerExtension_5=is missing required enablement expression and will be removed
 UpdatingConsoleState=Updating console state
+TurnOnAutoPinDialogMessage=To see updates of different console pages in parallel, both current and new Console view will be pinned now.
+TurnOnAutoPinDialogTitle=Pin Console views?
+TurnOnAutoPinRememberDecision=Remember my decision
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleUIPreferenceInitializer.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleUIPreferenceInitializer.java
@@ -29,6 +29,8 @@ public class ConsoleUIPreferenceInitializer extends AbstractPreferenceInitialize
 		IPreferenceStore prefs = ConsolePlugin.getDefault().getPreferenceStore();
 		prefs.setDefault(IConsoleConstants.P_CONSOLE_AUTO_SCROLL_LOCK, true);
 		prefs.setDefault(IConsoleConstants.P_CONSOLE_WORD_WRAP, false);
+		prefs.setDefault(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME, true);
+		prefs.setDefault(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, false);
 	}
 
 }
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
@@ -461,7 +461,7 @@ public class ConsoleView extends PageBookView implements IConsoleView, IConsoleL
 		fDisplayConsoleAction = new ConsoleDropDownAction(this);
 		ConsoleFactoryExtension[] extensions = ((ConsoleManager)ConsolePlugin.getDefault().getConsoleManager()).getConsoleFactoryExtensions();
 		if (extensions.length > 0) {
-			fOpenConsoleAction = new OpenConsoleAction();
+			fOpenConsoleAction = new OpenConsoleAction(this);
 		}
 	}
 
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
@@ -13,6 +13,12 @@
  *******************************************************************************/
 package org.eclipse.ui.internal.console;
 
+import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.jface.dialogs.MessageDialogWithToggle;
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.IViewPart;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PartInitException;
@@ -24,22 +30,70 @@ import org.eclipse.ui.console.IConsoleFactory;
 public class ConsoleViewConsoleFactory implements IConsoleFactory {
 
 	int counter = 1;
+	private ConsoleView currentConsoleView;
 
 	@Override
 	public void openConsole() {
 		IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
-		if (window != null) {
-			IWorkbenchPage page = window.getActivePage();
-			if (page != null) {
-				try {
-					String secondaryId = "Console View #" + counter; //$NON-NLS-1$
-					page.showView(IConsoleConstants.ID_CONSOLE_VIEW, secondaryId, 1);
-					counter++;
-				} catch (PartInitException e) {
-					ConsolePlugin.log(e);
-				}
-			}
+		if (window == null) {
+			return;
 		}
+		IWorkbenchPage page = window.getActivePage();
+		if (page == null) {
+			return;
+		}
+		boolean shouldPin = handleAutoPin();
+		try {
+			String secondaryId = "Console View #" + counter; //$NON-NLS-1$
+			IViewPart view = page.showView(IConsoleConstants.ID_CONSOLE_VIEW, secondaryId, 1);
+			if (view instanceof ConsoleView newConsoleView) {
+				newConsoleView.setPinned(shouldPin);
+			}
+			counter++;
+		} catch (PartInitException e) {
+			ConsolePlugin.log(e);
+		}
+	}
+
+	/**
+	 * This handler checks if the remember auto-pin decision state <b>not true</b>
+	 * and asks the user if auto pin of the view content should be enabled.
+	 * Afterwards it checks if remember auto-pin decision was checked and sets the
+	 * preference according to that
+	 *
+	 * If the remember auto-pin decision state is <b>true</b> it gathers the auto
+	 * pin preference value and sets this to the current view.
+	 */
+	private boolean handleAutoPin() {
+		if (currentConsoleView == null) {
+			return false;
+		}
+		IPreferenceStore store = ConsolePlugin.getDefault().getPreferenceStore();
+		if (!store.getBoolean(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME)) {
+			Shell shell = Display.getDefault().getActiveShell();
+			MessageDialogWithToggle toggleDialog = MessageDialogWithToggle.openOkCancelConfirm(shell,
+					ConsoleMessages.TurnOnAutoPinDialogTitle, ConsoleMessages.TurnOnAutoPinDialogMessage,
+					ConsoleMessages.TurnOnAutoPinRememberDecision, false, null, null);
+
+			store.setValue(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,
+					toggleDialog.getReturnCode() == IDialogConstants.OK_ID);
+
+			store.setValue(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, toggleDialog.getToggleState());
+		}
+
+		if (store.getBoolean(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME)) {
+			// To avoid if pinned manually and unpin due to preference..
+			currentConsoleView.setPinned(true);
+			return true;
+		}
+		return false;
+	}
+
+	/**
+	 * Sets the console view, on which the open new console action was called.
+	 */
+	void setConsoleView(ConsoleView consoleView) {
+		this.currentConsoleView = consoleView;
 	}
 
 }
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
@@ -39,9 +39,14 @@ public class OpenConsoleAction extends Action implements IMenuCreator {
 
 	private ConsoleFactoryExtension[] fFactoryExtensions;
 	private Menu fMenu;
+	private ConsoleView consoleView;
 
-	public OpenConsoleAction() {
+	/**
+	 * @param consoleView the view which this action belongs to.
+	 */
+	public OpenConsoleAction(ConsoleView consoleView) {
 		super(ConsoleMessages.OpenConsoleAction_0, AS_DROP_DOWN_MENU);
+		this.consoleView = consoleView;
 		fFactoryExtensions = getSortedFactories();
 		setToolTipText(ConsoleMessages.OpenConsoleAction_1);
 		setImageDescriptor(ConsolePluginImages.getImageDescriptor(IInternalConsoleConstants.IMG_ELCL_NEW_CON));
@@ -66,6 +71,7 @@ public class OpenConsoleAction extends Action implements IMenuCreator {
 
 	@Override
 	public void dispose() {
+		consoleView = null;
 		fFactoryExtensions = null;
 		if (fMenu != null) {
 			fMenu.dispose();
@@ -98,7 +104,7 @@ public class OpenConsoleAction extends Action implements IMenuCreator {
 			if (!WorkbenchActivityHelper.filterItem(extension) && extension.isEnabled()) {
 				String label = extension.getLabel();
 				ImageDescriptor image = extension.getImageDescriptor();
-				addActionToMenu(fMenu, new ConsoleFactoryAction(label, image, extension), accel);
+				addActionToMenu(fMenu, new ConsoleFactoryAction(label, image, extension, consoleView), accel);
 				accel++;
 				if (extension.isNewConsoleExtenson()) {
 					new Separator("new").fill(fMenu, -1); //$NON-NLS-1$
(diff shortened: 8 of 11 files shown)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/DebugPreferencesMessages.properties
```diff
@@ -23,6 +23,7 @@ ConsolePreferencePage_Wrap_text_1=Fixed &width console
 ConsolePreferencePage_Console_width=&Maximum character width:
 ConsolePreferencePage_Limit_console_output_1=&Limit console output
 ConsolePreferencePage_Console_buffer_size__characters___2=Console &buffer size (characters):
+ConsolePreferencePage_ConsoleAutoPinEnable=Auto pin current console view, if new console is opened
```
trancexpress: > Auto pin current console view, if new console is opened

->

>Auto pin current Console view, when new Console view is opened

mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/messages.properties
```diff
@@ -0,0 +1,3 @@
+ConsoleViewConsoleFactory_TurnOnAutoPinDialogMessage=If a new console is opened, it's possible to auto pin the old console view to keep its content. Would you like to turn on the auto pin?
```
trancexpress: Please capitalize Console here.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -28,17 +33,54 @@ public class ConsoleViewConsoleFactory implements IConsoleFactory {
...
+			store.setValue(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,
+					toggleDialog.getReturnCode() == IDialogConstants.YES_ID);
+
+			store.setValue(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, toggleDialog.getToggleState());
+		}
+
+		if (store.getBoolean(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME)
+				&& page.getActivePart() instanceof ConsoleView currentConsoleView) {
+			// To avoid if pinned manually and unpin due to preference..
+			currentConsoleView.setPinned(true);
```
trancexpress, line 86: This behaves somewhat weird. I don't think using the view menu activates the console for which the menu is invoked.

So if I have two Console views open already, click on the 2nd console to activate it, open a 3rd console from the view menu of the 1st console, the auto-pinned console is the 2nd console, not the 1st...

Is that what we want? It seems odd.

mehmet-karaman: I was thinking that if the user uses the menu, the view gets as the active part.. But tested it now and it seems you're right..

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/messages.properties
mehmet-karaman: done.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
mehmet-karaman: Forwarding the Console View to the action to solve this problem

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
```diff
@@ -147,7 +157,9 @@ public void run() {
 				if (fFactory == null) {
 					fFactory = fConfig.createFactory();
 				}
-
+				if (fFactory instanceof ConsoleViewConsoleFactory bla) {
+					bla.setConsoleView(consoleView);
```
trancexpress: Please use a proper name for the variable here.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/DebugPreferencesMessages.properties
```diff
@@ -23,6 +23,7 @@ ConsolePreferencePage_Wrap_text_1=Fixed &width console
 ConsolePreferencePage_Console_width=&Maximum character width:
 ConsolePreferencePage_Limit_console_output_1=&Limit console output
 ConsolePreferencePage_Console_buffer_size__characters___2=Console &buffer size (characters):
+ConsolePreferencePage_ConsoleAutoPinEnable=Auto pin current Console view, when new Console view is opened
```
trancexpress: "when new Console view is opened" -> "when **a** new Console view is opened"

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
```diff
@@ -64,6 +65,15 @@ private ConsoleFactoryExtension[] getSortedFactories() {
 		return factoryExtensions;
 	}
 
+	/**
+	 * @param consoleView the view which this action belongs to.
+	 *
+	 * @since 3.14
```
trancexpress: Please remove the `@since`.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +29,68 @@
...
+		}
+
+		if (store.getBoolean(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME)) {
+			// To avoid if pinned manually and unpin due to preference..
+			currentConsoleView.setPinned(true);
+		}
+	}
+
+	/**
+	 * @since 3.14
```
trancexpress: Please remove.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +29,68 @@
...
+	/**
+	 * This handler checks if the remember auto-pin decision state <b>not true</b>
+	 * and asks the user if auto pin of the view content should be enabled.
+	 * Afterwards it checks if remember auto-pin decision was checked and sets the
+	 * preference according to that
+	 *
+	 * If the remember auto-pin decision state is <b>true</b> it gathers the auto
+	 * pin preference value and sets this to the current view.
+	 *
+	 * @since 3.14
```
trancexpress: Please remove `@since`.

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
```diff
@@ -163,6 +163,21 @@ public interface IConsoleConstants {
...
+	/**
+	 * The preference name for the auto pin question to avoid that opening for every
+	 * new console opening.
+	 *
+	 * @since 3.15
+	 */
+	String REMEMBER_AUTO_PIN_DECISION_PREF_NAME = ConsolePlugin.getUniqueIdentifier() + ".AUTO_PIN_ASKED"; //$NON-NLS-1$
+
+	/**
+	 * The preference name for the auto pin.
```
trancexpress: "The preference name for the auto pin." -> "The preference name for automatically pinning current Console view when opening a new Console view."

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
```diff
@@ -163,6 +163,21 @@ public interface IConsoleConstants {
 	 */
 	String P_CONSOLE_WORD_WRAP = ConsolePlugin.getUniqueIdentifier() + ".P_CONSOLE_WORD_WRAP"; //$NON-NLS-1$
 
+	/**
+	 * The preference name for the auto pin question to avoid that opening for every
+	 * new console opening.
+	 *
+	 * @since 3.15
+	 */
+	String REMEMBER_AUTO_PIN_DECISION_PREF_NAME = ConsolePlugin.getUniqueIdentifier() + ".AUTO_PIN_ASKED"; //$NON-NLS-1$
```
trancexpress, line 172: Please add defaults for the two new preferences in: `org.eclipse.ui.internal.console.ConsoleUIPreferenceInitializer.initializeDefaultPreferences()`

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
```diff
@@ -177,6 +177,10 @@ public void widgetSelected(SelectionEvent e) {
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_OUT, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_output_3, SWT.NONE, getFieldEditorParent()));
 		addField(new BooleanFieldEditor(IDebugPreferenceConstants.CONSOLE_OPEN_ON_ERR, DebugPreferencesMessages.ConsolePreferencePage_Show__Console_View_when_there_is_program_error_3, SWT.NONE, getFieldEditorParent()));
 
+		BooleanFieldEditor editor = new BooleanFieldEditor(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,
```
trancexpress: Probably best to call this `autoPinEditor` or similar.

mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
mehmet-karaman: good idea.. done..

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/IConsoleConstants.java
mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
mehmet-karaman: done.

trancexpress:
LGTM.

@iloveeclipse anything to add or can I merge this?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +29,66 @@
...
+		}
+		IPreferenceStore store = ConsolePlugin.getDefault().getPreferenceStore();
+		if (!store.getBoolean(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME)) {
+			Shell shell = Display.getDefault().getActiveShell();
+			MessageDialogWithToggle toggleDialog = MessageDialogWithToggle.openYesNoQuestion(shell,
+					Messages.ConsoleViewConsoleFactory_TurnOnAutoPinDialogTitle,
+					Messages.ConsoleViewConsoleFactory_TurnOnAutoPinDialogMessage,
+					Messages.ConsoleViewConsoleFactory_TurnOnAutoPinRememberDecision, false, null, null);
+
+			store.setValue(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,
```
iloveeclipse, line 78: `store.setValue(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME,`

this wouldn't be needed if one would pass the store and the key to the dialog?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +29,66 @@
...
+					toggleDialog.getReturnCode() == IDialogConstants.YES_ID);
+
+			store.setValue(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, toggleDialog.getToggleState());
```
iloveeclipse, line 81: This decision can't be reverted anymore once set. I would reset this value in `org.eclipse.debug.internal.ui.preferences.ConsolePreferencePage.performDefaults()`

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/Messages.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.eclipse.osgi.util.NLS;
+
+/**
+ * @since 3.14
+ */
+public class Messages extends NLS {
+
+	/**
+	 * @since 3.14
```
iloveeclipse: For the future: since tags are not needed for internal packages

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/Messages.java
```diff
@@ -0,0 +1,37 @@
+package org.eclipse.ui.internal.console;
```
iloveeclipse: Why this class is added if we have `org.eclipse.ui.internal.console.ConsoleMessages` ?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/messages.properties
```diff
@@ -0,0 +1,3 @@
+ConsoleViewConsoleFactory_TurnOnAutoPinDialogMessage=If a new Console view is opened, it's possible to auto pin the old Console view to keep its content. Would you like to turn on the auto pin?
```
iloveeclipse: Same as above, we have `ConsoleMessages.properties` already

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
```diff
@@ -462,6 +462,7 @@ protected void createActions() {
 		ConsoleFactoryExtension[] extensions = ((ConsoleManager)ConsolePlugin.getDefault().getConsoleManager()).getConsoleFactoryExtensions();
 		if (extensions.length > 0) {
 			fOpenConsoleAction = new OpenConsoleAction();
+			fOpenConsoleAction.setOwnerView(this);
```
iloveeclipse: Why do we need an extra method only to provide reference to the console, if we can simply add constructor argument to OpenConsoleAction?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
```diff
@@ -64,6 +65,13 @@ private ConsoleFactoryExtension[] getSortedFactories() {
...
+	/**
+	 * @param consoleView the view which this action belongs to.
+	 */
+	void setOwnerView(ConsoleView consoleView) {
+		this.consoleView = consoleView;
+	}
+
 	@Override
 	public void dispose() {
 		fFactoryExtensions = null;
```
iloveeclipse, line 75: `consoleView = null;`

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
iloveeclipse: The class should have been static from very beginning. If you would make it static now you will see the problem.
Ideally the `consoleView` field from enclosing class used here should be passed as argument to this constructor.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
iloveeclipse, line 330: I miss reset of the value below:  
`ConsolePlugin.getDefault().getPreferenceStore().setValue(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME, false);`
can be done in a dedicated method here. Note, the other preference is managed by the visible field.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +29,66 @@
...
+			return;
 		}
+		IWorkbenchPage page = window.getActivePage();
+		if (page == null) {
+			return;
+		}
+		handleAutoPin();
+		try {
+			String secondaryId = "Console View #" + counter; //$NON-NLS-1$
+			page.showView(IConsoleConstants.ID_CONSOLE_VIEW, secondaryId, 1);
```
iloveeclipse: I wonder why shouldnt we pin both consoles, the old and new one?
Something like this:
```
diff --git a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
index 0891aeb..a8ca182 100644
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
@@ -20,2 +20,3 @@
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.IViewPart;
 import org.eclipse.ui.IWorkbenchPage;
@@ -43,6 +44,9 @@
 		}
-		handleAutoPin();
+		boolean shouldPin = handleAutoPin();
 		try {
 			String secondaryId = "Console View #" + counter; //$NON-NLS-1$
-			page.showView(IConsoleConstants.ID_CONSOLE_VIEW, secondaryId, 1);
+			IViewPart view = page.showView(IConsoleConstants.ID_CONSOLE_VIEW, secondaryId, 1);
+			if (view instanceof ConsoleView newConsoleView) {
+				newConsoleView.setPinned(shouldPin);
+			}
 			counter++;
@@ -62,5 +66,5 @@
 	 */
-	private void handleAutoPin() {
+	private boolean handleAutoPin() {
 		if (currentConsoleView == null) {
-			return;
+			return false;
 		}
@@ -83,3 +87,5 @@
 			currentConsoleView.setPinned(true);
+			return true;
 		}
+		return false;
 	}
```

Background: 
- user started program that constantly prints to the console.
- He wants to clone the console to see output from a different page.
- On opening, the same console page is shown as the old one
- By trying to swith to a new page in the new console the not pinned output causes constant console switches until one manually pins the new console too.

If both consoles are pinned, no problems with selecting other page in either console.

If we do so, text in the dialog and preference page has to be updated too.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
mehmet-karaman: No, thats necessary.. If remember decision is selected and the user clicks on no button, it won't save that.. The user can only use remember me + yes button to save that decision..

iloveeclipse: > If remember decision is selected and the user clicks on no button, it won't save that..

Please double check. I would expect that the preference change is always recorded by the dialog, and if no preference change happened, we also don't need to remember anything except the question state itself.

mehmet-karaman: Was recalling wrong: It was setting the preference values as string "always" / "never" and not the boolean value. Thats why I had chosen this way.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
mehmet-karaman: Applied patch and updated the the texts in dialog and preference too..

mehmet-karaman: applied patch + adjusted texts

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/ConsolePreferencePage.java
mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
mehmet-karaman: done

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
mehmet-karaman: done

iloveeclipse:
@mehmet-karaman : I see you've pushed new commit but some of the comments are still not resolved. Does it mean, you plan to push yet another change, should I wait with the review till then?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/preferences/DebugPreferencesMessages.properties
```diff
@@ -23,6 +23,7 @@ ConsolePreferencePage_Wrap_text_1=Fixed &width console
 ConsolePreferencePage_Console_width=&Maximum character width:
 ConsolePreferencePage_Limit_console_output_1=&Limit console output
 ConsolePreferencePage_Console_buffer_size__characters___2=Console &buffer size (characters):
+ConsolePreferencePage_ConsoleAutoPinEnable=Auto pin current Console view and the new Console view, when a new Console view is opened
```
iloveeclipse: `Pin current and the new Console views when a new Console view is opened`

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
```diff
@@ -58,4 +58,7 @@ TextConsolePage_PasteDescrip=Paste
 PatternMatchListenerExtension_3=Console Pattern Match Listener
 PatternMatchListenerExtension_4=contributed by
 PatternMatchListenerExtension_5=is missing required enablement expression and will be removed
-UpdatingConsoleState=Updating console state
\ No newline at end of file
+UpdatingConsoleState=Updating console state
+TurnOnAutoPinDialogMessage=If a new Console view is opened, it's possible to auto pin the old Console view and new Console view to keep its content. Would you like to turn on the auto pin?
```
iloveeclipse: What about this:
```
Should we pin the old and new Console views (so different console pages could be updated independently)?
```

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
```diff
@@ -130,15 +136,18 @@ public Menu getMenu(Menu parent) {
 
 	private class ConsoleFactoryAction extends Action {
```
iloveeclipse: Please make this class **static**

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
mehmet-karaman: done.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/OpenConsoleAction.java
mehmet-karaman: done.

mehmet-karaman:
> @mehmet-karaman : I see you've pushed new commit but some of the comments are still not resolved. Does it mean, you plan to push yet another change, should I wait with the review till then?

No.. I wanted to recheck before resolving them.. After amend + push I couldn't see your comments.. It looks like there is a new feature of github, which is working differently.. Switched to classic and now I can see all again..

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
iloveeclipse: Original PR version:
image

Proposed one:
image

Improved proposal:
image

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
```diff
@@ -58,4 +58,7 @@ TextConsolePage_PasteDescrip=Paste
 PatternMatchListenerExtension_3=Console Pattern Match Listener
 PatternMatchListenerExtension_4=contributed by
 PatternMatchListenerExtension_5=is missing required enablement expression and will be removed
-UpdatingConsoleState=Updating console state
\ No newline at end of file
+UpdatingConsoleState=Updating console state
+TurnOnAutoPinDialogMessage=If a new Console view is opened, it's possible to auto pin the old Console view and new Console view to keep its content. Should we pin the old and new Console views (so different console pages could be updated independently)?
+TurnOnAutoPinDialogTitle=Enable auto pin
```
iloveeclipse: `Pin Console views?`

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleMessages.properties
```diff
@@ -58,4 +58,7 @@ TextConsolePage_PasteDescrip=Paste
 PatternMatchListenerExtension_3=Console Pattern Match Listener
 PatternMatchListenerExtension_4=contributed by
 PatternMatchListenerExtension_5=is missing required enablement expression and will be removed
-UpdatingConsoleState=Updating console state
\ No newline at end of file
+UpdatingConsoleState=Updating console state
+TurnOnAutoPinDialogMessage=If a new Console view is opened, it's possible to auto pin the old Console view and new Console view to keep its content. Should we pin the old and new Console views (so different console pages could be updated independently)?
```
iloveeclipse: `To see updates of different console pages in parallel, both current and new Console view will be pinned now.`

Note, this message change implies change from yes/no to OK/Cancel.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleUIPreferenceInitializer.java
```diff
@@ -29,6 +29,8 @@ public void initializeDefaultPreferences() {
 		IPreferenceStore prefs = ConsolePlugin.getDefault().getPreferenceStore();
 		prefs.setDefault(IConsoleConstants.P_CONSOLE_AUTO_SCROLL_LOCK, true);
 		prefs.setDefault(IConsoleConstants.P_CONSOLE_WORD_WRAP, false);
+		prefs.setDefault(IConsoleConstants.AUTO_PIN_ENABLED_PREF_NAME, false);
```
iloveeclipse: We should enable by default, it makes no sense to keep it as is. 
The dialog shown is enough to explain the user the new behavior.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleViewConsoleFactory.java
```diff
@@ -24,22 +30,70 @@
...
+	 * pin preference value and sets this to the current view.
+	 */
+	private boolean handleAutoPin() {
+		if (currentConsoleView == null) {
+			return false;
+		}
+		IPreferenceStore store = ConsolePlugin.getDefault().getPreferenceStore();
+		if (!store.getBoolean(IConsoleConstants.REMEMBER_AUTO_PIN_DECISION_PREF_NAME)) {
+			Shell shell = Display.getDefault().getActiveShell();
+			MessageDialogWithToggle toggleDialog = MessageDialogWithToggle.openYesNoQuestion(shell,
```
iloveeclipse: Ok/Cancel now.

iloveeclipse:
@mehmet-karaman : please provide N&N for the change.
N&N repo is https://github.com/eclipse-platform/www.eclipse.org-eclipse

mehmet-karaman:
The N&N PR: https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/438

Issue #1690, opened 2025-01-16
javax.net.ssl.trustStoreType=Windows-ROOT is not working with jgit


markusschaefer:
Let's make sure issue is not already fixed in latest builds first.

- [X] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

From a fresh installation and clean workspace:

* Import an existing git repository as eclipse project from the locale filesystem
* Run a git operation against the remote origin

I expected: the certificate from the git server is trusted. When I remove the new VM args -Djavax.net.ssl.trustStoreType=Windows-ROOT -Djavax.net.ssl.trustStore=NONE everything is fine.

But got: I got an SunCertPathBuilderException: unable to find valid certification path to requested target

## Here is some relevant log output

From `<workspace>/.metadata/.log`
```
Caused by: org.eclipse.jgit.errors.TransportException: https://myGitHost/myRepo.git: Secure connection to https://myGitHost/myRepo.git could not be established because of SSL problems
	at org.eclipse.jgit.transport.TransportHttp.handleSslFailure(TransportHttp.java:840)
	at org.eclipse.jgit.transport.TransportHttp.connect(TransportHttp.java:741)
	at org.eclipse.jgit.transport.TransportHttp.openFetch(TransportHttp.java:465)
	at org.eclipse.jgit.transport.FetchProcess.executeImp(FetchProcess.java:153)
	at org.eclipse.jgit.transport.FetchProcess.execute(FetchProcess.java:105)
	at org.eclipse.jgit.transport.Transport.fetch(Transport.java:1458)
	at org.eclipse.jgit.api.FetchCommand.call(FetchCommand.java:238)
	... 3 more
Caused by: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:130)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:383)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:326)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:321)
	at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1318)
[...]

PR #2241, opened 2025-11-03
Use a merged JVM+OS trust store as default SSLContext

sratz:
Resolves:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=567504
https://github.com/eclipse-platform/eclipse.platform/issues/1690

Obsoletes (to-be-reverted):
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/929
https://github.com/eclipse-packaging/packages/pull/224

Replaces:
https://github.com/eclipse-equinox/equinox/pull/1176

See also:
https://github.com/eclipse-simrel/.github/pull/38


laeubi:
@sratz you need suppress the illegal access here, the testfailure look suspicious...

sratz:
> @sratz you need suppress the illegal access here, the testfailure look suspicious...

The problem is
https://github.com/eclipse-platform/eclipse.platform/blob/879c3bdc9f65ba869f790e82d8bfbfdfd5e373a1/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/runtime/PlatformURLSessionTest.java#L50-L52
which uses `SessionTestExtension` with `CustomSessionConfigurationImpl`
https://github.com/eclipse-platform/eclipse.platform/blob/879c3bdc9f65ba869f790e82d8bfbfdfd5e373a1/runtime/tests/org.eclipse.core.tests.harness/src/org/eclipse/core/tests/harness/session/customization/CustomSessionConfigurationImpl.java#L41

I have added `org.Mockito` as a dependency, which now cannot be resolved with the default of set bundles configured by `CustomSessionConfigurationImpl`.

Where should I add those additional bundles? customize just `PlatformURLSessionTest` or adjust the default in `CustomSessionConfigurationImpl`?

laeubi:
Mockito might be used in more tests in the future so I think you can decide what works best for you here.

akurtakov:
I have just read https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4587 which somehow made me more cautious about mocking frameworks.

HeikoKlare:
> Where should I add those additional bundles? customize just `PlatformURLSessionTest` or adjust the default in `CustomSessionConfigurationImpl`?

If we can assume that every consumer of the session test framework has Mockito available, it would probably make sense to add it to the default bundles. Note that they are used in other projects as well (such as Equinox), but I guess all of them will have Mockito available.

sratz:
> I have just read [eclipse-jdt/eclipse.jdt.core#4587](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4587) which somehow made me more cautious about mocking frameworks.

True, mocking concrete classes (especially final ones) / static mocks / etc. can be dangerous and are often result of a bad design. Rules of good design are typically:
- Limit mocking to interfaces
- Limit mocking to first-order / direct dependencies of the class-under-test.

In this case, `X509Certificate` is abstract with a bunch of abstract methods. Writing `CollectionTrustManagerTest` without mocks would be quite the overhead - especially since the tests does not care about the implementation at all.

sratz:
I added the needed bundles to just the one test class needing them.

Alternatively, use custom stubs instead of mocks and the need for Mockito:

With vs. without mocks:
https://github.com/eclipse-platform/eclipse.platform/compare/e4ca19088c598b00785f35d3ca05a83d940a004b...c72d4b3681e2b20b38e02c3c5c25e9721671e4fc

I have no real preference, but the latter would avoid the dependency.

laeubi:
@sratz I would say use what works best for you, do you plan to finish this for inclusion in M3?

sratz:
> @sratz I would say use what works best for you,

Then I'd say let's go with the latest state of the PR (explicit stubs) which avoids the hassle with new dependencies.

> do you plan to finish this for inclusion in M3?

This is ready for review now.

It would be great to get it into 2025-12 - but of course if considered a too impactful of a change for M3 we can aim for 2026-03 M1.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -585,6 +586,14 @@ private void initializeAuthorizationHandler() {
 		}
 	}
 
+	private void initializeSSLContext() {
+		try {
+			new KeyStoreUtil(getOS()).setUpSslContext();
+		} catch (Exception e) {
+			RuntimeLog.log(Status.error("Exception setting up SSLContext", e)); //$NON-NLS-1$
+		}
+	}
```
laeubi, line 595: This could be wrapped into 

```suggestion
	private void initializeSSLContext() {
	   if (!Boolean.getBoolean("eclipse.platform.mergeTrust")) {
	       return;
	   }
		try {
			new KeyStoreUtil(getOS()).setUpSslContext();
		} catch (Exception e) {
			RuntimeLog.log(Status.error("Exception setting up SSLContext", e)); //$NON-NLS-1$
		}
	}
```

or similar so one needs to explicitly enable it then EPPs have a choice to include this feature and users can disable it if it causes any issues.

sratz: Good idea. But by moving the system property check outside of the impl. I can't so easily unit-test that switch anymore in `KeyStoreUtilTest`.

sratz:
As suggested, switched to an **opt-in** boolean system property `eclipse.platform.mergeTrust` instead, hence, the feature is **disabled by default**.

@merks @HeikoKlare @iloveeclipse @akurtakov:
- Any opinions on whether to get this into 2025-12 M3 or wait for 2026-03 M1?
- Any opinions on the property name?

iloveeclipse:
> * Any opinions on whether to get this into 2025-12 M3 or wait for 2026-03 M1?

I would prefer to wait for M1.

merks:
Personally I'd rather see this move forward.  Given it's opt-in, it seems safe.  Note that I can test this in an Eclipse Installer as soon as it's available in an I-build, i.e., well before Friday.

akurtakov:
As it's opt-in and has been one of the most often reported problems it's good to have it in this cycle IMO too.

merks:
@sratz 

I'd push the "update branch" button to rebase, but maybe that would be disruptive?

laeubi:
+1 for have it in M3

merks (approved):
It looks property guarded.

laeubi:
@sratz I think squash into one commit would be good (just in case).

sratz:
OK, rebased + squashed locally to fix the 'committer' metadata and cleaned up some comments

laeubi:
@sratz you have approval from two committers now an an implicit one from @akurtakov so I think you should merge this as soon as you feel comfortable with it.

HeikoKlare (approved):
+1 for M3 also from my side.

I cannot review the actual implementation, but it's properly implemented as full opt-out with no side effects other than instantiating the `KeyStoreUtil` class and evaluating one system property, so it should be safe to do.

Issue #2246, opened 2025-11-05
TipImageURLTest fails since I20251101-0220


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20251101-0220/testresults/html/org.eclipse.tips.tests_ep438I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

The `TipImageURLTest` test uses image from external site which is not accessible anymore: http://remainsoftware.com/img.png

We should fix that.

iloveeclipse:
I have a trivial fix.

PR #2247, opened 2025-11-05
Use image from test resources on github instead of using external one

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2246

### DIFF
--- a/ua/org.eclipse.tips.tests/src/org/eclipse/tips/core/TipImageURLTest.java
+++ b/ua/org.eclipse.tips.tests/src/org/eclipse/tips/core/TipImageURLTest.java
@@ -25,7 +25,7 @@ import org.junit.jupiter.api.Test;
 
 public class TipImageURLTest {
 
-	private static final String URL = "http://remainsoftware.com/img.png";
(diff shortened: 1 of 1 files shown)


akurtakov:
Thanks @iloveeclipse !

ua/org.eclipse.tips.tests/src/org/eclipse/tips/core/TipImageURLTest.java
```diff
@@ -25,7 +25,7 @@
 
 public class TipImageURLTest {
 
-	private static final String URL = "http://remainsoftware.com/img.png";
+	private static final String URL = "https://github.com/eclipse-platform/eclipse.platform/blob/master/ua/org.eclipse.tips.tests/icons/dummy.png";
```
laeubi, line 28:
```suggestion
	private static final String URL = "https://github.com/eclipse-platform/eclipse.platform/blob/fef82c98cd89373f08ad5c3a5014aae8acf52ee4/ua/org.eclipse.tips.tests/icons/dummy.png";
```

Would even survive if we delete the image one day but just just for information...

Issue #2251, opened 2025-11-05
Project import spends most of the time in AliasManager.computeAliases()


iloveeclipse:
While working on https://github.com/eclipse-platform/eclipse.platform/issues/2220 with our internal data set I stumbled upon the fact that most of the time (~4/5) on creating (refreshing) huge project is spent in `AliasManager.computeAliases()` code, in particular in this method: `AliasManager$LocationMap.matchingResourcesDo().

The reason is that the `AliasManager.LocationMap.map` that is involved in the metod is `TreeMap` (used on purpose because `AliasManager.LocationMap.matchingPrefixDo()` tries to improve iteration by visiting only a subtree for given key).

Changing `AliasManager.LocationMap.map` to be a usual `HashMap` improved time by a magnitude (from ~500 seconds to ~5 seconds).

I have to investigate if the fix I have in mind is safe one, mode measurements & draft PR follows.

PR #2252, opened 2025-11-05
LocationMap should use HashMap for better performance

iloveeclipse:
The `TreeMap` data retrieval is slower compared to `HashMap` and `TreeMap` "sorted" behavior is only used in `matchingPrefixDo()` and `overLappingResourcesDo()` while updating resources.

Reading unchanged data will be faster if using `HashMap`, updating the data will use `TreeSet` and still be as fast as before.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2251

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
@@ -24,10 +24,9 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map.Entry;
 import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import org.eclipse.core.filesystem.EFS;
@@ -127,14 +126,55 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 	}
 
 	/**
+	 * A HashMap that also maintains its keys in sorted order. It combines the fast
+	 * access to the content with the ability to get sorted keys and key subsets.
+	 */
+	private static final class HashMapWithSortedKeys {
+		private final SortedSet<IFileStore> keys;
+		private final HashMap<IFileStore, Object> map;
+
+		HashMapWithSortedKeys() {
+			keys = new TreeSet<>(IFileStore::compareTo);
+			map = new HashMap<>(100);
+		}
+
+		Object put(IFileStore key, Object value) {
+			keys.add(key);
+			return map.put(key, value);
+		}
+
+		Object remove(Object key) {
+			keys.remove(key);
+			return map.remove(key);
+		}
+
+		Object get(IFileStore location) {
+			return map.get(location);
+		}
+
+		void clear() {
+			keys.clear();
+			map.clear();
+		}
+
+		Set<IFileStore> keys() {
+			return keys;
+		}
+
+		SortedSet<IFileStore> keysSubSet(IFileStore fromKey, IFileStore toKey) {
+			return keys.subSet(fromKey, toKey);
+		}
+	}
+
+	/**
 	 * Maintains a mapping of FileStore-&gt;IResource, such that multiple resources
 	 * mapped from the same location are tolerated.
 	 */
-	class LocationMap {
+	private static final class LocationMap {
 		/**
 		 * Map of FileStore-&gt;IResource OR FileStore-&gt;ArrayList of (IResource)
 		 */
-		private final SortedMap<IFileStore, Object> map = new TreeMap<>(IFileStore::compareTo);
+		private final HashMapWithSortedKeys map = new HashMapWithSortedKeys();
 
 		/**
 		 * Adds the given resource to the map, keyed by the given location.
@@ -177,17 +217,18 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 		 * given location as a prefix.
 		 */
 		public void matchingPrefixDo(IFileStore prefix, Consumer<IResource> doit) {
-			SortedMap<IFileStore, Object> matching;
+			Set<IFileStore> matching;
 			IFileStore prefixParent = prefix.getParent();
 			if (prefixParent != null) {
 				//endPoint is the smallest possible path greater than the prefix that doesn't
 				//match the prefix
 				IFileStore endPoint = prefixParent.getChild(prefix.getName() + "\0"); //$NON-NLS-1$
-				matching = map.subMap(prefix, endPoint);
+				matching = map.keysSubSet(prefix, endPoint);
 			} else {
-				matching = map;
+				matching = map.keys();
 			}
-			for (Object value : matching.values()) {
+			for (IFileStore key : matching) {
+				Object value = map.get(key);
 				if (value == null) {
 					return;
 				}
@@ -230,11 +271,10 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 		public void overLappingResourcesDo(Consumer<IResource> doit) {
 			IFileStore previousStore = null;
 			IResource previousResource = null;
-			for (Entry<IFileStore, Object> current : map.entrySet()) {
+			for (IFileStore currentStore : map.keys()) {
 				//value is either single resource or List of resources
-				IFileStore currentStore = current.getKey();
 				IResource currentResource = null;
-				Object value = current.getValue();
+				Object value = map.get(currentStore);
 				if (value instanceof List) {
 					for (Object element : ((List<?>) value)) {
 						if (element instanceof IResource) {


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -177,13 +178,20 @@ public void clear() {
...
 				//endPoint is the smallest possible path greater than the prefix that doesn't
 				//match the prefix
 				IFileStore endPoint = prefixParent.getChild(prefix.getName() + "\0"); //$NON-NLS-1$
-				matching = map.subMap(prefix, endPoint);
+				if (map instanceof SortedMap<IFileStore, Object> sorted) {
+					matching = sorted.subMap(prefix, endPoint);
+				} else {
+					TreeMap<IFileStore, Object> newMap = new TreeMap<>(IFileStore::compareTo);
+					newMap.putAll(map);
+					map = newMap;
```
laeubi: is this really intentional to replace the map here? It looks this has not happens in the old code.

iloveeclipse:
TODO:
- more measurements
- check corner cases who/how often needs the sorted map
- can we get rid of "faster" iteration on `subMap` completely (it was written ~2002), how much aliases can we have at all, shouldn't be a problem today to iterate over few entries more?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -134,7 +135,7 @@ class LocationMap {
 		/**
 		 * Map of FileStore-&gt;IResource OR FileStore-&gt;ArrayList of (IResource)
 		 */
-		private final SortedMap<IFileStore, Object> map = new TreeMap<>(IFileStore::compareTo);
+		private Map<IFileStore, Object> map = new HashMap<>(10_000); // new TreeMap<>(IFileStore::compareTo);
```
laeubi: Why not final anymore?!?

iloveeclipse: See line 192

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
iloveeclipse: the old code didn't used HashMap

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -177,13 +178,20 @@ public void clear() {
...
 		 */
 		public void matchingPrefixDo(IFileStore prefix, Consumer<IResource> doit) {
-			SortedMap<IFileStore, Object> matching;
+			Map<IFileStore, Object> matching;
 			IFileStore prefixParent = prefix.getParent();
 			if (prefixParent != null) {
 				//endPoint is the smallest possible path greater than the prefix that doesn't
 				//match the prefix
 				IFileStore endPoint = prefixParent.getChild(prefix.getName() + "\0"); //$NON-NLS-1$
-				matching = map.subMap(prefix, endPoint);
```
iloveeclipse, line 186: `map.subMap` is only available in SortedMap, not in HashMap. So **if** we can be sure there is no performance or functional issues with iterating over **entire** map, we can get rid of the SortedMap completely.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
laeubi: Sure but here you update the field(!) `map = newMap;` and then it would be slower again?!?

iloveeclipse:
> * can we get rid of "faster" iteration on `subMap` completely (it was written ~2002), how much aliases can we have at all, shouldn't be a problem today to iterate over few entries more?

This would be https://github.com/eclipse-platform/eclipse.platform/pull/2253. Let see what tests say.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
iloveeclipse: Sure. See my PR message, quote: 
> it seems that at least for startup / refresh / build case we can save lot of time.

fedejeanne:
> more measurements

If you mean _performance_ measurements, I'll be happy to test in our big WS and provide you with a _before_ and _after_ VisualVM samples. Just say the word and tell me what to test (steps to follow) :-)

iloveeclipse:
> If you mean _performance_ measurements, I'll be happy to test in our big WS and provide you with a _before_ and _after_ VisualVM samples. Just say the word and tell me what to test (steps to follow) :-)

**Many** thanks for the offer. 

What I see in the case of our application where a **very** large tree is imported into the single project, AnyEdit says `3226615 files in 492033 folders and 4 projects`, but 3 projects are "hidden".

We have ~40 entries in the `LocationMap` and ~60.000.000 calls to `matchingResourcesDo()` / `map.get()`.
AliasManager needs ~70 seconds during project import / refresh and most time spent here is because of the tree comparator used on `TreeMap.get()`, see
image

Same state, with the current approach (use HashMap for get and TreeSet for ordered iteration over keys or ordered subset of keys) results in the AliasManager time dropped to 6 seconds (from 70!) because the entire comparator overhead is gone:
image

The measured use case is reading the tree on import, doing first build (very small number of files involved) & refreshing the tree afterwards.

I'm still trying to create similar case where our application code is not involved. With small number of aliases (< 10) there seem to be no difference. We have ~40 because we use custom classpath container with ~40 folder entries pointing outside the workspace tree, so JDT "imports" them into hidden "external" project if project is imported.

So if you could simply try the patch here with your "usual" big WS, things to check would be:
1) Ful WS refresh (you can enable it on startup in preferences) as reported by `org.eclipse.core.internal.refresh.RefreshJob` should be better or same (to see numbers, one has to enable tracing with "org.eclipse.core.resources/refresh" option).
2) Full WS build numbers would be interesting (haven't measured that yet). Should be better or same.
3) If you have something like "initial project set import", try that too, as there is lot of work happens while discovering new resources. Should be better or same.
4) Close all projects and open them all again - this will automatically trigger refresh. Should be better or same.

iloveeclipse:
> - can we get rid of "faster" iteration on `subMap` completely (it was written ~2002),

Yes and no. We need ordered iteration (see below), but we can use sorted keyset for that.

> how much aliases can we have at all

It really depends on projects structure / content. I see huge performance improvement with only ~40 aliases used but a huge workspace tree.

> shouldn't be a problem today to iterate over few entries more?

The way how `AliasManager.LocationMap.overLappingResourcesDo(Consumer<IResource>)` and  `AliasManager.LocationMap.matchingPrefixDo(IFileStore, Consumer<IResource>)` implemented **require** iteration over **sorted** keys or entries on entire map or **sorted** subset of it.

iloveeclipse:
So far I see most improvements in the "big project import" case. There is lot of work in AliasManager in that case, and most of the time spent there disappears with the patch. Anything else I've tried in workspace operations seem to be mostly not affected by the patch.

I plan to submit this PR before M3, I believe the proposed solution should be safe from both functional and  performance point of view.

fedejeanne:
> So if you could simply try the patch here with your "usual" big WS, things to check would be:
> 
>     1. Ful WS refresh (you can enable it on startup in preferences) as reported by `org.eclipse.core.internal.refresh.RefreshJob` should be better or same (to see numbers, one has to enable tracing with "org.eclipse.core.resources/refresh" option).
> 
>     2. Full WS build numbers would be interesting (haven't measured that yet). Should be better or same.
> 
>     3. If you have something like "initial project set import", try that too, as there is lot of work happens while discovering new resources. Should be better or same.
> 
>     4. Close all projects and open them all again - this will automatically trigger refresh. Should be better or same.

Hi, sorry for the late response.

I tried some of the cases you mentioned but I must be testing it wrong because I barely get any activity from the `AliasManager`.

Here's what I tried:

I enabled these 2
image

image

I have **2096** projects in my WS.

Regarding point Nr. (initial import): I haven't tested that because that currently involves downloading everything from SVN and importing the projects (we do it automatically right after installing a new IDE with our Oomph). I can test that later if it's relevant.

I tested these scenarios in the I-Builds **20251024-0020** and **20251113-1020**:
- Opening the WS (full refresh enabled)
- Select all projects > Right click > Refresh
- Close all projects
- Open all projects (with and without full build)

Here are the samples: [Samples.zip](https://github.com/user-attachments/files/23546570/Samples.zip)

As I said: almost no activity from the  `AliasManager` in them :-\

iloveeclipse:
I saw the most work on project import case, and (at least in ou case) with projects that have classpath entries pointing outside the project/workspace locations (so ExternalFolderManager links them and aliases are created). During import of such projects AliasManager is (was) one of the hot spots.

fedejeanne:
> [...] with projects that have classpath entries pointing outside the project/workspace locations [...]

Hm, our WS is actually pretty self-contained: every project is in a (sub-)folder inside the workspace. Same as with the "contributors" IDE where everything goes inside the `git` folder in the workspace:

```
ws 
  '- git
      '- eclipse
      '- eclipse.jdt
      '- ...
      '- eclipse.platform
      '- ...
```

I guess that's why I didn't get much activity from the `AliasManager`.

PR #2253, opened 2025-11-05
DO NOT MERGE, brute force change in LocationMap

iloveeclipse:
LocationMap should use HashMap for better performance

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2251

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
@@ -24,10 +24,9 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import org.eclipse.core.filesystem.EFS;
@@ -134,7 +133,7 @@ public class AliasManager implements IManager, ILifecycleListener, IResourceChan
 		/**
 		 * Map of FileStore-&gt;IResource OR FileStore-&gt;ArrayList of (IResource)
 		 */
-		private final SortedMap<IFileStore, Object> map = new TreeMap<>(IFileStore::compareTo);
+		private final Map<IFileStore, Object> map = new HashMap<>(10_000);
 
 		/**
 		 * Adds the given resource to the map, keyed by the given location.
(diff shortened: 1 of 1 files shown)


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -177,16 +176,7 @@ public void clear() {
...
-			if (prefixParent != null) {
-				//endPoint is the smallest possible path greater than the prefix that doesn't
-				//match the prefix
-				IFileStore endPoint = prefixParent.getChild(prefix.getName() + "\0"); //$NON-NLS-1$
-				matching = map.subMap(prefix, endPoint);
-			} else {
-				matching = map;
-			}
+			Map<IFileStore, Object> matching = map;
 			for (Object value : matching.values()) {
```
laeubi, line 180: Don't we need to iterate over the entryset here and check that key is actually greater than endpoint if prefixParent != null?

Before we have only passed a subset to the consumer now its always the whole map what makes it feel wrong given the name is "matchingPrefixDo"

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
```diff
@@ -134,7 +133,7 @@ class LocationMap {
 		/**
 		 * Map of FileStore-&gt;IResource OR FileStore-&gt;ArrayList of (IResource)
 		 */
-		private final SortedMap<IFileStore, Object> map = new TreeMap<>(IFileStore::compareTo);
+		private final Map<IFileStore, Object> map = new HashMap<>(10_000);
```
laeubi, line 136: Form the javadoc

>  Iteration over collection views requires time proportional to the "capacity" of the HashMap instance (the number of buckets) plus its size (the number of key-value mappings). Thus, **it's very important not to set the initial capacity too high** (or the load factor too low) **if iteration performance is important**. 

Maybe we should start smaller (or use a system property to control the inital size)

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/AliasManager.java
iloveeclipse: **Please** don't review draft PR's with the prefix "DON'T MERGE".
This one is simply to see if there are any tests available that validate exactly this removed code.
See https://github.com/eclipse-platform/eclipse.platform/pull/2252#issuecomment-3492279775

iloveeclipse:
Good to know there **are** tests for this code and they also fail if something is wrong :-)
With that, this PR is not needed anymore, the solution is https://github.com/eclipse-platform/eclipse.platform/pull/2252

PR #2256, opened 2025-11-07
Improve copilot instructions and setup

laeubi:


laeubi:
Setup steps successful tested, the rest will not be tested by our builds.

vogella:
Better to use Agenda.md as discussed before

Issue #2261, opened 2025-11-10
'About Eclipse SDK' dialog has duplicate icons, can this be merged into 1 icon showing full list


deepika-u:
Description
When i open the "About Eclipse SDK" -> Help menu, i see the below dialog

<img width="688" height="606" alt="Image" src="https://github.com/user-attachments/assets/7d98daf0-20cd-4d8c-be3b-6d996c3bec9a" />

The first 2 icons both show the same tooltip as "Eclipse.org". 

1.  When i click on the 1st Icon(not so good looking among both) i see below features listed like SDK, RCP, JDT, Platform, Help.

<img width="873" height="990" alt="Image" src="https://github.com/user-attachments/assets/243a0e62-38ca-4a1d-a4cb-c213bf3d5127" />

2. When i click on the 2nd Icon(good looking icon) i see PDE feature and its details in the similar lines.

<img width="871" height="988" alt="Image" src="https://github.com/user-attachments/assets/780a48bf-f0e7-4e77-b691-fbaa515dae66" />

Can we we have 2nd one good looking icon and have all the list populated in it?
Is this expected behavior or can this be merged into 1 entry?
I see this behavior on Windows 11 and on below build ```
Version: 2025-12 (4.38)
Build id: I20251108-1800
```

deepika-u:
I just tried on 4.32 and see as below

<img width="582" height="530" alt="Image" src="https://github.com/user-attachments/assets/00c79d3a-035f-4ece-800f-fa96f4c8a4d8" />

and when clicked on 1st icon i see as below

<img width="1127" height="914" alt="Image" src="https://github.com/user-attachments/assets/ebe819f1-1d48-4b23-b7fe-d823f787f166" />

I see all the features and their versions listed here. I also see 1 other "Eclipse User Storage" here.

iloveeclipse:
@merks, @laeubi, @HannesWell : any idea why PDE entry is separated now? Some magic p2 instructions missing / too much / wrong? What I also observe, the icon for PDE is different - is **that** the root cause?

laeubi:
Last time I looked into this it is dependent on the branding plugin used for a feature, that is offering the icon and the about text.

merks:
Yes, it's on my TODO list to figure out why these two are not merged.   I'll try to get to this as soon as possible; this is not yet at the top of the queue, but getting there...

laeubi:
[...]

PR #2269, opened 2025-11-12
Use exactly the same eclipse32.svg as from org.eclipse.platform

merks:
https://github.com/eclipse-platform/eclipse.platform/issues/2261


merks:
The quality gate problem is unrelated to this change.

PR #2271, opened 2025-11-13
Add comprehensive README.md for Terminal feature

laeubi:


laeubi:
Only documentation, so build will not check this.

PR #2276, opened 2025-11-14
Enhance 'Find' context in Expression View

SougandhS:
Make 'Find' context in Expressions view appears only when valid expressions exist.

Before :
image

After :
image

image

### DIFF
--- a/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.ui; singleton:=true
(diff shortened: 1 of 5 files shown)


laeubi:
@SougandhS also not directly related but can we maybe add an icon to the find menu item= it looks a bit misplaced as it is the only one without one.

SougandhS:
> @SougandhS also not directly related but can we maybe add an icon to the find menu item= it looks a bit misplaced as it is the only one without one.

Sure, will add one 👍

SougandhS:
Added find icon 
image

iloveeclipse:
Not sure why didn't you also disable "Copy Expressions" for same reason?

SougandhS:
> Not sure why didn't you also disable "Copy Expressions" for same reason?

All the `VirtualCopyToClipboardActionDelegate` implementations have the same issue, showing copy bp, copy stack.. etc on empty selection, I will fix that in another PR.

SougandhS:
Hi @laeubi, could you please merge this too ?

SougandhS:
Thank you !

PR #2277, opened 2025-11-14
JDT Clean-up Use String.replace() instead of String.replaceAll() when

vogella:
possible

In https://github.com/eclipse-platform/eclipse.platform.ui/issues/3517 we discuss which automatic clean-ups we should add. This is the result of a manual run in eclipse-platform to see if this clean-up has issues. Also by running it manually before enabling the clean-ups we avoid creating PR per bundle.

### DIFF
--- a/ant/org.eclipse.ant.ui/META-INF/MANIFEST.MF
+++ b/ant/org.eclipse.ant.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ant.ui; singleton:=true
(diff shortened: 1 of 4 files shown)


vogella:
Clean-up has issues https://github.com/eclipse-platform/eclipse.platform.ui/pull/3544

vogella:
> Clean-up has issues [eclipse-platform/eclipse.platform.ui#3544](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3544)

No this one is fine, thanks to verification from @merks

vogella:
Converrting to draft, as IMHO we should not apply such clean-ups shortly before a release.

vogella:
These changes also look fine to me.

akurtakov:
Do you plan to finish this one?

PR #2279, opened 2025-11-14
Enhance stack trace information for rule conflicts in BuildManager

### LINKED ISSUE #2278: Attempted to beginRule: MultiRule[], does not match outer scope rule: R/
I sometimes see Attempted to beginRule: MultiRule[], does not match outer scope rule: R/ in the log but from the stack trace it is not clear who is doing something wrong here. 

````
java.lang.IllegalArgumentException: Attempted to beginRule: MultiRule[], does not match outer scope rule: R/
	at org.eclipse.core.runtime.Assert.isLegal(Assert.java:68)
	at org.eclipse.core.internal.jobs.ThreadJob.illegalPush(ThreadJob.java:144)
	at org.eclipse.core.internal.jobs.ThreadJob.push(ThreadJob.java:416)
	at org.eclipse.core.internal.jobs.ImplicitJobs.begin(ImplicitJobs.java:67)
	at org.eclipse.core.internal.jobs.JobManager.beginRule(JobManager.java:358)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:291)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:355)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:449)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:452)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:564)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:512)
	at org.eclipse.core.internal.events.BuildManager.build(BuildManager.java:594)
	at org.eclipse.core.internal.events.AutoBuildJob.doBuild(AutoBuildJob.java:208)
	at org.eclipse.core.internal.events.AutoBuildJob.run(AutoBuildJob.java:309)
````
If possible we should add more information here, who is responsible for this.

### PR DESCRIPTION
**Pull request by vogella**
Fixes #2278

When a builder attempts to begin or end a scheduling rule that doesn't match the outer scope rule, the error message now includes:
- Builder name and label
- Builder class name
- Plugin ID
- Project name
- The conflicting rule

This makes it much easier to identify which builder is causing the rule mismatch, especially in complex build scenarios with multiple builders.

The enhancement catches IllegalArgumentException at the point where beginRule() and endRule() are called in BuildManager.basicBuild() and wraps it with detailed context before re-throwing.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
@@ -288,7 +288,11 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 				depth = getWorkManager().beginUnprotected();
 				// Acquire the rule required for running this builder
 				if (rule != null) {
-					Job.getJobManager().beginRule(rule, monitor);
+					try {
+						Job.getJobManager().beginRule(rule, monitor);
+					} catch (IllegalArgumentException e) {
+						throw handleRuleConflict(true, currentBuilder, e);
+					}
 					// Now that we've acquired the rule, changes may have been made concurrently, ensure we're pointing at the
 					// correct currentTree so delta contains concurrent changes made in areas guarded by the scheduling rule
 					if (currentTree != null) {
@@ -303,7 +307,11 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 					getWorkManager().endUnprotected(depth);
 				}
 				if (rule != null) {
-					Job.getJobManager().endRule(rule);
+					try {
+						Job.getJobManager().endRule(rule);
+					} catch (IllegalArgumentException e) {
+						throw handleRuleConflict(false, currentBuilder, e);
+					}
 				}
 				// Be sure to clean up after ourselves.
 				if (clean || currentBuilder.wasForgetStateRequested()) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java [hunk 1]
```diff
@@ -288,7 +288,29 @@ private void basicBuild(int trigger, IncrementalProjectBuilder builder, Map<Stri
 				depth = getWorkManager().beginUnprotected();
 				// Acquire the rule required for running this builder
 				if (rule != null) {
-					Job.getJobManager().beginRule(rule, monitor);
+					try {
```
iloveeclipse, line 291: Looks like two blocks differ only in string arguments. Can you please extract them to a static method `handleRuleConflicts(boolean beginRule, InternalBuilder currentBuilder, IllegalArgumentException e)`?

Could you please also add some example or just a test how the result error would look like?

vogella:
Old (without the patch what the JobManager throws raw):                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                        
Begin-rule conflict:                                                                                                                                                                                                                                                                  
  Attempted to beginRule: P/Foo/bar, does not match outer scope rule: P/Other                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                        
End-rule conflict:                                                                                                                                                                                                                                                                    
  Attempted to endRule: P/Foo/bar, does not match most recent begin: P/Foo.  See log for trace information if rule tracing is enabled.                                                                                                                                                  
                                                                                                                                                                                                                                                                                        
\
Nothing in the message or stack reveals which builder caused it. You see "some builder requested a wrong rule" with no way to find out which one without enabling rule tracing.    
                                                                                                                                                                                                                                                                                         New (with the patch):                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                        Begin-rule conflict:
  beginRule failed for builder com.example.MyBuilder ('My Builder', plugin com.example) on project /Foo: Attempted to beginRule: P/Foo/bar, does not match outer scope rule: P/Other

End-rule conflict:                                                                                                                                                                                                                                                                    
  endRule failed for builder com.example.MyBuilder ('My Builder', plugin com.example) on project /Foo: Attempted to endRule: P/Foo/bar, does not match most recent begin: P/Foo.  See log for trace information if rule tracing is enabled.

\
The original IllegalArgumentException is preserved as the cause, so the original stack and message are still fully accessible. The visible win: builder class, label, plugin id, and project path.

PR #2281, opened 2025-11-15
Replace com.ibm.icu dependency with org.eclipse.ant.core in expression

vogella:
tests

The ExpressionTestsPluginUnloading test was using com.ibm.icu bundle for testing plugin loading/unloading and classloader isolation. This change replaces it with org.eclipse.ant.core bundle to support removal of com.ibm.icu from the platform.

Changes:
- Replace fake com.ibm.icu.text.DecimalFormat with org.eclipse.ant.core.AntRunner
- Update test to use org.eclipse.ant.core bundle instead of com.ibm.icu
- Adjust instanceof checks to match new bundle's class interfaces
- org.eclipse.ant.core is a suitable replacement as it can be safely stopped/restarted without affecting test execution


merks:
Of course I have nothing again this change to a test.

There are some actual uses outside of test:

image

akurtakov:
@vogella Would you please take care of the ECA failure?

PR #1962, opened 2025-06-16
Add grouping for breakpoints by Active/In-Active status

SougandhS:
This PR adds a new grouping feature in the Eclipse Debug view that organizes breakpoints based on their enabled or disabled status. Enabled breakpoints are shown at the top as Active, and disabled ones are listed below as In-Active. This makes it easier for developers to find and manage breakpoints during debugging. When working with many breakpoints, it can be hard to find disabled ones again to enable, especially after scrolling through a long list. Users often end up deleting and re-adding breakpoints just to bring them to the top, which can cause loss of settings like hit counts or conditions. This feature helps avoid that by keeping enabled breakpoints easily visible and improving overall breakpoint management

https://github.com/user-attachments/assets/adf16e40-0605-4e0e-9872-d86cec2c35aa

Screens : 
image <br>

image

Supports advance grouping too
image <br>

image

PR #2300, opened 2025-12-08
Improve quality of breakpoint enablement icon

HeikoKlare:
The icon for breakpoint enablement is currently composed of three PNG images, which leads to blurry results when scaling it. There exist similar images (with the contained circles just colored differently) as properly vectorized SVGs. As a side note: the blue tone does also not perfectly fit to the blue tone used in the other similar icons.

This change proposes a replacement of the breakpoint enablement icon which adapts the existing, fully vectorized icon for breakpoint groups.

@SougandhS the SVG was introduced with:
- https://github.com/eclipse-platform/eclipse.platform/pull/1962

I guess it was just created like that by accident and not on purpose, was it? Could you please check whether the proposed replacement fits to how you wanted it to look like?

### Screenshots
Appearance at 100% and 175%

#### Before
image
image

#### After
image
image


SougandhS:
Looks great now 👍, 
Thanks for the update !

PR #2307, opened 2025-12-10
Fix encoding for 'compare with clipboard' if default encoding is not UTF-8

wahlbrink:
Fixes encoding of the contents for 'compare with clipboard' if default encoding (of workspace) is not UTF-8.

Before (workspace default is set to CP-1252):
before

Fixed:
after


SougandhS:
Hi @vogella, could you please merge this PR ?

PR #2311, opened 2025-12-12
Honor expression context when creating Watches

SougandhS:
When selecting a field under an object, both the Watch command and Drag & Drop previously created a non-contextual expression using only the field name, which causes evaluation failures. This fix ensures that both Watch and Drag & Drop now generate proper context-aware expressions such as myobj.myfield instead of just the field name.

Fixes : https://bugs.eclipse.org/bugs/show_bug.cgi?id=321289

##  Before [Watch by context and by drag-drop] : 

https://github.com/user-attachments/assets/54c6145c-2fd5-4847-abed-cc2b67e3c819

## After [Watch by context and by drag-drop] : 

https://github.com/user-attachments/assets/e742ab33-e4af-4a5c-b640-ae43b8e5ff13

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2015 Wind River Systems and others.
+ * Copyright (c) 2008, 2026 Wind River Systems and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,10 +10,12 @@
  *
  * Contributors:
  *     Wind River Systems - initial API and implementation
+ *     IBM Corporation - Improved expression creation
  *******************************************************************************/
 package org.eclipse.debug.internal.ui.actions.expressions;
 
-import java.util.Iterator;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
@@ -33,6 +35,8 @@ import org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapter;
 import org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapter2;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.viewers.TreePath;
+import org.eclipse.jface.viewers.TreeSelection;
 import org.eclipse.ui.IViewPart;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.PartInitException;
@@ -49,12 +53,39 @@ public class WatchHandler extends AbstractHandler {
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		ISelection selection = HandlerUtil.getCurrentSelection(event);
-		if (selection instanceof IStructuredSelection) {
-			Iterator<?> iter = ((IStructuredSelection)selection).iterator();
-			while (iter.hasNext()) {
-				Object element = iter.next();
-				createExpression(element);
+		if (selection instanceof IStructuredSelection structuredSelection) {
+			if (structuredSelection instanceof TreeSelection treeSelection) {
+				for (TreePath path : treeSelection.getPaths()) {
+					List<IVariable> variables = new ArrayList<>();
+					if (path.getSegmentCount() > 1) {
+						for (int e = 0; e < path.getSegmentCount(); e++) {
+							IVariable variable = (IVariable) path.getSegment(e);
+							variables.add(variable);
+						}
+						IWatchExpressionFactoryAdapter2 factory = getFactory2(variables);
+						if (factory != null) {
+							boolean canCreate = factory.canCreateWatchExpression(variables);
+							if (canCreate) {
+								try {
+									String expression = factory.createWatchExpression(variables);
+									createWatchExpression(expression);
+								} catch (CoreException e) {
+									DebugPlugin.log(e);
+									break;
+								}
+							}
+						}
+					} else {
+						Object element = path.getFirstSegment();
+						createExpression(element);
+					}
+				}
+			} else {
+				for (Object element : structuredSelection.toArray()) {
+					createExpression(element);
+				}
 			}
+			showExpressionsView();
 		}
 		return null;
 	}
@@ -96,9 +127,12 @@ public class WatchHandler extends AbstractHandler {
 			DebugUIPlugin.errorDialog(DebugUIPlugin.getShell(), ActionMessages.WatchAction_0, ActionMessages.WatchAction_1, e); //
 			return;
 		}
+		createWatchExpression(expressionString);
+	}
 
+	private void createWatchExpression(String expressionString) {
 		IWatchExpression expression;
-			expression = DebugPlugin.getDefault().getExpressionManager().newWatchExpression(expressionString);
+		expression = DebugPlugin.getDefault().getExpressionManager().newWatchExpression(expressionString);
 		DebugPlugin.getDefault().getExpressionManager().addExpression(expression);
 		IAdaptable object = DebugUITools.getDebugContext();
 		IDebugElement context = null;
@@ -108,9 +142,7 @@ public class WatchHandler extends AbstractHandler {
 			context = ((ILaunch) object).getDebugTarget();
 		}
 		expression.setExpressionContext(context);
-		showExpressionsView();
 	}
-
 
 	/**
 	 * Returns the factory adapter for the given variable or <code>null</code> if none.
@@ -131,6 +163,16 @@ public class WatchHandler extends AbstractHandler {
 	static IWatchExpressionFactoryAdapter2 getFactory2(Object element) {
 		if (element instanceof IAdaptable) {
 			return ((IAdaptable)element).getAdapter(IWatchExpressionFactoryAdapter2.class);
+		}
+		if (element instanceof List<?> ExpressionList) {
+			for (Object obj : ExpressionList) {
+				if (!(obj instanceof IAdaptable)) {
+					return null;
+				}
+			}
+			if (ExpressionList.getFirst() instanceof IVariable variable) {
+				return variable.getAdapter(IWatchExpressionFactoryAdapter2.class);
+			}
 		}
 		return null;
 	}
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/expression/ExpressionDropAdapter.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/expression/ExpressionDropAdapter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2013 IBM Corporation and others.
+ * Copyright (c) 2007, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -40,6 +40,8 @@ import org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapter2;
 import org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapterExtension;
 import org.eclipse.jface.util.LocalSelectionTransfer;
 import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.viewers.ITreeSelection;
+import org.eclipse.jface.viewers.TreePath;
 import org.eclipse.jface.viewers.ViewerDropAdapter;
 import org.eclipse.swt.dnd.DND;
 import org.eclipse.swt.dnd.DropTargetEvent;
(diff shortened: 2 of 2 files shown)


SougandhS:
Hi @merks, could you please check this change when you get a moment ?

SougandhS:
Hi @iloveeclipse, could you please trigger a co-pilot review for this PR ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,11 +50,26 @@ public class WatchHandler extends AbstractHandler {
...
+							DebugUIPlugin.log(e1);
+						}
+					}
+					expressionString.deleteCharAt(expressionString.length() - 1);
+					createWatchExpression(expressionString.toString());
+				} else {
+					Object element = path.getFirstSegment();
+					createExpression(element);
+				}
+				showExpressionsView();
```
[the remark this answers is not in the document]
SougandhS: moved 👍

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,11 +50,26 @@ public class WatchHandler extends AbstractHandler {
...
 			}
 		}
 		return null;
```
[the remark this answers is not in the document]
SougandhS: done

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/expression/ExpressionDropAdapter.java
```diff
@@ -373,19 +376,28 @@ private boolean performExpressionDrop(IStructuredSelection selection) {
...
+						IVariable variable = (IVariable) path.getSegment(e);
+						try {
+							expressionString.append(variable.getName());
+							expressionString.append("."); //$NON-NLS-1$
+						} catch (DebugException e1) {
+							DebugUIPlugin.log(e1);
+						}
+					}
+					expressionString.deleteCharAt(expressionString.length() - 1);
+					expression = createExpression(expressionString.toString());
```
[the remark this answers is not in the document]
SougandhS: IndexedVariablePartition cant be used for creating watch expressions and cant be dragged onto expressions, so no need to handle it.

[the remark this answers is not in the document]
SougandhS: Added guards

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/expression/ExpressionDropAdapter.java
```diff
@@ -373,19 +376,28 @@ private boolean performExpressionDrop(IStructuredSelection selection) {
...
+					Object element = path.getFirstSegment();
+					String expressionText = createExpressionString(element);
+					expression = createExpression(expressionText);
 				}
-			} else {
-				return false;
+				expressions.add(expression);
 			}
 		}
 		if (expressions.size() == selection.size()){
```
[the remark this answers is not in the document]
SougandhS: Added 👍

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,11 +50,26 @@ public class WatchHandler extends AbstractHandler {
...
+						IVariable variable = (IVariable) path.getSegment(e);
+						try {
+							expressionString.append(variable.getName());
+							expressionString.append("."); //$NON-NLS-1$
+						} catch (DebugException e1) {
+							DebugUIPlugin.log(e1);
+						}
+					}
+					expressionString.deleteCharAt(expressionString.length() - 1);
+					createWatchExpression(expressionString.toString());
```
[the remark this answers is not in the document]
SougandhS: done

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,13 +51,41 @@ public class WatchHandler extends AbstractHandler {
...
+						boolean hasError = false;
+						for (int e = 0; e < path.getSegmentCount(); e++) {
+							IVariable variable = (IVariable) path.getSegment(e);
+							try {
+								expressionString.append(variable.getName());
+								expressionString.append("."); //$NON-NLS-1$
+							} catch (DebugException e1) {
+								DebugUIPlugin.log(e1);
+								hasError = true;
+								break;
```
trancexpress: This is probably more simple?

```
	try {
		StringBuilder expressionString = new StringBuilder();
		for (int e = 0; e < path.getSegmentCount(); e++) {
			IVariable variable = (IVariable) path.getSegment(e);
			expressionString.append(variable.getName());
			expressionString.append("."); //$NON-NLS-1$
		}
		expressionString.deleteCharAt(expressionString.length() - 1);
		createWatchExpression(expressionString.toString());
	} catch (DebugException e1) {
		DebugUIPlugin.log(e1);
	}
```

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,13 +51,41 @@ public class WatchHandler extends AbstractHandler {
...
+			if (structuredSelection instanceof TreeSelection treeSelection) {
+				for (TreePath path : treeSelection.getPaths()) {
+					if (path.getSegmentCount() > 1) {
+						StringBuilder expressionString = new StringBuilder();
+						boolean hasError = false;
+						for (int e = 0; e < path.getSegmentCount(); e++) {
+							IVariable variable = (IVariable) path.getSegment(e);
+							try {
+								expressionString.append(variable.getName());
+								expressionString.append("."); //$NON-NLS-1$
```
trancexpress: One problem I see with this is that the code is in platform, not in JDT. So I'm not sure about `.` here, technically the platform code is a framework to support any language. Are there other cases where such concatenation is used? Do they also hard-code `.`?

Maybe there is a way to check we are in a Java launch / context?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,13 +51,41 @@ public class WatchHandler extends AbstractHandler {
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		ISelection selection = HandlerUtil.getCurrentSelection(event);
-		if (selection instanceof IStructuredSelection) {
-			Iterator<?> iter = ((IStructuredSelection)selection).iterator();
-			while (iter.hasNext()) {
-				Object element = iter.next();
-				createExpression(element);
+		if (selection instanceof IStructuredSelection structuredSelection) {
```
trancexpress, line 56: If we continue with this (see below, I'm not sure we can), it would be nice to extract the code (and in the DND handler too) and add a test for it. This entire class is not tested, from what I can tell.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: Yes, it was like this before then Co-pilot suggest -> https://github.com/eclipse-platform/eclipse.platform/pull/2311#discussion_r2930447704 & https://github.com/eclipse-platform/eclipse.platform/pull/2311#discussion_r2930447646

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: How about we provide a preference option for that with default as ".", then JDT or CDT can modify that in debug preference UI or from their preference initilializer itself by accessing that constant ? --> this way clients dont need to re-implement this again..

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: if possible could you please share such sample tests in platform for reference ? I dont have much idea in writing tests for drag/drop UI changes... :)

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
trancexpress: I'm not sure this helps, since I can have both CDT and JDT installed in an Eclipse and debug with both.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
trancexpress: The Copilot suggestion was due to the missing `break` statement, I assume. With the above the code will not do anything after any `variable.getName()` throws.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
trancexpress: I'm not sure about the DND handler, looking at it I'm not sure it has the same code for creating expressions (for whatever reason).

For this handler, I was thinking about extracting the code and testing the extracted code. But I'm not sure that can be set up, even that will require to select something in the variables view... And I'm not seeing even one test use `ID_VARIABLE_VIEW`.

Lets leave tests out then.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
trancexpress: What about implementing this at `org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapter2.createWatchExpression(Object)`?

Try calling `org.eclipse.debug.ui.actions.IWatchExpressionFactoryAdapter2.canCreateWatchExpression(Object)` with `IVariable[]` gathered from the tree selection. If it returns `true`, call `createWatchExpression` with the array and implement this on the JDT side? The bulk of the new code will remain here, the concatenation will be delegated to the respective implementation.

For CDT I'm not sure, at some point debugging was re-implemented there. I'm not sure how much they use from the platform at this point.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: but once the exception is thrown won't the context flow be lost ?

eg. 
take --> obj1.obj2.obj3

got exception while processing obj2.
then this will create wrong expression..

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: That sounds great. I'll check this way

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
trancexpress: Nothing will be done with the expression, since the rest of the code (which uses the expression) wont run.

SougandhS: Okay then... will go with the initial way (same you suggested)

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: Hi @trancexpress,
I have implemented the way you suggested. Now its clean - Thanks!
 https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/926

trancexpress: While trying it out, I noticed this:

image

Snippet is:

```
package tst;

public class Tst2 {

	private class X {
		private int x = 0;
	}

	private class Y {
		private X x = new X();
		private int y = 0;
	}

	private class Z {
		private Y y = new Y();
		private int z = 0;
	}

		new Tst2().foo();
	}

	public void foo() {
		X x = new X();
		Y y = new Y();
		Z z = new Z();
		System.out.println(x + " " + y + " " + z);
	}
}
```

Breakpoint at the `System.out` line, then select `y.y` and `z.y.y` in the Variables view.

Could you take a look?

I'll check the JDT debug PR, this one looks great now!

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
```diff
@@ -49,12 +53,39 @@ public class WatchHandler extends AbstractHandler {
...
+			if (structuredSelection instanceof TreeSelection treeSelection) {
+				List<IVariable> variables = new ArrayList<>();
```
trancexpress: This list needs to be moved inside the `for (TreePath path : treeSelection.getPaths())` loop, otherwise we keep adding variables of different selected paths to it (see my previous comment and the screenshot).

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: Fixed 👍

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/actions/expressions/WatchHandler.java
SougandhS: Thanks, fxd this 👍

SougandhS:
Thank you @iloveeclipse & @trancexpress for the review & approval 👍

SougandhS:
Hi @trancexpress Build looks good, could u please merge this PR ?

SougandhS:
Thanks again !

trancexpress:
> Thanks again !

Thank YOU for all the improvements :)

PR #2316, opened 2025-12-19
Move bundles to Java 21

akurtakov:
They already depend on bundles that have BREE Java 21.


akurtakov:
Merging despite new warnings for Java 21 deprecations. They will have to be handled independently.

PR #2317, opened 2025-12-19
Bump bundles to Java 21

akurtakov:
They already depend on bundles with Java 21 BREEs.


akurtakov:
New warnings for Java 21 deprecations have to be handled separately. Merging.

Issue #2323, opened 2025-12-22
IFileTest#testIssue2290() fails sporadically on Windows


HeikoKlare:
Test was recently introduced by:
- https://github.com/eclipse-platform/eclipse.platform/pull/2291

It fails on Windows in ...
- GHA runs, e.g.,https://github.com/eclipse-platform/eclipse.platform/pull/2321/checks?check_run_id=58703733928
- I-Build test runs, e.g., https://download.eclipse.org/eclipse/downloads/drops4/I20251220-1800/testresults/html/org.eclipse.core.tests.resources_ep439I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html
- Locally: 
<img width="357" height="192" alt="Image" src="https://github.com/user-attachments/assets/86cc8b0b-a867-463a-97f9-f6912c40cd9f" />

```
org.eclipse.core.runtime.CoreException:
at org.eclipse.core.tests.resources.regression.IFileTest.testIssue2290(IFileTest.java:176)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

@basilevs can you please have a look?

iloveeclipse:
Interesting. 
On my Windows 11 (whatever is latest provided by Microsoft), test run in latest SDK is just fine (tried with Java 21 / 25).

HeikoKlare:
For me it fails sporadically when executed locally (like 10% of the runs or so).

iloveeclipse:
I see it. The delay of 1000 ms is simply too small for Windows.

iloveeclipse:
Removing delay results immediately in the test failure like on eclipse.org SDK build.

basilevs:
> Removing delay results immediately in the test failure like on eclipse.org SDK build.

Thank you for the analysis. I'm removing a dependency on job startup timing in #2324.

Sorry for the disruption.

basilevs:
Even with fixed timings, [the test still fails](https://github.com/eclipse-platform/eclipse.platform/actions/runs/20443567836/job/58741947675#step:5:19675). I've enhanced the failure message.

basilevs:
@akurtakov this is not fixed yet. Sorry about incorrect commit message.

basilevs:
[...]

PR #2324, opened 2025-12-22
Ensure the background payload starts #2323

basilevs:
If background job with concurrent access payload does not have a chance to start, the test harness kills it prematurely and produces a false positive.

This change avoids a dependency on a job startup timing.
As the test is failing even with these fixes, additional details are added to failure message.

Related to #2323.


basilevs:
The new failure message may help to localize the problem. Please merge.

basilevs:
Job cancellation was occasionally reported as a failure. I've fixed that in #2328

PR #2329, opened 2025-12-24
Delete IHelp API

akurtakov:
Deprecated since 3.0 and marked for removal since 2023-12. One should use org.eclipse.help.HelpSystem#getTocs() and
org.eclipse.help.HelpSystem#getContext(String).


akurtakov:
Starting I-build now to be on the safe side.

PR #2332, opened 2025-12-28
Remove useless assert messages

akurtakov:


akurtakov:
Test succeeded on second run but published results are not updated. Merging.

PR #2334, opened 2025-12-29
Fix feature copyright properties

merks:
-Remove license details that come from the license feature.


merks:
The windows-specific failure is unrelated to the changes.

PR #2363, opened 2026-01-02
Clean up resource tests

HeikoKlare:
- Remove obsolete assertion messages
- Replace assertTrue(!condition) with assertFalse(condition)
- Remove commented out code / long commented out test methods


HeikoKlare:
image

This is expected as the test case is removed due to the implementation being commented out for more than a decade.

PR #2373, opened 2026-01-02
Suppress warnings in resource tests explicitly testing deprecated API

HeikoKlare:
Several resource tests explicitly test API that is deprecated. Since those tests are expected to access deprecated methods, there should be no warnings that mess up the workspace. This change suppresses the according warnings for several methods that by intent call deprecated methods.


HeikoKlare:
Resolves 64 useless warnings:
image

PR #2377, opened 2026-01-02
Migrate localstore resource tests to JUnit 5 #903

HeikoKlare:
- Switch to JUnit 5 test annotations and assertions
- Use test extensions replacing JUnit 4 test rules


HeikoKlare:
image

The test case is intentionally removed as the implementation has been commented out for a very long time.

PR #2379, opened 2026-01-02
Wrap different ways of IResource read-only property access in tests

HeikoKlare:
Some tests still access the deprecated IResource#isReadyOnly() and IResource#setReadOnly(). Some do that by intent to test the deprecated functionality, some have not been adapted to (also) test the new way of accessing that property. This leads to potentially missed test paths and also produces useless warnings.
This change parameterizes the existing callers in tests to test both the legacy API as well as the recommended one..

### DIFF
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/IFileTest.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/IFileTest.java
@@ -19,7 +19,9 @@ import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createInWorkspace;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createRandomContentsStream;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.createTestMonitor;
+import static org.eclipse.core.tests.resources.ResourceTestUtil.isReadOnly;
 import static org.eclipse.core.tests.resources.ResourceTestUtil.isReadOnlySupported;
+import static org.eclipse.core.tests.resources.ResourceTestUtil.setReadOnly;
 import static org.junit.Assume.assumeFalse;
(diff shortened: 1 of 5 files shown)


resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/resources/regression/IFileTest.java
```diff
@@ -103,13 +110,13 @@ public void testBug25662() throws CoreException {
 		IFile file = folder.getFile("file.txt");
 
 		try {
-			folder.setReadOnly(true);
-			assertThat(folder).matches(IResource::isReadOnly, "is read only");
+			setReadOnly(folder, true, apiVersion);
+			assertThat(folder).matches(res -> isReadOnly(res, apiVersion), "is read only");
 			CoreException exception = assertThrows(CoreException.class,
 					() -> file.create(createRandomContentsStream(), true, createTestMonitor()));
 			assertEquals(IResourceStatus.PARENT_READ_ONLY, exception.getStatus().getCode());
```
HeikoKlare, line 117: The test failure caused by this test extension indicates that the behavior on Linux expected by the test is not given when setting the read-only property for a folder via ResourceAttributes instead of using the deprecated `setReadOnly()` method:
```
AutomatedResourceTests AllRegressionTests IFileTest.testBug25662(ReadOnlyApi)[1] CURRENT
expected: <277> but was: <273>
```
The test is about expecting a failure with status `IResourceStatus.PARENT_READ_ONLY` when creating a file in a folder that was marked as read only. But this only seems to happen when the folder was marked as read only via `IResource#setReadOnly()` and not via `IResource#setResourceAttributes()`.

@akurtakov do you think this is something that should be fixed for the Linux implementation or should I just revert the test extension here and keep it limited to the deprecated method `IResource#setReadOnly()` as it was before?

vogella: I suggest to fix the Linux implementation

PR #2396, opened 2026-01-04
Mark test methods for deprecated API as deprecated

HeikoKlare:
For methods testing deprecated API, the according warnings have been suppressed. However, it seems more reasonable to mark those methods as deprecated themselves.
This adapts the annotations of those methods and applies deprecations to further test methods that explicitly test deprecated API.


HeikoKlare:
ftr: this fixed 27 warnings
image

PR #2403, opened 2026-01-04
Migrate main test suite of terminal tests to JUnit 5

HeikoKlare:
Also make Tycho build execute that test suite instead of individual tests.


HeikoKlare:
image

Caused by `VT100EmulatorTest` missing in the test suites. Added with latest commit.

Issue #1033, opened 2023-12-18
Random failing ProcessConsoleTests


HeikoKlare:
`ProcessConsoleTests#testProcessTerminationNotificationWithInputFile()` (and maybe other tests in `ProcessConsoleTests`) are randomly failing when checking for logged errors during teardown, e.g.: https://github.com/eclipse-platform/eclipse.platform/pull/1031/checks?check_run_id=19741209418

```
Test triggered errors. expected:<0> but was:<1>
java.lang.AssertionError: Test triggered errors. expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.debug.tests.console.ProcessConsoleTests.tearDown(ProcessConsoleTests.java:111)
```

This issue is different from #770, in which some assertions during test run failed.

HeikoKlare:
Logging the error instead of the numer has now reveled the actual error: 
```
Status ERROR: org.eclipse.debug.ui code=120 Error logged from Debug UI:  java.io.IOException: Stream Closed
```

See this build: https://github.com/eclipse-platform/eclipse.platform/actions/runs/8705248821/job/23875244761
```
2024-04-16T12:24:06.9724917Z org.eclipse.debug.tests.console.ProcessConsoleTests.testProcessTerminationNotificationWithInputFile -- Time elapsed: 0.074 s <<< FAILURE!
2024-04-16T12:24:06.9726649Z java.lang.AssertionError: 
2024-04-16T12:24:06.9727144Z [logged errors] 
2024-04-16T12:24:06.9728308Z Expecting empty but was: [Status ERROR: org.eclipse.debug.ui code=120 Error logged from Debug UI:  java.io.IOException: Stream Closed]
2024-04-16T12:24:06.9730279Z 	at org.eclipse.debug.tests.console.ProcessConsoleTests.tearDown(ProcessConsoleTests.java:111)
```

HeikoKlare:
The recent failure shows that now the error is logged, but the stack trace that would allow to identify in which code the error occurred is missing. There are multiple places in org.eclipse.debug.ui (and also the potential source `ProcessConsole`) where an IOException because of a closed stream might occur. I have provided #1317 to improve log output with the strack trace to identify the line of code in which the error occurs.

HeikoKlare:
[...]

PR #2406, opened 2026-01-04
Add test.xml to org.eclipse.terminal.test for execution in I-Build

HeikoKlare:
The terminal tests currently lack a `test.xml` required for their nightly execution in the I-Build tests. This change adds them. A follow up change that integrates the test in the aggregation is required (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3590).

### DIFF
--- a/terminal/tests/org.eclipse.terminal.test/build.properties
+++ b/terminal/tests/org.eclipse.terminal.test/build.properties
@@ -16,7 +16,8 @@ bin.includes = META-INF/,\
                .,\
                plugin.properties,\
                plugin.xml,\
-               about.html
+               about.html,\
+               test.xml
 src.includes = teamConfig/,\
                about.html
 pom.model.property.testClass = org.eclipse.terminal.test.AutomatedIntegrationSuite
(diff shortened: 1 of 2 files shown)


HannesWell:
Thanks for this.

In general I think for simple cases we should provide less verbose means to activate I-build test execution, e.g. by a special custom manifest header or alike.

HeikoKlare:
This PR is not validated by CI as it only adds an Ant configuration. The failing test is unrelated and known to be flaky: https://github.com/eclipse-platform/eclipse.platform/issues/1033

> In general I think for simple cases we should provide less verbose means to activate I-build test execution, e.g. by a special custom manifest header or alike.

Yes, that would definitely be good. At best, the I-Build should take the same tests as also taken by GHA builds via Tycho instead of having the additional Ant configurations, but it would be an even longer way to go.

laeubi:
> e.g. by a special custom manifest header or alike

bndlib uses `Test-Cases` header for that purpose. I once looked into it but it seems we are requiring this because the test are actually executed by ant and there are some categories like performance tests. I think most of this could these days be replaced by JUnit Tags and a special test discovery strategy anyways.

Issue #2412, opened 2026-01-07
Failing terminal tests in I-Build


HeikoKlare:
Several (175) terminal tests are failing in I-Build test execution because they are expected to be run with assertions enabled, but the Ant-based execution does not seem to do that by default:
```
No Assertions! Run this code with assertions enabled! (vmargs: -ea)

java.lang.Error: No Assertions! Run this code with assertions enabled! (vmargs: -ea)
at org.eclipse.terminal.internal.model.AbstractITerminalTextDataTest.setUp(AbstractITerminalTextDataTest.java:34)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

It affects every test class derived from `AbstractITerminalTextDataTest`.

See, e.g., https://download.eclipse.org/eclipse/downloads/drops4/I20260106-1800/testresults/html/org.eclipse.terminal.test_ep439I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

akurtakov:
Shouldn't this check be removed? AFAIK JUnit doesn't have such requirements.

HeikoKlare:
Probably, yes. I wanted to take a look whether there is something in the production code or tests that requires JVM assertions to enabled, but we should see when just removing that check.

akurtakov:
A quick look points to cases like https://github.com/eclipse-platform/eclipse.platform/blob/53226b13c1a368e8d32ebb693af7e74c74ce7d2a/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/model/TerminalTextDataStore.java#L65 and many other such cases . 

IMO this is extremely bad pattern as if the JVM is not started with -ea there will be no RuntimeException with all possible problems in calculations coming from that. Even throwing RuntimeException with no further details is a bad idea, what should be done is throwing IllegalArgumentException with a clear message what is wrong and adjust tests.

HeikoKlare:
> IMO this is extremely bad pattern as if the JVM is not started with -ea there will be no RuntimeException with all possible problems in calculations coming from that. Even throwing RuntimeException with no further details is a bad idea, what should be done is throwing IllegalArgumentException with a clear message what is wrong and adjust tests.

[...]

PR #2413, opened 2026-01-07
Drop -ea requirement from AbstractTerminalTextDataTest

akurtakov:
JUnit doesn't require such param be passed to the JVM. Work towards https://github.com/eclipse-platform/eclipse.platform/issues/2412


akurtakov:
I'll merge this one - nothing changes with this change for surefire (due to https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#enableAssertions ) . It will make obvious the cases that have to handled better in terminal tests ( probably reducing failing tests). 

Additionally, it can be considered to run I-build tests with -ea but it would be best after terminal is fixed in order to not hide these problems.

PR #2414, opened 2026-01-08
Replace asserts with in terminal model

akurtakov:
Make the checks happen always and throw detailed
IllegalArgumentException in case of problematic parameter value.

One more step towards https://github.com/eclipse-platform/eclipse.platform/issues/2412


laeubi:
> IllegalArgumentException in case of problematic parameter value.

Just as a note, you can also throw `new AssertionError` if you want to retain the `assert` semantics here. But `IllegalArgument` seem more appropriate here for the cases I reviewed.

akurtakov:
I'll take @laeubi 's comment as approval and merge it.

PR #2418, opened 2026-01-12
Synchronize Console view icon with selected console

SougandhS:
The Console view icon now reflects the currently selected console. This helps users quickly identify different consoles when multiple consoles are present, without having to inspect each one.

Before : 

image

After : 

image

### DIFF
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
@@ -279,18 +279,19 @@ public class ConsoleView extends PageBookView implements IConsoleView, IConsoleL
 	protected void updateIcon() {
 		IConsole console = getConsole();
 		if (console == null) {
+			// Check and restore default console icon if last page is closed
+			if (currentIcon != defaultIcon) {
+				currentIcon = defaultIcon;
+				setTitleImage(currentIcon);
+			}
 			return;
 		}
-		Image newImage = null;
+		Image newImage;
 
-		if (console != null) {
-			ImageDescriptor desc = console.getImageDescriptor();
-			if (desc != null) {
-				newImage = localResManager.create(desc);
-			}
-		}
-
-		if (newImage == null) {
+		ImageDescriptor desc = console.getImageDescriptor();
+		if (desc != null) {
+			newImage = localResManager.create(desc);
+		} else {
 			newImage = defaultIcon;
 		}
 


debug/.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
iloveeclipse: Please remove

iloveeclipse:
Interesting idea. I remember there was a time the default JDT Java Console page had no icon at all. What would happen in such a case?

debug/.project
SougandhS: My bad, this was added by accident - Will remove this

SougandhS:
> Interesting idea. I remember there was a time the default JDT Java Console page had no icon at all. What would happen in such a case?

If no icon, then default `IConsoleConstants.IMG_VIEW_CONSOLE` is used

SougandhS:
@iloveeclipse Build looks good now, could you please merge this change ?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
```diff
@@ -269,6 +273,29 @@ protected void updateHelp() {
 		PlatformUI.getWorkbench().getHelpSystem().setHelp(getPageBook().getParent(), helpContextId);
 	}
 
+	protected void updateIcon() {
+		IConsole console = getConsole();
+		if (console == null) {
+			return;
+		}
+
+		Image newImage = null;
```
iloveeclipse: `= null` is not needed, but there is much bigger issue.

iloveeclipse:
I would probably manage images created for console pages with a map and probably update it in doCreatePage / doDestroyPage, but I havent looked deep enough into the ode to be sure.

laeubi (commented):
You should likely use a `LocalResourceManager` for images.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
SougandhS: It is initialized to detect consoles with no icons

SougandhS:
I have applied `LocalResourceManager` and tested few cases, didn't see any exceptions on logs

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
```diff
@@ -269,6 +276,30 @@ protected void updateHelp() {
...
 	}
 
+	protected void updateIcon() {
+		IConsole console = getConsole();
+		if (console == null) {
+			return;
+		}
+		Image newImage = null;
+
+		if (console != null) {
```
iloveeclipse: The `if (console != null)` condition is always true now. 
Please change to 

```
		Image newImage;
		ImageDescriptor desc = console.getImageDescriptor();
		if (desc != null) {
			newImage = localResManager.create(desc);
		} else {
			newImage = defaultIcon;
		}
```

iloveeclipse (changes requested):
Looks better now, resources problem seem to be fixed, beside small code change left is only this one:

> closing all pages leaves last selected page icon on the console.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
iloveeclipse: Fixed with https://github.com/eclipse-platform/eclipse.platform/pull/2418/commits/659e494f5f5805e43f77bcdc56abde61056db37a

iloveeclipse:
> beside small code change left is only this one:
> 
> > closing all pages leaves last selected page icon on the console.

Fixed remaining issues with https://github.com/eclipse-platform/eclipse.platform/pull/2418/commits/659e494f5f5805e43f77bcdc56abde61056db37a & rebased on master.

iloveeclipse:
@SougandhS : thanks for the idea, I really like it.
Could you please provide N&N entry ?

iloveeclipse:
A nice (unexpected) side effect is that now Console also changes icon if the JUnit tests are running or Ant tasks etc, so I see something like

image



image

Cool. I really like this small change.

SougandhS:
Thank you @iloveeclipse @laeubi for the review

>Could you please provide N&N entry ?

Sure 👍

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
SougandhS: > Fixed with [659e494](https://github.com/eclipse-platform/eclipse.platform/commit/659e494f5f5805e43f77bcdc56abde61056db37a)

Thanks for the commit and apologies for the response delay

Issue #2423, opened 2026-01-14
org.eclipse.terminal.feature does not include org.eclipse.debug.terminal bundle


iloveeclipse:
While validating https://github.com/eclipse-platform/eclipse.platform/issues/2420 I've stumbled over the fact that even if we have `org.eclipse.terminal.feature` installed in the SDK, it doesn't contain `org.eclipse.debug.terminal` bundle (which contributes also the console page).

I assume this is just an oversight and will push a PR adding the bundle to the feature.

laeubi:
> I assume this is just an oversight and will push a PR adding the bundle to the feature.

This is intentional, it does not belongs to the classical terminal feature and is an own functionality just building up on it.

Both feature can be used independently and only share the name and some common bundles.

iloveeclipse:
> This is intentional, it does not belongs to the classical terminal feature and is an own functionality just building up on it.

I'm confused why a "Terminal" console contibution shouldn't be a part of the "Terminal" feature. It depends on Terminal and provides Terminal related functionality.

laeubi:
The terminal feature provides the terminal view + connectors. It can be used anywhere and even without the console feature of eclipse.debug.

The terminal console is an extension to the Eclipse Console View and can also be used independently of the terminal view.

Just because two things are related does not mean they belong together.

merks:
Probably part of the reason for the question is related to whether or not it's in the Eclipse SDK product...

iloveeclipse:
Actually I see a big issue on Windows with this bundle (https://github.com/eclipse-platform/eclipse.platform/issues/2428), so I don't want to add it to the platform anymore :-.)
Closing as won't fix.

PR #2424, opened 2026-01-14
Add org.eclipse.debug.terminal bundle to org.eclipse.terminal.feature

iloveeclipse:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2423

### DIFF
--- a/terminal/features/org.eclipse.terminal.feature/feature.xml
+++ b/terminal/features/org.eclipse.terminal.feature/feature.xml
@@ -6,6 +6,7 @@
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
+
    <description>
       %description
    </description>
@@ -54,6 +55,10 @@
 
    <plugin
          id="org.eclipse.terminal.connector.ssh"
+         version="0.0.0"/>
+
+   <plugin
+         id="org.eclipse.debug.terminal"
          version="0.0.0"/>
 
 </feature>


laeubi:
As a side note: It is not required to have an issue + PR (and if then likely one want to have at least a basic PR description instead of just "fixes XYZ")

iloveeclipse:
> As mentioned this bundles does not belongs to the platform.terminal feature.
> 
> This does (as the name indicates) belong to `eclipse.debug` part.

Well, why shouldn't be Terminal related bundle be part of the Terminal feature? Do you really want to add this bundle to the `org.eclipse.platform` feature which includes generic debug plugins? This makes less sense in my understanding as to include Terminal specific (debug related) bundle to Terminal feature.

laeubi:
> Well, why shouldn't be Terminal related bundle be part of the Terminal feature? 

See https://github.com/eclipse-platform/eclipse.platform/issues/2423#issuecomment-3748665194 in general it might be easier to discuss at one place.

Beside that, just compare the dependency closure before and after the addition of that bundle.

> Do you really want to add this bundle to the org.eclipse.platform feature which includes generic debug plugins? 

I don't want to add it anywhere (now) as it can be easily installed by anyone who is interested in this feature as a standalone bundle.

> This makes less sense in my understanding as to include Terminal specific (debug related) bundle to Terminal feature.

If we later find it is extremely useful one *might* add it to the platform feature, or create an own feature or ... until now it is an option but no one is obliques to use that and its not part of any EPP until now as far as I know nor of the SDK.

>

iloveeclipse:
> I don't want to add it anywhere (now) as it can be easily installed by anyone who is interested in this feature as a standalone bundle.

How? 

I've tried to follow https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/news/4.37/platform.md#support-for-launching-a-run-inside-a-terminal-session but there is no "Terminal Session Support" feature:

image

laeubi:
It is included here https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/97792fd0512bf261b5e4fd71bdb521ec5ef8ebb9/sites/eclipse-platform-repository/category.xml#L42 and was installable at the time where I wrote the N&N entry.

I also can see it with repository explorer

grafik

The latest change I can see is that it was changed how the update-site is build, maybe something was lost there ( @HannesWell ? ) and now P2 do not recognize it anymore as a root unit that can be installed...

HannesWell:
> The latest change I can see is that it was changed how the update-site is build, maybe something was lost there ( @HannesWell ? ) and now P2 do not recognize it anymore as a root unit that can be installed...

Actually nothing should have changed content-wise, at least it wasn't intended. And as you showed, the bundle is still in the repo.
But I wonder how one can generally install single bundles directly via the P2 UI? Until know I thought only features could be installed? I don't know if assigning a category would make a difference?
But if it has ever worked, then through some means I haven't been aware of until now.

laeubi:
> But I wonder how one can generally install single bundles directly via the P2 UI, Until know I thought only features could be installed?

You can install anything that is an IU, including bundles, e.g. m2e Update site is another example where we select bundles by a filter.

>  I don't know if assigning a category would make a difference?

Funny enough if one enables "Group items by category" it is shown under the "Uncategorized" so it seems to be a bug in the P2 UI that it is not shown if grouping is *disabled*:

grafik

merks:
> so it seems to be a bug in the P2 UI that it is not shown if grouping is _disabled_:

I expect it's a design **feature** not to list EVERYTHING.

laeubi:
> > so it seems to be a bug in the P2 UI that it is not shown if grouping is _disabled_:
> 
> I expect it's a design **feature** not to list EVERYTHING.

From a users perspective its a strange design... grouping of items should not make a difference. And especially if I choose to NOT group things I expect to see everything (if not even more) that I would see if grouping is not used.

I created 
- https://github.com/eclipse-equinox/p2/issues/990

iloveeclipse:
Closing because of https://github.com/eclipse-platform/eclipse.platform/issues/2428

PR #2426, opened 2026-01-16
Fix empty description for Prototype Label Decorator in preferences

elsazac:
The Prototype label decorator was missing a description, which resulted in an empty description area in the Label Decorations preference page when selection was made. 
This PR adds the description for the Prototype label decorator so that it is displayed correctly in the description box upon selection like the rest of the entries.


elsazac:
test looks good; merging this now.

PR #2427, opened 2026-01-18
Console does not need to start UI job every second

iloveeclipse:
In order to update console status line, ProcessConsole.computeName() calls resetName() every second once console is created. This causes new UI job to be started every second via timerExec(), with code already being executed on UI thread. It is a misuse of Jobs framework, it makes use of jobs dispatching where no jobs are needed. The code can use Display.execute() for a more lightweight console status update.

As a nice side effect no jobs are flashing in the Progress view every second (if system jobs are shown).

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/StringVariableSelectionDialog.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/StringVariableSelectionDialog.java
@@ -165,7 +165,7 @@ public class StringVariableSelectionDialog extends ElementListSelectionDialog {
 		final Display display = DebugUIPlugin.getStandardDisplay();
 		BusyIndicator.showWhile(display, () -> {
 			final IStringVariable[] elements = VariablesPlugin.getDefault().getStringVariableManager().getVariables();
-			display.asyncExec(() -> setListElements(elements));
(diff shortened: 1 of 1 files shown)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/console/ProcessConsole.java
```diff
@@ -649,18 +648,12 @@ private synchronized void resetName(boolean changed) {
...
 		if (!name.equals(newName)) {
-			UIJob job = new UIJob("Update console title") { //$NON-NLS-1$
-				@Override
-				public IStatus runInUIThread(IProgressMonitor monitor) {
-					ProcessConsole.this.setName(newName);
-					if (changed) {
-						warnOfContentChange();
-					}
-					return Status.OK_STATUS;
+			DebugUIPlugin.getStandardDisplay().execute(() -> {
```
jonahgraham, line 651: Changing from UIJob to execute had the unintended side effect of possible SWT disposed Exceptions - please see #2554

iloveeclipse: @jonahgraham : thanks, I'm aware about this issue, I will check this later.

PR #2430, opened 2026-01-19
Move Terminal launch action closer to Run/Debug toolbar actions

SougandhS:
Align Terminal launch toolbar action with Run/Debug launch workflow

Before : 
Screenshot 2026-01-19 at 2 24 13 PM

After : 
Screenshot 2026-01-19 at 2 23 39 PM

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/plugin.xml
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <!--
-    Copyright (c) 2018 Red Hat and others.
+    Copyright (c) 2018, 2026 Red Hat and others.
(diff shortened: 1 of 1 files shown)


terminal/bundles/org.eclipse.terminal.view.ui/plugin.xml
```diff
@@ -120,7 +120,7 @@
       </delegate>
    </extension>
    <extension point="org.eclipse.ui.menus">
-      <menuContribution locationURI="toolbar:org.eclipse.ui.main.toolbar?after=additions">
+      <menuContribution locationURI="toolbar:org.eclipse.ui.main.toolbar?after=org.eclipse.debug.ui.launchActionSet">
```
laeubi, line 123: This might not be available in all products where the terminal is used.

SougandhS: Is there a preferred approach for the fix, or would you prefer to leave this unchanged ?

PR #2443, opened 2026-01-28
Add Missing icons for Debug view launch context

SougandhS:
Before :

image
image

After : 

image
image


iloveeclipse:
@SougandhS : FYI very good article about menu icons https://tonsky.me/blog/tahoe-icons/

Not sure we should add more, maybe better remove some.

SougandhS:
>Not sure we should add more, maybe better remove some.

Removing some icons would definitely help keep the Eclipse UI clean, similar to how IntelliJ and VS Code approach it.

SougandhS:
This was added only because this `context menu` option was the only one missing an icon.

iloveeclipse:
I see. This case can't be fixed :-)

PR #2448, opened 2026-01-30
Enable compiler release in e4.core.di.extensions.supplier

akurtakov:
Fixes couple of warnings in the workspace if one doesn't have Java 17 installed anymore.


akurtakov:
Test failure unrelated.

PR #2453, opened 2026-02-09
Add missing descriptions to Colours and Fonts theme entries in Preferences

elsazac:
**Consistent use of descriptions in Colours and Fonts preferences**

Some entries in the Colours and Fonts preference page already provide descriptions, while others leave the description area empty. This change makes use of the existing descriptions and applies them to theme entries so that the description box is consistently populated and available descriptions are properly utilised. This improves clarity and aligns the presentation with the approach used with the subtree elements in Eclipse preference pages and gives more clarity on what each entity is used for. The Colours and Fonts preference page shows a description panel which is empty now. There are already descriptions available for some and not available for rest and they are not being used so as to display in the box. This change adds the existing description usage to theme entries so that users always see meaningful information when navigating these settings.

The main change pertaining to this PR has been merged via https://github.com/eclipse-platform/eclipse.platform.ui/pull/3666 and some of the remaining changes are addressed here.

### DIFF
--- a/debug/org.eclipse.debug.ui/plugin.properties
+++ b/debug/org.eclipse.debug.ui/plugin.properties
@@ -213,6 +213,8 @@ ActionDefinition.toggleStepFilters.description= Toggles enablement of debug step
 ActionDefinition.stepInto.name= Step Into
 ActionDefinition.stepInto.description= Step into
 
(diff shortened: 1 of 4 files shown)


elsazac:
Build looks good, merging this now.

Issue #2455, opened 2026-02-11
[Terminal] Opening a terminal with multiple Terminal Views does not work


Philipp0205:
## Steps to reproduce

From a fresh installation and clean workspace:
1. Open One Terminal View 
2. Open a second terminal view from the first by using the "Open new Terminal view" button
3. Press the "open a new terminal" button in the second view

I expected: The new terminal opening in the terminal view where it is called 

But got:  The new terminal opens in the first termina view

## Tested under this environment:
* OS & version: RHEL9
* Eclipse IDE/Platform version:  2025-12 (4.38)
* I used the newest version of the Terminal Plugin I could find:   
   * Terminal Feature    1.0.100.v20251015-0620	org.eclipse.terminal.feature.feature.group	Eclipse.org from the update site https://download.eclipse.org/releases/2025-12

Note: I committed to the terminal a few months ago, I can look into this and also fix it.

iloveeclipse:
Fix: https://github.com/eclipse-platform/eclipse.platform/pull/2456

iloveeclipse:
Verified, works with I20260216-1800 build.

Philipp0205:
Thanks!

PR #2456, opened 2026-02-16
Properly register ConsoleManager listeners on workbench windows

iloveeclipse:
`ConsoleManager` service is activated on
`org.eclipse.terminal.view.ui.internal.UIPlugin.start(BundleContext)` **before** workbench creates any window. This prevents that it can register his window & perspective listeners to any window and so it can't properly manage terminals if they are opened from multiple different terminal view instances existing in parallel.

This happens because terminal ui bundle contributes `ExternalExecutablesState` service that is activated very early at workbench startup, **before** window creation.

Beside this, `ConsoleManager` service, if ever connected to a window, would only listen to that single window events and would never support other windows.

Fixed both issues by moving window & perspective listener registration (and de-registration) of `ConsoleManager` service to already existing `org.eclipse.terminal.view.ui.internal.listeners.AbstractWindowListener`.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2455

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/ConsoleManager.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/ConsoleManager.java
@@ -47,6 +47,7 @@ import org.eclipse.ui.IViewSite;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchPart;
 import org.eclipse.ui.IWorkbenchPartReference;
+import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PerspectiveAdapter;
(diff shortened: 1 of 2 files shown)


laeubi:
> This happens because terminal ui bundle contributes `ExternalExecutablesState` service that is activated very early at workbench startup, **before** window creation.

Maybe the creation of the service should then be more lazy as well?

iloveeclipse:
> Maybe the creation of the service should then be more lazy as well?

I haven't checked which services (there are three as of today) and how should be or should be not activated at which time in terminal view ui bundle. However, I saw that there was already `AbstractWindowListener` code added (probably for similar reason) which one could reuse there to fix the current issue without refactoring whole bundle.

I would prefer to have this small fix integrated in RC1, which shouldn't be a big deal. For the possible optimization of terminal services one can create a dedicated ticket to work on in next release.

iloveeclipse:
@laeubi : any objections to merge it for RC1?

iloveeclipse:
Thanks.

Issue #2457, opened 2026-02-18
Retire org.eclipse.core.tools.resources and (maybe) core.tools from Git master


vogella:
org.eclipse.core.tools.resources are not part of the build.

@merks suggested to retire it, see https://github.com/eclipse-platform/ui-best-practices/issues/151#issuecomment-3914525563

Not sure about org.eclipse.core.tools. Maybe this one should also be retired?

@akurtakov @merks WDYT?

merks:
I think it’s included in a feature also not in the sdk repo. Garbage collecting content not in the repo makes sense.

akurtakov:
Same from me - best to remove unused things. After March release is out of course.

vogella:
@iloveeclipse WDYT?

iloveeclipse:
Not sure why should `org.eclipse.core.tools.resources` be deleted from git, also not sure why it is not built. I believe it is built but not included in the SDK?

Anyway, the `org.eclipse.core.tools.resources` bundle provides few interesting low level views useful if one hunts resource related problems:

<img width="300" height="253" alt="Image" src="https://github.com/user-attachments/assets/4d5c5936-4d40-494a-956f-4e141f8801e9" />

Definitely "no go" to remove that, on contraire, if there is no feature including that it would be cool to add it, so one could install views into SDK from update site and does not need to run Eclipse from debugger to use the views. 

What is meant by `core.tools`?

merks:
Maybe there is confusion with this?

<img width="718" height="244" alt="Image" src="https://github.com/user-attachments/assets/946c8c6d-91cf-4de2-98d7-967ad3fdeef5" />

I do not see it anywhere:

<img width="954" height="262" alt="Image" src="https://github.com/user-attachments/assets/48a01fe3-5279-40b0-b168-6c48add95e81" />

<img width="1126" height="1362" alt="Image" src="https://github.com/user-attachments/assets/ca7dd459-d73a-4270-ad91-f5734294ccac" />

These appear to be zombies that don't go anywhere.  I'm not saying they're not useful, but there certainly not useful if we don't build then and don't redistribute them.

iloveeclipse:
> These appear to be zombies that don't go anywhere.

See my comment above: 

>  if there is no feature including that it would be cool to add it, so one could install views into SDK from update site and does not need to run Eclipse from debugger to use the views.

[...]

PR #2458, opened 2026-02-19
Build org.eclipse.core.tools feature

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform/issues/2457


HannesWell:
This is basically a RelEng change, not affecting any consumer code, thus merging this despite the freeze-period.

Issue #2295, opened 2025-11-30
Terminal and CDT Debugger Console are blurry with fractional scaling


gvl2023:
Running on a 4k monitor at native resolution and 125% scaling, editors and other views look sharp but the Debugger Console and Terminal are blurry. They are seemingly set to use the same default text font (Monospace 10) as the main Console view (left on the screenshot), which renders sharp. This is Eclipse 2025-09, 4.37.0/CDT12.2.0 running on RHEL 9.6/Wayland. The screenshot will likely be scaled on the web page introducing additional blurring, but you can still see the font on the left is sharper than on the right, the CDT debugger console is shown but the Terminal suffers from the same issue, my understanding is that the Debugger Console is based on the Terminal.

<img width="671" height="263" alt="Image" src="https://github.com/user-attachments/assets/62de6955-60a8-4e87-9d2b-9a2d5169e3f4" />

rdin777:
"This rendering issue is likely tied to how SWT handles Cairo surfaces when fractional scaling is active on Wayland.

Suggested checks for the reporter:

Does setting the environment variable GDK_BACKEND=x11 resolve the blurriness (by forcing XWayland, though it might disable other Wayland features)?

Have you tried starting Eclipse with -Dsun.java2d.uiScale=1 or other HiDPI-related JVM flags?

Technical Note for Maintainers: It seems that the Terminal and CDT Debugger Console might be using fixed-pixel buffers that don't account for the devicePixelRatio correctly when it's not an integer. We should verify if the org.eclipse.swt.widgets.Canvas used in these views is DPI-aware and if it's using the modern getImageDataAtCurrentDPI logic."

iloveeclipse:
I see similar issue on Windows too, with display zoom set to 125%. Terminal font (whatever it is, it seem to be same as in the COnsole) is rendered not sharp, some letters are really bad looking:

<img width="1578" height="457" alt="Image" src="https://github.com/user-attachments/assets/797143f2-9af8-4dba-95c6-3f58a3f17f39" />

iloveeclipse:
@HeikoKlare : this one is interesting.
I see in my SDK build terminal font not properly rendered, but if I start Eclipse in debugger, it looks fine (I'm currently on Windows). 

Only if I **explicitly** set antialiasing to `SWT.DEFAULT`, it looks broken again (at least on Windows) (see PR https://github.com/eclipse-platform/eclipse.platform/pull/2463 for the code in question & the fix).

[...]

PR #2463, opened 2026-02-24
Enable anti-aliasing for Terminal text

iloveeclipse:
<s>Fixes</s> Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/2295

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextLineRenderer.java
+++ b/terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextLineRenderer.java
@@ -18,6 +18,7 @@ package org.eclipse.terminal.internal.textcanvas;
 
 import java.util.Map;
 
+import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.GC;
@@ -63,6 +64,7 @@ public class TextLineRenderer implements ILinelRenderer {
 		}
 		Image buffer = new Image(gc.getDevice(), width, height);
 		GC doubleBufferGC = new GC(buffer);
+		doubleBufferGC.setAntialias(SWT.ON);
 		if (line < 0 || line >= getTerminalText().getHeight() || colFirst >= getTerminalText().getWidth()
 				|| colFirst - colLast == 0) {
 			fillBackground(doubleBufferGC, 0, 0, width, height);


HeikoKlare:
@iloveeclipse just to be sure: you don't want to merge this as a last-minute fix for RC2 but wait for 2026-06, do you?

iloveeclipse:
> you don't want to merge this as a last-minute fix for RC2 but wait for 2026-06, do you?

I believe this could be merged as it should be small & innocent fix, but I believe it is too late for RC2, isn't?
@merks : how are we in time?

If there will be RC2a (usually we have one), this would be good candidate  :-)

merks:
@iloveeclipse  @HeikoKlare 

The RC2/final build is tonight (I think 6:00PM EST) so we have all day for any **safe** final changes.  I'll leave it to you guys' good judgement as to what is appropriate and safe to merge on the last day of the cycle.

iloveeclipse:
OK, with that I believe we can merge for RC2. 
Heiko, do we need more approvals?

HeikoKlare:
> OK, with that I believe we can merge for RC2.
> Heiko, do we need more approvals?

I don't think so. You have my PL+1.

Just for the sake of safety I have tested the change on MacOS and it is fine there as well (in the sense that it does not change any behavior as the text was antialiased already before).

merks:
You guys are fully empowered to do the right thing in the best interests of the project.

iloveeclipse:
OK, thanks Heiko & Ed. Merging now.

HannesWell:
I was looking forward for a nicer rendering of the Terminal content since it was bugging me quiet some time.
And while it looks a little bit better now, unfortunately it became quite laggy for me on Windows-11 with a 4K monitor and 175% scaling. My graphics card is actually powerful, but I don't know if something else has to be configured.
Not only is there a noticeable lag when typing text into the console, it also noticeably slows down the SWT entire layout if I e.g. resize the view containers and takes noticeable time to redraw the content.

iloveeclipse:
OMG. I see the slower UI repaints on view resize, slower output and slightly slower typing speed in the terminal. The question is whether it is worth to have crisp text inside or not?

@trancexpress : could you please test on Linux with latest SDK build?

Should we revert the change?

akurtakov:
I would say revert it and have the next cycle to have time to improve. We shouldn't be shipping such regressions as Eclipse is not a speed monster already.

HannesWell:
> OMG. I see the slower UI repaints on view resize, slower output and slightly slower typing speed in the terminal. The question is whether it is worth to have crisp text inside or not?

At least in my setup the situation became a bit better with respect to sharpness of the rendered text but it's not perfect and I still have other artifacts like ghost lines after resize.
And while the terminal does not become unusable, for me personally the gain in sharpness doesn't outweigh the loss in performance.

> I would say revert it and have the next cycle to have time to improve. We shouldn't be shipping such regressions as Eclipse is not a speed monster already.

That's my opinion too.

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform/pull/2480

trancexpress:
> @trancexpress : could you please test on Linux with latest SDK build?


Checked with:

```
Eclipse SDK
Version: 2026-03 (4.39)
Build id: I20260225-1800
```

~I don't see a notable difference to:~

```
Eclipse SDK
Version: 2026-03 (4.39)
Build id: I20260224-2200
```

Actually no, when I start in my developer workspace resizing is a lot slower. I didn't notice anything in a new workspace, but with some views open and a Java editor, the UI is slower.

PR #2468, opened 2026-02-24
Add splash to o.e.sdk Plug-in and update it in new dev-cycle preparation

HannesWell:
The `org.eclipse.sdk` branding-plugin was missing the splash-screen, similar to the other branding-plugin `org.eclipse.platform`. This enables the splash-screen when running the `org.eclipse.sdk.ide` product in launch configurations for secondary Eclipse instances.
Furthermore, the `org.eclipse.sdk` branding-plugin already duplicates all other icons and images also in `org.eclipse.platform` and it also defines splash-screen locations/sizes, adding the splash just completes this.
https://github.com/eclipse-platform/eclipse.platform/blob/8594fad9c3b068d97d18abc3b8a6ef714c754023/platform/org.eclipse.sdk/plugin.xml#L46-L59

Additionally the about-dialog branding-icon and the (now new) splash-screen in `org.eclipse.sdk` weren't updated during the preparation of a new development cycle, again like `org.eclipse.platform`.
This also becomes simpler if the `org.eclipse.sdk` project also contains a splash.

This handles issues like the one noted in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2856#issuecomment-2710541402

With this applied we can also change the `eclipse-sdk` product to reference the `o.e.sdk` plugin for the splash screen, increasing the separation:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/7005d3970a86f75dd50aa56d19473e9fbe26ddf0/products/eclipse-sdk/sdk.product#L24


HannesWell:
This should be submitted before the next development cycle is prepared (otherwise we have to sync the about-images in `org.eclipse.sdk` manually).

Still we could this submit this now to have the splash-screen addition in RC2.
@akurtakov, @iloveeclipse, @HeikoKlare, as PLs, should this be done?

Otherwise I'd submit this immediately after RC2 is declared and before the preparation of the next dev-cycle runs.

> With this applied we can also change the `eclipse-sdk` product to reference the `o.e.sdk` plugin for the splash screen, increasing the separation:
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/7005d3970a86f75dd50aa56d19473e9fbe26ddf0/products/eclipse-sdk/sdk.product#L24

On the other hand, I think that part should definitively wait for next cycle.

HannesWell:
And btw., the `org.eclipse.sdk` plugin has a dependency on `org.eclipse.platform` due to [Bug 533285](https://bugs.eclipse.org/bugs/show_bug.cgi?id=533285):
https://github.com/eclipse-platform/eclipse.platform/blob/8594fad9c3b068d97d18abc3b8a6ef714c754023/platform/org.eclipse.sdk/META-INF/MANIFEST.MF#L13

Maybe that could be cleaned-up/unified in general, but that's another task.

iloveeclipse:
I have no objections, this can be merged for RC2 IMO.

merks:
Yes, do what makes things easiest.  PMC +1.

HannesWell:
Thank you all.
Then lets submit this now to have one more build in case something unexpected happens.

> The script adaptations looks sound as well.

My test with a dry-run of the preparation pipeline was also successful.

PR #2469, opened 2026-02-24
Add Splash Screens of 2026 for eclipse 4.40-4.43

HannesWell:
Add the splash-screens from https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6906#note_6427094

This is currently a draft since we're awaiting some last-minute enhancements:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6906#note_6427826

This has to be submitted before the preparation of the next development cycle on Friday (after RC2 promotion).

@MohananRahul FYI

PR #2470, opened 2026-02-24
Adjust buildId rectangle for 2026 Splash Screen

HannesWell:
The new splash-screens have their logo text in the center so the injected buildId text has to be moved to the bottom above the progress-bar, next to the product name.

Has to be submitted in the next development cycle and after
- https://github.com/eclipse-platform/eclipse.platform/pull/2469

The final splashes look as shown in
- https://github.com/eclipse-platform/eclipse.platform/pull/2470#issuecomment-3964790641

With the earlier proposal the splash-screen looked like
- For the `Eclipse Platform` Product
grafik

- For the `Eclipse SDK` Product
grafik

To clarify the borders of the `startupMessage` and `buildId` rectangle I temporarily set their background colors with the new sizes in
- https://github.com/eclipse-platform/eclipse.platform.ui/blob/89a19bf9c20a55490d281e0a97cd0a43bf505085/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/splash/EclipseSplashHandler.java#L93-L97
- https://github.com/eclipse-platform/eclipse.platform.ui/blob/89a19bf9c20a55490d281e0a97cd0a43bf505085/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/ProgressMonitorPart.java#L244-L245

 grafik

In case the final new splash-screens have the logo text moved up a bit and there is sufficient space, the buildId can also be moved to the very right.

Also note that only the Eclipse Platform and SDK product seem show the `buildID` and only for I-builds, not for milestones or RC or releases. So IMO it's not a big problem if the `buildID` isn't displayed optimally.

### DIFF
--- a/platform/org.eclipse.platform/plugin.xml
+++ b/platform/org.eclipse.platform/plugin.xml
@@ -46,17 +46,17 @@
 				value="ffffff"/>
           <property
                 name="startupMessageRect"
-                value="7,265,320,25"/>
+                value="7,265,250,25"/>
           <property
                 name="startupProgressRect"
                 value="2,290,450,10"/>
           <property
                 name="buildIdLocation"
-                value="108,222">
+                value="257,265">
           </property>
           <property
                 name="buildIdSize"
-                value="293,40">
+                value="188,25">
           </property>
 
       </product> 
--- a/platform/org.eclipse.sdk/plugin.xml
+++ b/platform/org.eclipse.sdk/plugin.xml
@@ -45,17 +45,17 @@
 				value="ffffff"/>
           <property
                 name="startupMessageRect"
-                value="7,265,320,25"/>
+                value="7,265,250,25"/>
           <property
                 name="startupProgressRect"
                 value="2,290,450,10"/>
           <property
                 name="buildIdLocation"
-                value="108,222">
+                value="257,265">
           </property>
           <property
                 name="buildIdSize"
-                value="293,40">
+                value="188,25">
           </property>
 
        </product>


BeckerWdf:
> grafik
> In case the final new splash-screens have the logo text moved up a bit and there is sufficient space, the buildId can also be moved to the very right.
> 
> Also note that only the Eclipse Platform and SDK product seem show the `buildID` and only for I-builds, not for milestones or RC or releases. So IMO it's not a big problem if the `buildID` isn't displayed optimally.

Can't we move up the build information text a bit so that the bottom of this text and the 2026-06 text are aligned (that they both "stand" on a common invisible line)?

HeikoKlare:
We should keep in mind that the build information pops up later in the application's start process (just like the status message and the progress bar) and having information pop up at multiple places of the splash screen is maybe not the best experience. And you would have to adapt the build information position every year for every new splash screen (or ensure that the text inside the splash screen image is always placed at the exact same position).
I would be in favor of just  "reserving" the bottom area of the splash (progress bar + a text line above) for information that shows up later (status to the left and build information to the right).

Also note that the final 2026 splash screens will hopefully have an improved layout of the text (see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6906#note_6427826) as the one that Hannes has integrated here does not look that appealing with the progress bar anyway.

HannesWell:
We now do have the splash screen images that I consider to be the final ones. I'll update the other PR this evening or tomorrow as time permits.

> I would be in favor of just "reserving" the bottom area of the splash (progress bar + a text line above) for information that shows up later (status to the left and build information to the right).

Sounds good to me. In general we can of course mandate any kind of format we like for the splashes of Eclipse. In custom products with custom splashes, this can again be adjusted to custom needs.

The new splashs also have sufficient space so it should fit to move the buildId to the very right, on the same height as the product name on the left, I suggest. Is everyone fine with that? 

I assume this also makes Matthias request obsolete to align the text with the SimRel name (to which I would have agreed for the previous splashes)?

BeckerWdf:
> I assume this also makes Matthias request obsolete to align the text with the SimRel name (to which I would have agreed for the previous splashes)?

Yes

HannesWell:
Since we now have the final splash screen images and they are submitted via https://github.com/eclipse-platform/eclipse.platform/pull/2469, I've now adapted the location of the buildId and it now looks like

grafik

again with highlighted areas for startupMessage (green) and buildId (blue):
grafik

To me this looks good now (and much better than before).
Are you fine with it too?

PR #2481, opened 2026-02-26
Make Terminal operate on and cache colors instead of RGB values

HeikoKlare:


terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/emulator/VT100TerminalControl.java
laeubi, line 687: should this here not be an color already?

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/emulator/VT100TerminalControl.java
```diff
@@ -694,7 +695,7 @@ private void onTerminalColorsChanged() {
 			if (rgb == null) {
 				rgb = TerminalColorPresets.INSTANCE.getDefaultPreset().getRGB(terminalColor);
```
laeubi, line 695: here as well ...

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextLineRenderer.java
```diff
@@ -86,8 +85,8 @@ public void drawLine(ITextCanvasModel model, GC gc, int line, int x, int y, int
 				int colStart = line == hsStart.y ? hsStart.x : 0;
 				int colEnd = line == hsEnd.y ? hsEnd.x : getTerminalText().getWidth();
 				if (colStart < colEnd) {
-					RGB defaultFg = fStyleMap.getForegrondRGB(null);
-					doubleBufferGC.setForeground(new Color(doubleBufferGC.getDevice(), defaultFg));
+					Color defaultFg = fStyleMap.getForegroundColor(null);
+					doubleBufferGC.setForeground(defaultFg);
```
laeubi: as we have `Color getDefaultBackgroundColor()` should the not better be a `Color getForegroundColor()` as well?

This would prevent repeating the code used here and in setupGC

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/emulator/VT100TerminalControl.java
HeikoKlare: It's an RGB value that is later wrapped into a color.

terminal/bundles/org.eclipse.terminal.control/src/org/eclipse/terminal/internal/textcanvas/TextLineRenderer.java
HeikoKlare: It's not a repetition as here it's about the default color (style == null) and in setupGC its the style-specific color. We can of course introduce additional methods like `getDefaultForegroundColor()` or the like, but it's will only be for consistency, not for reuse.

PR #2483, opened 2026-02-26
Avoid cut-off symbols in Terminal on monitor zoom != 100%

HeikoKlare:
On monitors with zoom != 100% on Windows, symbols in the Terminal appear cut-off. This can be avoided by enabled advanced mode for the GC used for rendering the text.

This PR includes and should thus be processed after:
- https://github.com/eclipse-platform/eclipse.platform/pull/2473

This is a replacement for the PR enabling antialiasing which had to be reverted because of performance issues:
- https://github.com/eclipse-platform/eclipse.platform/pull/2463

The change still has a performance impact, but it's much less severe than with the previous PR. We should test if the degradation is acceptable.

### Before
image

### After
image


HeikoKlare:
Closing this as obsolete due to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3101

PR #2495, opened 2026-03-02
Increment version for included o.e.equinox.core.feature increment

merks:
- org.eclipse.equinox.core.feature was incremented by https://github.com/eclipse-equinox/equinox/pull/1249


merks:
Incrementing a feature version cannot cause ant test failures.  I don't know if that failure is know. I know I should record it, but I'm at my wits end for the day.

Issue #2558, opened 2026-03-10
NPE when opening context menu in Breakpoints view


mx990:
## Steps to reproduce

From a fresh installation and clean workspace:

* Create a Java breakpoint

I tried

* Select the Java breakpoint in the *Breakpoints* view
* Select the *Conditional* checkbox in the detail pane
* Right-click on the breakpoint

I expected: no exception

But got: an exception

## Here is some relevant log output

From `<workspace>/.metadata/.log`
```
!ENTRY org.eclipse.ui 4 0 2026-03-10 16:51:42.472
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.jface.action.IAction.setImageDescriptor(org.eclipse.jface.resource.ImageDescriptor)" because "action" is null
	at org.eclipse.debug.internal.ui.views.breakpoints.BreakpointsView.fillContextMenu(BreakpointsView.java:205)
	at org.eclipse.jface.action.MenuManager.fireAboutToShow(MenuManager.java:337)
[...]
```

It looks like `getAction(FIND_ACTION)` in `BreakpointsView.fillContextMenu` returns `null`, since the global actions are cleared due to the focus loss when selecting something in the detail pane.

PR #2559, opened 2026-03-10
Fix NPE when Find action is unavailable in Breakpoints view

mx990:
The Find action may be unavailable when the focus is outside the Breakpoints view, which leads to an NPE in fillContextMenu.

This adds a null and enablement check for the Find action to resolve the problem.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2558

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/breakpoints/BreakpointsView.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/breakpoints/BreakpointsView.java
@@ -202,8 +202,10 @@ public class BreakpointsView extends VariablesView implements IBreakpointManager
 			menu.add(action);
 		}
 		action = getAction(FIND_ACTION);
-		action.setImageDescriptor(DebugPluginImages.getImageDescriptor(IDebugUIConstants.IMG_FIND_ACTION));
-		menu.add(action);
+		if (action != null && action.isEnabled()) {
+			action.setImageDescriptor(DebugPluginImages.getImageDescriptor(IDebugUIConstants.IMG_FIND_ACTION));
+			menu.add(action);
+		}
 		action = getAction(ACTION_REMOVE_FROM_GROUP);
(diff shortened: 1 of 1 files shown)


elsazac:
Merging this now;

fedejeanne:
Hi @mx990, thank you for the fix!

Question: there are other usages of the method `getAction(...)` and also other calls with the exact same parameter here:

image

Do you know if those are safe?

mx990:
Hi @fedejeanne, thank you for pointing me to those other locations!

I was able to produce a similar exception in the *Expressions* view:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jface.action.IAction.setImageDescriptor(org.eclipse.jface.resource.ImageDescriptor)" because "action" is null
	at org.eclipse.debug.internal.ui.views.expression.ExpressionView.fillContextMenu(ExpressionView.java:87)
	at org.eclipse.jface.action.MenuManager.fireAboutToShow(MenuManager.java:337)
```

I was also able to produce an error in the *Variables* view, which does have a null check but logs an exception:
```
java.lang.IllegalStateException: No action found for id: find, fGlobalActionMap: {selectAll=null, find=org.eclipse.debug.internal.ui.viewers.model.VirtualFindAction@5cca5ff3, copy=org.eclipse.ui.internal.ViewPluginAction@2fcad739, paste=null}
	at org.eclipse.debug.internal.ui.views.variables.VariablesView.fillContextMenu(VariablesView.java:1109)
	at org.eclipse.jface.action.MenuManager.fireAboutToShow(MenuManager.java:337)
```

I was unable to produce an exception or error in either the *Modules* or *Launch* view, but it may still be sensible to also add corresponding null checks there.

mx990:
I have created an additional PR #2562 to fix the NPE in the *Expressions* view. Please let me know if I should also change the other locations.

fedejeanne:
Great, thank you for looking into it! No need to change the one in the _Variables View_ on my account. @iloveeclipse since you made that change back then (e583bf6933e54467302239a27e5aad0554fa77ab) WDYT?

PR #2564, opened 2026-03-13
Show structure compare in Outline view by default

### LINKED ISSUE #3776: Improve compare defaults
There are few default compare in Eclipse which I always change in every workspace since years.
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3775 reminded me about them.

I believe the defaults can be improved.

1) **Show structural changes in the Outline** (feature introduced via https://bugs.eclipse.org/bugs/show_bug.cgi?id=45247).
By default is is off, which wastes almost 1/3 of the compare editor. That is exact the reason why it makes sense to enable the feature, so the space is not wasted.

**Default:**

Image

**Proposed:**

Image

2) **Connect ranges with single line**. This is very old, it seem that the option was introduced at 2.1 Eclipse via https://bugs.eclipse.org/bugs/show_bug.cgi?id=31357 and just no one cared to enable it by default. Single line is harder to see, so the "line ranges" (no idea how that is officially named) is visually more appealing and helps better identify flow of the changes.

**Default:**

Image

**Proposed:**

Image

3) **Change default diff panes order** to follow "usual" diff order - left is base/original, right is local/modified.
Default in Eclipse is the opposite, I believe from the CVS times where there were no new/old but local/server, and someone arbitrary decided that local should be left. That was probably OK at that time, but now where everyone uses git it is awkward to see diff panes shown in "wrong" order by default in Eclipse, where rest of the world does the opposite. See for example https://www.jetbrains.com/help/idea/comparing-file-versions.html.
The possibility to change the order was introduced via https://bugs.eclipse.org/bugs/show_bug.cgi?id=213780). The request to change the default is https://bugs.eclipse.org/bugs/show_bug.cgi?id=501752, but there were some test fixes needed, no time and no one continued the work. I believe now we should finally do that change.

**Default:**

Image

**Proposed:**

Image

**All together**

**Default:**

Image

**Proposed:**

Image

iloveeclipse:

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ComparePreferencePage.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ComparePreferencePage.java
@@ -162,7 +162,7 @@ public class ComparePreferencePage extends PreferencePage implements IWorkbenchP
 
 	public static void initDefaults(IPreferenceStore store) {
 		store.setDefault(OPEN_STRUCTURE_COMPARE, true);
-		store.setDefault(USE_OUTLINE_VIEW, false);
+		store.setDefault(USE_OUTLINE_VIEW, true);
 		store.setDefault(SYNCHRONIZE_SCROLLING, true);
 		store.setDefault(SHOW_PSEUDO_CONFLICTS, false);
 		store.setDefault(INITIALLY_SHOW_ANCESTOR_PANE, false);


basilevs:
This change had visible effect in RCPTT and affected UI tests.
The problem is double-click in structure tree on outline does not affect the compare editor. This results in inability to compare model elements one-by-one.

PR #2565, opened 2026-03-13
Use line ranges in compare editor by default

### LINKED ISSUE #3776: Improve compare defaults

iloveeclipse:

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ComparePreferencePage.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/ComparePreferencePage.java
@@ -170,7 +170,7 @@ public class ComparePreferencePage extends PreferencePage implements IWorkbenchP
 		store.setDefault(PREF_SAVE_ALL_EDITORS, false);
 		store.setDefault(ADDED_LINES_REGEX, ""); //$NON-NLS-1$
 		store.setDefault(REMOVED_LINES_REGEX, ""); //$NON-NLS-1$
-		store.setDefault(USE_SINGLE_LINE, true);
+		store.setDefault(USE_SINGLE_LINE, false);
 		store.setDefault(HIGHLIGHT_TOKEN_CHANGES, true);
 		store.setDefault(CAPPING_DISABLED, false);
 		store.setDefault(PATH_FILTER, ""); //$NON-NLS-1$


iloveeclipse:
```
artifacts/test-results-ubuntu-latest/resources/tests/org.eclipse.core.tests.resources/target/surefire-reports/TEST-org.eclipse.core.tests.resources.AutomatedResourceTests.xml [took 0s]
project 'P/a06e67f9e71e00111d63b712e335a2fb' is not expected to have nature: org.eclipse.core.tests.resources.earthNature
org.opentest4j.AssertionFailedError: project 'P/a06e67f9e71e00111d63b712e335a2fb' is not expected to have nature: org.eclipse.core.tests.resources.earthNature
	at org.eclipse.core.tests.resources.NatureTest.assertDoesNotHaveNature(NatureTest.java:125)
	at org.eclipse.core.tests.resources.NatureTest.testBug297871(NatureTest.java:311)

!ENTRY org.eclipse.core.tests.resources 1 0 2026-03-13 14:21:49.577
!MESSAGE [NatureTest.testBug297871] setUp

!ENTRY org.eclipse.core.resources 2 2 2026-03-13 14:21:49.582
!MESSAGE Save operation warnings.
!SUBENTRY 1 org.eclipse.core.resources 2 234 2026-03-13 14:21:49.582
!MESSAGE The project description file (.project) for 'a06e67f9e71e00111d63b712e335a2fb' was missing.  This file contains important information about the project.  A new project description file has been created, but some information about the project may have been lost.

!ENTRY org.eclipse.core.tests.resources 1 0 2026-03-13 14:21:49.586
!MESSAGE [NatureTest.testBug297871] tearDown 
```
Test failure above is unrelated, only on Linux/Ubuntu machine, so must be unstable test.

HannesWell:
While I find the new default visually appealing, it has one major functional draw-back:
For local changes the 'move'/undo buttons are not available:

Single line connection:
Image
'Range' connection:
Image

Without that the diff-view loses a tool, that at least I consider very important.
I assume these buttons can be added even for 'Range' connections, but until then I'm against making it the default.

HannesWell:
> For local changes the 'move'/undo buttons are not available:

@iloveeclipse what's the status on this issue?
Is there any plan to add those buttons or should this be reverted instead?

iloveeclipse:
> > For local changes the 'move'/undo buttons are not available:
> 
> @iloveeclipse what's the status on this issue? Is there any plan to add those buttons or should this be reverted instead?

No revert is needed, there are buttons on the top of the compare editor for same operation, and it is the preference that ome can change. I do not plan to add extra buttons on lines.

HannesWell:
Using the buttons in the toolbar is not really a satisfying alternative, at least for me.
One has to move the mouse much more, which becomes really annoying, especially on large diffs.
But maybe that's just my opinion...

Sure I can revert that in each workspace or configure my Oomph setup to do it, but I'm not sure if everyone who's not happy with the new default wants to do that.

akurtakov:
I was hit by this problem too. The buttons in the top are context sensitive so one can very easily have the wrong effect if haven't clicked in the right place before that. I second the request to add the buttons in the diff view back.

merks:
I'm generally concerned about changing defaults and that concern grows when even two committers are not happy with the new default.  It makes me concerned that maybe a large number of end-users will also not be happy, and that they will not read the N&N to find out they can change it back.   The reaction might be, "Oh, something new and improved that I don't like!"   I'm not sure we have a way to reconcile such a thing?  I wonder how hard it would be to add the button?  (Personally I generally use the toolbar buttons, but I also very often press the wrong one and it's super annoying.)

fedejeanne:
I also noticed this loss of functionality just recently and found it unpleasant and since there is no preference in UI (`Window > Preferences > ...`) other "regular" users will be hit by it too.

I would like to see this change reverted.

iloveeclipse:
> I also noticed this loss of functionality

2222ff69a218ccb7d08dcdf2dce8848ba7a6ddab / https://github.com/eclipse-platform/eclipse.platform/pull/2583 was supposed to fix that. 

> and since there is no preference in UI

Sure there is a preference in UI:
image

fedejeanne:
Thank you very much for the hints! I totally missed them before :-\

I withdraw then my comment...
> I would like to see this change reverted.

👍

merks:
I still keep thinking it might be nice to paint little buttons too. Maybe circular ones.  But this only seems to matter for large changes.

iloveeclipse:
> I still keep thinking it might be nice to paint little buttons too. Maybe circular ones

Or maybe just "dots" with darker color. COntributions are welcome :-)

fedejeanne:
I'm quite happy with the current arrows but don't stop innovation on my account (I am a self-proclaimed "dinosaur" when it comes to UI/UX improvements 😅 )

HannesWell:
> I still keep thinking it might be nice to paint little buttons too. Maybe circular ones. But this only seems to matter for large changes.

For larger diffs the move buttons are indeed hard to find. I created a PR to show the markers for larger diffs:
- https://github.com/eclipse-platform/eclipse.platform/pull/2643

PR #2566, opened 2026-03-13
Use "natural" diff order by default in compare editor

iloveeclipse:
Left is base/original, right is local/modified


iloveeclipse:
Test failures are to expect, let see how many.

Issue #2570, opened 2026-03-14
Improve the default layout of the debug perspective


merks:
Generally the first I do in every IDE when I first use the debug perspective is to change the area on the right to cover the entire height of the window.  That's because generally like to see more of the variables and like to see more of the toString values of those variables.  This could be change in .DebugPerspectiveFactory.createInitialLayout(IPageLayout)

Currently this is the default layout

<img width="961" height="529" alt="Image" src="https://github.com/user-attachments/assets/1b177224-67d7-4ba2-af6b-8f2ade152620" />

This is the propose new default layout:

<img width="691" height="527" alt="Image" src="https://github.com/user-attachments/assets/16904c2d-63dd-4d50-ab3b-04bd415f168b" />

vogella:
+1, great to see these requests to improve the default settings of our IDE

PR #2577, opened 2026-03-16
Improve the default layout of the debug perspective

merks:
- Configure the section with the Variables view so that it spans the full height of the window.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2570


vogella:
LGTM

PR #2579, opened 2026-03-18
Add font zoom functionality to console view

### LINKED ISSUE #2578: Add zoom (Ctrl + / Ctrl -) support for Console View
Currently zoom in/out functionality using Ctrl + / Ctrl - is supported in text editors (e.g., Java editor), but not consistently available in the Console view.
This creates an inconsistent user experience, especially for developers who frequently switch between editor and console while debugging or running applications.

Currently we can change the font size of the console view using below preference.
`Preferences → General → Appearance → Colors and Fonts → Console font`

raghucssit:
-Add key listener on text widget of console view which listens ctrl plus and control minus(including numpad +/-)
-Zoom in/out of the console view text in steps based on the keys pressed.

see https://github.com/eclipse-platform/eclipse.platform/issues/2578

### DIFF
--- a/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
+++ b/debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
@@ -16,6 +16,7 @@ package org.eclipse.debug.tests.console;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
@@ -24,22 +25,31 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 
 import org.eclipse.core.commands.Command;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.debug.tests.DebugTestExtension;
 import org.eclipse.debug.tests.TestUtil;
 import org.eclipse.jface.text.IDocument;
+import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.graphics.FontData;
 import org.eclipse.ui.IViewPart;
 import org.eclipse.ui.IWorkbenchCommandConstants;
 import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.commands.ICommandService;
 import org.eclipse.ui.console.ConsolePlugin;
 import org.eclipse.ui.console.IConsole;
 import org.eclipse.ui.console.IConsoleConstants;
 import org.eclipse.ui.console.IConsoleManager;
+import org.eclipse.ui.console.IConsoleView;
 import org.eclipse.ui.console.IOConsole;
 import org.eclipse.ui.console.IOConsoleOutputStream;
 import org.eclipse.ui.console.MessageConsole;
+import org.eclipse.ui.console.TextConsole;
+import org.eclipse.ui.handlers.IHandlerService;
 import org.eclipse.ui.internal.console.ConsoleManager;
+import org.eclipse.ui.internal.console.ConsoleZoomHandler;
 import org.eclipse.ui.texteditor.IWorkbenchActionDefinitionIds;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInfo;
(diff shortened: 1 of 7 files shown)


debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
```diff
@@ -909,4 +952,117 @@ public boolean getAutoScrollLock() {
...
+		Font oldCustomFont = fCustomFont;
+
+		// Create and set the new font
+		fCustomFont = new Font(styledText.getDisplay(), newFontData);
+		styledText.setFont(fCustomFont);
+
+		// Dispose the old custom font after setting the new one
+		if (oldCustomFont != null && !oldCustomFont.isDisposed()) {
+			oldCustomFont.dispose();
+		}
```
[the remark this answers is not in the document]
raghucssit: Fixed.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleView.java
```diff
@@ -410,6 +436,18 @@ public void handleException(Throwable exception) {
 			});
 		}
 
+		// Install font zoom key listener on the StyledText widget
+		StyledText textWidget = null;
+		if (page instanceof TextConsolePage textPage) {
```
iloveeclipse: I did not do full review, but looking at this palce here wonder if the code change should be done in `TextConsolePage` and all init in `createControl()` there? Because all other pages are irrelevant for the new code.

iloveeclipse:
Have you thought about persisting the current zoom **over** the current session? That will be likely next immediate customer request after using this feature for the first time & IDE restart.

Assuming the code moved to the page (from the console), we would always know "where we are" and should be able to persist the zoom in the preference store by using the `IConsole.getType()` as part of the key.

raghucssit:
@trancexpress Please check this PR.

trancexpress:
With:

```
Eclipse SDK
Version: 2026-06 (4.40)
Build id: I20260323-1800
```

The Console view can be zoomed when I'm in the Java editor and I press: `Ctrl` and `=`

The change here makes me press `Ctrl+` though, i.e. `Ctrl` and `Shift` and `=`. That seems odd. For `Ctrl -` its the same.

The Java editor zoom in / zoom out resets whatever the zoom level was in the console, which is probably OK.

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsolePage.java
```diff
@@ -393,4 +438,160 @@ public TextConsoleViewer getViewer() {
...
+		changeFontSize(-FONT_SIZE_STEP);
+	}
+
+	/**
+	 * Changes the font size of the console by the given delta.
+	 *
+	 * @param delta the amount to change the font size (positive to increase,
+	 *              negative to decrease)
+	 */
+	private void changeFontSize(int delta) {
```
trancexpress: It seems odd that I can do this from the Java editor without this code. Can you check if some existing code can be re-used here?

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsolePage.java
```diff
@@ -393,4 +438,160 @@ public TextConsoleViewer getViewer() {
...
+		int newHeight = Math.max(MIN_FONT_SIZE, Math.min(MAX_FONT_SIZE, currentHeight + delta));
+
+		if (newHeight == currentHeight) {
+			return;
+		}
+
+		// Create new font data with the new height
+		FontData[] newFontData = new FontData[fontData.length];
+		for (int i = 0; i < fontData.length; i++) {
+			newFontData[i] = new FontData(fontData[i].getName(), newHeight, fontData[i].getStyle());
```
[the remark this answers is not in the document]
raghucssit: Fixed.

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/TextConsolePage.java
```diff
@@ -145,6 +179,11 @@ protected void updateSelectionDependentActions() {
 	@Override
 	public void createControl(Composite parent) {
 		fViewer = createViewer(parent);
+		// Restore font for this console type (if saved)
+		Font prefFont = loadFontFromPreferences();
+		if (prefFont != null) {
+			fViewer.setFont(prefFont);
+		}
```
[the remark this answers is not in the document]
raghucssit: Fixed.

raghucssit:
@iloveeclipse Please check this PR.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+		Display display = Display.getCurrent();
+		if (display == null) {
+			return;
+		}
+		display.syncExec(() -> applyZoom(display, delta));
+	}
+
+	public static void applyZoom(Display display, int delta) {
+		// Capture the focused control.
+		Control focus = display.getFocusControl();
```
iloveeclipse: I don't know if `display.getFocusControl()` is not working as expected, but I consistently see various issues with various consoles.

Try to play with 3 diffeent consoles:
1) Run a simple Java code that outputs some stack trace - this will be the first console.
2) Copy/paste this trace to "Java stack trace" console - second one.
3) Open "Host OSGI Console" and type "help" there - third one.

Now open a text editor (Java code or just any plain txt file).

1) If you "Ctrl +" or "Ctrl+-" in the process output console, it works fine and editor is not affected.
2) If you "Ctrl +" or "Ctrl+-" in Java Stack Trace console, it automatically increases/decreases editor font size.
3) If you "Ctrl +" or "Ctrl+-" in "Host OSGI Console", it increases font size in the process output console too, but not otherwise.
4) And so on, just try with different consoles.
5) If I open a **second** Console view, font sizes seem to be properly updated in all consoles, however "Java Stack Trace" page seem to also always change font size in the editor.

My expectations are: 

1) The editor font size should **never** change if we change font in Console via "Ctrl +" or "Ctrl+-". **I guess** the problem is that the Java Stack Trace does not properly sets the context and the handler "thinks" it is the editor context. This can7should be fixed in a separated JDT UI fix.
2) The changes inside different Console pages should **ideally** work per page only, or (not so nice but acceptable) in all pages, but then at same time, and not via some random logic. So if I change the font size in the Console view, it is always applied consistently to either one or to all opened pages. I guess that switching Console pages doesn't always take the font size change into account if showing the previously hidden page.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+		changeFocusedFont(STEP);
+		return null;
+	}
+
+	private void changeFocusedFont(int delta) {
+		Display display = Display.getCurrent();
+		if (display == null) {
+			return;
+		}
+		display.syncExec(() -> applyZoom(display, delta));
```
iloveeclipse: if `Display.getCurrent() != null` we are in UI thread already. Why do we need `syncExec()` then?

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+				if (z != null && !z.isDisposed())
+					z.dispose();
+			});
+		}
+		if (oldZoom != null && !oldZoom.isDisposed()) {
+			oldZoom.dispose();
+		}
+
+		// Update shared JFace registry so other listeners can observe the change
+		JFaceResources.getFontRegistry().put(DEBUG_CONSOLE_FONT_REGISTRY_KEY, newFontData);
```
iloveeclipse: This line is responsible for really awkward behavior of different consoles I've observed.

I've tested few consoles and they all behave differently. Looks like consoles from Platform are OK and consistent (process console & ant), but JDT is different (it reads the font set for Console but it also changes the font for the editor) and the Host OSGI Console (PDE) doesn't care at all about default console font, same as SpotBugs (not from eclipse.org).

On one side it makes sense, every contributed console can have their own font defined, on the other side, they could derive their font from the base one.

So ideally we should fix JDT console (which propagates font changes to the editor), and the PDE (which should derive the font from Console preferences). It would make sense to check if CDT build console need some fix and fix that too, but first let fix the platform.

All other 3rd party plugins should fix their code by themselves.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+			return;
+		}
+		int currentHeight = fontData[0].getHeight();
+		int newHeight = Math.max(MIN_FONT_SIZE, Math.min(MAX_FONT_SIZE, currentHeight + delta));
+		if (newHeight == currentHeight) {
+			return;
+		}
+		FontData[] newFontData = fontData.clone();
+		for (FontData fd : newFontData) {
+			if (fd != null)
```
iloveeclipse: Please **always** use blocks for if/for/while statements, independently how many lines are inside.

iloveeclipse: Please **always** use blocks for if/for/while statements, independently how many lines are inside.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomOutHandler.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Command handler to decrease the font size of the focused console StyledText.
+ *
+ * @since 3.17
+ */
+public class ConsoleZoomOutHandler extends AbstractHandler {
+	private static final int STEP = 1;
+
+	@Override
+	public Object execute(ExecutionEvent event) throws ExecutionException {
```
iloveeclipse: Instead of overriding two methods, you can specify `protected int getStep()` in the super class and only override it here, or do what `AbstractTextZoomHandler` is doing and have an abstract class that does the job and two strivial handler classes specifying the increment in the constructor.

Also I wonder if the code similar to code in  `org.eclipse.ui.texteditor.AbstractTextZoomHandler.execute(ExecutionEvent)` can be used here (instead of directly modifying  StyledText`)?

debug/org.eclipse.ui.console/plugin.xml
```diff
@@ -157,6 +187,29 @@ M4 = Platform-specific fourth key
             id="org.eclipse.ui.commands.toggleState">
         </state>
       </command>
+      <command
```
iloveeclipse, line 190: These commands can be executed via "Ctrl+3", and guess what ... They change editor font, if it is currently active, or "Debug Shell" view font etc.

We need `<enabledWhen>` on the handler to restrict it to Console view only.

iloveeclipse, line 190: These commands can be executed via "Ctrl+3", and guess what ... They change current editor font, if editor is currently active, or "Debug Shell" view font etc.

We need `<enabledWhen>` on the handler to restrict it to Console view only.

iloveeclipse, line 190: These commands can be executed via "Ctrl+3", and guess what ... They change current editor font, if editor is currently active, or "Debug Shell" view font etc.

We need `<enabledWhen>` on the handler to restrict it to Console view only.

iloveeclipse, line 190: These commands can be executed via "Ctrl+3", and guess what ... They change current editor font, if editor is currently active, or "Debug Shell" view font etc.

We need `<enabledWhen>` on the handler to restrict it to Console view only.

iloveeclipse (changes requested):
Note: **before** pushing fixes, please rebase on master first.

iloveeclipse (changes requested):
Note: **before** pushing fixes, please rebase on master first.

iloveeclipse (changes requested):
Note: **before** pushing fixes, please rebase on master first.

iloveeclipse:
@raghucssit : I don't see changes **in the patch itself** after last rebase you did?

raghucssit:
I have fixed all the review comments and issues present at last review..
Current changes includes:
1. Change in font size at one console page is reflected at all the pages that console has ( and also pages that later added to console). This is done with the help of page listener.
2. Persistence of font between eclipse sessions. This is done using a non UI preference.
3. Command is restricted using `<enabledWhen>` tag of the Handler.
4. Using the async execution is needed because org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.launchChanged(ILaunch) starts a new ordinary Job(Not UIJob) which creates a new Console and add it to ConsoleManager. And ConsoleManager notifies listeners in the same non UI thread.

raghucssit:
Regarding bug that activates editor content on selection of Stacktrace Console has the below fix. I will create a separate issue at jdt debug repo and contribute below patch.

```
diff --git a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/console/JavaStackTracePageParticipant.java b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/console/JavaStackTracePageParticipant.java
index a46005d..ffe5977 100644
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/console/JavaStackTracePageParticipant.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/console/JavaStackTracePageParticipant.java
@@ -24,8 +24,6 @@
 import org.eclipse.ui.console.IConsoleConstants;
 import org.eclipse.ui.console.IConsolePageParticipant;
 import org.eclipse.ui.console.actions.CloseConsoleAction;
-import org.eclipse.ui.contexts.IContextActivation;
-import org.eclipse.ui.contexts.IContextService;
 import org.eclipse.ui.handlers.IHandlerActivation;
 import org.eclipse.ui.handlers.IHandlerService;
 import org.eclipse.ui.part.IPageBookViewPage;
@@ -38,7 +36,6 @@
     private CloseConsoleAction fCloseAction;
     private FormatStackTraceActionDelegate fFormatAction;
     private IHandlerActivation fHandlerActivation;
-    private IContextActivation fContextActivation;
 
     /* (non-Javadoc)
      * @see org.eclipse.ui.console.IConsolePageParticipant#init(org.eclipse.ui.part.IPageBookViewPage, org.eclipse.ui.console.IConsole)
@@ -87,9 +84,11 @@
         };
 
         fHandlerActivation = handlerService.activateHandler("org.eclipse.jdt.ui.edit.text.java.format", formatHandler); //$NON-NLS-1$
-
-        IContextService contextService = workbench.getAdapter(IContextService.class);
-        fContextActivation = contextService.activateContext("org.eclipse.jdt.ui.javaEditorScope"); //$NON-NLS-1$
+        // Note: intentionally not activating the "org.eclipse.jdt.ui.javaEditorScope" context here.
+        // That context is a child of "org.eclipse.ui.textEditorScope", so activating it would make
+        // generic text editor key bindings (e.g. Ctrl+=/Ctrl+- zoom) take priority over this
+        // console's own key bindings (e.g. the console zoom commands) while this page is showing.
+        // The format action remains reachable via its context menu entry and the handler above.
 	}
 
 	/* (non-Javadoc)
@@ -104,11 +103,6 @@
 			handlerService.deactivateHandler(fHandlerActivation);
 			fHandlerActivation = null;
 		}
-        if (fContextActivation != null){
-        	IContextService contextService = workbench.getAdapter(IContextService.class);
-        	contextService.deactivateContext(fContextActivation);
-        	fContextActivation = null;
-        }
 	}
 
 }
```

raghucssit:
Stacktrace of the Exception if we don't use Display UI execution. Async is used to avoid any possible deadlocks.

```
org.eclipse.swt.SWTException: Unable to create font "org.eclipse.jface.textfont" in a non-UI thread. Using default font instead.
	at org.eclipse.jface.resource.FontRegistry.getFontRecord(FontRegistry.java:700)
	at org.eclipse.jface.resource.FontRegistry.get(FontRegistry.java:631)
	at org.eclipse.jface.resource.JFaceResources.getFont(JFaceResources.java:318)
	at org.eclipse.ui.console.TextConsole.getDefaultFont(TextConsole.java:271)
	at org.eclipse.ui.console.TextConsole.getFont(TextConsole.java:260)
	at org.eclipse.ui.internal.console.ConsoleZoomInHandler.applyHeight(ConsoleZoomInHandler.java:278)
	at org.eclipse.ui.internal.console.ConsoleZoomInHandler$1.consolesAdded(ConsoleZoomInHandler.java:132)
	at org.eclipse.ui.internal.console.ConsoleManager$ConsoleNotifier.run(ConsoleManager.java:151)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.ui.internal.console.ConsoleManager$ConsoleNotifier.notify(ConsoleManager.java:172)
	at org.eclipse.ui.internal.console.ConsoleManager.fireUpdate(ConsoleManager.java:248)
	at org.eclipse.ui.internal.console.ConsoleManager.addConsoles(ConsoleManager.java:215)
	at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager$ConsoleCreation.run(ProcessConsoleManager.java:80)
```

debug/org.eclipse.ui.console/src/org/eclipse/ui/console/ConsolePlugin.java
```diff
@@ -183,6 +184,14 @@ public static ImageDescriptor getImageDescriptor(String key) {
 		return ConsolePluginImages.getImageDescriptor(key);
 	}
 
+	@Override
```
iloveeclipse: Please remove this, just init the ZoomHandler on ConsoleManager startup

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+ * <li>the zoom never affects an unrelated part (e.g. a text editor), since the
+ * command is only enabled/invoked while a console view is the active part;</li>
+ * <li>all pages of all Console views (and all consoles, whether currently shown
+ * or not) end up with the same font size, since they all react to font changes
+ * fired by their own console model;</li>
+ * <li>a console/page created after a zoom step already starts with the current
+ * zoom level, instead of resetting to the default size.</li>
+ * </ul>
+ *
+ * @since 3.17
```
iloveeclipse: `@since` is only needed on API. This class is not API.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+	 */
+	public static void applyZoom(IWorkbenchPart part, int delta) {
+		if (!(part instanceof IConsoleView consoleView)) {
+			return;
+		}
+		IConsole console = consoleView.getConsole();
+		if (!(console instanceof TextConsole textConsole)) {
+			return;
+		}
+		int baseHeight;
```
iloveeclipse: Please check for display here and run any operation whic accesses Font & Co from UI thread only (call asyncExec() as you do it below).

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+	 * {@link TextConsole#getFont()}, which lazily creates one) can only be touched
+	 * on the UI thread. If called from any other thread, this re-dispatches itself
+	 * onto the UI thread asynchronously and returns immediately.
+	 * </p>
+	 *
+	 * @param textConsole the console to update
+	 * @param height      the font height to apply, in points
+	 */
+	private static void applyHeight(TextConsole textConsole, int height) {
+		Display display = Display.getDefault();
```
iloveeclipse: Instead of this 
```
Display display = Display.getDefault();
if (display.getThread() != Thread.currentThread()) 
```
please use `if (Display.getCurrent() == null)`, here and in other places

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+	 *
+	 * @param textConsole the console whose zoom font should be disposed
+	 */
+	private static void disposeZoomFont(TextConsole textConsole) {
+		Object attribute = textConsole.getAttribute(ZOOM_FONT_ATTRIBUTE);
+		if (!(attribute instanceof Font font) || font.isDisposed()) {
+			return;
+		}
+		Display display = Display.getDefault();
+		if (display.getThread() != Thread.currentThread()) {
```
iloveeclipse: Please start the method with this check.

iloveeclipse (changes requested):
Quick review: have not tested yet, but I see some pieces that need to be fixed.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+ *
+ * @since 3.17
+ */
+public class ConsoleZoomInHandler extends AbstractHandler {
+
+	/**
+	 * Key used to remember, on the console itself, the custom font created for
+	 * zooming, so it can be reused/replaced and eventually disposed.
+	 */
+	static final String ZOOM_FONT_ATTRIBUTE = ConsoleZoomInHandler.class.getName() + ".zoomFont"; //$NON-NLS-1$
```
iloveeclipse: I would not store the "global" `zoom font size` but the `zoomStep` value **as delta**, **and/or** store the delta value **per console type**.

Every console can define its own font & size, and font sizes defined by different consoles may differ. Consoles may have individual font preferences settings which can be configured.

Current proposal increases font size for all consoles and sets it to the font size defined by the **current** console +- zoom step value. Also it enforces **last zoomed value** even if user changes console font size in preferences.

This all resuls in an unexpected behavior:

1) If the current console has "unusual" font size and/or other consoles font size differ from the current one and if the user only wanted to change font size in the current console only. Here I would expect that if we apply zoom, we apply zoom delta (+-1 step) to the currently set font size in each console.
2) If the user changes Console font size for some console to "20" in preferences but we already had "zoomed" to font size "10", it will stay at 10. I would expect that we rezet zoom delta once user changes the font size via preferences, so that whatever is chosen by user in preferences "wins".

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,329 @@
...
+				return;
+			}
+			baseHeight = fontData[0].getHeight();
+		}
+		int newHeight = Math.max(MIN_FONT_SIZE, Math.min(MAX_FONT_SIZE, baseHeight + delta));
+		if (currentZoomHeight != null && newHeight == currentZoomHeight.intValue()) {
+			return;
+		}
+		sZoomHeight = Integer.valueOf(newHeight);
+		persistZoomHeight(newHeight);
```
iloveeclipse: Actually the code above can be removed entirely and we can simply iterate over all consoles below, applying the **zoom step** to each console font size (not the font size of the current console)

raghucssit:
1. Now there is a per console page persisted font height(base height + delta). All the pages don't share common height. So If we zoom in ProcessConsole it applies only there and if we go to Stacktrace Console then it has it's own font. If we zoom at Stacktrace Console its zoom is persisted for it.
2. If User changes the font in preference page then zoomed value is reset to preference value.
3. persisted values are restored per console type.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,412 @@
...
+	 * {@link #FONT_ENFORCER} to them, and disposes each console's custom zoom
+	 * font when it is removed.
+	 */
+	private static final IConsoleListener ZOOM_FONT_LISTENER = new IConsoleListener() {
+		@Override
+		public void consolesAdded(IConsole[] consoles) {
+			for (IConsole console : consoles) {
+				if (console instanceof TextConsole textConsole) {
+					textConsole.addPropertyChangeListener(FONT_ENFORCER);
+					ZoomState state = sZoomByType.get(typeKey(textConsole));
```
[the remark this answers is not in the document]
iloveeclipse: Incorrect, ListenerList#add checks for duplicated listeners added.

iloveeclipse (commented):
Looks good.

Please resolve not yet addressed Copilot comments (test, font creation & disposal) and we are good to go.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,412 @@
...
+		}
+		FontData[] fontData = currentFont.getFontData();
+		if (fontData == null || fontData.length == 0 || fontData[0].getHeight() == height) {
+			return;
+		}
+		FontData[] newFontData = new FontData[fontData.length];
+		for (int i = 0; i < fontData.length; i++) {
+			FontData fd = fontData[i];
+			newFontData[i] = new FontData(fd.getName(), height, fd.getStyle());
+		}
```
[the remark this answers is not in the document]
raghucssit: Fixed

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomHandler.java
```diff
@@ -0,0 +1,412 @@
...
+	 * that type. Must run on the UI thread; re-dispatches itself otherwise.
+	 *
+	 * @param part  the active part; must be an {@link IConsoleView} for anything to happen
+	 * @param delta the font size delta to apply, in points
+	 */
+	public static void applyZoom(IWorkbenchPart part, int delta) {
+		if (Display.getCurrent() == null) {
+			Display.getDefault().asyncExec(() -> applyZoom(part, delta));
+			return;
+		}
```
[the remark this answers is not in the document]
raghucssit: Tests added.

debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
```diff
@@ -0,0 +1,412 @@
...
+				return;
+			}
+			if (!Boolean.TRUE.equals(textConsole.getAttribute(REENFORCED_ATTRIBUTE))) {
+				textConsole.setAttribute(REENFORCED_ATTRIBUTE, Boolean.TRUE);
+				applyHeight(textConsole, state.height());
+				return;
+			}
+		}
+		sZoomByType.put(type, new ZoomState(currentHeight.intValue(), 0));
+		persistZoomStates();
```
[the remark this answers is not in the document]
raghucssit: Fixed.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
```diff
@@ -233,4 +244,409 @@ public void testIOConsoleAvailable() throws Exception {
...
+	}
+
+	private static void processUiEvents() {
+		Display display = Display.getCurrent();
+		if (display != null) {
+			while (display.readAndDispatch()) {
+				// drain pending UI work (e.g. asyncExec calls) triggered by the zoom handler
+			}
+		}
+	}
```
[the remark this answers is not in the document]
raghucssit: Fixed

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
```diff
@@ -233,4 +243,400 @@ public void testIOConsoleAvailable() throws Exception {
...
+		return new Font(font.getDevice(), fontData.getName(), height, fontData.getStyle());
+	}
+
+	/**
+	 * Reads the raw, persisted per-console-type zoom state string via
+	 * reflection, since {@code ConsoleZoomInHandler} intentionally does not
+	 * expose its preference key as public API.
+	 */
+	private static String getPersistedZoomState() throws ReflectiveOperationException {
+		Field prefKeyField = ConsoleZoomInHandler.class.getDeclaredField("PREF_ZOOM_FONT_HEIGHTS"); //$NON-NLS-1$
```
iloveeclipse: Please if you already created helper function `getStaticIntField`, use it consistently.

But I would ask here instead: remove all the reflection and make all the required constants in `ConsoleZoomInHandler` public.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
```diff
@@ -233,4 +243,400 @@ public void testIOConsoleAvailable() throws Exception {
...
+	/**
+	 * When a console is removed, its custom zoom font must be disposed so it
+	 * is not leaked (see also the {@code SWT Resource was not properly
+	 * disposed} regression this guards against).
+	 */
+	@Test
+	public void testRemovingConsoleDisposesZoomFont(TestInfo testInfo) throws Exception {
+		String type = uniqueConsoleType("disposeOnRemove"); //$NON-NLS-1$
+
+		IWorkbenchPage activePage = PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage();
```
iloveeclipse: The code which opens console view below repeats in every test. Please extract method for that.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
```diff
@@ -233,4 +243,400 @@ public void testIOConsoleAvailable() throws Exception {
...
+			TestUtil.waitForJobs(testInfo.getDisplayName(), ConsoleManager.CONSOLE_JOB_FAMILY, 100, 3000);
+			TestUtil.processUIEvents();
+
+			int naturalHeight = getFontHeight(console);
+
+			IHandlerService handlerService = PlatformUI.getWorkbench().getService(IHandlerService.class);
+			handlerService.executeCommand("org.eclipse.ui.console.command.fontZoomIn", null); //$NON-NLS-1$
+			TestUtil.processUIEvents();
+
+			assertEquals(naturalHeight + 1, getFontHeight(console),
```
iloveeclipse, line 501: The second command (zoom out) is not verified, but should

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
```diff
@@ -233,4 +243,400 @@ public void testIOConsoleAvailable() throws Exception {
...
+	 */
+	@Test
+	public void testRemovingConsoleDisposesZoomFont(TestInfo testInfo) throws Exception {
+		String type = uniqueConsoleType("disposeOnRemove"); //$NON-NLS-1$
+
+		IWorkbenchPage activePage = PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage();
+		IViewPart consoleViewPart = activePage.showView(IConsoleConstants.ID_CONSOLE_VIEW);
+		IConsoleView consoleView = (IConsoleView) consoleViewPart;
+
+		MessageConsole console = new MessageConsole("Zoom Dispose Console", type, null, //$NON-NLS-1$
```
iloveeclipse: The code which creates and adds console to the manager below repeats in every test, just with different console names. Please extract method for that.

debug/org.eclipse.debug.tests/src/org/eclipse/debug/tests/console/ConsoleTests.java
iloveeclipse: And while writing this above, I've asked myself why do we need **two** handler classes at all, where second one doesn't do anything but just redirects to the first one with a different step?


Instead, we should have one `ConsoleZoomHandler` handler and parametrize that. Please check patch below which provides minimal changes (without rename to  `ConsoleZoomHandler` for simplicity). Please apply it and rename the one handler to `ConsoleZoomHandler` & update Console tests to the new name.

```
diff --git a/debug/org.eclipse.ui.console/plugin.xml b/debug/org.eclipse.ui.console/plugin.xml
index 48cd9f4..a6a8aef 100644
--- a/debug/org.eclipse.ui.console/plugin.xml
+++ b/debug/org.eclipse.ui.console/plugin.xml
@@ -214,3 +214,3 @@
       <handler
-            class="org.eclipse.ui.internal.console.ConsoleZoomOutHandler"
+            class="org.eclipse.ui.internal.console.ConsoleZoomInHandler:-1"
             commandId="org.eclipse.ui.console.command.fontZoomOut">
diff --git a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
index b7b2026..c3361fd 100644
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
+++ b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomInHandler.java
@@ -22,2 +22,5 @@
 import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExecutableExtension;
 import org.eclipse.core.runtime.Status;
@@ -41,4 +44,4 @@
 /**
- * Command handler to increase the font size of the console currently shown in
- * the active console view.
+ * Command handler to increase or decrease the font size of the console
+ * currently shown in the active console view.
  * <p>
@@ -50,4 +53,11 @@
  * </p>
+ * <p>
+ * This single class backs both the zoom-in and zoom-out commands: the
+ * direction is supplied as executable extension data in {@code plugin.xml},
+ * e.g. {@code class="...ConsoleZoomInHandler:-1"} for zooming out. Each
+ * {@code <handler>} registration gets its own instance, so there is no shared
+ * mutable state between the two directions.
+ * </p>
  */
-public class ConsoleZoomInHandler extends AbstractHandler {
+public class ConsoleZoomInHandler extends AbstractHandler implements IExecutableExtension {
 
@@ -70,2 +80,10 @@
 	/**
+	 * The zoom direction/step size for this particular handler instance, in
+	 * points; positive to zoom in, negative to zoom out. Set from the
+	 * executable extension data configured in {@code plugin.xml} (defaults to
+	 * {@link #STEP}, i.e. zoom in, if none is provided).
+	 */
+	private int step = STEP;
+
+	/**
 	 * Key used, in the per-type persisted zoom string, for consoles that report a
@@ -269,5 +287,17 @@
 	@Override
+	public void setInitializationData(IConfigurationElement config, String propertyName, Object data)
+			throws CoreException {
+		if (data instanceof String stepData) {
+			try {
+				step = Integer.parseInt(stepData.trim());
+			} catch (NumberFormatException e) {
+				// keep default (zoom in) if the configured data is not a valid integer
+			}
+		}
+	}
+
+	@Override
 		IWorkbenchPart part = HandlerUtil.getActivePart(event);
-		applyZoom(part, STEP);
+		applyZoom(part, step);
 		return Status.OK_STATUS;
diff --git a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomOutHandler.java b/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomOutHandler.java
deleted file mode 100644
index aef9296..0000000
--- a/debug/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ConsoleZoomOutHandler.java
+++ /dev/null
@@ -1,38 +0,0 @@
-/*******************************************************************************
- * Copyright (c) 2026 Advantest Europe GmbH and others.
- *
- * This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License 2.0
- * which accompanies this distribution, and is available at
- * https://www.eclipse.org/legal/epl-2.0/
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- * 				Raghunandana Murthappa
- *******************************************************************************/
-package org.eclipse.ui.internal.console;
-
-import org.eclipse.core.commands.AbstractHandler;
-import org.eclipse.core.commands.ExecutionEvent;
-import org.eclipse.core.commands.ExecutionException;
-import org.eclipse.core.runtime.Status;
-import org.eclipse.ui.IWorkbenchPart;
-import org.eclipse.ui.handlers.HandlerUtil;
-
-/**
- * Command handler to decrease the font size of the console currently shown in
- * the active console view.
- *
- * @see ConsoleZoomInHandler
- */
-public class ConsoleZoomOutHandler extends AbstractHandler {
-	private static final int STEP = 1;
-
-	@Override
-	public Object execute(ExecutionEvent event) throws ExecutionException {
-		IWorkbenchPart part = HandlerUtil.getActivePart(event);
-		ConsoleZoomInHandler.applyZoom(part, -STEP);
-		return Status.OK_STATUS;
-	}
-}

```

iloveeclipse: `org.eclipse.ui.console.command.fontZoomOut` command/handler is still not tested

iloveeclipse:
```
			handlerService.executeCommand("org.eclipse.ui.console.command.fontZoomOut", null); //$NON-NLS-1$
			TestUtil.processUIEvents();

			assertEquals(naturalHeight, getFontHeight(console),
					"executing the zoom-out command should decrease the font size of the active console"); //$NON-NLS-1$
```

iloveeclipse:
I've pushed test change and rebased on master.

iloveeclipse (approved):

PR #2582, opened 2026-03-20
Initialize Step Filters button state from DebugUITools preference

SougandhS:
see https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/914

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/DebugPreferenceInitializer.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/DebugPreferenceInitializer.java
@@ -26,7 +26,7 @@ public class DebugPreferenceInitializer extends AbstractPreferenceInitializer {
 	@Override
 	public void initializeDefaultPreferences() {
 		// Step filter preferences
-		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), StepFilterManager.PREF_USE_STEP_FILTERS, false);
+		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), StepFilterManager.PREF_USE_STEP_FILTERS, true);
 		//launch configurations preferences
 		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), DebugPlugin.PREF_DELETE_CONFIGS_ON_PROJECT_DELETE, false);
 		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), IInternalDebugCoreConstants.PREF_ENABLE_STATUS_HANDLERS, true);


iloveeclipse:
Tigether with this PR & https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/921 I see now that the Java -> Debug -> Step filtering checkbox is **always** enabled, even if I uncheck it & do apply & close. Next tme preferences page is opened is is again enabled.

SougandhS:
> Tigether with this PR & [eclipse-jdt/eclipse.jdt.debug#921](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/921) I see now that the Java -> Debug -> Step filtering checkbox is **always** enabled, even if I uncheck it & do apply & close. Next tme preferences page is opened is is again enabled.

Fixed this in JDT

iloveeclipse:
> Fixed this in JDT

Thanks, it works now.

After that I've checked the code and I wonder if you've seen this line

https://github.com/eclipse-platform/eclipse.platform/blob/2222ff69a218ccb7d08dcdf2dce8848ba7a6ddab/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/commands/actions/ToggleStepFiltersAction.java#L40

coming from e6341a18a8aad7f331cd68ff65ee8a7d53ee5d3c.

Also most of the code in `ToggleStepFiltersAction` seem to be added by that commit and to deal with the right action state. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=162320.

This need to be understood what the mentioned bug fixed and why we need to fix it again.

iloveeclipse:
BTW, while reviewing this PR I've found https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/924.

debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/DebugPreferenceInitializer.java
```diff
@@ -26,7 +26,7 @@ public class DebugPreferenceInitializer extends AbstractPreferenceInitializer {
 	@Override
 	public void initializeDefaultPreferences() {
 		// Step filter preferences
-		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), StepFilterManager.PREF_USE_STEP_FILTERS, false);
+		Preferences.setDefaultBoolean(DebugPlugin.getUniqueIdentifier(), StepFilterManager.PREF_USE_STEP_FILTERS, true);
```
iloveeclipse, line 29: The problem with this is that it will change defaults for every bundle that uses the preference, even if no JDT is included. So if some code used this preference, it will be now affected, and that is not nice (ideally we enable it only if JDT is there).

See https://github.com/search?q=PREF_USE_STEP_FILTERS&type=code&p=1

What really surprises me, is that if I add exact same code like here in `JDIDebugUIPreferenceInitializer.initializeDefaultPreferences()`, it will **not** work with product customization, but here it works. 

What is also strange, is that if I debug here and see what is reported by the code below in the debug shell

```
org.eclipse.core.runtime.Platform.getPreferencesService().getBoolean(
		org.eclipse.debug.core.DebugPlugin.getUniqueIdentifier(), 
		org.eclipse.debug.internal.core.StepFilterManager.PREF_USE_STEP_FILTERS, false, null);
```

the value is shown as "true" after this line (even if product property is set to "false"), but once devbugger stops at `JDIDebugUIPreferenceInitializer.initializeDefaultPreferences()`, the same evaluation shows "false" again.

This is really strange.

SougandhS: Do you have any suggestions/approach for the fix ?

PR #2583, opened 2026-03-21
Show mouse move control even when not showing single line

merks:
With this simple change, the controls to do the merging still appear when you hover over the area that's connecting the changes on the two sides:

image

@HannesWell @akurtakov 

I think this addresses your concerns about lost of functionality.

@iloveeclipse 

Given there is no visual appearance until you hover, I assume this is not a controversial addition.


merks:
The windows tests passed when I restarted it, so the summary is incorrect, i.e., not updated.

image

merks:
FYI, I also experiment with commenting out the `fUseSingleLine ` guard here:

https://github.com/eclipse-platform/eclipse.platform/blob/2222ff69a218ccb7d08dcdf2dce8848ba7a6ddab/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java#L4580-L4592

It's more visual clutter but better (familiar) affordance:

image

I though maybe circles would look prettier... But then I thought it seemed more likely to raise objects if there visual changes.

Hopefully folks will feel motivated to make further improvements.

PR #2586, opened 2026-03-23
Set Creation Time as Default Sorting Order in Breakpoints View

SougandhS:
This change updates the default ordering of breakpoints in the Breakpoints view to sort them based on their creation time. At present, the default ordering does not emphasize when a breakpoint was added, which can make it less intuitive for users to locate the most recently created breakpoints during an active debugging session. By setting creation time as the default sorting criteria, the view better aligns with common debugging workflows where recently added breakpoints are typically the most relevant and frequently accessed.

Screenshot 2026-03-23 at 3 45 41 PM


SougandhS:
Hi @iloveeclipse, WDYT ?

iloveeclipse:
Makes sense, this is my default setting too.

SougandhS:
Thank you @iloveeclipse

PR #2587, opened 2026-03-24
Improve tracing options in JobManager and DeadlockDetector

fedejeanne:
Add the following to the .options file so that one can choose to trace these events via the preferences (`General > Tracing`):
- jobs/yielding
- jobs/yielding/detailed
- jobs/blockedunblocked

Do not log to sysout for the preference "jobs/locks", use the same method as other tracers.

Show the name of system jobs that block other jobs in `JobManager.reportBlocked(...)`


fedejeanne:
If there are no objections, I plan to merge this week

Issue #2588, opened 2026-03-24
Provide options for disabling history on some files


trancexpress:
We have some sensitive files in the workspace, for which we want to avoid Eclipse saving history content on the disk. We would like a way to disable history for such files. We don't require UIs to mark the files, our product knows which files are sensitive at runtime.

trancexpress:
Note that `org.eclipse.core.resources.IResource.clearHistory(IProgressMonitor)` is not exactly what we want. We want no history to be written for the files.

PR #2589, opened 2026-03-25
Add preference for disabling history on restricted files

trancexpress:
This change adds a new preference to `org.eclipse.core.resources`:
```
disableRestrictedFileHistory
```
If the preference is set to 'true' and a file is restricted, setting the contents of the file will result in no new history entry for the edit.

Example preference for product customization:
```
org.eclipse.core.resources/disableRestrictedFileHistory=true
```
Fixes: https://github.com/eclipse-platform/eclipse.platform/issues/2588

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -84,6 +84,8 @@ import org.xml.sax.InputSource;
  */
 public class FileSystemResourceManager implements ICoreConstants, IManager {
 
+	private static final String DISABLE_RESTRICTED_FILE_HISTORY_PREFERENCE = "disableRestrictedFileHistory"; //$NON-NLS-1$
+
 	/**
 	 * The history store is initialized lazily - always use the accessor method
 	 */
@@ -91,11 +93,15 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 	protected Workspace workspace;
 
 	private volatile boolean lightweightAutoRefreshEnabled;
+	private volatile boolean disableRestrictedFileHistory;
 
-	private final IPreferenceChangeListener lightweightAutoRefreshPrefListener = event -> {
-		if (ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH.equals(event.getKey())) {
+	private final IPreferenceChangeListener prefListener = event -> {
+		String preferenceName = event.getKey();
+		if (ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH.equals(preferenceName)) {
 			lightweightAutoRefreshEnabled = Platform.getPreferencesService().getBoolean(ResourcesPlugin.PI_RESOURCES,
 					ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH, false, null);
+		} else if (DISABLE_RESTRICTED_FILE_HISTORY_PREFERENCE.equals(preferenceName)) {
+			setDisableRestrictedFileHistory();
 		}
 	};
 
@@ -1203,15 +1209,16 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 			_historyStore.shutdown(monitor);
 		}
 		InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES)
-				.removePreferenceChangeListener(lightweightAutoRefreshPrefListener);
+				.removePreferenceChangeListener(prefListener);
 	}
 
 	@Override
 	public void startup(IProgressMonitor monitor) {
 		InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES)
-				.addPreferenceChangeListener(lightweightAutoRefreshPrefListener);
+				.addPreferenceChangeListener(prefListener);
 		lightweightAutoRefreshEnabled = Platform.getPreferencesService().getBoolean(ResourcesPlugin.PI_RESOURCES,
 				ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH, false, null);
+		setDisableRestrictedFileHistory();
 	}
 
 	/**
@@ -1497,7 +1504,31 @@ public class FileSystemResourceManager implements ICoreConstants, IManager {
 
 	public boolean storeHistory(IResource file) {
 		WorkspaceDescription description = workspace.internalGetDescription();
-		return description.isKeepDerivedState() || !file.isDerived();
+		return (description.isKeepDerivedState() || !file.isDerived()) && !disableHistory(file);
 	}
 
+	private void setDisableRestrictedFileHistory() {
+		disableRestrictedFileHistory = Platform.getPreferencesService().getBoolean(ResourcesPlugin.PI_RESOURCES,
+				DISABLE_RESTRICTED_FILE_HISTORY_PREFERENCE, false, null);
+	}
+
+	private boolean disableHistory(IResource resource) {
+		if (disableRestrictedFileHistory) {
+			if (resource instanceof IFile file) {
+				try {
+					return file.isContentRestricted();
+				} catch (CoreException e) {
+					Policy.log(e.getStatus());
+					/*
+					 * The preference 'disableRestrictedFileHistory' indicates we should skip
+					 * restricted files, but we ran into an exception while checking if the file is
+					 * restricted. Disable history for the file, since we don't know if the file is
+					 * restricted or not.
+					 */
+					return true;
+				}
+			}
+		}
+		return false;
+	}
 }
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/DisableHistoryTests.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/DisableHistoryTests.java
@@ -1,0 +1,214 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.tests.internal.localstore;
+
+import static org.eclipse.core.resources.ResourcesPlugin.getWorkspace;
+import static org.eclipse.core.tests.resources.ResourceTestUtil.createInWorkspace;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFileState;
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.core.tests.resources.util.WorkspaceResetExtension;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.osgi.service.prefs.BackingStoreException;
+
+/**
+ * Tests disabling file history based on a workspace instance preference and a
+ * per-file restriction flag. The tests set the preference
+ * {@code org.eclipse.core.resources/disableRestrictedFileHistory} to
+ * {@code true} and mark specific files as having restricted content (via
+ * {@link IFile#setContentRestricted(boolean)}). After the preference is enabled
+ * and the file is marked restricted, subsequent edits to that file are not
+ * expected to create new file history states.
+ */
+@ExtendWith(WorkspaceResetExtension.class)
+public class DisableHistoryTests {
+
+	private static final String DISABLE_PREFERENCE_NAME = "disableRestrictedFileHistory";
+
+	private IProject project;
+
+	@BeforeEach
+	public void createTestProject() throws Exception {
+		project = getWorkspace().getRoot().getProject("Project");
+		createInWorkspace(project);
+	}
+
+	@AfterEach
+	public void cleanUp() throws Exception {
+		setHistoryDisablePreference(false);
+	}
+
+	@Test
+	public void testDisableHistoryBeforeEdit() throws Exception {
+		IFile file = project.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(0, history.length, "Unexpected history after disable: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryAfterEdit() throws Exception {
+		IFile file = project.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history before disable: " + toString(history));
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 3");
+		history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after disable: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryBeforeMove() throws Exception {
+		IFile file = project.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history before disable: " + toString(history));
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		IFile movedFile = project.getFile("test2.txt");
+		file.move(movedFile.getFullPath(), IResource.FORCE, null);
+		history = movedFile.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after move: " + toString(history));
+		writeContents(movedFile, "edit 3");
+		history = movedFile.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after edit: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryBeforeMovingParentFolder() throws Exception {
+		IFolder folder = project.getFolder("test_folder");
+		folder.create(true, true, null);
+		IFile file = folder.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history before disable: " + toString(history));
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		IFolder movedFolder = project.getFolder("test_folder2");
+		folder.move(movedFolder.getFullPath(), true, null);
+		IFile movedFile = movedFolder.getFile("test.txt");
+		history = movedFile.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after move: " + toString(history));
+		writeContents(movedFile, "edit 3");
+		history = movedFile.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after edit: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryBeforeDelete() throws Exception {
+		IFile file = project.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history before disable: " + toString(history));
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 3");
+		file.delete(true, null);
+		history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after delete: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryBeforeDeletingParentFolder() throws Exception {
+		IFolder folder = project.getFolder("test_folder");
+		folder.create(true, true, null);
+		IFile file = folder.getFile("test.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history before disable: " + toString(history));
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 3");
+		folder.delete(true, null);
+		history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history after delete: " + toString(history));
+	}
+
+	@Test
+	public void testHistoryNotDisabledForUnrestrictedFile() throws Exception {
+		IFile file = project.getFile("unrestricted.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		setHistoryDisablePreference(true);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(2, history.length, "Unexpected history for unrestricted file: " + toString(history));
+	}
+
+	@Test
+	public void testHistoryReenable() throws Exception {
+		IFile file = project.getFile("unrestricted.txt");
+		file.create("initial".getBytes(), IResource.FORCE, null);
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(0, history.length, "Unexpected history after disable: " + toString(history));
+		file.setContentRestricted(false);
+		writeContents(file, "edit 3");
+		history = file.getHistory(null);
+		assertEquals(1, history.length, "Unexpected history after unrestricting file: " + toString(history));
+	}
+
+	private void writeContents(IFile file, String contents) throws CoreException {
+		file.setContents(contents.getBytes(), IResource.KEEP_HISTORY, null);
+	}
+
+	private static String toString(IFileState... states) {
+		return Arrays.toString(states);
+	}
+
+	private static void setRestricted(IFile file) throws CoreException {
+		file.setContentRestricted(true);
+	}
+
+	private static void setHistoryDisablePreference(boolean enabled) throws BackingStoreException {
+		IEclipsePreferences node = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES);
+		if (enabled) {
+			node.putBoolean(DISABLE_PREFERENCE_NAME, true);
+		} else {
+			node.remove(DISABLE_PREFERENCE_NAME);
+		}
+		node.flush();
+	}
+}
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/AllLocalStoreTests.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/AllLocalStoreTests.java
@@ -35,6 +35,7 @@ import org.junit.platform.suite.api.Suite;
 		SafeFileInputOutputStreamTest.class, //
 		SymlinkResourceTest.class, //
 		UnifiedTreeTest.class, //
+		DisableHistoryTests.class, //
 })
 public class AllLocalStoreTests {
 


trancexpress:
Undo seems to work even without the history, so no problem with that part of editing.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1497,7 +1506,22 @@ public void writeSilently(IProject target) throws CoreException {
 
 	public static boolean storeHistory(IResource file) {
 		WorkspaceDescription description = ((Workspace) file.getWorkspace()).internalGetDescription();
-		return description.isKeepDerivedState() || !file.isDerived();
+		return (description.isKeepDerivedState() || !file.isDerived()) && !disableHistory(file);
 	}
 
+	private static boolean disableHistory(IResource resource) {
+		if (resource.getType() == IResource.FILE && resource.getWorkspace() instanceof Workspace ws) {
+			FileSystemResourceManager manager = ws.getFileSystemManager();
```
iloveeclipse: I had to read it twice to understand why do we ask ws for filesystem manager while being inside filesystem manager.

Also the existing code above looks weird for same reason, asking for a workspace while being inside fs manager for exact this workspace.

Can you please apply this patch on top, which makes it explicit that we are not in some static context but work in the only file system manager created for specific workspace:

```
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/DeleteVisitor.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/DeleteVisitor.java
index b19cc97..b42d2e7 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/DeleteVisitor.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/DeleteVisitor.java
@@ -23,2 +23,3 @@
 import org.eclipse.core.internal.resources.Resource;
+import org.eclipse.core.internal.resources.Workspace;
 import org.eclipse.core.internal.utils.Messages;
@@ -128,3 +129,3 @@
 			}
-			if (FileSystemResourceManager.storeHistory(node.getResource())) {
+			if (((Workspace) target.getWorkspace()).getFileSystemManager().storeHistory(target)) {
 				store.addState(target.getFullPath(), node.getStore(), info, true);
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
index 771bb54..e30dbd3 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
@@ -1367,4 +1367,3 @@
 		// add entry to History Store.
-		if (BitMask.isSet(updateFlags, IResource.KEEP_HISTORY) && fileInfo.exists()
-				&& FileSystemResourceManager.storeHistory(target)) {
+		if (BitMask.isSet(updateFlags, IResource.KEEP_HISTORY) && fileInfo.exists() && storeHistory(target)) {
 			// never move to the history store, because then the file is missing if write
@@ -1506,4 +1505,4 @@
 
-	public static boolean storeHistory(IResource file) {
-		WorkspaceDescription description = ((Workspace) file.getWorkspace()).internalGetDescription();
+	public boolean storeHistory(IResource file) {
+		WorkspaceDescription description = workspace.internalGetDescription();
 		return (description.isKeepDerivedState() || !file.isDerived()) && !disableHistory(file);
@@ -1511,6 +1510,5 @@
 
-	private static boolean disableHistory(IResource resource) {
-		if (resource.getType() == IResource.FILE && resource.getWorkspace() instanceof Workspace ws) {
-			FileSystemResourceManager manager = ws.getFileSystemManager();
-			QualifiedName property = manager.disableHistoryProperty;
+	private boolean disableHistory(IResource resource) {
+		if (resource.getType() == IResource.FILE) {
+			QualifiedName property = disableHistoryProperty;
 			if (property != null) {
diff --git a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
index 4d78eab..cac927a 100644
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
@@ -63,3 +63,3 @@
 	public void addToLocalHistory(IFile file) {
-		if (!FileSystemResourceManager.storeHistory(file)) {
+		if (!localManager.storeHistory(file)) {
 			return;
```

trancexpress: Lets make a separate PR for that and merge it before this one. I'll prepare it in a moment. IMO Its not related to this change, only triggered by it.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/DisableHistoryTests.java
```diff
@@ -0,0 +1,113 @@
...
+		createInWorkspace(project);
+	}
+
+	@AfterEach
+	public void cleanUp() throws Exception {
+		setHistoryDisablePreference(null);
+	}
+
+	@Test
+	public void testDisableHistoryBeforeEdit() throws Exception {
```
iloveeclipse, line 65: The test miss the cases of deleted / moved files and same for files in deleted / moved folders (trees).

Which also opens another question, whether the session property on a moved file is preserved or not (it should be IMO, but who knows).

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
iloveeclipse: OK, as you like.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/DisableHistoryTests.java
trancexpress: I don't understand what we want to test with deleted files/trees?

iloveeclipse: That there is no local history preserved for these files, independently how they were modified.

So file a.txt moved (renamed) to b.txt or file a.txt is removed. or directory a/b/c.txt is moved / deleted?

1) Does local history contain the a.txt / c.txt entries or not?
2) In case of moved file: is the target file still marked as "not to be tracked" by the local history or not? IMO it should be.

trancexpress: Please check now, I'm not sure if I understand what you want for the delete case.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1497,7 +1505,31 @@ public void writeSilently(IProject target) throws CoreException {
...
+				DISABLE_HISTORY_PREFERENCE_NAME, "", null); //$NON-NLS-1$
+		disableHistoryProperty = propertyName.isEmpty() ? null : new QualifiedName(null, propertyName);
 	}
 
+	private boolean disableHistory(IResource resource) {
+		if (resource.getType() == IResource.FILE) {
+			QualifiedName property = disableHistoryProperty;
+			if (property != null) {
+				try {
+					return resource.getSessionProperty(property) != null;
```
laeubi: I think a session property is not really sufficient.

Instead it should be stored like we do for encoding already for example:

```
eclipse.preferences.version=1
encoding/<project>=UTF-8
encoding//target/maven-archiver=UTF-16
encoding/testme.txt=ISO-8859-1
```

that way it would be relative easy to even make this available to users on the resource setting page and prevents the history even if no specialized plugin that "knows". Further it will allow to possible have wildcards for folders in a future and so on.

trancexpress: From my POV out-of-scope for: https://github.com/eclipse-platform/eclipse.platform/issues/2588

Our goal is to mark specific files, for the duration of the Eclipse session (for now, that might change in future). Our application knows what to mark, there is no common file name pattern or extension.

We also don't plan on additing UIs for this, since its nothing that the user should be able to control.

@iloveeclipse does Advantest plan on contributing any of the above? If so, I suggest opening a new issue for it.

laeubi: > From my POV out-of-scope for: https://github.com/eclipse-platform/eclipse.platform/issues/2588

What part of it?

> Our goal is to mark specific files, for the duration of the Eclipse session (for now, that might change in future).

That's fine but Eclipse Platform is a generic platform for many people not tailored for a specific product. So it always should have extensibility and reuse in mind.

> We also don't plan on additing UIs for this, since its nothing that the user should be able to control.

Also if you do not plan it at the moment others might choose to do so in the future.

> @iloveeclipse does Advantest plan on contributing any of the above? 

As mentioned no one will force you to implement additional features, but the foundation of such thing must be solid and in the scope and spirit of Eclipse platform and a session property seems not fitting very well.

trancexpress:
> @trancexpress please don't dismiss my review here just because you think it is out of scope from your point of view!

I don't think its out-of-scope, it is out-of-scope.

> This is not how Eclipse Platform development works!

Blocking PRs for arbitrary reasons is also not how platform development has worked, at least not in the last 10 years.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
trancexpress: > What part of it?

Here is the description, and what we need:

> We have some sensitive files in the workspace, for which we want to avoid Eclipse saving history content on the disk. We would like a way to disable history for such files. We don't require UIs to mark the files, our product knows which files are sensitive at runtime.

> That's fine but Eclipse Platform is a generic platform for many people not tailored for a specific product. So it always should have extensibility and reuse in mind.

Sure, the new preference can be used by any client that has similar needs.

> As mentioned no one will force you to implement additional features, but the foundation of such thing must be solid and in the scope and spirit of Eclipse platform and a session property seems not fitting very well.

The session property is the result of what was requested by @iloveeclipse here: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3816#discussion_r2981407897

We don't need to disable just one piece of functionality, we must disable multiple. Specifying the name of a session property seems the most simple option.

laeubi: > Here is the description, and what we need

To make it clear, Eclipse Platform is not a collection of things someone "needs" and in the past we have requested to show at least one usage of a new feature in Platform itself or a strong indication that it will be useful for integrator before adding a new feature.

> The session property is the result of what was requested by @iloveeclipse here

The very nature of opensource collaboration is that there are different ideas and different viewpoints and we have all agreed as per the EDP process to be open for different things so it is natural to discuss and has often lead to better designs in the end instead of rushing in the first idea that cam in mind.

> Specifying the name of a session property seems the most simple option.

What is debatable and would require a much more elaborated description, so as it now "spreads" already it makes me thing that the design was not the best initially and could benefit from an architectural planing instead.

akurtakov:
First of all, this project produced stable platform because people working on it cared to really discuss and spend the time to view things from another angle. The fact that people don't feel the need to do that anymore is the "Root of all Evil" (tm) and I'm extremely disappointed by the level of communication lately. No technical reason matters when people don't act with the presumption that everyone involved is acting in the best interest of the project but rather that others are acting just to make it harder. At some point this might even become self full-filling prophecy!

Now to the issue itself - As per https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines#committer-disagreement-resolution I'm asking you to prepare 2 clear statements/proposals, create a poll and ask for general vote on the topic. As long as this lack of good faith communication continues we need to engage the whole community to not allow this problem to spiral down even further.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
trancexpress: > in the past we have requested to show at least one usage of a new feature in Platform

The usage is clearly stated in the linked issue.

> What is debatable and would require a much more elaborated description, so as it now "spreads" already it makes me thing that the design was not the best initially and could benefit from an architectural planing instead.

See:

* https://github.com/eclipse-platform/eclipse.platform/issues/2588
* https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815
* https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4970

Potentially other places too.

One preference to disable all? I believe this was the initial suggestion from @iloveeclipse. IMO it will be hard to describe what this preference actually does.

The need to mark per file is a requirement here. As said earlier, there is no common file naming or pattern we can use.

Its not something the user should be able to configure, this is also a requirement.

> so it is natural to discuss and has often lead to better designs in the end instead of rushing in the first idea that cam in mind

@iloveeclipse please discuss this further, I'm not sure how much of the Advantest requirements can be disclosed here.

From my POV this is a fairly simple change, add a preference to disable a specific behavior in Eclipse.

trancexpress:
> First of all, this project produced stable platform because people working on it cared to really discuss and spend the time to view things from another angle. The fact that people don't feel the need to do that anymore is the "Root of all Evil" (tm) and I'm extremely disappointed by the level of communication lately. No technical reason matters when people don't act with the presumption that everyone involved is acting in the best interest of the project but rather that others are acting just to make it harder. At some point this might even become self full-filling prophecy!

Alright, I'm sorry if I've acted in bad faith.

> Now to the issue itself - As per https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines#committer-disagreement-resolution I'm asking you to prepare 2 clear statements/proposals, create a poll and ask for general vote on the topic. As long as this lack of good faith communication continues we need to engage the whole community to not allow this problem to spiral down even further.

@iloveeclipse, please take this over. I don't know what of the Advantest feature can be disclosed.

trancexpress:
Also, are we reverting the already merged change? https://github.com/eclipse-platform/eclipse.platform.ui/commit/f04b19d4aa6b3c3d8276966baf64eb72de9a56f8

If we are looking for an over-arching design, this is a part of what we need.

I've prepared a PR for this: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3824

akurtakov:
FWIW, Advantest feature is irrelevant here . It's the Eclipse Platform feature that has to be discussed here and what matters to me most is the process of expressing the disagreement in an easy to understand way for every committer of the project. In the process of doing so I hope that reading/listening to each other will be the outcome as each side will have to prepare (at least) one of the options.

trancexpress:
> FWIW, Advantest feature is irrelevant here . It's the Eclipse Platform feature that has to be discussed here

I've signed an NDA, for me its not irrelevant. I'd like to write a full description and explain what is actually needed, but I'm not actually able to.

> and what matters to me most is the process of expressing the disagreement in an easy to understand way for every committer of the project. In the process of doing so I hope that reading/listening to each other will be the outcome as each side will have to prepare (at least) one of the options.

Sorry, I overreacted on the merge-blocking.

akurtakov:
> > FWIW, Advantest feature is irrelevant here . It's the Eclipse Platform feature that has to be discussed here
> 
> I've signed an NDA, for me its not irrelevant. I'd like to write a full description and explain what is actually needed, but I'm not actually able to.

You do not have to share anything about Advantest or under NDA. What I need from both sides is how/when/why/what Eclipse Platform benefits from such a feature implemented in the way the given side proposes.

iloveeclipse:
> I think a session property is not really sufficient.
> 
> Instead it should be stored like we do for encoding already for example:
> 
> ```
> eclipse.preferences.version=1
> encoding/<project>=UTF-8
> encoding//target/maven-archiver=UTF-16
> encoding/testme.txt=ISO-8859-1
> ```
> 
> that way it would be relative easy to even make this available to users on the resource setting page and prevents the history even if no specialized plugin that "knows". Further it will allow to possible have wildcards for folders in a future and so on.

Whatever you have in mind seem to be a completely different feature of what Simeon was working on. Also in your case this feature will need changes in other modules like search too. Persistent resource property, similar to session is just something in low level resources API. If any IDE functionality is going to use that, it will be unavoidable to change various places outside resources plugin.

I believe you misunderstood the proposed changes and so you think about some different use case. So whatever you have in mind needs a clear use case, so far I miss to see it. Let assume there are "sensible data" (passwords? bank account information?) is saved in the project. This is against any security basics, and not something we should "hide" from local history - it should be **encrypted** instead. "Hiding" persisted local files from local history is nonsense - they are already on disk and there is already a way to do so without touching local history at all. Beside this, pointing to that specific file in the project properties is the hint for every "bad" actor: look, there is a sensible information stored, don't miss it!

What you probably want is not hiding files from local history or search, but hiding it completely - and there is already API and UI for that, it is resource filters and configuration dialog to do exactly that:

image

There is no need to reinvent the wheel. If you mean something like this but **encrypted** for sensible information, it would be implemented in a very similar way and it would get a similar UI.

But that is all not what we want & work on.

The use case we have in mind here is different, and doesn't require **any** persistent data or doesn't touch any persisted data on purpose, because it is about sensible data that is not persisted at all on disk, but exposed to user as "virtual" file system. The data is "plugged in" in resource tree on demand and exists no longer as during the current session. This works flawlessly with existing APi's and all infrastructure except few places where the "virtual" resource is **persisted** by IDE (unintentionally) in some way on disk, like local history.

And that is what this change about. Not introducing new UI concepts or features but limiting some existing features to read or persist data that shoulnd't be persisted.

> The very nature of opensource collaboration is that there are different ideas and different viewpoints and we have all agreed as per the EDP process to be open for different things so it is natural to discuss and has often lead to better designs in the end instead of rushing in the first idea that cam in mind.

We've discussed different approaches here, so it is not the first idea:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/3816#discussion_r2981407897

> instead of rushing in the first idea that cam in mind.

This is not appropriate and offending communication style. Please avoid such language.

laeubi:
Thank you @iloveeclipse for the detailed explanation of the actual use case.
I think I now have a much better understanding of what you're solving: virtual files that exist only during a session and must not leave traces on disk.

I accept that persistent preferences (like encoding) are not the right fit for this — the data is session-scoped by nature, so the mechanism should be too.

However, I still have a concern about the **architectural pattern**, not the persistence model.

### The problem I see

Right now the same pattern is being replicated independently across multiple plugins:

- `org.eclipse.core.resources/disable_history_property` → skip local history
- `org.eclipse.search.core/search_exclusion_property` → skip search indexing
- `eclipse.jdt.core` (planned, #4970) → skip JDT indexing

Each plugin reads its own preference, resolves it to a session property name, and checks `resource.getSessionProperty(...)` independently.
This works, but:

- Every new subsystem that needs this must reinvent the same boilerplate.
- Client code must configure N preferences to fully protect a file.
- There is no discoverable API — integrators have to know about each hidden preference.
- If the property name isn't kept in sync across preferences, files get partial protection.

-----

@akurtakov here is my proposal as requested

-----

I'd like to propose the two options for the community vote as follows:

**Option A — Per-plugin preferences (current PR approach):**
Each plugin defines its own preference whose value names a session property.
Subsystems check independently. Client configures N preferences and sets 1 session property.
Minimal change per subsystem, no new API.

**Option B — Centralized session-based mechanism at the resources level:**
A single well-known concept at the `IFile` level (e.g. a method like `isContentRestricted()` backed by a single session property) that all subsystems query.
One property to set, one concept to learn, self-documenting through API.
This is conceptually similar to what @iloveeclipse originally proposed in the [PR #3816 discussion](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3816#discussion_r2981407897) before the per-plugin approach was adopted.

I believe Option B benefits Eclipse Platform more because:
- It provides a clean, discoverable contract for all current and future subsystems.
- It reduces the burden on integrators (set one property instead of configuring N preferences).
- It follows the spirit of existing resource-level concepts like `isDerived()` and `isHidden()`.
- It scales naturally as more subsystems need to respect the flag.
- It would be possible to enhance the concept to persisted resources as well if ever found useful

iloveeclipse:
> I believe Option B benefits Eclipse Platform more

We've discussed with Simeon this before and we haven't heard any other opinions, so it was 1:1 and Simeon's proposal won. Now with your vote we have 1:2 for the Option B, so let follow this path, outlined in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3816#discussion_r2981407897.

In `IFile` interface following would be added (with proper javadoc etc of course):
```
	/**
	 * Session property used to mark a file as containing restricted content
	 * 
	 * @since 3.24
	 */
	QualifiedName RESTRICTED_CONTENT = new QualifiedName(ResourcesPlugin.PI_RESOURCES, "restrictedContent"); //$NON-NLS-1$

	/**
	 * Returns whether the current file is marked as containing restricted
	 * (sensible) content, where some IDE functionality related to the file content
	 * might be limited.
	 * 
	 * @return whether the current file is marked as containing sensible content.
	 *         This flag is not persisted and is false by default.
	 * @since 3.24
	 */
	default boolean isContentRestricted() throws CoreException {
		return getSessionProperty(RESTRICTED_CONTENT) != null;
	}

	/**
	 * Marks the current file as containing restricted (sensible) content. Some IDE
	 * functionality related to the file content might be limited as long as the
	 * file is marked as restricted. This flag is not persisted and is false by
	 * default.
	 * 
	 * @since 3.24
	 */
	default void setContentRestricted(boolean restricted) throws CoreException {
		setSessionProperty(RESTRICTED_CONTENT, Boolean.toString(restricted));
	}
```

I think we can discuss further details of wording etc on the PR. Already merged changes can be easily adopted to use the proposed API.

trancexpress:
@iloveeclipse one of your requirements was having a preference and reading the session property only if the preference is set - to avoid overhead for when the session property is not needed.

Does that stay and if so, what is the name of the preference?

merks:
A general comment.  The word sensible in German does not mean the same thing as word sensible in English where that English word is probably better translated to the German word sinnvol. You should use the word sensitive as the translation in the documentation that I see...

iloveeclipse:
> one of your requirements was having a preference and reading the session property only if the preference is set

Good point. To be precise, the requirement was that the "normal" workspace / IDE work shouldn't be affected / there shouldn't be measurable performance impact for existing applications, and the implied solution was to guard that by a preference that can be set/unset at runtime. For the name: if it would be preference, it would be in the resources bundle, value something like `enableRestrictedContentSupport`. But then the implementation would go into the `File`, the `IFile` would have only default method stubs doing nothing.

Another possibility (no storage based) would be to have new state exposed as API (via `IWorkspace.isRestrictedContentSupportEnabled()` + setter), which would avoid using preference store completely and `IFile` default methods could have full implementation, because iFile has always access to the `getWorkspace()`.
 
> A general comment. The word sensible in German does not mean the same thing as word sensible in English 

Thanks, that all sounds same to me, being native russian :-).

trancexpress:
Considering there are votes in favor of new API, what about adding a new flag, like the derived flag?

1. For the history change here, the code already checks the derived flag - it can fetch all the flags and that won't change performance noticeably.
2. For the text search, `TextSearchVisitor.TextSearchJob.processFile()` will call `Resource.exists()` during `TextSearchRequestor.acceptFile(IFile)` - before the file is processed. Here we would be adding one more flag check on top of this, which is probably not much.
3. I'm not sure about the Java indexer, but there I'm also not sure if we continue with the change (due to evolving requirements).

iloveeclipse:
> what about adding a new flag, like the derived flag?

Such flags are usually persisted in the workspace tree, but we use session property / no need to "touch" workspace.

trancexpress:
I assume the current voting is:

In favor of adding APIs to support https://github.com/eclipse-platform/eclipse.platform/issues/2588:

* @iloveeclipse 
* @laeubi 

In favor of the preference approach:

* @trancexpress 

So I've opened a PR where we can discuss and add the API: https://github.com/eclipse-platform/eclipse.platform/pull/2597

Interested contributors please add yourselves to the review list, I've added only @iloveeclipse and @laeubi.

Once the API is added, we can continue here using it. As well as adjust: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815

trancexpress:
OK, lets continue here.

One question is, do we want to control the history disabling with a preference? Currently the change has this.

If we do want this, what is the default value for the preference?

I assume we want the same for the other places we need to disable?

* https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815
* https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4970

iloveeclipse:
> If we do want this, what is the default value for the preference?

Yes, we want preference, default would be "don't check for restricted content", we don't want affect resources performance by default. Whoever wants checks must enable them explicitly.

trancexpress:
> Yes, we want preference, default would be "don't check for restricted content", we don't want affect resources performance by default. Whoever wants checks must enable them explicitly.

@laeubi please provide feedback on this when you can.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1497,7 +1504,26 @@ public void writeSilently(IProject target) throws CoreException {
...
 	public boolean storeHistory(IResource file) {
 		WorkspaceDescription description = workspace.internalGetDescription();
-		return description.isKeepDerivedState() || !file.isDerived();
+		return (description.isKeepDerivedState() || !file.isDerived()) && !disableHistory(file);
+	}
+
+	private void setDisableHistoryProperty() {
+		String preferenceValue = Platform.getPreferencesService().getString(ResourcesPlugin.PI_RESOURCES,
+				DISABLE_RESTRICTED_FILE_HISTORY_PREFERENCE, "", null); //$NON-NLS-1$
+		disableRestrictedFileHistory = Boolean.TRUE.toString().equals(preferenceValue);
```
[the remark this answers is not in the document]
trancexpress: @iloveeclipse do boolean preferences work with product customization?

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1497,7 +1504,26 @@ public void writeSilently(IProject target) throws CoreException {
...
 	}
 
+	private boolean disableHistory(IResource resource) {
+		if (resource.getType() == IResource.FILE) {
+			if (disableRestrictedFileHistory) {
+				IFile file = (IFile) resource;
+				try {
+					return file.isContentRestricted();
+				} catch (CoreException e) {
+					Policy.log(e.getStatus());
```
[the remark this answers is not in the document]
trancexpress: @iloveeclipse same question as here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4973#discussion_r3020069073

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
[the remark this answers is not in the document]
iloveeclipse: Sure, there is no difference

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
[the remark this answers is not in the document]
iloveeclipse: Same answer, keep on `true` return, for same reason.

iloveeclipse:
@laeubi : please re-review or remove negative review on this PR.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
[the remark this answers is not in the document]
iloveeclipse: I believe current state is not what I've meant with my answer?
Beside returning true on error, can we also slightly change the code to check the main condition first and avoid extra cast:

```
	private boolean disableHistory(IResource resource) {
		if (disableRestrictedFileHistory) {
			if (resource instanceof IFile file) {
				try {
					return file.isContentRestricted();
				} catch (CoreException e) {
					Policy.log(e.getStatus());
					return true;
				}
			}
		}
		return false;
	}
```

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/localstore/DisableHistoryTests.java
```diff
@@ -0,0 +1,198 @@
...
+		setHistoryDisablePreference(true);
+		setRestricted(file);
+		writeContents(file, "edit 1");
+		writeContents(file, "edit 2");
+		IFileState[] history = file.getHistory(null);
+		assertEquals(0, history.length, "Unexpected history after disable: " + toString(history));
+	}
+
+	@Test
+	public void testDisableHistoryAfterEdit() throws Exception {
```
iloveeclipse, line 77: Tests seem to be almost complete, but I miss one test that validates that the history is **working again** after `IFile.setContentRestricted(false);` is executed.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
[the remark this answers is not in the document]
trancexpress: `if (resource instanceof IFile file) {`

I thought `IResource.getType()` is how we are supposed to check if something is a file?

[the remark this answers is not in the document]
merks: > I thought `IResource.getType()` is how we are supposed to check if something is a file?

Yes, I would bet money getType is faster. 😳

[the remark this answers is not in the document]
iloveeclipse: It doesn't really matter in this case, we want cast anyway, it can only be file here.

iloveeclipse:
Dear reviewers, please review. I plan to submit it tomorrow.

iloveeclipse:
> @laeubi : please re-review or remove negative review on this PR.

Last reminder, please re-review. I plan to merge today.

akurtakov:
Cursory read only - nothing obvious to block inclusion from my side.

iloveeclipse:
Test failures are likely caused by recent update of ant (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3741) and are unrelated:

```
Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:5.0.3-SNAPSHOT:test (default-test) on project org.eclipse.ant.tests.ui: There are test failures.
```

image


I've created https://github.com/eclipse-platform/eclipse.platform/issues/2605

iloveeclipse:
> I've created #2605

There are even more ant errors on Windows, but all unrelated to this PR. Merging.

PR #2597, opened 2026-03-30
Add API for restricting operations for sensitive files

trancexpress:
This change adds the following API:

* `org.eclipse.core.resources.IFile.isContentRestricted()`
* `org.eclipse.core.resources.IFile.setContentRestricted(boolean)`

The goal of this API is to allow Eclipse-based IDEs to restrict certain
platform functionality on a per-file basis, for sensitive files.

The flags are not persisted, to enable restricted handling
these flags must be set for each session.

The restrictions will apply to following functionality:

* File search
* Storing file history

Potentially more platform functionality will be restricted.

See: https://github.com/eclipse-platform/eclipse.platform/issues/2588

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
@@ -41,6 +41,7 @@ import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceStatus;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
@@ -60,6 +61,12 @@ import org.eclipse.osgi.util.NLS;
  * The standard implementation of {@link IFile}.
  */
 public class File extends Resource implements IFile {
+
+	/**
+	 * Session property used to mark a file as containing restricted content
+	 */
+	private static final QualifiedName RESTRICTED_CONTENT =
+			new QualifiedName(ResourcesPlugin.PI_RESOURCES, "restrictedContent"); //$NON-NLS-1$
 
 	protected File(IPath path, Workspace container) {
 		super(path, container);
@@ -670,4 +677,14 @@ public class File extends Resource implements IFile {
 		return checkParent ? getProject().getDefaultLineSeparator() : null;
 	}
 
+	@Override
+	public boolean isContentRestricted() throws CoreException {
+		Object sessionProperty = getSessionProperty(RESTRICTED_CONTENT);
+		return Boolean.TRUE.equals(sessionProperty);
+	}
+
+	@Override
+	public void setContentRestricted(boolean restricted) throws CoreException {
+		setSessionProperty(RESTRICTED_CONTENT, restricted ? Boolean.TRUE : null);
+	}
 }
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2000, 2015 IBM Corporation and others.
+ *  Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -21,7 +21,9 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.net.URI;
 import java.util.Objects;
+import org.eclipse.core.internal.resources.ResourceException;
 import org.eclipse.core.internal.utils.FileUtil;
+import org.eclipse.core.internal.utils.Messages;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IPath;
@@ -32,6 +34,7 @@ import org.eclipse.core.runtime.QualifiedName;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.content.IContentDescription;
 import org.eclipse.core.runtime.content.IContentTypeManager;
+import org.eclipse.osgi.util.NLS;
 
 /**
  * Files are leaf resources which contain data.
@@ -1413,5 +1416,38 @@ public interface IFile extends IResource, IEncodedStorage, IAdaptable {
 	 */
 	public default String getLineSeparator(boolean checkParent) throws CoreException {
 		return getProject().getDefaultLineSeparator();
+	}
+
+	/**
+	 * Returns whether the current file is marked as containing restricted
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited. The file must exist.
+	 *
+	 * @return whether the current file is marked as containing sensitive content.
+	 *         This flag is not persisted and is {@code false} by default
+	 * @throws CoreException if the file doesn't exist or if this method fails
+	 * @since 3.24
+	 */
+	default boolean isContentRestricted() throws CoreException {
+		if (!exists()) {
+			String message = NLS.bind(Messages.resources_mustExist, getFullPath());
+			throw new ResourceException(IResourceStatus.RESOURCE_NOT_FOUND, getFullPath(), message, null);
+		}
+		return false;
+	}
+
+	/**
+	 * Marks the current file as containing restricted (sensitive) content. Some IDE
+	 * functionality related to the file content might be limited as long as the
+	 * file is marked as restricted. The file must exist. This flag is not persisted
+	 * and is {@code false} by default.
+	 *
+	 * @param restricted <code>true</code> if the file should be marked restricted,
+	 *                   <code>false</code> if the file should be unmarked
+	 * @throws CoreException if the file doesn't exist or if this method fails
+	 * @since 3.24
+	 */
+	default void setContentRestricted(boolean restricted) throws CoreException {
+		throw new CoreException(Status.error("Not implemented")); //$NON-NLS-1$
 	}
 }
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.core.tests.internal.resources;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.IWorkspaceRoot;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.tests.resources.util.WorkspaceResetExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(WorkspaceResetExtension.class)
+public class RestrictedFileTests {
+
+	@Test
+	public void testRestrictedFile() throws Exception {
+		IWorkspace workspace = ResourcesPlugin.getWorkspace();
+		IWorkspaceRoot root = workspace.getRoot();
+		IProject project = root.getProject(RestrictedFileTests.class.getSimpleName());
+		try {
+			project.create(null);
+			project.open(null);
+
+			IFile file1 = project.getFile("test1.txt");
+			IFile file2 = project.getFile("test2.txt");
+
+			try {
+				file1.isContentRestricted();
+				fail("Should not work on not existing files");
+			} catch (CoreException e) {
+				// expected, file should not exist
+			}
+
+			file1.create("line 1".getBytes(), IResource.FORCE, null);
+			file2.create("line 1".getBytes(), IResource.FORCE, null);
+
+			assertFalse(file1.isContentRestricted(), "Expected file to not be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+
+			file1.setContentRestricted(true);
+			assertTrue(file1.isContentRestricted(), "Expected file to be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+
+			file1 = project.getFile("test1.txt");
+			file2 = project.getFile("test2.txt");
+			assertTrue(file1.isContentRestricted(), "Expected file to be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+		} finally {
+			project.delete(true, null);
+		}
+	}
+}
--- a/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
+++ b/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.23.300.qualifier
+Bundle-Version: 3.24.0.qualifier
 Bundle-Activator: org.eclipse.core.resources.ResourcesPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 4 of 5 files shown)


trancexpress:
@merks if you can help with the API documentation wording, that would be great!

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/WorkspaceDescription.java
```diff
@@ -227,4 +230,14 @@ public boolean isKeepDerivedState() {
 	public void setKeepDerivedState(boolean keepDerivedState) {
 		this.keepDerivedState = keepDerivedState;
```
trancexpress, line 228: Essentially we mimic this, but:

* we don't want to persist the new file state/flag, so we use a session property
* we don't want the user to be able to turn off the handling of restricted files, so no preference with UI

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1422,30 @@ public default String readString() throws CoreException {
...
+	 * Returns whether the current file is marked as containing restricted
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited.
+	 *
+	 * @return whether the current file is marked as containing sensitive content.
+	 *         This flag is not persisted and is {@code false} by default.
+	 * @since 3.24
+	 */
+	default boolean isContentRestricted() throws CoreException {
+		return getWorkspace().getDescription().isRestrictedContentEnabled()
```
iloveeclipse: I see two issue with that:
- `getWorkspace()` can be null (tests or special code that needs to mock/wrap IFile interface)
- `getWorkspace().getDescription()` will create a copy of the ws description every time. This would likely affect performance if feature is switched off.

To solve that, null check is needed and similar to `IWorkspace.isAutoBuilding()` implementation, the flag should be part of the `IWorkspace` API

merks:
> @merks if you can help with the API documentation wording, that would be great!

FYI. I’m on vacation mostly monitoring from my iPhone with slow internet.  Ping me when you think it’s done and I can look at it on my computer later.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/WorkspaceDescription.java
```diff
@@ -227,4 +227,5 @@ public boolean isKeepDerivedState() {
...
 	}
+
```
iloveeclipse: This new line can be reverted now.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/Workspace.java
```diff
@@ -271,6 +271,8 @@ public class Workspace extends PlatformObject implements IWorkspace, ICoreConsta
 	 */
 	protected IFileModificationValidator validator = null;
 
+	private boolean enableRestrictedContent;
```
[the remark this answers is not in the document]
trancexpress: @iloveeclipse are we expecting multi-threading for the flag?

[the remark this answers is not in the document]
iloveeclipse: Well, you never know in which thread the code was running setting this flag and in which thread the other code will run that needs to see this flag, so yes, I would say we expect it.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
```diff
@@ -0,0 +1,58 @@
...
+			project.open(null);
+
+			IFile file = project.getFile("test.txt");
+			file.create("line 1".getBytes(), IResource.FORCE, null);
+
+			assertFalse(file.isContentRestricted(), "Expected file to not be restricted");
+			file.setContentRestricted(true);
+			assertTrue(file.isContentRestricted(), "Expected file to be restricted");
+			workspace.setRestrictedContentEnabled(false);
+			assertFalse(file.isContentRestricted(), "Expected file to not be restricted");
```
iloveeclipse: I miss negative test for file that was not created yet or for the file that was never "touched".

```
diff --git a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
index 34ce7ea..d41b5a6 100644
--- a/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
+++ b/resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/resources/RestrictedFileTests.java
@@ -17,2 +17,3 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
@@ -24,2 +25,3 @@
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.tests.resources.util.WorkspaceResetExtension;
@@ -44,10 +46,27 @@
 
-			IFile file = project.getFile("test.txt");
-			file.create("line 1".getBytes(), IResource.FORCE, null);
+			IFile file1 = project.getFile("test1.txt");
+			IFile file2 = project.getFile("test2.txt");
 
-			assertFalse(file.isContentRestricted(), "Expected file to not be restricted");
-			file.setContentRestricted(true);
-			assertTrue(file.isContentRestricted(), "Expected file to be restricted");
+			try {
+				file1.isContentRestricted();
+				fail("Should not work on not existing files");
+			} catch (CoreException e) {
+				// expected, file should not exist
+			}
+
+			file1.create("line 1".getBytes(), IResource.FORCE, null);
+			file2.create("line 1".getBytes(), IResource.FORCE, null);
+
+			assertFalse(file1.isContentRestricted(), "Expected file to not be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+
+			file1.setContentRestricted(true);
+			assertTrue(file1.isContentRestricted(), "Expected file to be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+
 			workspace.setRestrictedContentEnabled(false);
-			assertFalse(file.isContentRestricted(), "Expected file to not be restricted");
+
+			assertFalse(file1.isContentRestricted(), "Expected file to not be restricted");
+			assertFalse(file2.isContentRestricted(), "Expected file to not be restricted");
+
 		} finally {
```

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1422,36 @@ public default String readString() throws CoreException {
 	public default String getLineSeparator(boolean checkParent) throws CoreException {
 		return getProject().getDefaultLineSeparator();
 	}
+
+	/**
+	 * Returns whether the current file is marked as containing restricted
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited.
```
iloveeclipse: After looking at the test, we should add a note that this API only works on **existing** files and will throw an exception on non existing files.

iloveeclipse:
From my POV this PR is ready to merge, so 

* @laeubi : please check the proposed API.
* @merks :  if you could look at wording, would be great.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IWorkspace.java
```diff
@@ -1861,4 +1861,37 @@ public default void write(Map<IFile, byte[]> contentMap, boolean force, boolean
...
+	 * Restricted files are marked with: {@link IFile#setContentRestricted(boolean)}
+	 * </p>
+	 *
+	 * @return <code>true</code> if restricted (sensitive) files handling is enabled
+	 *         by the IDE, <code>false</code> otherwise
+	 * @see IFile#setContentRestricted(boolean)
+	 * @see IFile#isContentRestricted()
+	 * @since 3.24
+	 */
+	boolean isRestrictedContentEnabled();
```
laeubi: Do we really need this API? What would be the use-case for disabling this? I would more see this on an plugin level or even component level.

e.g. the classic search dialog has a setting to search in derived files, and it could have a similar for restricted files.

Also JDT might have a preference weather or not restricted files should be indexed or not.

Having a way to enable/disable it completely for everything seems too broard and at this stage bloat the API. Also it makes the definition of `isRestricted()` more complicated as it not only depends on the state.

Further, as likely all resources today return `false` anyways I don't see any backward compatibility issues here that would require this.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -53,6 +53,14 @@
  * @noextend This interface is not intended to be extended by clients.
  */
 public interface IFile extends IResource, IEncodedStorage, IAdaptable {
+
+	/**
+	 * Session property used to mark a file as containing restricted content
+	 *
+	 * @since 3.24
+	 */
+	QualifiedName RESTRICTED_CONTENT = new QualifiedName(ResourcesPlugin.PI_RESOURCES, "restrictedContent"); //$NON-NLS-1$
```
laeubi: Is it needed to have it public API? Seems like an implementation detail that is better moved to `File`

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1422,38 @@ public default String readString() throws CoreException {
...
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited. The file must exist.
+	 *
+	 * @return whether the current file is marked as containing sensitive content.
+	 *         This flag is not persisted and is {@code false} by default.
+	 * @throws CoreException if the file doesn't exist or if its session properties
+	 *                       cannot be read
+	 * @since 3.24
+	 */
+	default boolean isContentRestricted() throws CoreException {
```
laeubi, line 1431: this does not seems a good candidate for a default method, as the interface is marked as noextend and noimplement better move it to the implementation (see previous comment) and make this a plain interface method.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1422,38 @@ public default String readString() throws CoreException {
...
+	 * by default. The file must exist.
+	 *
+	 * @param restricted <code>true</code> if the file should be a marked
+	 *                   restricted, <code>false</code> if the file should be
+	 *                   unmarked
+	 * @throws CoreException if the file doesn't exist or if its session properties
+	 *                       cannot be read
+	 * @since 3.24
+	 */
+	default void setContentRestricted(boolean restricted) throws CoreException {
```
laeubi, line 1450: same as before, implementation should be moved to `File`

laeubi (commented):
See comments below, I think the API exposure can be reduced a bit.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IWorkspace.java
trancexpress: The concern from @iloveeclipse is that we are adding extra cost to the functionality we want to disable, by checking for the session property. And the expectation is that the session property will be set in very few cases, but we would be checking it every time we want to add history for a file, or when we want to search in a file.

From my POV this can be a preference too, though again no UIs for it, as the user should not be able to disable it.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
trancexpress: @iloveeclipse ? The proposal is yours here.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IWorkspace.java
trancexpress: > I would more see this on an plugin level or even component leve

> Also JDT might have a preference weather or not restricted files should be indexed or not.

@laeubi is your proposal to add preferences for each functionality we want to disable? Or a preference per plug-in? I think this would mean, one preference for `org.eclipse.core.resources` and one preference for `org.eclipse.jdt.core`?

laeubi: > The concern from @iloveeclipse is that we are adding extra cost to the functionality we want to disable, by checking for the session property.

Checking a session property should be really fast and I would assume this function is not called quite frequently enough that it has any impact. As we have getter/setter here the implementation can even cache the value in a field (no session property needed at all).

At least we should not do any premature optimization here.

For the same reason a preference does not really looks suitable, if code relies on this check we should not allow to disable it and we have nothing similar for `isHidden()` or `isDerived()`

> @laeubi is your proposal to add preferences for each functionality we want to disable? Or a preference per plug-in? I think this would mean, one preference for org.eclipse.core.resources and one preference for org.eclipse.jdt.core?

I'm not directly proposing something, but if it is seen useful for users to control then it could be done better on this level than on workspace.

Just think about I want to find something ANYWHERE derived, binary, secret, I don't care and then can tick the option in the search dialog. Still I might not want these things to be indexed or whatever else.

trancexpress: OK, lets discuss on the following PRs if there should be preferences or no.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
iloveeclipse: > this does not seems a good candidate for a default method, as the interface is marked as noextend and noimplement 

@laeubi : in the ideal world I would agree. However, the `IFile` interface is very old (one of the oldest in Eclipse I guess), and the chances are high that there are multiple "illegal" implementations flying around for different reasons, mostly for tests, but not only.

There is at least one use in the SDK, see `org.eclipse.e4.tools.emf.ui.internal.common.component.dialogs.ContributionDataFile`.

Therefore I would propose to provide default methods that would return false / do nothing by default.

laeubi: > Therefore I would propose to provide default methods that would return false / do nothing by default.

This sounds reasonable for me, maybe one would better throw `NotSupportedOperation` Exception for the setter at least.

iloveeclipse:
@laeubi : anything else? If no, let's merge that.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -61,6 +61,8 @@
  */
 public class File extends Resource implements IFile {
 
+	private volatile boolean isRestricted;
+
```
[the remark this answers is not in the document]
iloveeclipse: Arrgh. This is a very good point.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1414,34 @@ public default String readString() throws CoreException {
...
+	 * and is {@code false} by default.
+	 *
+	 * @param restricted <code>true</code> if the file should be a marked
+	 *                   restricted, <code>false</code> if the file should be
+	 *                   unmarked
+	 * @throws CoreException if the file doesn't exist or if this method fails
+	 * @since 3.24
+	 */
+	default void setContentRestricted(boolean restricted) throws CoreException {
+		throw new UnsupportedOperationException();
```
laeubi: Thinking more about it it likely is better to throw `CoreException` her as it is already documented 

> if the file doesn't exist **or if this method fails**

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1414,34 @@ public default String readString() throws CoreException {
...
+	 * Returns whether the current file is marked as containing restricted
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited. The file must exist.
+	 *
+	 * @return whether the current file is marked as containing sensitive content.
+	 *         This flag is not persisted and is {@code false} by default
+	 * @throws CoreException if the file doesn't exist or if this method fails
+	 * @since 3.24
+	 */
+	default boolean isContentRestricted() throws CoreException {
```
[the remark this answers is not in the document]
trancexpress: Makes me wonder if we should throw "not implemented" here too...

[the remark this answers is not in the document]
laeubi: I think returning false is fine and adding the suggested additional check would not harm.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/File.java
```diff
@@ -670,4 +677,14 @@ public String getLineSeparator(boolean checkParent) throws CoreException {
 		return checkParent ? getProject().getDefaultLineSeparator() : null;
 	}
 
+	@Override
+	public boolean isContentRestricted() throws CoreException {
+		Object sessionProperty = getSessionProperty(RESTRICTED_CONTENT);
+		return sessionProperty != null;
```
laeubi:
```suggestion
		if(getSessionProperty(RESTRICTED_CONTENT) instanceof Boolean value) {
		   return value;
		}
		return false;
```

that would address the concern from the copilot bot raised earlier and would not hurt much...

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1417,37 @@ public default String readString() throws CoreException {
...
+		if (!exists()) {
+			String message = NLS.bind(Messages.resources_mustExist, getFullPath());
+			throw new ResourceException(IResourceStatus.RESOURCE_NOT_FOUND, getFullPath(), message, null);
+		}
+		return false;
+	}
```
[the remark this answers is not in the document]
iloveeclipse: No reason to do so.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/resources/IFile.java
```diff
@@ -1414,4 +1417,37 @@ public default String readString() throws CoreException {
 	public default String getLineSeparator(boolean checkParent) throws CoreException {
 		return getProject().getDefaultLineSeparator();
 	}
+
+	/**
+	 * Returns whether the current file is marked as containing restricted
+	 * (sensitive) content, where some IDE functionality related to the file content
+	 * might be limited. The file must exist.
```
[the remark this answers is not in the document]
iloveeclipse: PR's will follow.

iloveeclipse:
Thanks all (including Copilot :-)), merging.

Issue #2605, opened 2026-04-02
Various Ant tests fail after ant update


iloveeclipse:
See https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-2589/15/, most likely caused by ant update in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3741.

```
Expected size: 1 but was: 2 in:
[mklink, mkdir]
	at org.eclipse.ant.tests.ui.editor.CodeCompletionTest.testTaskProposals(CodeCompletionTest.java:696)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.ant.tests.ui.testplugin.TestAgainExtension.lambda$0(TestAgainExtension.java:37)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at org.eclipse.ant.tests.ui.testplugin.TestAgainExtension.handleTestExecutionException(TestAgainExtension.java:34)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

merks:
@iloveeclipse 

I'm having a look at this.  It does look like just a new `mklink` task is available in the latest version of ant.   But I need to resolve a newer target platform and that takes FOREVER with my bad internet.  Chugging away now...

I think the fix is probably just to expect that new result:
```
		// "<project><target><mk"
		proposals = processor.getTaskProposals("<project><target><mk", "target", "mk"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
		assertThat(proposals).hasSize(2);
		proposal = proposals[0];
		assertEquals("mklink", proposal.getDisplayString()); //$NON-NLS-1$
		proposal = proposals[1];
		assertEquals("mkdir", proposal.getDisplayString()); //$NON-NLS-1$
		processor.dispose();
```

But of course I want to verify it fixes the test failure before I create a PR...

akurtakov:
Looking at the changes at https://downloads.apache.org/ant/ there is:
" * added a Windows specific <mklink> task that can be used to create  hard links, symbolic links and junctions."

It is probably best to make the request for task proposals more concrete - e.g. `mkd` instead of `mk`, that should allow everything else to stay the same.

iloveeclipse:
[...]

PR #2606, opened 2026-04-02
Make Ant completion test more detailed

akurtakov:
Test for mkdir is to start with `mkd` before requesting complete instead of just `mk` to prevent finding 1.10.16 added mklink tag. That way tests will continue to work proper on both older and newer And versions.


akurtakov:
Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/2605

akurtakov:
Merge despite windows test failures as this change makes the test less fragile to changes in Ant .

PR #2608, opened 2026-04-03
Ensure o.a.tools.ant.util FileUtils is initialized before JavaEnvUtils

merks:
https://github.com/eclipse-platform/eclipse.platform/issues/2605


akurtakov:
I still believe we can replace this line with with 
```java
return Integer.parseInt(System.getProperty("java.specification.version")) < 18;
```
and not need this workaround at all in these cases.

akurtakov:
Let's merge without extra changes and have further cleanups later on.

PR #2610, opened 2026-04-03
Simplify intro framework: deprecate StandbyPart and IntroLaunchBar

### LINKED ISSUE #2598: Welcome screen should be handled like a regular editor
IMHO currently the welcome screen is useless to me and the users I know of.

It blocks the full screen with information which does not feel important. I'm not aware that Eclipse IDE users interact with this view other than closing it to remove the "block" of the IDE.

Image

Also if I minimize it adds to the toolbar in a way I find not adding value to my IDE experience. 

Image

I also do not know how to get back to the main Welcome screen once I clicked on of the buttons.

Image

Vscode handles its "News" as a regular editor which I find much more engaging as it does not block the IDE or clutters the toolbar with new items.

Therefore, I suggest we also open the Welcome screen in the editor area. This would allow the user to use it without blocking the whole IDE if it is open.

cc @merks @HeikoKlare WDYT?

vogella:
## Summary

Treat the welcome screen like a regular editor by removing the special standby mode and launch bar handling from `CustomizableIntroPart`.

## Changes

- **Simplify `CustomizableIntroPart`**: Remove `StackLayout`, `StandbyPart` handling, and standby memento logic. Use `FillLayout` to show the presentation directly.
- **Deprecate `IntroLaunchBar`** (`forRemoval=true`): The toolbar launch bar is no longer created when the intro is minimized.
- **Deprecate `StandbyPart`** (`forRemoval=true`): Standby content parts are no longer used by the intro framework.
- **Clean up `IntroURL`**: Deprecate `SET_STANDBY_MODE`, `SHOW_STANDBY`, and `SWITCH_TO_LAUNCH_BAR` actions. Remove `switchToLaunchBar()` and `handleStandbyState()` methods; deprecated actions now log warnings and fall back to showing/closing the intro.
- **Deprecate `closeLaunchBar()` and `setLaunchBar()`** in `IntroPlugin`.
- **Remove `<launchBar>` element** from universal intro `plugin.xml`.
- **Bump `org.eclipse.ui.intro` version** to 3.9.0.

## Testing

- `org.eclipse.ui.intro` compiles with 0 errors and no new warnings.

Fixes #2598

### DIFF
--- a/ua/org.eclipse.ui.intro.universal/plugin.xml
+++ b/ua/org.eclipse.ui.intro.universal/plugin.xml
@@ -54,12 +54,7 @@
             <implementation
                   kind="swt">
             </implementation>
-            <launchBar
-               	location="fastview"
-            	bg="$launchbarBackground$"
-            	computed="true">
-            </launchBar>
-         </presentation>
+            </presentation>
       </config>
    </extension>
    <extension
--- a/ua/org.eclipse.ui.intro/META-INF/MANIFEST.MF
+++ b/ua/org.eclipse.ui.intro/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %plugin_name
 Bundle-SymbolicName: org.eclipse.ui.intro; singleton:=true
-Bundle-Version: 3.8.100.qualifier
+Bundle-Version: 3.9.0.qualifier
 Bundle-Activator: org.eclipse.ui.internal.intro.impl.IntroPlugin
 Bundle-Vendor: %provider_name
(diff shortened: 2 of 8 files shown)


merks:
Was there an issue in which I expressed my reservation about the welcome page changes you proposed?  Is this part of that?

vogella:
> Was there an issue in which I expressed my reservation about the welcome page changes you proposed?  Is this part of that?

Sorry, not sure what you are asking? This is draft pr to see if my suggestion is feasible. If concerns are continued to remain raised and not resolved this change will be dropped.

merks:
As I mentioned in the issue just now, it's really a challenge to keep up with the floods of emails.  I can only hope this is determined not to be feasible and is dropped.

PR #2609, opened 2026-04-03
Drop Ant 1.5 and 1.6 support

akurtakov:
Drop Ant 1.5/1.6 support in InternalAntRunner classes. Last Ant 1.6 release is from 2006. It is high time to simplify the codebase a bit.

PR #2612, opened 2026-04-05
Ensure remote.jar is updated the fix for JavaEnvUtils

merks:
https://github.com/eclipse-platform/eclipse.platform/issues/2605


merks:
The jar was updated by. 

https://github.com/eclipse-platform/eclipse.platform/pull/2609

PR #2614, opened 2026-04-07
fix: Draw terminal characters individually to prevent clipping with f…

astrograph:
…ractional DPI scaling

On Windows with fractional DPI scaling (125%, 150%), SWT/GDI accumulates sub-pixel rounding when drawing a whole string, causing glyphs to overflow their grid cells. Draw each character at its exact cell position instead.

### DIFF
--- a/instructions.md
+++ b/instructions.md
@@ -1,0 +1,96 @@
+# Fix: Terminal character clipping on Windows after Eclipse 2024-06 update
+
+## Root cause
+
+Eclipse 2024-06 (SWT 4.32) switched the Windows text rendering backend from
+**GDI to Direct2D (D2D)** by default. D2D uses sub-pixel precision for font
+glyph positioning:
+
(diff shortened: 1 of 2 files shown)


tomaswolf:
Is this related to #2617? Does drawing characters individually have any (negative?) effect on performance?

HeikoKlare:
Please add a proper description of the issue and how to reproduce it. To me this looks like a solution to an already solved issue, see: https://github.com/eclipse-platform/eclipse.platform/issues/2617#issuecomment-4211986789.

HeikoKlare:
@astrograph can you please add information on reproduction or the issue and check if the issue is already resolved in the latest development state?

astrograph:
thank you for the feedback, I will reassess after we consumed the current release!

merks:
> thank you for the feedback, I will reassess after we consumed the current release!

Testing before a release is a much better idea than testing after a release.

PR #2629, opened 2026-04-22
Add initial LLM model support

MatthewKhouzam:
This creates a way to have a default llm provider so plugins can make simple inference.

The intent is to have a way for all llms to work in a extensions that need inference. E.g. "explain this anomaly" 

This was assisted by Claude Opus 4.7

### DIFF
--- a/runtime/bundles/org.eclipse.core.llm/.classpath
+++ b/runtime/bundles/org.eclipse.core.llm/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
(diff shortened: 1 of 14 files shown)


MatthewKhouzam:
I want to use this to have a one place to query for events for trace compass. It could be in the wrong place, but I strongly doubt anyone can have a wider need for AI than this,

vogella:
@MatthewKhouzam I don't understand what your target is here. 

I recently developed an MCP server for Eclipse so that LLMs can request informations from the Eclipse IDE. For example, my MCP server provides the problems in the workspace, can trigger builds and check references in Eclipse using the JDT AST.

Can you explain your use case here? Your PR and commit message are unfortunately to short to understand what you mean.

MatthewKhouzam:
Hi, not an issue, I was just added support for eclipse to natively query llms via the OpenAI API, this is so that plug-ins don't need to re-implement the wheel everytime. It was a patch I made at OCX to get a discussion/ball rolling.

The way I see it, eclipse would be the place you configure which model you want to use, then the plug-ins would just connect directly to the shared infrastructure. I made this to help out the Eclipse IDE Bof, it is not something I needed personally.

Your MCP is the opposite, this was so Eclipse can call MCPs without using a third party... eventually.

vogella:
I re-open so that you could continue if you want

PR #2631, opened 2026-04-25
Retire outdated Migrate Welcome page

vogella:
The "Migrate" page on the Eclipse Welcome screen has not been updated since 4.29 and no user has complained, so this removes it together with its CSS, properties, icon, and intro data wiring. The actual porting guides remain available through the platform documentation and the help system, which is the more durable place to surface them.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2630


vogella:
Releng team FYI, since a ~week all verifcation builds fail in eclipse-platform cc @HannesWell cc @merks

iloveeclipse:
For the problem, see the solution in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5016.
This must be applied to all affected bindles like `org.eclipse.core.tests.resources.saveparticipant1/2/3`

HannesWell:
> Releng team FYI, since a ~week all verifcation builds fail in eclipse-platform

Fixed via
- https://github.com/eclipse-platform/eclipse.platform/pull/2632

Was harder to spot with the inconsistent state back then. Sorry for the inconvenience.

vogella:
Good catch — applied the same removal to `platform/org.eclipse.sdk/plugin_customization.ini` (drop `migrate` from `INTRO_ROOT_PAGES`) and `platform/org.eclipse.sdk/introData.xml` (drop the `org.eclipse.ui.workbench.migration` extension reference and the `migrate` page) so the SDK welcome no longer points at the now-removed wiring.

PR #2635, opened 2026-04-28
Remove unwanted single quote from links

sweettypdevassy:
This commit removes unwanted single quotes from label links in the Ant editor’s Preference pages and makes links across Preference pages consistent.
Before:
image
After:
image


elsazac (approved):
LGTM

PR #2646, opened 2026-05-03
Fix progress overshoot >100% in Synchronize refresh jobs

### LINKED ISSUE #2645: RefreshSubscriberParticipantJob.doRefresh reports >100% progress (monitor passed to two sequential consumers)
## Summary

`RefreshSubscriberParticipantJob.doRefresh` reports >100% progress in the Synchronize view because the same `IProgressMonitor` is passed to two sequential consumers (`subscriber.refresh` and `getCollector().waitForCollector`) without splitting.

## Environment

- Eclipse 2026-03 (4.39) on macOS arm64, JDK 21
- Affects every Team Synchronize provider that goes through `RefreshSubscriberParticipantJob`. Repro is via Subversive in my workspace; Git/CVS/Mercurial subscribers share the same code path.

## Symptoms

In the Eclipse status bar / Progress view, the `Synchronizing <provider>` job consistently overshoots 100%. Observed values: 115–160% on a typical workspace, scaling roughly with the number of resource mappings included in the sync.

The functional outcome of the sync is correct; this is a cosmetic but persistent and confusing display issue. Triggered every time the Synchronize view is refreshed.

## Root cause

In `team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/RefreshSubscriberParticipantJob.java`:

```java
protected void doRefresh(...) {
    subscriber.refresh(resources, IResource.DEPTH_INFINITE, monitor);
    getCollector().waitForCollector(monitor);
}
```

Both `subscriber.refresh(...)` and `getCollector().waitForCollector(...)` receive the **same** `IProgressMonitor` instance and each treats it as their own.

`RefreshParticipantJob.initialize` sets up the parent group monitor with `group.beginTask(taskName, 100)` plus `setProgressGroup(group, 80)` for the job itself and `handleProgressGroupSet(group, 20)` delegated elsewhere. The outer 80/20 budget is therefore over-consumed: the subscriber typically uses up its full slice, and `waitForCollector` then reports additional ticks against the same monitor.

I instrumented this end-to-end against a fork of Subversive: the subscriber consumes exactly its allocated budget (a wrapped logging monitor reports `total=100.0` at `done`), yet the JobManager group still shows >100% — confirming the extra ticks come from the `waitForCollector` call after `subscriber.refresh` returns.

## Suggested fix

Split the monitor between the two consumers, e.g.:

```java
protected void doRefresh(...) {
    SubMonitor sub = SubMonitor.convert(monitor, 100);
    subscriber.refresh(resources, IResource.DEPTH_INFINITE, sub.split(80));
    getCollector().waitForCollector(sub.split(20));
}
```

The 80/20 split mirrors the existing `setProgressGroup(group, 80)` / `handleProgressGroupSet(group, 20)` allocation in `RefreshParticipantJob.initialize`. Any reasonable proportion would work — the key change is that each consumer gets its own bounded child monitor instead of sharing the parent.

## Why filed here, not against the team provider

I originally hit this through Subversive's `Synchronizing SVN` job and instrumented the subscriber-side progress reporting first. Subversive consumes exactly its allocated budget. The over-reporting comes from this base class in `eclipse.platform.team`, so the fix has to live here. Same pattern affects every team provider (Git, CVS, Mercurial) since they all flow through `RefreshSubscriberParticipantJob.doRefresh`.

Happy to send a PR with the fix if welcome.

FoxyBOA:
## Summary

Fixes #2645.

`RefreshSubscriberParticipantJob.doRefresh` (and the sibling `RefreshModelParticipantJob.doRefresh`) were handing the same `IProgressMonitor` instance to two sequential consumers — for the Subscriber job: `subscriber.refresh(...)` then `SubscriberSyncInfoCollector.waitForCollector(...)`; for the Model job: `context.refresh(...)` then `Job.getJobManager().join(...)`. Each consumer treated the monitor as exclusively its own and reported its full budget against the parent, which is how the JobManager group ends up displaying 115–160 % for Synchronize jobs (originally observed with the Subversive subscriber, but reproducible with any team provider).

Both methods now wrap the incoming monitor with `SubMonitor.convert(monitor, 100)` and split 80/20 between the two consumers, mirroring the existing 80/20 split already in use in `RefreshParticipantJob.initialize` (`setProgressGroup(group, 80)` + `handleProgressGroupSet(group, 20)`).

## Changes

- `RefreshSubscriberParticipantJob.doRefresh`: split monitor 80/20.
- `RefreshSubscriberParticipantJob`: new `protected void waitForCollector(IProgressMonitor)` seam delegating to `getCollector().waitForCollector(...)`. Production code shape is unchanged for clients (the seam is internal-only, x-friends).
- `RefreshModelParticipantJob.doRefresh`: same split applied to the sibling code path (same antipattern).
- New regression test `RefreshSubscriberParticipantJobProgressTests` in `org.eclipse.team.tests.core`, wired into `AllTeamTests`. It overrides the new seam in a `TestableJob` to imitate a greedy collector.

## Test plan

- The new test reports `parent.total = 2000` against the unfixed code and `parent.total ≈ 1000` with the fix; threshold is `≤ 1500`.
- All other tests in `AllTeamTests` still pass.
- Verified locally with `mvn -pl team/bundles/org.eclipse.team.core,team/bundles/org.eclipse.team.ui,team/tests/org.eclipse.team.tests.core,team/tests/org.eclipse.compare.tests,runtime/tests/org.eclipse.core.tests.harness,resources/tests/org.eclipse.core.tests.resources,resources/tests/org.eclipse.core.tests.filesystem.feature clean install -Dtest=AllTeamTests -DfailIfNoTests=false` → BUILD SUCCESS, 22 tests, 0 failures.

## Notes for review

- The 80/20 split is intentional, not arbitrary: it matches the existing budget split in `RefreshParticipantJob.initialize`. Empirically `subscriber.refresh` does the bulk of the wall time and `waitForCollector` the tail.
- Reported and diagnosed originally via the Subversive client, but the antipattern (and the fix) are entirely in `o.e.team.ui` — no SCM-provider-specific code involved.

### DIFF
--- a/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/RefreshModelParticipantJob.java
+++ b/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/RefreshModelParticipantJob.java
@@ -22,6 +22,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.team.core.diff.IDiff;
 import org.eclipse.team.core.diff.IDiffChangeEvent;
@@ -77,11 +78,12 @@ public class RefreshModelParticipantJob extends RefreshParticipantJob {
 		ISynchronizationContext context = ((ModelSynchronizeParticipant)getParticipant()).getContext();
 		try {
 			context.getDiffTree().addDiffChangeListener((ChangeDescription)changeListener);
+			SubMonitor sub = SubMonitor.convert(monitor, 100);
 			// TODO: finer grained refresh
-			context.refresh(mappings, monitor);
+			context.refresh(mappings, sub.split(80));
 			// Wait for any asynchronous updating to complete
 			try {
-				Job.getJobManager().join(context, monitor);
+				Job.getJobManager().join(context, sub.split(20));
 			} catch (InterruptedException e) {
 				// Ignore
 			}
--- a/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/RefreshSubscriberParticipantJob.java
+++ b/team/bundles/org.eclipse.team.ui/src/org/eclipse/team/internal/ui/synchronize/RefreshSubscriberParticipantJob.java
@@ -16,6 +16,7 @@ package org.eclipse.team.internal.ui.synchronize;
 
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.team.core.TeamException;
 import org.eclipse.team.core.subscribers.Subscriber;
 import org.eclipse.team.core.synchronize.SyncInfo;
@@ -119,11 +120,16 @@ public class RefreshSubscriberParticipantJob extends RefreshParticipantJob {
 		if (subscriber != null) {
 			try {
 				subscriber.addListener((RefreshChangeListener)changeListener);
-				subscriber.refresh(resources, IResource.DEPTH_INFINITE, monitor);
-				getCollector().waitForCollector(monitor);
+				SubMonitor sub = SubMonitor.convert(monitor, 100);
+				subscriber.refresh(resources, IResource.DEPTH_INFINITE, sub.split(80));
+				waitForCollector(sub.split(20));
 			} finally {
 				subscriber.removeListener((RefreshChangeListener)changeListener);
 			}
 		}
+	}
+
+	protected void waitForCollector(IProgressMonitor monitor) {
+		getCollector().waitForCollector(monitor);
 	}
 }
--- a/team/tests/org.eclipse.team.tests.core/src/org/eclipse/team/internal/ui/synchronize/RefreshSubscriberParticipantJobProgressTests.java
+++ b/team/tests/org.eclipse.team.tests.core/src/org/eclipse/team/internal/ui/synchronize/RefreshSubscriberParticipantJobProgressTests.java
@@ -1,0 +1,113 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Foxy BOA and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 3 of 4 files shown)


FoxyBOA:
Reverted the Bundle-Version bumps in the latest force-push. API Tools flagged both as MINOR_VERSION_CHANGE_NO_NEW_API on build #1, which is correct: the only production changes are inside internal packages (`o.e.team.internal.core.subscribers` and `o.e.team.internal.ui.synchronize`), so no public API was added or modified and no version increment is required. Build #2 is rerunning now.

vogella:
Please verify on your side that the mvn build works (see Agents.md for instructions). 

The Javadoc on the test class should be shorter. 

Rather than dropping final from                           
  SubscriberSyncInfoCollector just to enable the test, could you introduce a
   small protected seam on the job itself, e.g.
````                                                                            
  protected void waitForCollector(IProgressMonitor monitor) {
      getCollector().waitForCollector(monitor);
  }
````
  and override that in TestableJob? That keeps the production class shape   
  unchanged (it has been final since 3.0) and avoids any API-tools noise on
  an x-friends type.

merks:
> Reverted the Bundle-Version bumps in the latest force-push. API Tools flagged both as MINOR_VERSION_CHANGE_NO_NEW_API on build #1, which is correct: the only production changes are inside internal packages (`o.e.team.internal.core.subscribers` and `o.e.team.internal.ui.synchronize`), so no public API was added or modified and no version increment is required. Build #2 is rerunning now.

Note that any change to the binary contents of a bundle requires minimally a minor (service) version increment which is automatically generated.  So the version increment is correct.  You can force push changes and that commit will always be recreated...

FoxyBOA:
@vogella thanks for the review. Force-pushed an updated commit:

- **Seam over `final` removal**: introduced `protected void waitForCollector(IProgressMonitor)` in `RefreshSubscriberParticipantJob` exactly as you suggested; `SubscriberSyncInfoCollector` is back to `final`. The test's `TestableJob` overrides the seam.
- **Javadoc**: shortened the test class header to a single sentence.
- **mvn build verified**: `clean install` of the production bundles + dependent test bundles in a single reactor invocation, filtered with `-Dtest=AllTeamTests` and `-DfailIfNoTests=false`. BUILD SUCCESS, 22 tests, 0 failures, 0 errors. (Per-bundle `mvn clean verify -Pbuild-individual-bundles` works for `team.ui` and `team.core` but cannot be used for `team.tests.core` alone — that bundle requires sibling test bundles like `org.eclipse.compare.tests` and `org.eclipse.core.tests.resources` which aren't published to P2; this is the limitation noted in `AGENTS.md`.)
- Also reverted the `team.core` Bundle-Version bump from build #1 — with `final` restored the binary content is unchanged vs master, so the bump is no longer warranted. The bot will add a fresh bump if its analysis says one is needed.

vogella:
LGTM, planning to merge once the verification build finishes.

Thanks @FoxyBOA

FoxyBOA:
Thanks for the quick review — and especially for nudging me toward the seam approach over the `final` removal. Cleaner all around.

PR #2649, opened 2026-05-04
Remove remaining pseudo dependency of E4-Injector to jakarta annotations

HannesWell:
This ensures that the E4-Injector does not work with any specific annotation class.

The pseudo dependency was introduced when resolving [Bug 301462](https://bugs.eclipse.org/bugs/show_bug.cgi?id=301462), but just returning an empty array as dependencies of the ClassRequestor also passes the then added `DisposeClassLinkTest`.
I stepped through the code a bit when using the `ClassRequestor` and the `pseudoVariable` doesn't seem to have an effect.
Or does anyone know a reason to keep it?

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform/pull/1566

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.di/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.e4.core.di/META-INF/MANIFEST.MF
@@ -15,9 +15,7 @@ Export-Package: org.eclipse.e4.core.di;version="1.7.0",
  org.eclipse.e4.core.internal.di.osgi;x-internal:=true,
  org.eclipse.e4.core.internal.di.shared;x-friends:="org.eclipse.e4.core.contexts,org.eclipse.e4.core.di.extensions.supplier"
 Require-Bundle: org.eclipse.e4.core.di.annotations;bundle-version="[1.4.0,2.0.0)";visibility:=reexport
-Import-Package: jakarta.annotation;version="[2.0.0,4.0.0)",
- jakarta.inject;version="[2,3)",
- org.eclipse.osgi.framework.log;version="1.1.0",
+Import-Package: org.eclipse.osgi.framework.log;version="1.1.0",
  org.osgi.framework;version="[1.8.0,2.0.0)",
  org.osgi.util.tracker;version="[1.5.1,2.0.0)"
 Automatic-Module-Name: org.eclipse.e4.core.di
--- a/runtime/bundles/org.eclipse.e4.core.di/META-INF/MANIFEST.MF
+++ b/runtime/bundles/org.eclipse.e4.core.di/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.e4.core.di
-Bundle-Version: 1.9.800.qualifier
+Bundle-Version: 1.9.900.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/di/IInjector.java
+++ b/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/di/IInjector.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2017 IBM Corporation and others.
+ * Copyright (c) 2010, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,10 +14,6 @@
  *******************************************************************************/
 package org.eclipse.e4.core.di;
 
-import jakarta.annotation.PostConstruct;
-import jakarta.annotation.PreDestroy;
-import jakarta.inject.Scope;
-import jakarta.inject.Singleton;
 import java.lang.annotation.Annotation;
 import org.eclipse.e4.core.di.suppliers.PrimaryObjectSupplier;
 
@@ -37,7 +33,8 @@ import org.eclipse.e4.core.di.suppliers.PrimaryObjectSupplier;
  * The injection of values is generally done as a number of calls. User objects
  * that want to finalize the injected data (for instance, to perform
  * calculations based on multiple injected values) can place such calculations
- * in a method with the {@link PostConstruct} annotation.
+ * in a method with a {@code PostConstruct} annotation, e.g.
+ * {@code jakarta.annotation.PostConstruct}
  * </p>
  * <p>
  * When injecting values, all fields are injected prior to injection of methods.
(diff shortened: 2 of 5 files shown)


runtime/bundles/org.eclipse.e4.core.di/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -49,10 +49,10 @@ org.eclipse.jdt.core.compiler.problem.includeNullInfoFromAsserts=disabled
 org.eclipse.jdt.core.compiler.problem.incompatibleNonInheritedInterfaceMethod=warning
 org.eclipse.jdt.core.compiler.problem.incompleteEnumSwitch=ignore
 org.eclipse.jdt.core.compiler.problem.indirectStaticAccess=warning
-org.eclipse.jdt.core.compiler.problem.invalidJavadoc=error
+org.eclipse.jdt.core.compiler.problem.invalidJavadoc=warning
```
HannesWell: Even with `invalidJavadoc` turned to warning the javadoc links below report errors because the `jakarta.annotation/inject` annotations are not accessible anymore.
The only way I found to silence the errors was to add them back as `additional.bundles` in the `build.properties` below.
Does anyone know a JDT preference to ignore inaccessible types in javadoc? Or is this a bug?

akurtakov:
That way the DI will not make sure there is at least one annotation provider - it opens the door for products crafted with either javax or jakarta or both or none even. 
This might have the implication that target platforms will have neither thus projects need modifications in their target platforms in order to migrate to new eclipse version.

runtime/bundles/org.eclipse.e4.core.di/.settings/org.eclipse.jdt.core.prefs
laeubi: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4923

Apart from that I think the javadoc references are not that valuable so we can remove them if needed.

laeubi:
> The pseudo dependency was introduced when resolving [Bug 301462](https://bugs.eclipse.org/bugs/show_bug.cgi?id=301462), but just returning an empty array as dependencies of the ClassRequestor also passes the then added `DisposeClassLinkTest`.

Sadly the bugzilla link is inaccessible - do you have a copy of what this is talking about?

> I stepped through the code a bit when using the ClassRequestor and the pseudoVariable doesn't seem to have an effect.

Is this maybe somewhere else used (e.g. via reflection?) as it is a public static field ... it might be used for some kind of triggering a (reinjection) of dependencies.

HannesWell:
> > The pseudo dependency was introduced when resolving [Bug 301462](https://bugs.eclipse.org/bugs/show_bug.cgi?id=301462), but just returning an empty array as dependencies of the ClassRequestor also passes the then added `DisposeClassLinkTest`.
> 
> Sadly the bugzilla link is inaccessible - do you have a copy of what this is talking about?

Bugzilla seems to have had some temporary issues today. It was available when I wrote the comment and is now available again for me. But besides that the resolving commit might provide more context: https://github.com/eclipse-platform/eclipse.platform/commit/6cf3c7c9065674195f3238f061e5e062ff9951d7
 
> > I stepped through the code a bit when using the ClassRequestor and the pseudoVariable doesn't seem to have an effect.
> 
> Is this maybe somewhere else used (e.g. via reflection?) as it is a public static field ... it might be used for some kind of triggering a (reinjection) of dependencies.

I didn't find any such reflective reference. And something like that was also not done in the commit that introduced that class. My assumption is that it was just added to return a non empty array in `calcDependentObjects()`.
But as far as I can tell, just returning an empty array seems to have no drawbacks. The `DisposeClassLinkTest`, introduced with https://github.com/eclipse-platform/eclipse.platform/commit/6cf3c7c9065674195f3238f061e5e062ff9951d7, still passes.

HannesWell:
> That way the DI will not make sure there is at least one annotation provider - it opens the door for products crafted with either javax or jakarta or both or none even. This might have the implication that target platforms will have neither thus projects need modifications in their target platforms in order to migrate to new eclipse version.

That's in theory correct, but the other very basic e4-core/di Plug-ins still require it:
grafik

Therefore I'd assume that the chances are relatively low that one can really avoid the jakarta annotation bundles or that one would not have them in the in the TP transitively. But the chances are not zero.

My primary goal was to make the `o.e.e4.core.di` independent from any specific annotation package in order to ensure by design that it doesn't rely on specific class instances but just relies on names.

runtime/bundles/org.eclipse.e4.core.di/.settings/org.eclipse.jdt.core.prefs
HannesWell: Thanks for the pointer.

> Apart from that I think the javadoc references are not that valuable so we can remove them if needed.

Yes, they are not essential. But I think they're nice to have and if possible, I'd like to retain them

HannesWell:
I've now replaced the currently erroring JavaDoc links by code elements, which could be reverted when the JavaDoc processing in ECJ is less strict.



> > > I stepped through the code a bit when using the ClassRequestor and the pseudoVariable doesn't seem to have an effect.
> > 
> > 
> > Is this maybe somewhere else used (e.g. via reflection?) as it is a public static field ... it might be used for some kind of triggering a (reinjection) of dependencies.
> 
> I didn't find any such reflective reference. And something like that was also not done in the commit that introduced that class. My assumption is that it was just added to return a non empty array in `calcDependentObjects()`. But as far as I can tell, just returning an empty array seems to have no drawbacks. The `DisposeClassLinkTest`, introduced with [6cf3c7c](https://github.com/eclipse-platform/eclipse.platform/commit/6cf3c7c9065674195f3238f061e5e062ff9951d7), still passes.

My proposal is to submit this soon, so we would have sufficient time for fixes in case the conclusion is wrong that this is a save change.

akurtakov:
I agree, please merge as soon as you're comfortable with the change to get as big testing window as possible.

vogella:
I think this causes the failing test in the I-Build.

````
org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests	testBut336225()	Failure	The shell should not have been disposed first ==> expected: <false> but was: <true>

org.opentest4j.AssertionFailedError: The shell should not have been disposed first ==> expected: <false> but was: <true>
at org.junit.jupiter.api.AssertFalse.failNotFalse(AssertFalse.java:63)
at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:36)
at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:239)
at org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.testBut336225(PartRenderingEngineTests.java:1751)
````

I assume the pseudoVariable has a influence in the ordering of @PreDestroy

HannesWell:
> I assume the pseudoVariable has a influence in the ordering of @PreDestroy

Yes, that seems to be the case.

> The pseudo dependency was introduced when resolving [Bug 301462](https://bugs.eclipse.org/bugs/show_bug.cgi?id=301462), but just returning an empty array as dependencies of the ClassRequestor also passes the then added `DisposeClassLinkTest`.

At the same time, it seems that that test is not sufficient to cover all cases and should be enhanced. But I don't know all details from the test that failed...

Issue #2650, opened 2026-05-05
Wrong test changes reported in every PR


HeikoKlare:
For several days, the post of the test results bot in every PR lists changed (removed/added) tests (see [bot report example below](bot_report_example)). The report looks obivously broken as the actual test results do not indicate an issue, there is not such a high number of changed tests in any of the affected PRs, and log extract posted by the bot shows strange output. Still, that report is quite important for every PR as it tells us if we have accidentally removed any tests. But no one seems to care, as PRs are being merged without any note of the report.

I cannot properly trace the occurrence of the broken reports back to a PR merged in the Platform repository. 
One of the last PRs that had proper bot report was this:
- https://github.com/eclipse-platform/eclipse.platform/pull/2642

The following PR could have been the last with or without proper bot report. It was merged before CI results (and bot report) were available, which is why CI and bot report results are missing:
- https://github.com/eclipse-platform/eclipse.platform/pull/2641

That PR did affect console tests though, which are now listed in the strange output in the bot report. However, the report in https://github.com/eclipse-platform/eclipse.platform/pull/2642 states that it was compared against the commit of https://github.com/eclipse-platform/eclipse.platform/pull/2641 without showing the wrong bot report result.

The first PR that that contained the broken bot report and was merged with it was this:
- https://github.com/eclipse-platform/eclipse.platform/pull/2646

So the issue may not have been introduced by any change in this repository but by a change somewhere else.

I have also not found the log extracts posted by the bot in either of the CI logs (GHA or Jenkins), so I am currently not sure where they are extracted from.

### Bot Report Example

From https://github.com/eclipse-platform/eclipse.platform/pull/2643:

<details>
  <summary>This pull request <b>removes</b> 387 and <b>adds</b> 276 tests. <i>Note that renamed tests count towards both.</i></summary>

```
, nl=[Ljava.lang.String;@10bad903]
, nl=[Ljava.lang.String;@138eac9f]
, nl=[Ljava.lang.String;@14e1214a]
, nl=[Ljava.lang.String;@1a6c8679]
, nl=[Ljava.lang.String;@1c93d91c]
, nl=[Ljava.lang.String;@1d112dbf]
, nl=[Ljava.lang.String;@1fc74fd3]
, nl=[Ljava.lang.String;@221e21ee]
, nl=[Ljava.lang.String;@2350ee6d]
, nl=[Ljava.lang.String;@240e7fb]
…
```

```
[...]

PR #2651, opened 2026-05-05
Simplify display name in new parameterized tests for console

trancexpress:
This change adjusts display names in new parameterized tests, for console line wrapping. This hopefully avoids confusing GitHub jobs.

In addition, emojis in `IOConsoleTests` are replaced with unicode codepoints.

See: https://github.com/eclipse-platform/eclipse.platform/issues/2650


HeikoKlare:
I just further investigated the logs and saw that it also (even more likely) be the line breaks in the test name causing the issue. I am sure if it's a good idea to have such a complex test name representation. What about assigning proper descriptive names for the parameterization than just repeating all arguments?

trancexpress:
Descriptive will be difficult, but distinctive is possible. How about now?

HeikoKlare (approved):
> Descriptive will be difficult, but distinctive is possible. How about now?

Sounds good. I like the proposal, even in case it finally not fixes the issue with the test report. Let's give this a try. Thank you for quickly taking action!

PR #2652, opened 2026-05-05
Add active launch indicator to launch history menu

SougandhS:
Adds support to show an active launch indicator in the launch history menu, helping users quickly identify which launch configurations are currently running from launch history. Entries with associated non-terminated launches are marked with a visual indicator "●" ,Once the indicator is shown, users can immediately recognize that the configuration is already running, helping avoid unnecessary or unintended clicks from the launch history menu. 

With Active (Running)
image
image

<br>
With In-Active (Terminated or not launched)
image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
@@ -23,6 +23,7 @@ import java.util.concurrent.CompletableFuture;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.DebugPlugin;
+import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationType;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
@@ -347,8 +348,12 @@ public abstract class AbstractLaunchHistoryAction implements IActionDelegate2, I
 
 		// Add favorites
 		int accelerator = 1;
+		ILaunch[] launches = DebugPlugin.getDefault().getLaunchManager().getLaunches();
 		for (ILaunchConfiguration launch : favoriteList) {
 			LaunchAction action= new LaunchAction(launch, getMode());
+			if (checkIfLaunchActive(launch, launches)) {
+				action.setText(action.getText() + "  \u2699"); //$NON-NLS-1$
+			}
 			addToMenu(menu, action, accelerator);
 			accelerator++;
 		}
@@ -361,6 +366,9 @@ public abstract class AbstractLaunchHistoryAction implements IActionDelegate2, I
 		// Add history launches next
 		for (ILaunchConfiguration launch : historyList) {
 			LaunchAction action= new LaunchAction(launch, getMode());
+			if (checkIfLaunchActive(launch, launches)) {
+				action.setText(action.getText() + "  \u2699"); //$NON-NLS-1$
+			}
 			addToMenu(menu, action, accelerator);
 			accelerator++;
 		}
@@ -371,6 +379,24 @@ public abstract class AbstractLaunchHistoryAction implements IActionDelegate2, I
 			ActionContributionItem item= new ActionContributionItem(action);
 			item.fill(menu, -1);
 		}
+	}
+
+	/**
+	 * Returns whether the given launch configuration has been launched and active
+	 * now.
+	 *
+	 * @param launchConfiguration the launch configuration
+	 * @param launches            the current active launches
+	 * @return {@code true} if a matching launch exists, {@code false} otherwise
+	 */
+	private boolean checkIfLaunchActive(ILaunchConfiguration launchConfiguration, ILaunch[] launches) {
+		for (ILaunch launch : launches) {
+			ILaunchConfiguration activeLaunch = launch.getLaunchConfiguration();
+			if (activeLaunch != null && activeLaunch.equals(launchConfiguration) && !launch.isTerminated()) {
+				return true;
+			}
+		}
+		return false;
 	}
 
 	/**


iloveeclipse:
Since recently the "circle" was added to indicate **changed** file, I would not use this for **running** launch config, it would be confusing.

What about simple "(running)" label?

SougandhS:
> What about simple "(running)" label?

I went with this first, but It felt bit long...

image

SougandhS:
>Since recently the "circle" was added to indicate changed file, I would not use this for running launch config, it would be confusing.

How about `\u25B8` ?

image

iloveeclipse:
> How about `\u25B8` ?

How about \u2699  (  &#x2699; "gear" ) or \u1F3C3 ( &#x1F3C3; "running man" ) ?

SougandhS:
> How about \u2699 ( ⚙ "gear" ) or \u1F3C3 ( 🏃 "running man" ) ?

Gear looks great ! 
image
Pushed this change 

> \u1F3C3 ( 🏃 "running man" ) 

this doesn't render 😅

iloveeclipse:
The change is simple and nice, do we really need an option for that? If yes, please enable by default.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
```diff
@@ -342,13 +345,20 @@ private void setMenu(Menu menu) {
...
-		ILaunchConfiguration[] historyList= getHistory();
+		ILaunchConfiguration[] historyList = getHistory();
 		ILaunchConfiguration[] favoriteList = getFavorites();
 
 		// Add favorites
 		int accelerator = 1;
+		ILaunch[] launches = DebugPlugin.getDefault().getLaunchManager().getLaunches();
+		IPreferenceStore store = DebugUIPlugin.getDefault().getPreferenceStore();
+		boolean launchPref = store.getBoolean(IInternalDebugUIConstants.PREF_LAUNCHED_INDICATOR);
 		for (ILaunchConfiguration launch : favoriteList) {
```
iloveeclipse, line 352: Can you please guard the new code, including fetching of the launch configs by first check if `launchPref` is set? Same below.

akurtakov:
If unicode symbol is to be used - extra testing is needed that the chosen one is widely available in default fonts as per (https://eclipse.dev/eclipse/development/plans/eclipse_project_plan_4_40.xml#target_environments) and ideally in some popular fonts(whatever they are) and last but not least verify on all platform that nothing bad happens when a font that doesn't contain the symbol is used.

akurtakov:
> The change is simple and nice, do we really need an option for that? If yes, please enable by default.

Please no option for such things. We already have more options than one can find and manage.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
SougandhS: Added a length check for launches

SougandhS:
> > The change is simple and nice, do we really need an option for that? If yes, please enable by default.
> 
> Please no option for such things. We already have more options than one can find and manage.

Removed all the preference parts 👍

SougandhS:
> If unicode symbol is to be used - extra testing is needed that the chosen one is widely available in default fonts as per (https://eclipse.dev/eclipse/development/plans/eclipse_project_plan_4_40.xml#target_environments) and ideally in some popular fonts(whatever they are) and last but not least verify on all platform that nothing bad happens when a font that doesn't contain the symbol is used.

Thanks, I'll try to check in these platforms..

iloveeclipse:
> If unicode symbol is to be used - extra testing is needed that the chosen one is widely available in default fonts

See also https://stackoverflow.com/questions/48468088/unicode-character-uncompatibility

At least on RHEL 9.6 with "Sans Serif" dialog font (whatever system means with that name) it looks OK.

SougandhS:
Hi @iloveeclipse, @akurtakov, how about adding overlay image ?
image image

iloveeclipse:
> how about adding overlay image ?

Why not, but then maybe the one with the green "Run" arrow like on the Progress view icon?

akurtakov:
I have a preference to (widely spread) unicode symbol as (at least on Linux) these autoadjust contrast on changes of Gtk theme. Also they are less resource hungry and auto managed by OS instead of images that have to be destroyed and etc.

iloveeclipse:
So let's try with \u2699  (  &#x2699; "gear" ) and push it to M3 so hopefully more people could test it. If someone complains, we can switch it off for RC.

SougandhS:
> So let's try with \u2699 ( ⚙ "gear" ) and push it to M3 so hopefully more people could test it. If someone complains, we can switch it off for RC.

So include a preference option ?

iloveeclipse:
no, just merge & check for bug reports. worst case is some rectangle shown instead I assume.

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
```diff
@@ -373,6 +381,22 @@ protected void fillMenu(Menu menu) {
...
+	/**
+	 * Returns whether the given launch configuration has been launched.
+	 *
+	 * @param launchConfiguration the launch configuration
+	 * @param launches            the current active launches
+	 * @return {@code true} if a matching launch exists, {@code false} otherwise
+	 */
+	private boolean checkIfLaunched(ILaunchConfiguration launchConfiguration, ILaunch[] launches) {
+		for (ILaunch launch : launches) {
+			if (launch.getLaunchConfiguration().equals(launchConfiguration) && !launch.isTerminated()) {
```
iloveeclipse: launch.getLaunchConfiguration() can be null

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
```diff
@@ -361,6 +366,9 @@ protected void fillMenu(Menu menu) {
 		// Add history launches next
 		for (ILaunchConfiguration launch : historyList) {
 			LaunchAction action= new LaunchAction(launch, getMode());
+			if (launches.length > 0 && checkIfLaunched(launch, launches)) {
+				action.setText(action.getText() + "  \u2699"); //$NON-NLS-1$
```
iloveeclipse, line 370: Are **two** spaces between text and new character intended?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
iloveeclipse: This check is not needed, the `checkIfLaunched()` will not do much if it is empty, same below

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
SougandhS: Yes two spaces are required. else it wont look good

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
SougandhS: done

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/actions/AbstractLaunchHistoryAction.java
```diff
@@ -361,6 +366,9 @@ protected void fillMenu(Menu menu) {
...
+			}
 			addToMenu(menu, action, accelerator);
 			accelerator++;
 		}
```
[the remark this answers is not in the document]
SougandhS: ignoring this, as it will be few launches.

SougandhS:
Hi @iloveeclipse
if there are no more changes required can I merge this once the build is good ?

iloveeclipse (approved):

iloveeclipse:
Thanks @SougandhS. Please also add N&N entry.

SougandhS:
Thank you @iloveeclipse & @akurtakov 
>Please also add N&N entry.

Sure 👍

BeckerWdf:
Sorry for being late to the party: But I think we could improve UX of that feature. See https://github.com/eclipse-platform/eclipse.platform/issues/2933

PR #2653, opened 2026-05-05
Refactor launch configuration filtering to reuse preference store

SougandhS:
Cache the preference store in `doLaunchConfigurationFiltering(ILaunchConfiguration)` instead of repeatedly calling `getDefault().getPreferenceStore()`. This improves readability and reduces redundant method calls without changing existing behavior.


iloveeclipse (approved):

SougandhS:
Thank you @iloveeclipse & @MohananRahul

PR #2654, opened 2026-05-05
Reduce javax E4-Injector tests to the minimum and add Singleton test

HannesWell:
Reduce javax E4-Injector tests to the minimum as suggested in
- https://github.com/eclipse-platform/eclipse.platform/pull/1566#issuecomment-4371426287

I kept the separate `o.e.e4.core.javax.tests` project in order to keep dependency separation ensured.
In general it should also work to move the remaining tests into `o.e.e4.core.tests` but I worry a bit that then some issues maybe could went unnoticed (although I cannot think of one explicitly).

Additionally this adds a Singleton test, which didn't exist before (for jakarta and javax annotations).


runtime/tests/org.eclipse.e4.core.javax.tests/README.md
```diff
@@ -1,2 +1,2 @@
-This is a copy of `org.eclipse.e4.core.tests`, that uses javax annotations while the original uses jakarta annotations now.
+This is a reduced minimal copy of `org.eclipse.e4.core.tests`, that uses verifies javax annotations continue to work, not only the mainly used jakarta annotations.
```
HeikoKlare:
```suggestion
This is a reduced minimal copy of `org.eclipse.e4.core.tests`, that verifies javax annotations continue to work, not only the mainly used jakarta annotations.
```

HeikoKlare (approved):
This seems to be a reasonable reduction of the tests. Thank you! I only have a nit comment on an obsolete word in the readme.

HannesWell:
Great, thank you.

The test failure is unrelated, submitting.

PR #2664, opened 2026-05-18
Replace per-bit listener count fields with an AtomicIntegerArray

vogella:
Collapses the six volatile `count1`..`count32` fields and the four near-identical switch/if-bit helpers in `ResourceChangeListenerList` into a single `AtomicIntegerArray` indexed by `Integer.numberOfTrailingZeros(mask)`. The bookkeeping is now linear in the number of set bits, future event types are a no-op to support, and `hasListenerFor` no longer silently drops anything outside the hard-coded 1..32 range. Behaviour and thread-safety guarantees are preserved.

**Planned for 4.41**

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/ResourceChangeListenerList.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/ResourceChangeListenerList.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2008 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


iloveeclipse:
> The bookkeeping is now linear in the number of set bits, future event types are a no-op to support, and `hasListenerFor` no longer silently drops anything outside the hard-coded 1..32 range

These a valid points from AI point of view, but from human point of view:

1) There are no planned changes on listeners, `IResourceChangeEvent` API is stable since ~ 18 years.
2) Changed code is harder to understand for humans.

So while change "technically" itself is OK, it makes debugging / maintenance harder for humans. 
"no go" from me.

vogella:
I plan to add build events for individual projects and builder (requires events for projects), hence this change. This would allow clients to measure the build time of these builders on each project.

No problem if it is rejected, still possible to add new events just ugly to do so.

PR #2665, opened 2026-05-18
Drop stale event-type count from IResourceChangeEvent class Javadoc

vogella:
The class Javadoc on `IResourceChangeEvent` claims "There are currently five different types of resource change events", but the bullet list immediately below it enumerates six (`PRE_BUILD`, `POST_BUILD`, `POST_CHANGE`, `PRE_CLOSE`, `PRE_DELETE`, `PRE_REFRESH`). The count went stale when `PRE_REFRESH` was added and will keep drifting, so the fix replaces it with a count-free phrasing.


vogella:
Just a Javadoc change.

Issue #2667, opened 2026-05-19
JobGroupTest.testShouldCancel_4 failed in I-build


trancexpress:
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260518-1800/testresults/html/org.eclipse.core.tests.runtime_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

```
	expected: <true> but was: <false>

org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
at org.eclipse.core.tests.runtime.jobs.JobGroupTest.testShouldCancel_4(JobGroupTest.java:1280)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596) 
```

trancexpress:
This code seems to be called from multiple jobs:

```
numShouldCancelCalled[0]++;
```

Maybe it should be an `AtomicInteger` in that case. Currently its just an `int`.

PR #2668, opened 2026-05-19
Use AtomicInteger in JobGroupTest.testShouldCancel_4

trancexpress:
Adjusts `JobGroupTest.testShouldCancel_4()` to use an atomic data structure, to avoid concurrency issues with incrementing an integer from multiple jobs.

See: https://github.com/eclipse-platform/eclipse.platform/issues/2667

### DIFF
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
@@ -31,6 +31,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.concurrent.atomic.AtomicLong;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -1246,16 +1247,12 @@ public class JobGroupTest extends AbstractJobTest {
 	public void testShouldCancel_4() {
 		final int NUM_JOBS = 1000;
 		final int NUM_JOBS_LIMIT = 100;
-		final int numShouldCancelCalled[] = {0};
+		final AtomicInteger numShouldCancelCalled = new AtomicInteger(0);
 		final TestBarrier2 barrier = new TestBarrier2();
 		final JobGroup jobGroup = new JobGroup("JobGroup", 10, NUM_JOBS) {
 			@Override
 			protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFailedJobs, int numberOfCanceledJobs) {
-				numShouldCancelCalled[0]++;
-				if (numShouldCancelCalled[0] == NUM_JOBS_LIMIT) {
-					return true;
-				}
-				return false;
+				return numShouldCancelCalled.incrementAndGet() >= NUM_JOBS_LIMIT;
 			}
 		};
 		for (int i = 0; i < NUM_JOBS; i++) {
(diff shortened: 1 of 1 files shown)


runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1274,10 +1271,11 @@ public IStatus run(IProgressMonitor monitor) {
 		// Allow the jobs to proceed to run.
 		barrier.setStatus(TestBarrier2.STATUS_START);
 		waitForCompletion(jobGroup);
-		assertTrue(numShouldCancelCalled[0] >= NUM_JOBS_LIMIT);
+		int called = numShouldCancelCalled.get();
+		assertTrue(called >= NUM_JOBS_LIMIT);
```
fedejeanne:
```suggestion
		assertThat(called).isGreaterThanOrEqualTo(NUM_JOBS_LIMIT);
```

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1274,10 +1271,11 @@ public IStatus run(IProgressMonitor monitor) {
...
 		// Verify that the group is canceled in a reasonable time,
 		// i.e only 10 jobs are allowed to run after the shouldCancel method returned true.
-		assertTrue(numShouldCancelCalled[0] < NUM_JOBS_LIMIT + 10);
+		assertTrue(called < NUM_JOBS_LIMIT + 10, "Expected fewer jobs to run but got: " + called);
```
fedejeanne:
```suggestion
		assertThat(called).isLessThan(NUM_JOBS_LIMIT + 10);
```

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
```diff
@@ -1246,16 +1247,12 @@ public IStatus run(IProgressMonitor monitor) {
...
 		final TestBarrier2 barrier = new TestBarrier2();
 		final JobGroup jobGroup = new JobGroup("JobGroup", 10, NUM_JOBS) {
 			@Override
 			protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFailedJobs, int numberOfCanceledJobs) {
-				numShouldCancelCalled[0]++;
-				if (numShouldCancelCalled[0] == NUM_JOBS_LIMIT) {
-					return true;
-				}
-				return false;
+				return numShouldCancelCalled.incrementAndGet() >= NUM_JOBS_LIMIT;
```
fedejeanne, line 1255: Nice catch, checking for `==` was wrong because of the race condition (2 threads executing `numShouldCancelCalled[0]++;` before one of them reached the `if` below would skip numbers).

FTR it is now safe to simply do:

```java
return numShouldCancelCalled.incrementAndGet() == NUM_JOBS_LIMIT;
``` 

... because the comparison (`==`) uses the now stable value that comes from `incrementAndGet()` (no race condition there).

fedejeanne (commented):
Nice catch, thank you for the contribution! 

Just a nit: would it be possible to use AssertJ for more expressive assertions?

Other than that, the PR is fine.

runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/runtime/jobs/JobGroupTest.java
trancexpress: I'd rather be safe here with `>=`.

PR #2669, opened 2026-05-23
[RelEng] Handle absence of eclipse_lg images in release preparation

HannesWell:
The eclipse_lg*.png images are usually the same for multiple releases and a 'future' version therefore often does not exist.

This was missed in
- https://github.com/eclipse-platform/eclipse.platform/pull/2468


HannesWell:
Required RelEng change that's not checked in a verification build. Thus submitting immediately.

PR #2676, opened 2026-05-26
Allow customizing CommonTab

### LINKED ISSUE #2675: Allow customizing CommonTab
Our engineers would like to customize the common tab in launch configurations, to not show customization options that have no effect for some types of launches in our product. The respective code is in: `org.eclipse.debug.ui.CommonTab`

trancexpress:
Open up `CommonTab` for overriding methods, to customize what elements are shown in the UI. This allows to e.g. hide common tab elements that are not relevant for custom launch types.

Fixes: https://github.com/eclipse-platform/eclipse.platform/issues/2675

### DIFF
--- a/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.ui; singleton:=true
-Bundle-Version: 3.21.100.qualifier
+Bundle-Version: 3.22.0.qualifier
 Bundle-Activator: org.eclipse.debug.internal.ui.DebugUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
@@ -205,9 +205,9 @@ public class CommonTab extends AbstractLaunchConfigurationTab {
 	/**
 	 * Creates the favorites control
 	 * @param parent the parent composite to add this one to
-	 * @since 3.2
+	 * @since 3.22
 	 */
-	private void createFavoritesComponent(Composite parent) {
+	protected void createFavoritesComponent(Composite parent) {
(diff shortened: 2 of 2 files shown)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
```diff
@@ -205,9 +205,9 @@ IDialogSettings getDialogBoundsSettings(String id) {
 	/**
 	 * Creates the favorites control
 	 * @param parent the parent composite to add this one to
-	 * @since 3.2
+	 * @since 3.21
```
merks: Should this be 3.3?

I don’t see any version increment anywhere.

trancexpress: I did increment to 3.22 and tried using that for the `@since`, but the tooling complained:

```
Description	Resource	Location	Path	Type
The minor version is increased unnecessarily since either the major or minor version is already increased	MANIFEST.MF	line 5	/org.eclipse.debug.ui/META-INF	Version Numbering Problem
```

So I'm not sure. The current version is `3.21`.

merks: Goodness knows why a private method has a since annotation. In any case, when master opens there will be a new baseline and then the version will need to increase. The since annotations will need to match that new version.

debug/org.eclipse.debug.ui/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.debug.ui; singleton:=true
-Bundle-Version: 3.21.0.qualifier
+Bundle-Version: 3.21.100.qualifier
```
iloveeclipse: This should be `3.22.0` (please like here in an extra commit)

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
iloveeclipse: Please use `3.22` now everywhere, because before the tooling was checked against 4.39 platform, and now we have 4.40 as baseline, and this is the first API change in the 4.41 release.

iloveeclipse (changes requested):
Looks good except bundle version bump.

iloveeclipse:
@trancexpress : please use "since 3.22" now **everywhere** in this PR, "since 3.21" makes no sense.

iloveeclipse (approved):
LGTM, but please resolve merge conflicts with manifest version.

PR #2681, opened 2026-06-01
Allow reordering terminal tabs via drag and drop

### LINKED ISSUE #2679: Allow reordering terminal instances
Multiple terminals can be open in the Terminal view:
Image

It's not possible to reorder them with mouse click-and-drag like it is possible to reorder views or editor tabs. I suggest adding the same functionality there.

Build id: I20260518-1800
Windows 10

vogella:
Dragging a terminal tab within the same Terminals view now reorders it to the drop position, the same way editor and view tabs can be rearranged.
Previously a drop inside the originating view was rejected, so tabs could only be moved to a different Terminals view.
Only drops on the tab strip reorder; drops on the terminal content area are still rejected.

The reorder relies on the new `CTabFolder.moveItem(int, int)` API (SWT 3.135, eclipse-platform/eclipse.platform.swt#3350), which keeps the live terminal control attached instead of recreating it.
The existing cross-view move is left untouched because transferring a terminal to a different view still requires re-parenting.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2679

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/META-INF/MANIFEST.MF
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.terminal.view.ui;singleton:=true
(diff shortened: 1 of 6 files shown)


vogella:
Needs the change in SWT, which should be available tomorrow, after that this can be tested.

PR #2705, opened 2026-06-04
Increment o.e.core.runtime version for PluginVersionIdentifier removal

HannesWell:
The bundle `org.eclipse.core.runtime` re-exports
`org.eclipse.equinox.common` and therefore has to handle the API-tools consequences of the removal of `PluginVersionIdentifier` too.

Required due to:
- https://github.com/eclipse-equinox/equinox/pull/1308


HannesWell:
The expected API problems will only surface after the next I-build, so suppressing them now seems unused and generates a new warning.
In anticipation of the described problem I still submit this now so that verification builds continue to pass tomorrow (and workspaces stay error free).

PR #2707, opened 2026-06-04
Restore supplier-disposal link in ClassRequestor without jakarta

### LINKED ISSUE #2700: Test error in I-Build org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests	testBut336225()
https://download.eclipse.org/eclipse/downloads/drops4/I20260602-1800/testresults/html/org.eclipse.e4.ui.tests_ep441I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests testBut336225()

````

	The shell should not have been disposed first ==> expected: <false> but was: <true>

org.opentest4j.AssertionFailedError: The shell should not have been disposed first ==> expected: <false> but was: <true>
at org.junit.jupiter.api.AssertFalse.failNotFalse(AssertFalse.java:63)
at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:36)
at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:239)
at org.eclipse.e4.ui.tests.workbench.PartRenderingEngineTests.testBut336225(PartRenderingEngineTests.java:1751)

````

Looks to me like this is caused by https://github.com/eclipse-platform/eclipse.platform/pull/2649

@HannesWell please have a look

akurtakov:
PR #2649 decoupled the E4 injector from jakarta annotations by replacing the ClassRequestor pseudo-dependency descriptor with an empty array. That pseudo-link is what notifies an injected object when its supplier is disposed. With an empty descriptor the link is no longer established, so @PreDestroy of objects that only use constructor/@PostConstruct injection (no @Inject links) is no longer guaranteed to run before the supplying context is disposed.

This broke disposal ordering and surfaced as a failure of PartRenderingEngineTests.testBut336225 ("The shell should not have been disposed first"), reported in eclipse-platform/eclipse.platform#2700.

Restore a non-empty, optional descriptor without re-introducing jakarta: use a private marker interface InjectionLink that is never available from any supplier together with e4's own @Optional annotation. This keeps the injector free of specific jakarta annotation classes (the goal of #2649) while re-establishing the supplier-disposal link.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2700

### DIFF
--- a/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/ClassRequestor.java
+++ b/runtime/bundles/org.eclipse.e4.core.di/src/org/eclipse/e4/core/internal/di/ClassRequestor.java
@@ -13,8 +13,10 @@
  *******************************************************************************/
 package org.eclipse.e4.core.internal.di;
 
+import java.lang.reflect.Field;
 import org.eclipse.e4.core.di.IInjector;
 import org.eclipse.e4.core.di.InjectionException;
+import org.eclipse.e4.core.di.annotations.Optional;
 import org.eclipse.e4.core.di.suppliers.IObjectDescriptor;
 import org.eclipse.e4.core.di.suppliers.PrimaryObjectSupplier;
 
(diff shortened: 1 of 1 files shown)


HannesWell (approved):
Thank you Alex.

I was able to verify locally that this fixes the test in question.
Furthermore I tried to simplify this and to avoid the `PSEUDO_VARIABLE`, by creating the used types directly. But for the Annotation instantiation this was eventually not simpler.
So I just made the code and comments a bit less verbose, as we discussed in private.

Will submit this once the build is ready to have the test failure fixed with tonight’s I-build.
And since it's similar to the state from https://github.com/eclipse-platform/eclipse.platform/pull/2649, I think there is not much discussion necessary.

PR #2708, opened 2026-06-04
Remove unnecessary provision of disabled icons

HeikoKlare:
Disabled icons are generated by SWT on the fly in the same quality, thus explicitly providing them is not necessary


akurtakov:
As this is examples bundle - delete the icons too.

HeikoKlare:
Yes. Updated the commit with removal of the icon files.

PR #2717, opened 2026-06-05
Remove obsolete hover image descriptors

HeikoKlare:
For several actions, hover image descriptors that are equal to the ordinary image descriptors are set without any need. This change removes those obsolete descriptors.


vogella:
LGTM

PR #2725, opened 2026-06-05
Add Additional Space after <terminated> & <Disconnected> label

SougandhS:
Current space does not sync with what shown in console 
Screenshot 2026-06-05 at 5 58 36 PM

Current (No single space)
Screenshot 2026-06-05 at 5 58 20 PM

New (added a single space)
Screenshot 2026-06-05 at 6 01 07 PM

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DefaultLabelProvider.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/DefaultLabelProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -255,13 +255,13 @@ public class DefaultLabelProvider implements ILabelProvider {
 						int exit = process.getExitValue();
 						terminatedMessage = MessageFormat.format(DebugUIMessages.DefaultLabelProvider_16, Integer.toString(exit));
 					} else {
-						terminatedMessage= DebugUIMessages.DefaultLabelProvider_1;
+						terminatedMessage = DebugUIMessages.DefaultLabelProvider_1;
 					}
-					label.insert(0, terminatedMessage);
+					label.insert(0, terminatedMessage + " "); //$NON-NLS-1$
 				}
 			} else if (element instanceof IDisconnect) {
 				if (((IDisconnect) element).isDisconnected()) {
-					label.insert(0, DebugUIMessages.DefaultLabelProvider__disconnected__1);
+					label.insert(0, DebugUIMessages.DefaultLabelProvider__disconnected__1 + " "); //$NON-NLS-1$
 				}
 			}
 		} catch (DebugException e) {


tomaswolf:
How does that look in an RTL UI language?

merks:
> How does that look in an RTL UI language?

That's a trick question!  😁  A space is a direction-neutral character so how things render really depends on the directionality of the characters before and after the space.  

https://www.w3.org/International/articles/inline-bidi-markup/uba-basics

This brings back bad memories of translation testing at IBM where somehow it felt like one could never really quite get things correct for RTL languages.  And then there is the politically incorrect question, do developers really use the development environment itself in RTL mode?  (I think actually that's quite unlikely given talking to real developers back in those days.)  Note that's a completely different question whether people really develop RTL applications because that definitely should be well supported...

SougandhS:
Good point. I haven't verified the rendering in an RTL yet. Is there a recommended way to quickly test this in a child Eclipse ?

merks:
It's really confusing if you ask me. 

I copied character from the alphabet here:

For the following line I "typed" Alef, Bet, A, and B, in that order with spaces between.

א ב A B

Here they are on separate lines, in order of how I entered them:
- א
- ב
- A
- B

So you see how it rendered RTL and then LTF depending on the characters.  (It's really hard to select the text as well.)

Here how it looked before it was rendered:

image

All this stuff will make your head explode.  Mixing RTL and LTR is kind of a disaster (in my opinion) but space is neutral and it is the least of your concerns....

I cannot imagine a developer inflecting this pain upon themselves in their development IDE...

merks (approved):
So this looks fine to me.

I'm not sure if @tomaswolf as a serious concern.  I expect the space just to work, but I'm not sure if you are concerned about the translated text itself being translated to an RTL language or about the text after the `terminatedMessage` text?  But I don't think in either case it matters at all; the space will separate the two things and the order will be determined by the bidirectional rendering algorithm which might given you `foo bar` or `bar oof` or `foo rab` all depending the directionality of the characters involved...

tomaswolf:
It was a minor but serious concern. What does it show in RTL _now_? "ZYX&lt;terminated>"? If the RTL after the change shows this as "ZYX&lt;terminated> ", then just adding a blank doesn't work. If adding a blank to "&lt;terminated>" results in the correct spacing also in RTL, then fine. (And of course, the word "terminated" itself might be translated too, and complicate matters even more.)

Is there a translation bundle for debug.ui for an RTL language? If there isn't, then let's just forget about this. If there is, try it  and take a look at how it renders this before and after the change.

merks:
> It was a minor but serious concern. 

Serious, but as I suggest not actually realistic.  Because, having spoken to actual developers, the general reaction was, OMG I would never do that because it would be a confusing mess.

> What does it show in RTL _now_? "ZYX<terminated>"? 

You can literally answer this yourself, if you don't get brain damage trying to type it:

`<terminated>AB`
`<terminated> אב`
`<אב> AB`
`<אב> אב`

You see how unpleasant is the handling of the `<>`?

image

There are some invisible directional control characters one can use to "correct" the behavior.  It simply a nightmare.

> If the RTL after the change shows this as "ZYX<terminated> ", then just adding a blank doesn't work. 

We seem to have gone done the rabbit hole.   Please draw no conclusions without a basis for them.  You can copy Hebrew letters form an online alphabet and play with it.  It's just no pleasant.

> If adding a blank to "<terminated>" results in the correct spacing also in RTL, then fine. (And of course, the word "terminated" itself might be translated too, and complicate matters even more.)
> 

Yes, it's never ending is it.  And while `-` and ` ` are non-direcitonal `<` and `>` are not so neutral.

> Is there a translation bundle for debug.ui for an RTL language?

I don't think so:

https://download.eclipse.org/justj/?file=technology/babel/babel_language_packs/R0.20.0/2022-12

> If there isn't, then let's just forget about this. If there is, try it and take a look at how it renders this before and after the change.

I think we should forget about this.  It's bringing back really bad memories.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=179191
https://bugs.eclipse.org/bugs/show_bug.cgi?id=202752

There's more of those, many, many more.   Nothing works well without some serious semantic knowledge of what is being displayed..

SougandhS:
Thank you @merks for looking into the RTL issues :) , I was totally confused with it

PR #2741, opened 2026-06-07
Remove unwanted inner try block

SougandhS:


debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/groups/GroupLaunchConfigurationDelegate.java
```diff
@@ -402,12 +402,8 @@ public static void storeLaunchElements(ILaunchConfigurationWorkingCopy configura
 	public static void removeLaunchElements(ILaunchConfigurationWorkingCopy configuration) {
 		try {
 			for (String attr : configuration.getAttributes().keySet()) {
-				try {
-					if (attr.startsWith(MULTI_LAUNCH_CONSTANTS_PREFIX)) {
-						configuration.removeAttribute(attr);
-					}
-				} catch (Exception e) {
-					DebugPlugin.log(e);
+				if (attr.startsWith(MULTI_LAUNCH_CONSTANTS_PREFIX)) {
```
merks, line 405: What kind of exception might happen?  Do you really want the whole loop to terminate if there is an exception?

SougandhS: Since neither startsWith() nor removeAttribute() declares any exceptions, I initially thought the inner try/catch could be removed

merks: It all looks a little overly cautious given that nothing declares any thrown exceptions.   It looks like it's always been there so there's no real accounting for why either of those try blocks are there, neither the inner nor the outer.  Maybe it happened during some initial implementation and was then never cleaned up.  I'm not sure what to say about that.   😞

SougandhS: I checked other references of removeAttribute(), and they don't have a similar try/catch. Since neither API declares any exceptions, the inner try/catch may just be leftover from an older implementation.

merks: Do other places have the outer try block?  It all looks a bit bogus...

SougandhS: Only place I found a try block is in `org.eclipse.jdt.debug.ui.launchConfigurations.JavaDependenciesTab.performApply(ILaunchConfigurationWorkingCopy)` but that is only for catching `CoreException` (which is for `org.eclipse.debug.core.ILaunchConfiguration.getAttribute(String, boolean)` )
image

everywhere else is try-catch free..

merks (approved):
It’s okay then.  Thanks for checking.

SougandhS:
> It’s okay then. Thanks for checking.

Thanks!

elsazac (approved):
LGTM; merging this now.

PR #2745, opened 2026-06-08
Fix incorrect JavaDoc for ATTR_TERMINATE_TIMESTAMP

SougandhS:
Update the JavaDoc for ATTR_TERMINATE_TIMESTAMP. Previously, it incorrectly referred to the launch time instead of the termination time.


MohananRahul (approved):

SougandhS:
Thanks !

PR #2753, opened 2026-06-09
Remove requirement of inlined o.e.jdt.launching.macosx fragment

HannesWell:
Part of
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/855

### DIFF
--- a/ant/org.eclipse.ant.tests.ui/META-INF/p2.inf
+++ b/ant/org.eclipse.ant.tests.ui/META-INF/p2.inf
@@ -1,3 +1,0 @@
-requires.0.namespace = org.eclipse.equinox.p2.iu
-requires.0.name = org.eclipse.jdt.launching.macosx
-requires.0.filter = (osgi.os=macosx)


vogella:
I think that is what currently fails the aggregator build, https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3888

Thanks!

HannesWell:
> I think that is what currently fails the aggregator build, [eclipse-platform/eclipse.platform.releng.aggregator#3888](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3888)

Yes it is. But since there is some chicken-egg problem when resolving the removal/inlining of the previous needed fragment there is no (trivial) perfect way to do it and this will probably fail some tests until the inlining has passed an I-build. 
But since verification builds of multiple repositories will fail with this applied I decided to do it close to the I-build start.
The aggregator verification build failing over the day was IMO the smaller collateral damage.

HannesWell:
Test failures on Mac are expected because the inlined code of the fragment is only available in the new target bundle after tonight's I-build.

vogella:
Thanks for the update, yes a aggregator failing for a day or so is fine IMHO for such a nice improvement. Thanks for working on this

Issue #2680, opened 2026-05-31
Codex CLI in an Eclipse IDE terminal doesn't allow backward scrolling (also with --no-alt-screen option)


stefano-1973:
Codex CLI in an Eclipse IDE terminal window doesn't allow backward scrolling. Also, running Codex with the --no-alt-screen option doesn't change anything.
I know the Eclipse terminal is scrollable and Codex CLI running in Windows cmd or powershell is scrollable.
So there must be some problem between Codex CLI and Eclipse terminal. 

## Steps to reproduce

* Open a terminal in eclipse IDE
* Start Codex CLI with or without --no-alt-screen  (even setting "set TERM=dumb" before calling codex doesn't help)
* Give Codex the following prompt : write 50 rows (a enter b enter c etc...)
* Try scrolling backward and see if it appears

## Tested under this environment:
* Windows Server 2019
* Eclipse IDE/Platform version Version: 2026-03 (4.39.0) Build id: 20260305-0817
* Codex Cli version 135

## Workarounds

1. Use the internal Codex transcript directrly in the Eclipse terminal running Codex TUI, activable with **Ctrl+T**
2. **Tricks Codex to think it's running inside Zwellij**:
For linux:
```
ZELLIJ=1 codex --no-alt-screen -c tui.raw_output_mode=true
```
For windows:
```
set ZELLIJ=1
codex --no-alt-screen -c tui.raw_output_mode=true
```

stefano-1973:
Because the cause is Codex Cli only when used with Eclipse terminal (and vice-versa), I open also this: https://github.com/openai/codex/issues/25412

merks:
I would have extremely low expectations with respect to someone else doing something to fix whatever problem there might be here.   Whomever wants this to work will most likely need to invest the effort needed to do whatever is necessary to make it work themselves.

vogella:
@stefano-1973 What does Codex CLI say about a fix in this repo?

vogella:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3350

merks:
> [eclipse-platform/eclipse.platform.swt#3350](https://github.com/eclipse-platform/eclipse.platform.swt/pull/3350)

I guess you meant to link tab reordering to a different issue.

stefano-1973:
The problem is present also if I start linux elipse in WSL2 ubuntu. Eclipse installed in wsl2 with "sudo snap install eclipse --classic"

[...]

PR #2764, opened 2026-06-11
Move displaced line of top-anchored scroll region into scroll-back

vogella:
Full-screen CLIs such as Codex in --no-alt-screen mode render through DECSTBM scroll regions anchored at the top of the screen and scroll by emitting newlines at the region bottom. The emulator scrolled such a region in place and discarded the displaced top line, so no scroll-back history was created and the output could not be scrolled backwards.

Like xterm and VTE, a newline at the bottom of a scroll region whose top margin is the top of the screen now moves the displaced top line into the scroll-back history, while the lines below the region keep their position on the screen. Regions with a lower top margin still scroll in place, guarded by the characterization tests from #2701; the test documenting the old top-anchored behavior is flipped to assert the history is created.

This addresses the missing backward scrolling reported in #2680.


stefano-1973:
It seems to work:
image

PR #2766, opened 2026-06-12
Run terminal tests with the UI test harness

### LINKED ISSUE #2762: 5 terminal.test TerminalsViewReorderTest failures since I20260610-1800
Regression from 1a1fa26bf8031ec5fa610c178464acf3d6fc0c54 / https://github.com/eclipse-platform/eclipse.platform/issues/2679.

See https://download.eclipse.org/eclipse/downloads/drops4/I20260610-1800/reports.html#tests

Looks like test setup is not working properly on SDK builds.

vogella:
Since #2681 the terminal test suite contains TerminalsViewReorderTest, which loads the TerminalsView class and thereby lazily activates org.eclipse.terminal.view.ui. That activation fails in the headless core-test harness used by the SDK I-builds because UIPlugin.start() registers a workbench listener, causing the 5 failures reported in #2762. The Tycho build stayed green because it auto-detects the UI harness for this bundle and already runs the whole suite with a workbench.

This switches the test.xml suite target from core-test to ui-test so the SDK builds get the same environment, matching the other UI test bundles in this repository.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2762

### DIFF
--- a/terminal/tests/org.eclipse.terminal.test/test.xml
+++ b/terminal/tests/org.eclipse.terminal.test/test.xml
@@ -32,7 +32,7 @@
 	<target name="suite">
 		<property name="sniff-folder" value="${eclipse-home}/terminal_sniff_folder" />
 		<delete dir="${sniff-folder}" quiet="true" />
-		<ant target="core-test" antfile="${library-file}" dir="${eclipse-home}">
+		<ant target="ui-test" antfile="${library-file}" dir="${eclipse-home}">
 			<property name="data-dir" value="${sniff-folder}"/>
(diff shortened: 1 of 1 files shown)


HeikoKlare:
Seems like a proper correction of consistency. Because fwiw, this statement is not correct:

> The Tycho build stayed green because it auto-detects the UI harness for this bundle and already runs the whole suite with a workbench.

There is no auto-detection but the UI harness was explicitly set for the test bundle in the PR introducing the tests:
- https://github.com/eclipse-platform/eclipse.platform/pull/2681

PR #2767, opened 2026-06-12
Make terminal tab reorder drop handling platform robust

vogella:
The terminal tab reorder added for #2679 did not work on Windows and on Linux Wayland, while it worked on X11. The reorder mapped the drop event coordinates to the tab folder, but those coordinates are not display-relative on every platform (on Wayland they are relative to the shell), so the drop position check rejected every drop on the tab strip. In addition, drops over the tab folder are not necessarily delivered to a drop target registered on its parent.

The drop target is now also registered on the tab folder itself, and when the drop event coordinates do not map onto the tab strip the reorder retries with the cursor location, which uses the same coordinate base as Control.toControl and is what the workbench tab reorder relies on. Verified in a runtime IDE on Wayland; a retest on Windows with the next I-build would be appreciated.

See https://github.com/eclipse-platform/eclipse.platform/issues/2679

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/view/TerminalsView.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/view/TerminalsView.java
@@ -254,11 +254,19 @@ public class TerminalsView extends ViewPart implements ITerminalsView, IShowInTa
 	}
 
 	/**
-	 * Initialize the drop support on the terminals page book control.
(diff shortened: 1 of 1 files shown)


nlisker:
> a retest on Windows with the next I-build would be appreciated.

Ping me when it's there.

vogella:
Give it a try in tomorrows build

nlisker:
Build id: I20260616-0637 still can't reorder them on Win 10:

https://github.com/user-attachments/assets/394f518e-5c29-4128-b4c5-08b202fe9fb4

merks:
It works for me in a debug launch:


https://github.com/user-attachments/assets/151bf077-6464-484a-ab32-87e2256d6450

PR #2771, opened 2026-06-14
Remove unused org.eclipse.core.filesystem.linux.loongarch64 fragment

HannesWell:
It is not published and not used anywhere.

Complements
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3384


akurtakov:
This is related to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2187 . 
I can't make my mind on this issues as I have been on the other side back in the days for ~2 years with initial Linux/arm64 builds before it became part official build of Eclipse. On one side making it easier for people to build on other architectures helps growing the community but only if there is someone truly active to drive such a port.

HannesWell:
> This is related to [eclipse-platform/eclipse.platform.releng.aggregator#2187](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2187) .

Since downstream consumers that build the native fragments exists, lets keep this to support them.
When migrating to FFM all the CPU arch specific fragments become obsolete anyway.

PR #2785, opened 2026-06-29
Switch to `Platform.getDebugBoolean(String)` where appropriate

ptziegler:
With 3b76d20c6f6da97f8ba449d6303e8a47b4543331, a convenience method `getDebugBoolean(String)` was added, which returns `true` if the given tracing option is enabled. Using this method avoids a lot of the boilerplate code that comes with using `getDebugOption(String)` and which usually looks something like:

`"true".equals(Platform.getDebugOption("..."))`


vogella:
LGTM

ptziegler:
@vogella Thank you checking this PR! 🎉

PR #2788, opened 2026-07-01
Fix "Show Command Line" button clipping at larger font sizes

### LINKED ISSUE #462: Java Application Run Configuration "Show Command Line" button text display exceed its button border
# Operation environment
- win11 pro workstation
- eclipse 2024-06
- openjdk 9
- my locale

    zh_CN

- my preference
  - dialog font
    
    font name: `source code pro`

    font size: 11
  - display resolution

    `1920x1080`, intel display default settings.
# Case
A Java class which has main method, check its run configuration plane, the button `Show Command Line` locates at bottom right corner,its button text display exceed the button border, but the exceed part is unvisible.

screen shot as follows:
image

# Try

1. restore dialog font to default(Use System Font)

    by doing this,this button display correctly.

2. set font size greater than 9(default size)

    Use system font, and when size greater than 9, it is going to exceed the button border like what i described before.

I dont know if it works well on your operation environment.

SougandhS:
Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/462

Before:
image

After: 
image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/LaunchConfigurationTabGroupViewer.java
@@ -276,6 +276,7 @@ public class LaunchConfigurationTabGroupViewer {
 		fShowCommandLineButton = SWTFactory.createPushButton(buttonComp,
 				LaunchConfigurationsMessages.LaunchConfigurationDialog_ShowCommandLine, null,
 				GridData.HORIZONTAL_ALIGN_BEGINNING);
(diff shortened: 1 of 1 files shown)


deepika-u:
With the patch applied I am seeing -> 
image

In debug configuration -> 
image

With current preference -> 
image

Looks like your fix is working. Thanks.

SougandhS:
Thank you @deepika-u for confirming 👍

fedejeanne (commented):
Hi @SougandhS I noticed that `Apply` and `Revert` suffer from the same problem, maybe you could fix them too?


I set the font size (of the `Dialog font`) to 28 and it looks like this:

Image

SougandhS:
I did tested those with few high font sizes on Mac, it looked fine for me 
eg. 28 on mac
image

But will fix these too 👍

PR #2789, opened 2026-07-01
Show unified diff for compare inputs that carry an ancestor

vogella:
The experimental unified diff overlay was only offered for strict 2-way compare inputs, because `canShowInUnifiedDiff` bailed out whenever the input carried a common ancestor.
That excluded EGit history compares such as "Compare With Previous Version", which attach a git merge-base ancestor even for a conceptually 2-way diff.
Since the overlay is computed purely from the left and right sides, the ancestor can safely be ignored.
With this change those compares render inline diff markers when the experimental unified diff preference is enabled.


vogella:
It works from "Compare with -> Previous Revision".

image

Without this patch:

image

vogella:
Also works now from the History view for files -> Compare with Previous version

tobiasmelcher (approved):
thanks a lot Lars.

vogella:
> thanks a lot Lars.

Awesome feature thanks a bunch @tobiasmelcher for providing this.

PR #2790, opened 2026-07-02
Use ILog instead of CompareUIPlugin logging helpers

vogella:
This replaces the static `CompareUIPlugin.log`/`logErrorMessage` helpers with `org.eclipse.core.runtime.ILog` at every call site in `org.eclipse.compare` and removes the three helpers.
Severity and messages are preserved, so there is no behavior change; it just moves logging off the `CompareUIPlugin` activator and drops one more use of the `getDefault()` singleton.
All call sites in the bundle were migrated and the bundle compiles cleanly.


merks:
This actually looks more verbose, so an advantage seems unclear to me.  Also the log call needs to look up the bundle form the class so this will perform less well, not that it really much matters.  But in the end, I don't see the motivation nor advantage here.

PR #2791, opened 2026-07-02
Fold unchanged regions in the unified diff

vogella:
Folds the unchanged regions of the experimental unified diff overlay, similar to the collapsed context in GitHub's diff view.
The gaps between the displayed changes are collapsed with three context lines kept around each change, which is what makes a large file with few changes reviewable at a glance: `AbstractTextEditor.java` with four changes fits on one screen instead of 8072 lines.
Each collapsed region carries a clickable band that expands it in place and then reads "Hide n unchanged lines", so a region opened by mistake can be put back without going to the folding ruler.
The editor's projection model is reused and the folds are tagged, so tearing the overlay down leaves the editor's own folds alone; a fold the editor contributes inside a collapsed region is taken out while that region is collapsed, because its indicator would otherwise sit on a foreign line and do nothing when clicked.
A "Fold unchanged regions" preference on the Compare page turns the folding off; it is on by default.

Two known limitations.
Reusing projection support means this has no effect in an editor that does not install it, such as the default text editor.
And in a Java editor, revealing a line whose enclosing element is the whole type, a line in the class javadoc for instance, expands every region inside that type, because `JavaEditor.adjustHighlightRange` exposes the element's full source range.

Needs https://github.com/eclipse-platform/eclipse.platform.ui/pull/4361, which is merged.

Part of https://github.com/eclipse-platform/eclipse.platform.ui/issues/3771

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019P5XSUkajCGAiyxqAzgymD

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.java
@@ -148,6 +148,8 @@ public final class CompareMessages extends NLS {
 	public static String UnifiedDiff_openTwoWayCompare_tooltip;
 	public static String UnifiedDiff_cannotEditFile_title;
 	public static String UnifiedDiff_cannotEditFile_message;
+	public static String UnifiedDiff_expandUnchangedLine;
+	public static String UnifiedDiff_expandUnchangedLines;
 
 	static {
 		NLS.initializeMessages(BUNDLE_NAME, CompareMessages.class);
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.java
@@ -137,6 +137,10 @@ public final class CompareMessages extends NLS {
 	public static String UnifiedDiff_openTwoWayCompare_tooltip;
 	public static String UnifiedDiff_preparing;
 	public static String UnifiedDiff_computing;
+	public static String UnifiedDiff_showUnchangedLine;
+	public static String UnifiedDiff_showUnchangedLines;
+	public static String UnifiedDiff_hideUnchangedLine;
+	public static String UnifiedDiff_hideUnchangedLines;
 
 	static {
 		NLS.initializeMessages(BUNDLE_NAME, CompareMessages.class);
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.properties
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareMessages.properties
@@ -163,3 +163,5 @@ UnifiedDiff_revert=Revert
 UnifiedDiff_openTwoWayCompare_tooltip=Open in 2-way Compare Editor
 UnifiedDiff_cannotEditFile_title=Cannot edit file
 UnifiedDiff_cannotEditFile_message=Cannot edit file {0}: {1}
+UnifiedDiff_expandUnchangedLine=Expand 1 unchanged line
+UnifiedDiff_expandUnchangedLines=Expand {0} unchanged lines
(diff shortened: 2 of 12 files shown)


vogella:
This is a draft so that @tobiasmelcher can join in if he wants.

Current state: Folding works but I have no way to unfold.

image

tobiasmelcher:
Hi Lars,

thanks for adding code folding to the unified diff view! I tried it out and ran into a confusing interaction.

When I click on line 6 and line 46, the fold indicator appears to be clickable, but no folding actually happens. The lines remain fully expanded after the click. I recorded a short screencast showing the behavior:

https://github.com/user-attachments/assets/a7230fd6-6e6b-4516-b341-04295539d21c

I think the issue is related to the folding region starting at a line that is already collapsed by an "Expand ...
unchanged lines" header annotation. In other words, the fold indicator is shown, but the region it would fold is
already partly hidden by the diff context mechanism, so the click has no visible effect.

In general, I like the feature. Clicking the "Expand ... unchanged lines" header annotation works perfectly fine.

What are your plans for this? Should we merge the feature as-is, or would it make sense to first fix this interaction?
Alternatively, we could provide a preference so users can opt into the folding feature once it is more polished.

With best regards,
Tobias

vogella:
Hey Tobias thanks for testing and the feedback.

I also think it will be quite useful especially in case of big files with only small changes.

I left it in draft because I also found issues, for example if I increase the text size only parts of the unified display changes.

I would like to leave it for now in draft, I testing at the moment the functionality and will return to this pr in a few days.

During development I realized that compare was too slow and focused on this before working on new functionality. For example a 10 000 line file is or at least was too slow to be seen in unified diff.

vogella:
Still some issues with this one....

vogella:
Screenshot From 2026-09-07 20-01-06

@tobiasmelcher please re-test.

vogella:
Depends on https://github.com/eclipse-platform/eclipse.platform.ui/pull/4361

tobiasmelcher:
Thanks a lot Lars, looks really nice now. 

Unfortunately, I found a small issue after clicking the first "Show ... unchanged files" header annotation. The header mining jumps then back to the top of the document which doesn't feel right, it should have been disappeared from my point of view. Here a screencast:

https://github.com/user-attachments/assets/08f9636a-ea9b-4745-855a-273d4a1e9706

I tested the changes of this pull request together with https://github.com/eclipse-platform/eclipse.platform.ui/pull/4361 . Was that correct?

tobiasmelcher:
Hi Lars,
 
I retested the latest changes locally together with eclipse.platform.ui#4361. The folding itself works well, but after expanding a folded region the "Hide 4 unchanged lines" code mining remains visible. I think the header should be an expander only and disappear once its region is expanded; the folding ruler can still be used to collapse it again.
 
I tested the attached patch locally and it fixes this scenario. It creates unchanged-region code minings only for collapsed folds, removes the obsolete “Hide …” label, and adds a regression test which checks that the attached mining disappears after expansion.

If this matches your intended interaction, could you please apply it with:
git apply fix-unified-diff-expander-mining.patch

[fix-unified-diff-expander-mining.patch](https://github.com/user-attachments/files/32015955/fix-unified-diff-expander-mining.patch)

Thanks,
Tobias

vogella:
Thanks @tobiasmelcher I think the "Hide 4 unchanged lines" is a useful feature. Sometimes I want to expand a part to see the x lines above the change and afterwards I want to hide them again. 

Here are some screenshots:

 
shot-01-initial
shot-02-top
shot-03-expanded
shot-04-after-expand-band
shot-05-hide-band
shot-06-first-region-expanded
shot-07-hide241
shot-08-folded-overview

vogella:
@tobiasmelcher any additional feedback? Planning I merge it tomorrow.

tobiasmelcher:
> @tobiasmelcher any additional feedback? Planning I merge it tomorrow.

Sorry for the late reply. I was busy with other tasks. Thanks a lot for your contribution, Lars! This feature is a significant improvement. Great work!

PR #2793, opened 2026-07-02
Introduce faster file refresh on Linux (x86_64 only for now)

iloveeclipse:
The code is supposed to speed up NFS file access on Linux by an order of magnitude by fetching directory children data (IFileInfo) in one shot from JNI (via array of StructStat).

LocalFile.childInfos(int, IProgressMonitor) overrides now the FileStore.childInfos(int, IProgressMonitor) and calls (on Linux) LinuxFileHandler.listDirectoryAndGetFileInfos() which forwards to faster native implementation in
LinuxFileNatives.listDirectoryAndGetFileInfos().

- Introduced new libfastlinuxfile_1_0_0 library for x86_64 Linux architecture (which can be also built on other Linux architectures).
- Created dedicated Linux specific package org.eclipse.core.internal.filesystem.local.linux
- removed Mac code paths from new code to simplify Linux port
- Added system property to disable new binary if needed via `-Declipse.filesystem.useFastLinuxNatives=false`.
- With the flag above, the behavior on Linux should be unchanged compared with the previous code state.
- In next release we can try to remove native code and use FFM API if we will allow the bundle to use Java 25. 
- As of today, FFM use is not possible - multi-release based bundle attempt failed due missing PDE / Equinox support for debugging MR code (see https://github.com/eclipse-pde/eclipse.pde/issues/2391).

| Variant       | NFS (ms)   | Local (ms)  |
|---------------|---------------|-------------|
| Original code | 70000 - 90000 | 900 - 2500  |
| Updated code  | 800 - 5500    | 800 - 2500  |

I've measured data with the extra patch to trace the refresh data from Eclipse : https://github.com/eclipse-platform/eclipse.platform/pull/2792 on RHEL 9.6.

To drop local FS caches, I've used `sync && echo 3 | sudo tee /proc/sys/vm/drop_caches` .

Unfortunately this doesn't drop NFS server file caches, but interestingly, I couldn't see any caching effect on the NFS results for the **old** code anyway. Only new code is affected by the NFS server cache and produces sometimes NFS times comparable to the local FS performance.

Script below was used to create test directory structure with 50 folders a 5000 files each (250.000 files) used for performance tests above. Just create shell script and run it inside a project.

```
#!/bin/bash

TOTAL_FILES=250000
FILES_PER_DIR=5000
BASE_DIR="testdata"

mkdir -p "$BASE_DIR"

NUM_DIRS=$(( (TOTAL_FILES + FILES_PER_DIR - 1) / FILES_PER_DIR ))

for ((d=1; d<=NUM_DIRS; d++)); do
(
    dir="$BASE_DIR/dir_$(printf "%03d" $d)"
    mkdir -p "$dir"

    start=$(( (d-1)*FILES_PER_DIR + 1 ))
    end=$(( d*FILES_PER_DIR ))
    [ $end -gt $TOTAL_FILES ] && end=$TOTAL_FILES

    for ((i=start; i<=end; i++)); do
        : > "$dir/file_$(printf "%06d" $i).dat"
    done
) &
done

wait
echo "Done"
```


resources/bundles/org.eclipse.core.filesystem/natives/unix/fastlinux/fastlinuxfile.c
```diff
@@ -0,0 +1,578 @@
...
+		closedir(dir);
+	}
+	if (namesArray != NULL) {
+		(*env)->DeleteLocalRef(env, namesArray);
+	}
+	if (resultStatsArray != NULL) {
+		(*env)->DeleteLocalRef(env, resultStatsArray);
+	}
+	return result;
+}
```
HannesWell, line 385: It's too late for this PR, but I wonder if there is a real benefit in a native implementation of `NativeHandler.listDirectoryNames(String)` since the native implementation in the JDK of `File.list()` looks very similar to me:
https://github.com/openjdk/jdk/blob/6c48f4ed707bf0b15f9b6098de30db8aae6fa40f/src/java.base/unix/native/libjava/UnixFileSystem_md.c#L296-L358

But I'm not a native Linux expert and might have oversaw important details.

Maybe the benefit of this change is just in `listDirectoryAndGetFileInfos()` and `listDirectoryNames()` could just use the default Java implementation calling `File.list()`?

HannesWell: @iloveeclipse could you have a look at the previous questions? Thanks in advance.

iloveeclipse: Sorry, no time, we have internal release and must fix lot of issues before code freeze before Friday.

HannesWell: No problem. I don't think we're in a rush.

In the meantime I asked Copilot (it used Claude Sonnet 5) and it confirmed that only `listDirectoryAndGetFileInfos()` can have a significant gain on non-local file-systems.
But of course that answer can also be wrong, so if can check and if necessary follow-up on that, when you have time, that would be great.

<details><summary>Prompt</summary>
<p>

```
Check the PR @eclipse-platform/eclipse.platform/pull/2793 and explain why that native implementation is more performant than using existing Java APIs available in Java-25?
Additionally check which parts could be implement with existing Java APIs, preferable NIO, with comparable performance and which parts of the native implementation are significantly faster and should for example be ported to use the FFM API.
Mixing Java NIO and FFM native calls where suitable to get the maximal performance is fine.
```

</p>
</details> 

<details><summary>Copilot response</summary>
<p>

## Why the native implementation is faster

The performance win described in the PR (70,000–90,000 ms → 800–5,500 ms for 250,000 files over NFS) is **not** primarily about native code being "faster CPU-wise" than Java — it's about **eliminating round-trips per file**. That's the key architectural insight:

**The old path (still what plain Java NIO/`java.io` does today):**
For each file in a directory, Eclipse's resource refresh does the equivalent of:
1. `File.list()` / `Files.newDirectoryStream()` → one syscall sequence to enumerate names
2. For **each** entry: a separate `stat`/`lstat` call (via `File.exists()`, `isDirectory()`, `lastModified()`, etc., or `Files.readAttributes()`)

On local disks this is cheap because the metadata is cached in the local page cache/dentry cache. On **NFS**, every `stat` call is (or can be) a network round-trip to the NFS server. With 5,000 files per directory, that's 5,000 sequential round-trips — even at 1–2ms latency each, that's 5–10 seconds per directory, and it compounds across all directories, matching the observed 70–90 seconds for 250k files.

**The new native path (`listDirAndGetFileInfos` in `fastlinuxfile.c`):**
- Opens the directory once (`opendir`/`dirfd`)
- Uses `readdir()` to enumerate entries **and**, for each entry, immediately calls `fstatat(directoryFd, entry->d_name, ...)` **relative to the already-open directory file descriptor**, in the same JNI call, in a single tight loop, without returning to Java between each entry.
- Returns the **entire batch** of `LinuxStructStat` objects to Java in one JNI array transfer, instead of one JNI call per file.

The speedup comes from three combined effects:
1. **Batching stat calls into one JNI invocation** — avoids per-file JNI transition overhead (which is small but adds up at 5,000+ calls) and, more importantly, avoids Eclipse's higher-level code making one Java method call → one native call → one syscall per file.
2. **Using `fstatat(dirfd, name, ...)` instead of `stat(path)`** — resolves relative to an already-open directory descriptor rather than re-walking/re-resolving the full path from the root for every single file. On NFS this can avoid redundant lookup/attribute-cache revalidation per call.
3. **Because it's all in one native call, the NFS client can benefit from readahead/caching behavior within that single burst** of syscalls in ways that don't happen when Java-level code does one `stat` per iteration with intervening Java work (which is exactly what the PR author observed: the NFS server cache only kicked in for the *new* code, never for the old code).

This is fundamentally a **"vectorize/batch the syscalls" optimization**, not something achievable by just calling equivalent single-file NIO APIs — `Files.readAttributes()` is still one syscall per file with no batching primitive in java.nio.

## What HannesWell pointed out in review

Reviewer `HannesWell` raised exactly the right question inline (comment on `fastlinuxfile.c`):

> "I wonder if there is a real benefit in a native implementation of `NativeHandler.listDirectoryNames(String)` since the native implementation in the JDK of `File.list()` looks very similar... Maybe the benefit of this change is just in `listDirectoryAndGetFileInfos()` and `listDirectoryNames()` could just use the default Java implementation calling `File.list()`?"

`iloveeclipse` did not respond substantively (only noted lack of time before a release freeze), so this was never resolved in the PR.

## Breakdown: what can use plain NIO vs. what needs FFM/native

| Native function | Purpose | Can use existing Java API with comparable perf? | Recommendation |
|---|---|---|---|
| `listDir()` / `listDirectoryNames()` | Just enumerate directory entry names | **Yes** — `File.list()` / `Files.newDirectoryStream(Path)` already do a single `opendir`/`readdir` loop internally in the JDK's native `UnixFileSystem_md.c`. No batching benefit here since no per-entry `stat` is done. | Replace with `Files.list()`/`File.list()`. Matches HannesWell's review comment exactly — no native code needed. |
| `stat()` / `lstat()` (single file) | Get metadata for one path | **Yes** — `Files.readAttributes(path, PosixFileAttributes.class, LinkOption.NOFOLLOW_LINKS)` provides st_mode/size/mtime equivalents. | Use NIO `PosixFileAttributes`; no measurable difference for single-file calls. |
| `readlink()` | Read symlink target | **Yes** — `Files.readSymbolicLink(Path)`. | Use NIO. |
| `chmod()` | Set POSIX permissions | **Yes** — `Files.setPosixFilePermissions(Path, Set<PosixFilePermission>)`. | Use NIO. |
| **`listDirAndGetFileInfos()` (the batched directory+stat combo)** | Enumerate directory **and** stat every entry via one dirfd-relative loop, returned as one batch | **No** — there is no NIO/FFM equivalent of "give me a directory listing with per-entry `fstatat` already resolved in a single native call." `Files.newDirectoryStream` + per-entry `readAttributes()` reintroduces the N round-trips problem this PR exists to solve. | **This is the one piece that legitimately justifies a native/FFM implementation.** It should be the target for a Java 25 **FFM (Foreign Function & Memory API)** port: use `Linker`/`SymbolLookup` to bind directly to `opendir`, `readdir`, `dirfd`, `fstatat`, and lay out `struct dirent`/`struct stat` with `MemoryLayout`, batching the whole loop in Java-side FFM code without needing a compiled `.so` (eliminates the Makefile/native-build/architecture-packaging problems entirely — which is the current PR's actual complexity: a custom C library, JNI headers, per-arch fragments, a `Makefile` with a bug the reviewer already flagged pointing at the wrong 32-bit output path).

**Bottom line:** the described speedup is real and architecturally sound (batching `stat` calls under NFS latency), but it applies specifically to `listDirAndGetFileInfos`, not to the whole native surface added in this PR. `listDirectoryNames`/`stat`/`lstat`/`readlink`/`chmod` are unlikely to be meaningfully faster than JDK's own native implementations and add native-library maintenance burden (build system, `x86_64`-only packaging, the Makefile path bug flagged in review) for no real gain. A good "mixed NIO + FFM" design would drop the JNI library entirely and use plain NIO for the single-file operations, keeping only a small FFM-based `readdir`+`fstatat` batching routine for the directory-refresh hot path — which also solves the PR's stated goal of eventually removing native code once FFM/PDE debugging support (referenced in the PR body, blocked on eclipse-pde/eclipse.pde#2391) becomes viable.

</p>
</details>

PR #2794, opened 2026-07-03
Add breakpoint filtering support to Import/Export Breakpoint wizards

SougandhS:
Adds a filter text field to EmbeddedBreakpointsViewer to allow users to quickly locate breakpoints in the Import and Export Breakpoint wizards

image

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/importexport/breakpoints/EmbeddedBreakpointsViewer.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/importexport/breakpoints/EmbeddedBreakpointsViewer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2015 IBM Corporation and others.
+ * Copyright (c) 2005, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


MohananRahul (approved):
Looks Good

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/importexport/breakpoints/EmbeddedBreakpointsViewer.java
```diff
@@ -135,6 +150,28 @@ private void createControl(Composite parent, Object input, IStructuredSelection
...
+	private static class BreakpointPatternFilter extends ViewerFilter {
+
+		private String pattern = ""; //$NON-NLS-1$
+
+		public void setPattern(String pattern) {
+			this.pattern = pattern == null ? "" : pattern.toLowerCase(); //$NON-NLS-1$
+		}
+
+		@Override
+		public boolean select(Viewer viewer, Object parentElement, Object element) {
```
fedejeanne, line 162: Hi @SougandhS , I noticed while testing that the selection is set but not reset when changing the contents of the filter:

I tried this:
- I have 2 breakpoints

image

- I enter `4` and they are both selected

image

- I enter another `6` (resulting in `46`) and only one is selected

image

- I then click the `x` on the right side of the filter to clear the filter and they are both selected again

image

I understand the intent of the functionality (selecting exactly what matches the filter) but I personally find it a bit confusing. I prefer the kind of filter that behaves like the one in the launch configuration of an RCP application, in the _Plug-ins_ tab where the filter **does not change the selection** and you can still find and select stuff that are not called the same (or similar).

I would remove the automatic selection (this whole method) altogether and rely on the single selection and the _select/deselect all_ buttons.

SougandhS: Sure, I will update to not to change any selection on filtering

Issue #2801, opened 2026-07-06
Add per-builder build events to IResourceChangeEvent


vogella:
# Summary

Today the resources framework only fires workspace-level PRE_BUILD and POST_BUILD events.
Listeners learn that a build ran and get the aggregated delta, but they cannot see which builder ran, on which project, for which build kind, or how long each individual builder took.

This proposes two new event types, PRE_PROJECT_BUILD and POST_PROJECT_BUILD, fired once per builder execution and nested inside the surrounding workspace PRE_BUILD/POST_BUILD pair.

# Benefits:

- Per-builder profiling. Tools can attribute build time to a specific builder and project instead of a single opaque workspace total, making it easy to spot the slow builder in a large reactor.
- Monitoring and diagnostics tooling. The granularity enables build-monitor views and telemetry that show the real build sequence, which the aggregate events cannot support.

# Performance impact

Negligible by default.

When no listener is registered, dispatch is a single hasListenerFor bitmask check per event with no allocation, so normal builds are unaffected.
The per-builder path also skips resource-delta computation entirely, so even with a listener attached the added work is cheap metadata dispatch (project, builder id, build kind) rather than tree/delta building.

# Compatibility

The change is a binary-compatible API addition (new int constants plus a default method), so existing listeners are unaffected.

vogella:
Example view which would use these events: 

<img width="833" height="962" alt="Image" src="https://github.com/user-attachments/assets/42c4340b-f461-4835-8cfd-b6267300ecd0" />

merks:
I wonder if this is really such appropriate place for somehow gathering such information.  It seems to me something more appropriate for logging that gathers exactly measurement details one desired rather than needing to add a listener to an API that sudden produces a potentially thousands of events where it currently produces only two.  It seems nothing really needs the resource deltas for these events, which appears to be perhaps another indication that this is not really the appropriate API for this.

Perhaps something internal that builds a non-API internal data structure that might be used to display the information.  The information does seem interesting/useful...

vogella:
@merks I developed that a while ago for my patched Eclipse as I had multiple clients that had super long build cycles and no way to identify who is actually responsible for that. Some had "dead" builder still running and resulting in long build times others had custom builder which were not effective. I push my change and if you want you can have a look.

merks:
[...]

PR #2802, opened 2026-07-06
Make builder tracing work and switchable at runtime

vogella:
Per-builder tracing already existed as `org.eclipse.core.resources/perf/builders`, blaming the builder and using the project as context, but it never worked: `ResourceStats` kept the in-progress `PerformanceStats` in a single static field, so under parallel builds two builders overwrote each other's start time, and `startRun(context)` mutated the context of the shared instance, which is part of its `hashCode`, corrupting the key of an object already in the global stats map. Each start now returns a handle that the matching end consumes, and the elapsed time is recorded through `addRun`, which leaves the shared context alone.

Tracing enablement was also frozen at startup, which made it useless for tooling. `PerformanceStats.ENABLED` now follows the debug options and the resources trace flags are refreshed from the listener core.resources already had, so tracing can be switched on and off without a restart. `org.eclipse.core.runtime` now contributes a trace component like core.jobs and core.resources already do, so its `perf` options are reachable from the Tracing preference page instead of only from a launch configuration.

To validate: enable tracing for "Platform Core Runtime" (`perf`, `perf/success`) and "Platform Core Resources" (`perf/builders`), then trigger a build. A `perf/builders` threshold of 0 gathers statistics for every run without writing anything to the log, and the existing Event Spy view in `org.eclipse.core.tools` shows the resulting per-builder and per-project times.

`BuilderTracingTest` covers the runtime toggle, the per-builder and per-project attribution, and the listener notification, none of which was testable before enablement became dynamic.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
+++ b/resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.24.100.qualifier
+Bundle-Version: 3.25.100.qualifier
 Bundle-Activator: org.eclipse.core.resources.ResourcesPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/NotificationManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/NotificationManager.java
@@ -215,6 +215,20 @@ public class NotificationManager implements IManager, ILifecycleListener {
 	}
 
 	/**
+	 * Dispatches a per-builder build event (PRE_PROJECT_BUILD / POST_PROJECT_BUILD).
+	 * Unlike {@link #broadcastChanges}, this computes no resource delta and does not
+	 * touch {@code lastPostBuildTree}; the events are pure timing/metadata signals
+	 * nested inside the enclosing workspace build.
+	 */
+	public void broadcastProjectBuildEvent(IProject project, String builderName, int type, int buildKind) {
+		if (!listeners.hasListenerFor(type)) {
+			return;
+		}
+		ResourceChangeEvent event = new ResourceChangeEvent(project, type, buildKind, project, builderName);
+		notify(getListeners(), event, false);
+	}
+
+	/**
 	 * Helper method for the save participant lifecycle computation. */
 	public void broadcastChanges(IResourceChangeListener listener, int type, IResourceDelta delta) {
 		ResourceChangeListenerList.ListenerEntry[] entries;
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/BuildManager.java
@@ -49,6 +49,7 @@ import org.eclipse.core.resources.IIncrementalProjectBuilder2;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.IResourceStatus;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
@@ -262,6 +263,11 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 			currentTree = ((trigger == IncrementalProjectBuilder.FULL_BUILD) || clean) ? null : workspace.getElementTree();
 			int depth = -1;
 			ISchedulingRule rule = null;
+			IProject builderProject = builder.getProject();
+			ICommand builderCommand = builder.getCommand();
+			String builderName = builderCommand != null ? builderCommand.getBuilderName() : null;
+			workspace.broadcastProjectBuildEvent(builderProject, builderName,
+					IResourceChangeEvent.PRE_PROJECT_BUILD, trigger);
 			try {
 				//short-circuit if none of the projects this builder cares about have changed.
 				if (!needsBuild(currentBuilder, trigger)) {
@@ -274,9 +280,9 @@ public class BuildManager implements ICoreConstants, IManager, ILifecycleListene
 				String name = currentBuilder.getLabel();
 				String message;
 				if (name != null) {
-					message = NLS.bind(Messages.events_invoking_2, name, builder.getProject().getFullPath());
+					message = NLS.bind(Messages.events_invoking_2, name, builderProject.getFullPath());
 				} else {
-					message = NLS.bind(Messages.events_invoking_1, builder.getProject().getFullPath());
+					message = NLS.bind(Messages.events_invoking_1, builderProject.getFullPath());
 				}
 				monitor.subTask(message);
 				hookStartBuild(builder, trigger);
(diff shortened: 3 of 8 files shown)


resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.core.resources; singleton:=true
-Bundle-Version: 3.24.100.qualifier
+Bundle-Version: 3.25.100.qualifier
```
iloveeclipse: Bump seem to be wrong

iloveeclipse:
Have you considered providing tracing information instead? For the reasons mentioned in https://github.com/eclipse-platform/eclipse.platform/issues/2801 it should be sufficient. There is also "Event view" in runtime thac can be extended for that, or just create something similar based on that.

My main concern with the proposal here is that except for occasional performance checks it is not needed (it wasn't obviously requested by anyone since 20 years of resources API), however it will **always** generate lot more events (per builder & per project).

resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
vogella: New API, hence the bump.

vogella:
Thanks @iloveeclipse for the suggestion I look into tracing as an alternative. 

This might take a bit longer, as I soon will vanish into summer vacation and because I use this in a similar version already since a while in my custom aggregator build so my personal pressure is not that high to rework that approach.

resources/bundles/org.eclipse.core.resources/META-INF/MANIFEST.MF
iloveeclipse: The bump shouldn't include service segment!

merks: The point is that a bump of the minor version sets the service version back to zero so it should be 3.25.0 not 3.25.100

vogella: I think in the past we had the rule to use 100 in the end to allow downports. Is that obsolete? https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md#when-to-change-the-service-segment

vogella: Forget this comment, this is for the first change of the third segment. Not the other one, so 3.25.0 would be correct.

I leave this as it is for now as we anyhow want tracing and not new API.

vogella:
Implemented the suggestions from @merks and @iloveeclipse 

Please check again.

iloveeclipse:
> Please check again.

Please wait till next week, I'm getting old and need some rest on weekends :-)

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/PerformanceStats.java
```diff
@@ -342,10 +337,12 @@ public void addRun(long elapsed, String contextName) {
 		}
 		runCount++;
 		runningTime += elapsed;
-		if (elapsed > getThreshold(event)) {
+		// the threshold is the shortest duration that is still reported, so that a
+		// threshold of 0 reports every occurrence
+		if (elapsed >= getThreshold(event)) {
 			PerformanceStatsProcessor.failed(createFailureStats(contextName, elapsed), blamePluginId, elapsed);
 		}
```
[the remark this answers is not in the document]
vogella: Correct, `logFailure` wrote a warning with a stack trace per run. `PerformanceStatsProcessor.failed` now takes a flag and skips the performance log when the threshold is 0, while listeners are still notified about every occurrence. Commit message adjusted accordingly.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/BuilderTracingTest.java
```diff
@@ -0,0 +1,172 @@
...
+				} else {
+					debugOptions.setOption(option, value);
+				}
+			});
+		} else {
+			// disabling debug discards all options that were set
+			debugOptions.setDebugEnabled(false);
+		}
+		waitUntil(() -> !ResourceStats.isTracingBuilders(), "builder tracing was not switched off again");
+		PerformanceStats.clear();
```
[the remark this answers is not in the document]
vogella: Done, `tearDown` now waits for `ResourceStats.isTracingBuilders()` to return to the value captured in `setUp`.

vogella:
> > Please check again.
> 
> Please wait till next week, I'm getting old and need some rest on weekends :-)

Its been now a month. :-) Any new feedback? 

It is based on tracing as requested by you and @merks The tracing can be turned on and off at runtime. No additional event.

iloveeclipse:
> The tracing can be turned on and off at runtime.

Please provide steps to validate the PR. Which options need to be set? `org.eclipse.core.resources/perf/builders=0` or more / other option or value?

iloveeclipse:
At least for me I see builders info appearing in only after setting these tracing options:
```
org.eclipse.core.runtime/perf=true
org.eclipse.core.runtime/perf/success=true

org.eclipse.core.resources/perf/builders=0
```
Please confirm that this is the right way to validate.
This is what I see
image

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
```diff
@@ -772,9 +775,20 @@ private void startServices() {
 		customPreferencesService = context.registerService(IProductPreferencesService.class, new ProductPreferencesService(), new Hashtable<>());
 
 		legacyPreferencesService = context.registerService(ILegacyPreferences.class, new InitLegacyPreferences(), new Hashtable<>());
+
+		Hashtable<String, String> debugProperties = new Hashtable<>(2);
+		debugProperties.put(DebugOptions.LISTENER_SYMBOLICNAME, Platform.PI_RUNTIME);
```
iloveeclipse, line 780: I miss contribution to the extension point so that users can enable runtime performance tracing at runtime, something like this:
```
<extension
         point="org.eclipse.ui.trace.traceComponents">
      <component
            id="org.eclipse.core.runtime.traceComponent" label="%traceComponentName">
            <bundle name="org.eclipse.core.runtime" />
      </component>
   </extension>
```

iloveeclipse:
In general looks good, assuming I've used it as expected, see my comments above.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
vogella: Components can disable/ activate it via:

````
	private void disableTracing() {
		ServiceCaller.callOnce(getClass(), DebugOptions.class, options -> {
			if (!debugWasEnabled) {
				// disabling debug discards all options, including the ones set above
				options.setDebugEnabled(false);
				return;
			}
			replacedOptions.forEach((option, value) -> {
				if (value == null) {
					options.removeOption(option);
				} else {
					options.setOption(option, value);
				}
			});
		});
	}
````

Enable see here https://github.com/eclipse-platform/eclipse.platform/pull/2896/changes#diff-a6abe37830234d56ff13e2c61b8e7f026fd3ddc40adaa29bbba2e44e79471c2eR111

No need for plugin.xml. 

The above coding is copied from my example view https://github.com/eclipse-platform/eclipse.platform/pull/2896

File: runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/PerformanceStats.java [hunk 5]
```diff
@@ -342,10 +337,13 @@ public void addRun(long elapsed, String contextName) {
...
 		runCount++;
 		runningTime += elapsed;
-		if (elapsed > getThreshold(event)) {
-			PerformanceStatsProcessor.failed(createFailureStats(contextName, elapsed), blamePluginId, elapsed);
+		// the threshold is the shortest duration that is still reported, so that a
+		// threshold of 0 reports every occurrence to listeners without logging it
+		long threshold = getThreshold(event);
+		if (elapsed >= threshold) {
+			PerformanceStatsProcessor.failed(createFailureStats(contextName, elapsed), blamePluginId, elapsed, threshold > 0);
```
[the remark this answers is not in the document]
vogella: "Applied"

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/ResourceStats.java
```diff
@@ -26,70 +26,113 @@
...
-		currentStats = null;
+		try {
+			return Integer.parseInt(option.trim());
+		} catch (NumberFormatException e) {
+			return 0;
+		}
+	}
+
+	private static Run start(String event, Object blame, String context) {
+		return new Run(PerformanceStats.getStats(event, blame), context, System.currentTimeMillis());
```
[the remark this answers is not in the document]
vogella: "Applied"

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/events/NotificationManager.java
```diff
@@ -356,7 +352,7 @@ public void run() throws Exception {
 
 	public void removeListener(IResourceChangeListener listener) {
 		listeners.remove(listener);
-		if (ResourceStats.TRACE_LISTENERS) {
+		if (ResourceStats.isTracingListeners()) {
 			ResourceStats.listenerRemoved(listener);
 		}
```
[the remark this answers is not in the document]
vogella: "Applied"

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/internal/builders/BuilderTracingTest.java
```diff
@@ -0,0 +1,175 @@
...
+		setAutoBuilding(false);
+		project.create(createTestMonitor());
+		project.open(createTestMonitor());
+		updateProjectDescription(project).addingCommand(SortBuilder.BUILDER_NAME).withTestBuilderId(name).apply();
+		return project;
+	}
+
+	@Test
+	public void testTracingFollowsDebugOptionsAtRuntime() throws InterruptedException {
+		assertFalse(PerformanceStats.isEnabled(EVENT_BUILDERS), "tracing should start out disabled");
```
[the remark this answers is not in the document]
vogella: "Applied"

vogella:
@iloveeclipse let me know if you have more questions, I think I answered all.

runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/internal/runtime/InternalPlatform.java
iloveeclipse: > Components can disable/ activate it via:

I'm talking about enabling tracing by **users**, not code. It should be possible to activate it by users, like with other bundles which provide tracing?

iloveeclipse:
I miss tracing activation possibility by users, see above.

vogella:
These are existing trace options, which are part of the launch configuration.

Screenshot From 2026-09-04 16-41-29

Once this is merged, I plan to promote my build monitor from draft to real PR to make it easier for the user to see the build time of the individual builders. Another way for the user to activate this is not planned by me.

iloveeclipse:
User currently has no way to change **runtime** bundle, trwcing options, see my comment above which. Please simply add the contribution to plugin xml, it is fully unrelated to additional bundle or view or whatever.

vogella:
Why should this tracing option be special in how the user can activate it? We have tons of tracing options.

iloveeclipse:
> Why should this tracing option be special in how the user can activate it? We have tons of tracing options.

Exact. It shouldn't be special, but it **is special** today because it is **not exposed to user**. You can only set it from debugger, but not from IDE, unlike many other tracing options. Just try to find it in your IDE.

vogella:
Added, thanks.

org.eclipse.core.runtime now contributes an org.eclipse.ui.trace.traceComponents component, so its perf and perf/success options can be set from Preferences > General > Tracing instead of only from a launch configuration.

vogella:
Any additional feedback @iloveeclipse? Planning to merge early next week.

iloveeclipse:
I have no time for a real review, I've briefly scanned through the changes, looks OK.

Issue #2638, opened 2026-04-30
Race condition in ConsoleColorProvider.connect()


trancexpress:
Seen in: https://github.com/eclipse-platform/eclipse.platform/issues/596

Reproduction steps are similar to: https://github.com/eclipse-platform/eclipse.platform/issues/2637

1. Set a breakpoint at: `org.eclipse.debug.ui.console.ConsoleColorProvider.connect(IProcess, IConsole)`
2. Debug Eclipse, make sure the Console view is open.
3. In the debuggee, create a snippet:
```
public class Test3 {

	public static void main(String[] args) throws InterruptedException {
		Thread.sleep(60_000);
	}
}
```
4. Launch this snippet twice.
5. Observe two threads suspended at the breakpoint.
6. Step over in the 1st thread until `if (streamsProxy != null) {`, do the same for the 2nd thread.
7. Continue stepping until this state is reached:
```
"Worker-7: Creating console for /usr/lib/jvm/java-21-openjdk/bin/java (Apr 30, 2026, 11:56:11 AM) [pid: 70092]" #79 [69171] prio=5 os_prio=0 cpu=72.35ms elapsed=122.76s tid=0x00007fea50003e50 nid=69171 runnable  [0x00007feaecea0000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.debug.internal.ui.views.console.ProcessConsole.connect(ProcessConsole.java:797)
        - locked <0x0000000720e4cde8> (a org.eclipse.debug.internal.core.OutputStreamMonitor)
        at org.eclipse.debug.internal.ui.views.console.ProcessConsole.connect(ProcessConsole.java:760)
        at org.eclipse.debug.ui.console.ConsoleColorProvider.connect(ConsoleColorProvider.java:43)
        at org.eclipse.debug.internal.ui.views.console.ProcessConsole.<init>(ProcessConsole.java:274)
        at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager$ConsoleCreation.run(ProcessConsoleManager.java:74)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
```
[...]

PR #2805, opened 2026-07-07
Synchronize lazy initialization in ProcessConsoleManager.getColorProvider

### LINKED ISSUE #2637: Race condition when initializing ProcessConsoleManager.fColorProviders
Seen in the latest fail for https://github.com/eclipse-platform/eclipse.platform/issues/596, see: https://github.com/eclipse-platform/eclipse.platform/issues/596#issuecomment-4350678790

To reproduce:

1. Set a breakpoint at: `org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.getColorProvider(String)`
2. Debug Eclipse, make sure the Console view is open.
3. In the debuggee, create a snippet:
```
public class Test3 {

	public static void main(String[] args) throws InterruptedException {
		Thread.sleep(60_000);
	}
}
```
4. Launch this snippet twice.
5. Observe two threads suspended at the breakpoint:
```
"Worker-2: Creating console for /usr/lib/jvm/java-21-openjdk/bin/java (Apr 30, 2026, 11:28:57 AM) [pid: 53966]" #74 [51843] prio=5 os_prio=0 cpu=31.67ms elapsed=118.72s tid=0x00007f8f9c0068a0 nid=51843 at breakpoint [0x00007f8fc99fb000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.getColorProvider(ProcessConsoleManager.java:262)
        at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager$ConsoleCreation.run(ProcessConsoleManager.java:72)
```
```
"Worker-5: Creating console for /usr/lib/jvm/java-21-openjdk/bin/java (Apr 30, 2026, 11:28:51 AM) [pid: 53920]" #77 [51846] prio=5 os_prio=0 cpu=67.00ms elapsed=118.72s tid=0x00007f8fc00041b0 nid=51846 at breakpoint [0x00007f8fc9dfd000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.getColorProvider(ProcessConsoleManager.java:262)
        at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager$ConsoleCreation.run(ProcessConsoleManager.java:72)
```
6. Both are at the same object, e.g.: `org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager@2b14b28c`

See also screen recording from the reproduction:

https://github.com/user-attachments/assets/b6bd1501-6ef3-437a-93bc-b3b7f4a69340

vogella:
## Problem

`ConsoleCreation` jobs run in parallel but share the singleton `ProcessConsoleManager`. `getColorProvider(String)` lazily initialized `fColorProviders` (a plain `HashMap`) and the default `ConsoleColorProvider` **without synchronization**, so concurrent first calls could corrupt the map via parallel `put()` or create two different default providers (#2637).

## Fix

Initialize both lazily in a **lock-free** way, so no lock is held across the extension-registry calls (which could deadlock with 3rd-party code):

- `fColorProviders` is `volatile` and built into a local map before being published; a concurrent duplicate build is harmless because each thread fills its own local map.
- The default provider is stored in an `AtomicReference` and created via `compareAndSet`, guaranteeing a single shared instance.

Intentionally isolated to the init race (#2637); the shared stateful default provider (#2638 / #596) is left for a follow-up.

Fixes #2637

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/console/ProcessConsoleManager.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/console/ProcessConsoleManager.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IConfigurationElement;
@@ -105,15 +106,16 @@ public class ProcessConsoleManager implements ILaunchListener {
 	}
 
 	/**
-	 * Console document content provider extensions, keyed by extension id
+	 * Console document content provider extensions, keyed by extension id.
+	 * Published once; never mutated after initialization.
 	 */
-	private Map<String, IConfigurationElement> fColorProviders;
+	private volatile Map<String, IConfigurationElement> fColorProviders;
 
 	/**
 	 * The default color provider. Used if no color provider is contributed
 	 * for the given process type.
 	 */
-	private IConsoleColorProvider fDefaultColorProvider;
+	private final AtomicReference<IConsoleColorProvider> fDefaultColorProvider = new AtomicReference<>();
 
 	/**
 	 * Console line trackers; keyed by process type to list of trackers (1:N)
@@ -129,6 +131,7 @@ public class ProcessConsoleManager implements ILaunchListener {
 	 * Lock for fLineTrackers
 	 */
 	private final Object fLineTrackersLock = new Object();
+
 	/**
 	 * @see ILaunchListener#launchRemoved(ILaunch)
 	 */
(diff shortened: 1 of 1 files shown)


debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/console/ProcessConsoleManager.java
```diff
@@ -259,14 +264,20 @@ public void shutdown() {
 	 * @return IConsoleColorProvider
 	 */
 	public IConsoleColorProvider getColorProvider(String type) {
-		if (fColorProviders == null) {
-			fColorProviders = new HashMap<>();
-			IExtensionPoint extensionPoint= Platform.getExtensionRegistry().getExtensionPoint(DebugUIPlugin.getUniqueIdentifier(), IDebugUIConstants.EXTENSION_POINT_CONSOLE_COLOR_PROVIDERS);
-			for (IConfigurationElement extension : extensionPoint.getConfigurationElements()) {
-				fColorProviders.put(extension.getAttribute("processType"), extension); //$NON-NLS-1$
+		Map<String, IConfigurationElement> colorProviders;
+		synchronized (fColorProvidersLock) {
```
iloveeclipse: I would try to avoid adding a lock over extension class init / calls into 3rd party code, because that code might want to do something with console related code and also use locks, introducing potential deadlock case.

vogella: Agree, I changed it.

iloveeclipse:
Closing in favor of https://github.com/eclipse-platform/eclipse.platform/pull/2810. See https://github.com/eclipse-platform/eclipse.platform/issues/2638#issuecomment-4936831912 why.

PR #2806, opened 2026-07-07
Fix concurrent connect in ConsoleColorProvider

vogella:
## Problem

The default `ConsoleColorProvider` is shared across `ConsoleCreation` jobs that run in parallel. `connect(IProcess, IConsole)` stored both arguments in the `fProcess`/`fConsole` fields and then wired up via those fields, so concurrent calls could interleave and connect a console to another process' streams proxy. This caused an `ArrayIndexOutOfBoundsException` on the wrong console's stream listeners and consoles not being connected (#2638).

## Fix

Wire up using the method arguments instead of the shared fields.

Independent of the initialization-race fix (#2637); touches only `ConsoleColorProvider`.

Refs #2638


iloveeclipse:
Closing in favor of https://github.com/eclipse-platform/eclipse.platform/pull/2810. See https://github.com/eclipse-platform/eclipse.platform/issues/2638#issuecomment-4936831912 why.

PR #2808, opened 2026-07-07
Fix ReevaluateExpression enabled state on session termination

SougandhS:
selectionChanged() now checks whether the debug target is terminated or disconnected before enabling the action, preventing stale re-evaluate actions from being invocable after the debug session has ended.

before 
Screenshot 2026-07-07 at 5 24 40 PM

After
Screenshot 2026-07-07 at 5 23 48 PM


SougandhS:
Noticed while working on https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/976

SougandhS:
Thank you Ed

PR #2816, opened 2026-07-18
Bold Debug view tab on new launch when not focused

SougandhS:
When a new launch is added while the Debug view is not the active view, make the Debug view tab label bold to provide a subtle indication that a new launch has started/completed, allowing the user to notice it without manually switching to the Debug view.
debug

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/launch/LaunchView.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/views/launch/LaunchView.java
@@ -38,6 +38,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.debug.core.DebugPlugin;
 import org.eclipse.debug.core.ILaunch;
+import org.eclipse.debug.core.ILaunchListener;
 import org.eclipse.debug.core.commands.IRestartHandler;
 import org.eclipse.debug.core.model.IDebugTarget;
 import org.eclipse.debug.core.model.IProcess;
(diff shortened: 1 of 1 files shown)


SougandhS:
Hi @merks, can you check this PR too ?

merks:
It's simple.  I didn't even know of this feature of the progress service!

I wonder if one might want to be "notified" when a launch terminates as well?

@iloveeclipse 

Do you have an opinion on this?

SougandhS:
a small correction "started/completed," I meant launched and completed in a second (like launching sample prgms for testing).. so it gets only notified when the launch is added..

SougandhS:
> I wonder if one might want to be "notified" when a launch terminates as well?

Sounds great 👍

iloveeclipse:
> I wonder if one might want to be "notified" when a launch terminates as well?

If launches are automatically removed on termination (preference), there will be nothing to show in the Debug view, so no reason to update it's title.

The change itself seem to be OK, I just thinking that all the external builders which can be setup on projects (also in SDK), would also trigger Debug view updates, flashing the title for nothing. Therefore I would probably tend to let the Debug view to only listen for start of the "debug" launches, but then the next question would be - which benefit it would give to user? If a breakpoint is hit, Debug view will show up anyway.

So I'm kind of undecided why should we add more code here in general, how useful is it?

SougandhS:
> which benefit it would give to user? If a breakpoint is hit, Debug view will show up anyway. So I'm kind of undecided why should we add more code here in general, how useful is it?

One of the personal use cases I find is that when launches inside a group start after a delay or after a console condition is reached, this lets me know a launch has actually started, without needing to watch the Console or switch views manually.

iloveeclipse (approved):

iloveeclipse:
> One of the personal use cases I find ...

OK, let's merge then. If the new behavior will be annoying, we can revert.

PR #2817, opened 2026-07-18
Improve UX for file selection in the Common tab

SougandhS:
Set focus to the shared location text field when the "Shared file" option is selected in the launch configuration dialog. This allows users to immediately enter or browse for the shared configuration location without requiring an additional click.

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/ui/CommonTab.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -232,7 +232,12 @@ public class CommonTab extends AbstractLaunchConfigurationTab {
 		gd.horizontalSpan = 3;
 		fLocalRadioButton.setLayoutData(gd);
 		fSharedRadioButton = createRadioButton(comp, LaunchConfigurationsMessages.CommonTab_S_hared_4);
-		fSharedRadioButton.addSelectionListener(widgetSelectedAdapter(e -> handleSharedRadioButtonSelected()));
+		fSharedRadioButton.addSelectionListener(widgetSelectedAdapter(e -> {
+			handleSharedRadioButtonSelected();
+			if (fSharedRadioButton.getSelection()) {
+				fSharedLocationText.setFocus();
+			}
+		}));
 
 		fSharedLocationText = SWTFactory.createSingleText(comp, 1);
 		fSharedLocationText.getAccessible().addAccessibleListener(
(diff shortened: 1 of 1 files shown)


merks:
I don't think this can be quite right.  The enable argument might be false and still you would call set focus.  I also wonder how this makes sense when the common tab is not visible (is not the active tab) or what the impact of this is while using keyboard navigation to navigate all the launch configurations...

SougandhS:
>I don't think this can be quite right. The enable argument might be false and still you would call set focus

weird, I remember I added enabled condition and pushed that change.. let me push that again.

SougandhS:
> I also wonder how this makes sense when the common tab is not visible (is not the active tab) or what the impact of this is while using keyboard navigation to navigate all the launch configurations...

I'll check this

SougandhS:
>I also wonder how this makes sense when the common tab is not visible (is not the active tab) or what the impact of this is while using keyboard navigation to navigate all the launch configurations...

Handled this to activate only during selections.. 👍

merks:
That seems more sensible.  I wonder though, how does this affect keyboard users?  Maybe in a good way?  😀

SougandhS:
> how does this affect keyboard users? Maybe in a good way?

On mac I see no benefits with keyboards, using `TAB` I only get access to Text fields and menu tabs, not on any checkboxes and radios..

merks:
On Windows the up and down arrow keys change the radio button selection and also cycle to the browse button.  With this change, the focus moves to the entry field as soon as you navigate to the second radio button via down arrow whereas previous one could navigate to the browse button which is probably the more likely next thing you want to do...

Is there anyone on your team who can test the behavior on Windows?  I'm really not 100% sure this is an improvement.

Maybe @iloveeclipse  or @HeikoKlare  have an opinion?

iloveeclipse:
I've tested on Linux and it looks OK. So LGTM from my side.

HeikoKlare:
> On Windows the up and down arrow keys change the radio button selection and also cycle to the browse button. With this change, the focus moves to the entry field as soon as you navigate to the second radio button via down arrow whereas previous one could navigate to the browse button which is probably the more likely next thing you want to do...
> 
> Is there anyone on your team who can test the behavior on Windows? I'm really not 100% sure this is an improvement.

In my opinion, this introduces a new UX concept at a single place only. There are other places with the same pattern (selecting a radio option enables a text field in which you can define a location), e.g.:
image

In my opinion, establishing this concpet at a single place will lead to more confusion than improvement. The average user who does not heavily use this input field will not know and value that at this specific field the user interaction behaves different than anywhere else. They may rather be confused why the interaction is inconsistent to everywhere else.

So I would generally vote against this kind of behavior changes without a holistic (and consistently applied) UX concept for such a change. Though I think this specific place will not be that problematic due to being rarely used (which I why I would not block this kind of change), still making Eclipse UX more inconsistent. @BeckerWdf wdyt?

BeckerWdf:
I agree with Heiko. Consistency is the prio 1 topic in UX. Changing interaction patterns only in single places (and not in a holistic way) does worsen the UX of the overall product.

SougandhS:
Thank you everyone for the feedback , that's a fair point about consistency 👍 
So I'll close this PR for now. Happy to revisit if there's ever a holistic push to standardize this pattern across the UI.

PR #2825, opened 2026-07-25
Prompt to save changes when closing Organize Favorites dialog

SougandhS:
Prompt users to save changes before closing the Organize Favorites dialog when the favorites list has been modified. This prevents accidentally losing changes made to the favorites list when the dialog is closed without pressing OK.

Screenshot 2026-07-25 at 6 36 34 PM

### DIFF
--- a/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/FavoritesDialog.java
+++ b/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/FavoritesDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -15,6 +15,7 @@ package org.eclipse.debug.internal.ui.launchConfigurations;
 
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


iloveeclipse:
Why not simply save on OK, without dialog?

SougandhS:
> Why not simply save on OK, without dialog?

it saves on clicking `Ok` but on accidental `ESC` press or `Cancel` click it simply closes without saving, so added for those cases.

SougandhS:
Hi @iloveeclipse, 
build looks good, can this be merged ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/FavoritesDialog.java
```diff
@@ -155,6 +159,57 @@ protected void handleAddConfigButtonSelected() {
...
+
+		int option = dialog.open();
+		if (option == Window.OK) {
+			saveFavorites();
+			return true;
+		}
+		if (option == Window.CANCEL) {
+			return true;
+		}
+		return false;
```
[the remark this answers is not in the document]
SougandhS: Done

SougandhS:
Hi @iloveeclipse, 
if no other changes are required, can this be merged ?

debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/launchConfigurations/FavoritesDialog.java
```diff
@@ -155,6 +159,58 @@ protected void handleAddConfigButtonSelected() {
...
+				new String[] { LaunchConfigurationsMessages.FavoritesDialogPromptOnCloseSaveButton,
+						IDialogConstants.NO_LABEL, IDialogConstants.CANCEL_LABEL },
+				0);
+
+		int option = dialog.open();
+		switch (option) {
+		case Window.OK:
+			saveFavorites();
+			return true;
+		case 1:
```
iloveeclipse: `case 1:` what is `1`? Can you please use constant, I assume it is `org.eclipse.jface.window.Window.CANCEL`?

**If** this is `CANCEL`, why do we return `true` here? Shouldn't that be `false`, which should NOT close the dialog?

iloveeclipse: In the debugger one can see that the return code is coming from `org.eclipse.jface.dialogs.MessageDialog.buttonPressed(int)`. So it is **NOT** any of the `org.eclipse.jface.window.Window` constants but the button index, and above case is also misleading.

Please add 0,1,2 cases and comments on each which button it is supposed to be

SougandhS: Added comments for numeric cases 👍

iloveeclipse (approved):

iloveeclipse:
Thanks.

PR #2827, opened 2026-07-27
Do not show error dialog when validateEdit fails for Unified Diff

tobiasmelcher:
Unified Diff cannot be applied when the current editor content is read-only and not modifiable. In that case validateEdit fails and open() returns CANCEL_STATUS, so the caller falls back to opening the existing two-way diff editor. Showing an error dialog on top of that fallback was more irritating than helpful, so it is removed.


iloveeclipse (approved):
LGTM.

Issue #2795, opened 2026-07-03
Opening a compare editor blocks the UI and reads the input contents repeatedly


vogella:
Opening a compare editor (for example from the Git History or Git Staging view) takes long, even for small files. The same applies to the new unified diff display.

Tracing the open path in org.eclipse.compare shows the delay is structural, not in the diff algorithm (see also #2575, where a DocLineComparator micro-optimization did not help):

- The input contents are read repeatedly during a single open. A new counting test measures 15 `getContents()` calls per side for one editor open. With EGit inputs each call re-inflates the blob since `StorageTypedElement` only buffers the storage handle.
- After the "Comparing..." job, everything else runs on the UI thread inside `createContents`. Timing tests confirm time-to-text equals time-to-diff, so nothing is visible until the diff completes (about 190 ms for 5000 lines with 100 changes, about 310 ms for 50000 lines, measured on Linux/Xvfb).
- With the unified diff preference enabled, `CompareUIPlugin.canShowInUnifiedDiff` runs the full `prepareInput` synchronously on the UI thread with a `NullProgressMonitor`, builds a throwaway `TextMergeViewer` on an invisible `Shell`, and `UnifiedDiffManager.open` then computes a second diff inline on the UI thread.

I plan to address this incrementally: measurable tests first (deterministic call counters plus repeated-run timing medians, since the old `org.eclipse.test.performance`-based tests assert nothing without the performance database), then removing the redundant lookups and reads, then moving the unified path onto the same background-job pattern as the classic editor, and finally making the initial diff asynchronous so the text shows immediately.

See also #121 for a related problem after the compare editor opens.

PR #2834, opened 2026-07-29
Open the unified diff without blocking the UI thread

vogella:
With the unified diff preference enabled, opening a compare editor ran the full `prepareInput` synchronously on the UI thread with a `NullProgressMonitor`, and read the whole opposite side of the comparison there as well. The preparation now runs in a cancelable job, the same way the classic compare editor already works; only opening the text editor and applying the diff stay on the UI thread. Progress and cancellation are visible to the user instead of a frozen workbench, and when the input does not qualify the classic editor reuses the already prepared result rather than comparing a second time.

Obtaining the token comparator and the ignore-whitespace contributor no longer builds a throwaway `TextMergeViewer` on an invisible shell. The unified diff already falls back to exactly what the platform text merge viewer contributes, so that viewer is only created when a custom merge viewer is registered for the content type.

The two tests that documented the old on-the-UI-thread behavior now assert the opposite.


tobiasmelcher (approved):
thanks a lot Lars.

PR #2836, opened 2026-07-29
Cache the syntax highlighting of a unified diff mining

vogella:
Every draw of a unified diff code mining whose drawing cache had been dropped recomputed the syntax highlighting of the overlaid text, and that is far from cheap: `SourceViewer.computeStyleRanges` reconnects the document partitioner to a temporary document holding the whole file prefix plus the diff text, and reconnects it to the real document afterwards, so each call re-partitions the file twice. With many changes in a large file, a single font change or editor zoom step made every visible mining pay that again.

The style ranges depend on the document content alone and not on the font, so they are now cached per mining and survive the font driven cache invalidation. They are dropped when the mining is disposed, and the mining itself is recreated whenever the document changes, so the cache cannot go stale. The ranges are never mutated in place either: the one place that needs a font clones the range first.

This is the drawing path, so it is not covered by a unit test; the existing unified diff tests still pass.


tobiasmelcher (commented):
sorry, too late. I was on vacation last week. Thanks for providing the performance improvement.

PR #2838, opened 2026-07-30
Make the compare editor font zoom shortcut work off macOS

vogella:
The key listener that zooms the compare editor text checked `SWT.COMMAND`, a modifier that is only ever set on macOS, so Ctrl+= and Ctrl+- did nothing on Linux and Windows while the pinch gesture worked everywhere. It now checks `SWT.MOD1`, which resolves to Command on macOS and Ctrl elsewhere, and it accepts the keypad plus and minus keys as well.

Two things noticed in the same method: the ancestor pane kept its old font in a three-way compare, and every zoom step re-ran the whole comparison even though the differences do not depend on the font. Only the line geometry does, so the vertical scroll bar and the presentation are refreshed instead, which also makes zooming a large comparison responsive.


vogella:
Related to commit 1f6693ee04

vogella:
@SougandhS can you review?

SougandhS (approved):
LTGM 👍

PR #2857, opened 2026-08-07
Remove unused org.eclipse.team.internal.core.streams

### LINKED ISSUE #2856: Remove unused org.eclipse.team.internal.core.streams
Every caller of org.eclipse.team.internal.core.streams lived in org.eclipse.team.cvs.core or org.eclipse.team.cvs.ssh2, and they all disappeared in 7592b18cbb "Retire CVS from master (#10)", 2022-03-31.

I suggest we retire it.

vogella:
Every caller of `org.eclipse.team.internal.core.streams` lived in `org.eclipse.team.cvs.core` or `org.eclipse.team.cvs.ssh2`, and all of them disappeared with 7592b18cbb "Retire CVS from master" in 2022. The package had been moved into `org.eclipse.team.core` by 68ca7959a2 precisely so the CVS core bundle and the ssh fragments could share one copy of the stream filters, so serving CVS was its only reason to live here, and no non-CVS caller ever existed outside the package's own test.

The eight classes form a closed cluster, so the whole package goes: the five orphaned directly by the CVS removal, `PollingInputStream` which only `SizeConstrainedInputStream` used, and the two line-delimiter filters whose only remaining client was `StreamTests`. That test is removed with them, together with its entry in `AllTeamTests`, the `x-friends` export and five message keys nothing references any more. This drops about 1,200 lines.

Removing the export is not an API change: the package was exported only via `x-friends`, naming the three deleted CVS bundles plus `org.eclipse.team.ui` and `org.eclipse.team.tests.core`, and neither of the latter two imports it.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2856


iloveeclipse:
Ideally in 4.42, removing just before 4.41 M3 is too risky for downstream.

PR #2859, opened 2026-08-08
restore current working directory in terminal even if directory selected

danthe1st:
When creating a new terminal with a folder selected (by pressing Ctrl+Alt+T) with terminals to be restored (i.e. #2850 is enabled and there are terminals from a previous session), Eclipse currently resets the working directory of the folders to be restored to the selection (and the new terminal is also using that selection).

This PR changes this by only using the selection for new terminals. As the new terminal was previously added at the beginning of the `CTabFolder`/list of terminals (because the order of the terminals being opened was wrong), I also made sure that the newly opened terminal is at the very right where it should be.

I hope nobody [relies on this broken behaviour](https://xkcd.com/1172/) (otherwise I could add a preference for the first fix but I don't really see why anyone would want that).

### How to test

- Start Eclipse
- Open one or more terminals in arbitrary directories
- Navigate to another view such that the terminal view is no longer visible
- Restart Eclipse
- Without opening the terminal view manually, open a terminal using a selected folder (`Show In` > `Local Terminal` or similar)
- The new terminal with the new directory should be opened at the end. The old terminals should "keep" their working directories from before the restart.

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
+++ b/terminal/bundles/org.eclipse.terminal.connector.local/src/org/eclipse/terminal/connector/local/launcher/LocalLauncherDelegate.java
@@ -105,8 +105,10 @@ public class LocalLauncherDelegate extends AbstractLauncherDelegate {
 			properties.put(ITerminalsConnectorConstants.PROP_FORCE_NEW, Boolean.TRUE);
 		}
 
+		boolean hadWorkingDirectoryBefore = properties
+				.containsKey(ITerminalsConnectorConstants.PROP_PROCESS_WORKING_DIR);
 		// Initialize the local terminal working directory.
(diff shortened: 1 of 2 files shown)


iloveeclipse (approved):

iloveeclipse:
Test failure is unrelated, code looks good & I've verified the problem & the solution. Thanks!

iloveeclipse:
@danthe1st : not changed by the patch but related to the patch: it is possible that the remembered, previously existing directories are deleted before Terminal view restores terminals. At least on Linux I see that the terminal restores at the current working directory from which Eclipse was started. I don't know if that is what users expect, probably it would make sense to not restore terminal if the working directory is gone.

danthe1st:
>  I don't know if that is what users expect, probably it would make sense to not restore terminal if the working directory is gone.

One alternative might be to make sure to restore to the `Initial Working Directory` from the preferences (and possibly use the selection if it exists).
image


~~I also don't know what the best approach is here. I think I will create a discussion for this.~~ I created a discussion for this with #2871

PR #2864, opened 2026-08-10
Keep the unified diff scrollable when a file is emptied

vogella:
When the whole content of a file is removed, the unified diff shows every deleted line as a code mining on an empty document. That mining was created as a document footer mining, and a footer mining reserves no height in the text widget, so the viewer had nothing to scroll over and the deleted content was unreachable as soon as it exceeded the visible area.

A diff offset that starts a line now gets a line header mining instead, which reserves its height as the vertical indent of its line. This covers an emptied document as well as every document ending with a line delimiter; only content behind a last line without a trailing delimiter still needs a footer mining. The mining position is also clamped to the document, because the annotation model silently drops a position that reaches beyond the end.

Covered by a new test in `UnifiedDiffManagerTest` that empties the editor, opens an overlay diff of 200 removed lines and asserts the reserved space. It fails without the change.


HeikoKlare:
In case someone else stumbles upon this: from my understanding today's I-Build failure of the here-introduced `org.eclipse.team.tests.ui.UnifiedDiffManagerTest#testRemovingTheWholeContentStaysScrollable` should already be fixed by:
- https://github.com/eclipse-platform/eclipse.platform/pull/2867

PR #2866, opened 2026-08-10
Provide bundle groups without scanning the installation

### LINKED ISSUE #2863: Provide IBundleGroupProvider from a dedicated bundle instead of org.eclipse.update.configurator
IPlatformConfiguration and IPlatformConfigurationFactory are marked @Deprecated(forRemoval = true, since = "2025-06 (removal in 2027-06 or later)"). 

The only thing still depending on them outside org.eclipse.update.configurator is BundleGroupComponent.
The About dialog, the Welcome action and the Tips and Tricks action are wired to an implementation with a removal date. When that date arrives, removing the API silently empties the About dialog's feature list.

It is also expensive. Getting bundle groups from it builds PlatformConfiguration, which reconciles the install site and opens every plug-in jar on the UI thread during startup, several seconds on large installations. The data actually needed (feature id, version, branding bundle, plus about.ini and feature.xml content) never requires opening a jar.

Proposal: register an equivalent provider from a new small bundle reading only feature metadata, and delete BundleGroupComponent. org.eclipse.update.configurator itself stays until the announced removal.

vogella:
Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2863

The SDK's only `IBundleGroupProvider` was `BundleGroupComponent` in `org.eclipse.update.configurator`, which built the legacy `PlatformConfiguration` and opened every `plugins/*.jar` to read its manifest, on the UI thread during startup.
The new bundle `org.eclipse.core.runtime.bundlegroups` replaces it with a provider that reads the feature list from the `platform.xml` p2 writes at provisioning time, falling back to a scan of the `features` directory.
Branding comes from the branding bundle's `about.ini`, description and bundle list from `feature.xml` on first access, so no plug-in jar is opened at any point.

Besides the startup cost this takes the last consumer off `IPlatformConfigurationFactory`, so its announced removal becomes a no-op rather than a regression in the About dialog.
With `BundleGroupComponent` gone nothing reaches the branding reader in `org.eclipse.update.configurator` any more, so that package and the `IBundleGroup` half of `FeatureEntry` are deleted here too instead of being left as a duplicate of what the new bundle carries.
The deprecated `IPlatformConfiguration` API and its factory stay untouched.

The `platform.xml` model and parser are copied from the read path of `org.eclipse.equinox.internal.p2.update` and pruned to what a feature entry needs.
Copying rather than depending keeps p2 out of `org.eclipse.rcp`, which ships without it; PDE set the same precedent.
Consumers need no change, they all go through `Platform.getBundleGroupProviders()`.

Note for merging: this needs a matching `<plugin id="org.eclipse.core.runtime.bundlegroups" version="0.0.0"/>` entry in `org.eclipse.rcp/feature.xml` in `eclipse.platform.releng.aggregator`.
Until that lands there is no bundle group provider at all, since the old one is removed here.


vogella:
Something for early next release

merks:
Wow, that's a lot of code.  I guess it moved from somewhere?

vogella:
> org.eclipse.update.configurator

Yes, most comes from org.eclipse.equinox.internal.p2.update

merks:
I guess without history.  Oh well...

HeikoKlare:
I see that at least `AboutInfo` and `IniFileReader` are identical to those in `org.eclipse.update.configurator`. Isn't there a chance to move them (and maybe other) instead and have them referenced to avoid duplication?

vogella:
> I see that at least `AboutInfo` and `IniFileReader` are identical to those in `org.eclipse.update.configurator`. Isn't there a chance to move them (and maybe other) instead and have them referenced to avoid duplication?

The whole code in update.configurator can be deleted once this is in and the planned API deletion period for update.configurator has been reached.

HeikoKlare:
That code is not API, is it? So why not (re)move it right away?

vogella:
> That code is not API, is it? So why not (re)move it right away?

I check tomorrow, IIRC some API marked for deletion was using it.

vogella:
> > That code is not API, is it? So why not (re)move it right away?
> 
> I check tomorrow, IIRC some API marked for deletion was using it.

Lots of the non-API code can be deleted. I update the PR.

vogella:
The old, non-API code is deleted. Any additional concerns? Otherwise I plan to merge this tomorrow.

HannesWell:
Generally I don't think it's ideal to duplicate a lot of code into a new bundle.
Wouldn't it at least make sense/be possible to extract that shared bundle in P2 so that P2 can use it and it can be used for this purpose?
Because if fixing a deprecation requires adding over a 1000 lines of code, then I'd also question the deprecation again.

merks:
I wonder whether all this information isn't in the installation profile and is accessible via p2's existing APIs.  Oomph's repository explore can explore a profile and the installation profile is added to the repository choice drop-down automatically.

image

A profile is a metadata repository...

vogella:
The main motivation to work on this was the delay of startup by reading the plug-ins at startup, especially on Windows with active defender this took a long time. And not all clients can turn defender off for windows. The corresponding PR to read the plug-ins lazy on request is already merged, so the main problem is solved.

This PR was just a follow-up to ensure that the deprecated API can be deleted without huge changes once we reach the point. But I also do not see immediate benefit from it and as others also dislike it, I close it.

PR #2867, opened 2026-08-11
Assert the reserved diff height without requiring a paint

vogella:
`testRemovingTheWholeContentStaysScrollable` read the vertical indent of the widget line, but that indent is only written while the code mining annotation is actually painted. The I-build gives no guarantee that the editor is ever painted, so the test spun for its full ten second timeout and reported zero reserved pixels in I20260810-2300, even though it passes in well under a second under the Tycho UI harness locally.

The test now waits for the line header annotation in the annotation model and asserts its height, which is derived from the mining labels and needs no GC. That keeps the regression guarded: reverting the production fix makes it fail with "the removed content must be shown as a line header mining, not as a document footer", because a document footer annotation is no line header.

### DIFF
--- a/team/tests/org.eclipse.team.tests.core/src/org/eclipse/team/tests/ui/UnifiedDiffManagerTest.java
+++ b/team/tests/org.eclipse.team.tests.core/src/org/eclipse/team/tests/ui/UnifiedDiffManagerTest.java
@@ -42,6 +42,8 @@ import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.jface.text.Position;
 import org.eclipse.jface.text.source.Annotation;
 import org.eclipse.jface.text.source.IAnnotationModel;
+import org.eclipse.jface.text.source.ISourceViewer;
+import org.eclipse.jface.text.source.inlined.LineHeaderAnnotation;
 import org.eclipse.swt.custom.StyledText;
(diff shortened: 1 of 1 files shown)


vogella:
That should fix the failure in our nightly: https://download.eclipse.org/eclipse/downloads/drops4/I20260810-2300/testresults/html/org.eclipse.team.tests.core_ep441I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

org.eclipse.team.tests.ui.UnifiedDiffManagerTest	testRemovingTheWholeContentStaysScrollable()	Failure	the 200 removed lines must reserve scrollable space, but only 0 pixels were reserved ==> expected: <true> but was: <false>

org.opentest4j.AssertionFailedError: the 200 removed lines must reserve scrollable space, but only 0 pixels were reserved ==> expected: <true> but was: <false>
at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:214)
at org.eclipse.team.tests.ui.UnifiedDiffManagerTest.testRemovingTheWholeContentStaysScrollable(UnifiedDiffManagerTest.java:340)

Issue #2868, opened 2026-08-11
Unified diff for deleted elements does not work


vogella:
if I activate the unified diff and use it for deleted files I get the old compare editor.

<img width="743" height="578" alt="Image" src="https://github.com/user-attachments/assets/d58517b4-e29d-46a5-9910-25d21ac8c238" />

vogella:
@tobiasmelcher this is already fixed, please reopen if something is still missing.

PR #2874, opened 2026-08-13
Show the unified diff when a file exists on one side only

vogella:
The unified diff only opened when both sides of a compare input provided a document key and stream contents. A file that exists on one side only, an added file for example, fails that check because compare inputs represent the missing side with a bare element that just names the file, or with no element at all, so the classic compare editor opened instead.

Only the side that hosts the editor has to be a workspace file now; the other side merely supplies the text to diff against and contributes an empty document when it is absent, which renders the whole file as one change. In the candidate/source split this means UnifiedDiffCandidate carries the other side as an ITypedElement rather than an IStreamContentAccessor, so picking a candidate stays free of content reads and canShowAsUnifiedDiff remains cheap. Three new tests in UnifiedDiffOpenTest cover a null side, a side without contents, and the mirrored case where the left side is the absent one; all three fail without the change.

This is the first half of #2868. A file deleted in the workspace is handled in #2875. Stacked on #2854, whose commit appears here until that one merges.

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/contentmergeviewer/TextMergeViewer.java
@@ -43,6 +43,7 @@ import java.util.Optional;
 import java.util.ResourceBundle;
 
 import org.eclipse.compare.CompareConfiguration;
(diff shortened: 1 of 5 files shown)


vogella:
Without this unified diff feels broken, as it "randomly" uses the compare editor in certain situations.

PR #2875, opened 2026-08-13
Show the unified diff for a file that is gone from the workspace

vogella:
Comparing a file that was deleted in the workspace, its state against the git index for example, opened the classic compare editor, because neither side is a workspace file and there was nothing to overlay the diff onto. The version that still exists is now shown read-only instead, with its whole content marked as removed.

The reason such an editor came up without any diff was not the content, which loads fine, but the annotations: an editor on a storage editor input gets no annotation model, since StorageDocumentProvider.createAnnotationModel returns null by design. UnifiedDiffManager now falls back to the model installed on the source viewer and installs a plain AnnotationModel when the document provider has none, so the diffs have somewhere to go. When the remaining version comes up without content at all, the classic compare editor still takes over and the editor opened a moment earlier is closed again.

Which side is absent is decided from the element alone, an element without contents or one whose resource does not exist, so unifiedDiffCandidateOf performs no content reads and canShowAsUnifiedDiff stays cheap. Two sides that both exist keep opening the classic compare editor, so eclipse-platform/eclipse.platform#2876 is deliberately untouched. Three new tests in UnifiedDiffOpenTest use an element whose shared document adapter hands out a storage editor input, mirroring EGit's FileRevisionTypedElement, and cover the deleted file, two versions with content, and an empty remaining version.

Together with eclipse-platform/eclipse.platform.ui#2874 this covers eclipse-platform/eclipse.platform.ui#2868. Rebased on master now that #2854 and eclipse-platform/eclipse.platform.ui#2874 are merged, so this is a single commit.

One point for review: installing an annotation model on an editor the platform owns is a decoration AbstractTextEditor does not know about, so a later setDocument by the editor would drop it, and that editor has no markers or quick diff (it had none before either). @tobiasmelcher, does that seem acceptable to you?

### DIFF
--- a/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareUIPlugin.java
+++ b/team/bundles/org.eclipse.compare/compare/org/eclipse/compare/internal/CompareUIPlugin.java
@@ -88,6 +88,7 @@ import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.operation.IRunnableContext;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.util.IPropertyChangeListener;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.osgi.service.debug.DebugOptions;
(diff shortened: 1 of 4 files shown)


vogella:
image

tobiasmelcher:
Thanks so much for this improvement! I tested it on my machine and unified diff is now used for deleted files in the "Unstaged Changes" view, which is great.

One thing I noticed: if I move the deleted file to the "Staged Changes" view, unified diff is no longer used. Do you have plans to support that scenario as well?

tobiasmelcher (approved):

vogella:
Thanks @tobiasmelcher for the review and feedback. 

The stage situation is https://github.com/eclipse-platform/eclipse.platform/issues/2876 (I just moved it from eclipse.platform.ui to this repo) and has not yet been fixed. This PR is a prerequisite for such support.

PR #2877, opened 2026-08-14
Offer the switch to the unified diff only when it is enabled

vogella:
The compare editor showed the button that switches an open comparison over to the unified diff for every comparison the unified diff can display, no matter whether the experimental preference that enables the unified diff was set. The button slipped into master with #2874 earlier than intended, and @tobiasmelcher asked in #2854 to gate it on that preference.

The button now follows the preference and stays hidden while it is off, so users who did not opt into the experiment do not see it. The check reads the same preference store that decides which editor an input opens in, which means the button appears exactly when reopening the same comparison would bring up the unified diff. `canShowAsUnifiedDiff` stays free of the preference so that switching keeps working for an editor that was opened while the preference was on and had it turned off afterwards.

The existing test class covers both states of the preference for a qualifying comparison, plus a comparison the unified diff cannot display while the preference is on.


tobiasmelcher (approved):
Thanks a lot. I tested it manually on my machine and the feature works as described.

vogella:
Thanks @tobiasmelcher for the review.

PR #2878, opened 2026-08-14
skip restoring terminals with missing working directory

danthe1st:
Since we already have M3 now, this may be better for Eclipse 2026-12 but that's a decision for committers.

---

With restoring terminals now [properly restoring the working directory](https://github.com/eclipse-platform/eclipse.platform/pull/2859), it doesn't really make sense to restore terminals pointing to a directory that no longer exists so this PR skips restoring these terminals (see also #2859).

I also used this chance to extract the logic for restoring a single terminal from the stored connection properties into a method for better readability.

### DIFF
--- a/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/view/TerminalsViewMementoHandler.java
+++ b/terminal/bundles/org.eclipse.terminal.view.ui/src/org/eclipse/terminal/view/ui/internal/view/TerminalsViewMementoHandler.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.terminal.view.ui.internal.view;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


danthe1st:
@vogella Would you like to review this?

vogella:
> @vogella Would you like to review this?

Sure. Looks good, except:

Path.of(workingDirectory) can throw InvalidPathException. Why not using new File(workingDirectory).isDirectory()?

merks:
> Path.of(workingDirectory) can throw InvalidPathException. Why not using new File(workingDirectory).isDirectory()?

It is used frequently:

image

I'm not sure we should be generally suggesting "let's avoid using java.nio" where Path.of is a basic starting point for using those modern APIs.

vogella:
> > Path.of(workingDirectory) can throw InvalidPathException. Why not using new File(workingDirectory).isDirectory()?
> 
> It is used frequently:
> 
> image
> I'm not sure we should be generally suggesting "let's avoid using java.nio" where Path.of is a basic starting point for using those modern APIs.

As an alternative, @danthe1st you need to catch the exception.

merks:
> As an alternative, @danthe1st you need to catch the exception.

Of course that wouldn't hurt,  But I also don't think there is a compelling need because the value was used as a working directory.  The exception appears to be handled mostly only where user input is involved, i.e., where the user could type anything completely bogus:

image

vogella:
> > As an alternative, @danthe1st you need to catch the exception.
> 
> Of course that wouldn't hurt, But I also don't think there is a compelling need because the value was used as a working directory

I disagree, the value is read from persisted memory and so much can go wrong with re-reading as persisted value, encoding, OS updates, moved, etc. Checking the exception or avoiding it is mandatory here.

danthe1st:
I updated to use `File#isDirectory` as I think it's more readable without additional exception handling here.

merks:
> Oracle developed the improved NIO2 API. Developers should start new projects using this new java.nio package instead of legacy classes, where possible.

I find this outcome really disappointing.  I'd personally just use newer API.  I'd also not bother with the exception.  If one was expected to always check for it, then it would be a checked exception.  But hey, we're all entitled to opinions, and opinions neither right nor wrong...

vogella:
@danthe1st I suggest you use the API which @merks suggests. No need to disappoint him for this change.

HannesWell:
> > Of course that wouldn't hurt, But I also don't think there is a compelling need because the value was used as a working directory
> 
> I disagree, the value is read from persisted memory and so much can go wrong with re-reading as persisted value, encoding, OS updates, moved, etc. Checking the exception or avoiding it is mandatory here.

I would not be too cautious here. The Javadoc of Path#of says:
```
@throws  InvalidPathException if the path string cannot be converted to a {@code Path}
```
And given that a Path is a 'handle' the conversion only fails if the specified path string is structurally forbidden (e.g. use of forbidden symbols like `<>:\"|?*` on Windows). If a file/directory moved, isn't a problem and I don't expect OS updates to impose stricter rules on file system paths.
I'm not sure what you mean exactly with encoding, but if you mean values are written and read with different encoding, then it could of course happen that some other symbols end in the path accidentally.
But I'd consider that a bug in the reading/writing of the persisted paths.

danthe1st:
On Linux, I could see this happen with a `\0` character in the path (`Path.of("\0")`) but almost nothing else. Something like that could potentially happen when moving a workspace from Linux to Windows if the path is weird enough for Windows to take issue with it but is ok on Linux.
While I wouldn't recommend moving entire workspaces between OSes, there may be people doing that for one reason or another (I have been guilty of sharing an Eclipse workspace between Windows and Linux with dual booting but even things like `C:\` should be valid for `Path.of` on Linux).

I generally prefer NIO as well but if there's a reason that `java.io` is more readable in a context (at least where we don't use any operations like creating files that might fail with returning `false`), IMO we don't need to force NIO.

iloveeclipse:
Let's merge. Change is OK in either variant.

iloveeclipse:
I've rebased, just in case the change missed to update manifest version for 4.42.

vogella:
> @vogella Would you like to review this?

You ask for one reviewer and got 5. :-)

PR #2882, opened 2026-08-19
Fix InputStreamMonitor.writeNext() wait() not in loop

### LINKED ISSUE #2881: InputStreamMonitor.writeNext() wait() not in loop — vulnerable to spurious wakeups
In `InputStreamMonitor.writeNext()`, the `fLock.wait()` call at line 189 is guarded by an `if` check:

```java
synchronized(fLock) {
    // Queue could receive more input between last empty check and
    // lock acquire. See https://bugs.eclipse.org/550834
    if (fQueue.isEmpty()) {
        fLock.wait();
    }
}
```

Per the [Java specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.2.1), `Object.wait()` can return spuriously — without a corresponding `notify()`/`notifyAll()`. The standard pattern is to use a `while` loop instead of `if`:

```java
synchronized(fLock) {
    while (fQueue.isEmpty() && !fClosed) {
        fLock.wait();
    }
}
```

This ensures the condition is rechecked after any wakeup. The `!fClosed` guard also avoids waiting forever if the stream was closed while waiting.

Found via SpotBugs static analysis (WA_NOT_IN_LOOP).

robstryker:
## Summary
- Change `if (fQueue.isEmpty())` to `while (fQueue.isEmpty() && !fClosed)` around `fLock.wait()` in `writeNext()` to guard against spurious wakeups per [JLS 17.2.1](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.2.1)
- Also checks `!fClosed` to avoid waiting on a closed stream

The existing `if` guard was added for [bug 550834](https://bugs.eclipse.org/550834) but doesn't protect against spurious wakeups. Found via SpotBugs static analysis (WA_NOT_IN_LOOP).

Fixes #2881

## Test plan
- [ ] Existing `InputStreamMonitorTests` pass
- [ ] Manual verification: launch a process, write to its stdin, verify output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
+++ b/debug/org.eclipse.debug.core/core/org/eclipse/debug/internal/core/InputStreamMonitor.java
@@ -186,7 +186,8 @@ public class InputStreamMonitor {
 			synchronized(fLock) {
 				// Queue could receive more input between last empty check and
 				// lock acquire. See https://bugs.eclipse.org/550834
-				if (fQueue.isEmpty()) {
+				// Use while instead of if to guard against spurious wakeups.
+				while (fQueue.isEmpty() && !fClosed) {
 					fLock.wait();
 				}
 			}


laeubi (approved):
Good catch!

vogella (changes requested):
The added !fClosed check is wrong and introduces a loop. After the console EOF action calls closeInputStream(), the monitor thread keeps running until the process exits, and with this change writeNext() no longer waits but returns immediately, so the while (fThread != null) loop in write() spins at full CPU.

Existing tests pass because none of them keep the thread alive after closeInputStream() but our production code does that.

Correct would be `while (fQueue.isEmpty())`

vogella:
Please squash both commits into one.

robstryker:
Changes were made and squashed

vogella:
Thanks

PR #2884, opened 2026-08-26
Skip filter evaluation for projects without resource filters

vogella:
`Resource.filterChildren` runs for every directory scanned by a refresh or an `isSynchronized` walk. Even for a project with no resource filters at all, it allocated two lists and walked the project-relative path to the project root, allocating an `IPath` and taking a synchronized `getFilter` call per segment, only to discard everything.

This returns early when the project description holds no filters. `ProjectDescription.getFilters()` is null exactly in that case, so the check is precise rather than a heuristic, and `isFilteredWithException` already guards the same way. The gain shows up in the auto-refresh polling path, which walks whole projects continuously on Linux. `FilteredResourceTest` passes unchanged.


vogella:
Wrong click, will remove the merge commit

PR #2890, opened 2026-08-31
Remove unused internal classes from compare, team.core and team.ui

vogella:
Deletes 15 internal classes in org.eclipse.compare, org.eclipse.team.core and org.eclipse.team.ui that are no longer referenced anywhere in the platform, in plugin.xml, or by any known consumer of the x-friends packages. Most of them are leftovers from the CVS days, such as the resource mapping dialog areas, the generic history table provider and the batching change set manager. The Javadoc of ContentComparator loses its link to the removed ContentComparisonDiffFilter.

Validated in the IDE under Linux with no additional warnings or errors in the workspace, and with a full Maven build of the aggregator repository.


merks:
Let's hope the SVN support isn't doing bad things.  It's been broken before...

vogella:
They might have to copy the classes over for cvs I already notied the fork I could find. https://github.com/JAndrassy/org.eclipse.team.cvs/issues/13

vogella:
> Let's hope the SVN support isn't doing bad things. It's been broken before...

is SVN in the simrel?

merks:
No.  But you'd be surprised which "big clients" still use it, so I'm remind of the recent problems I've had.  

It's not always possible to implement rich extensions with just the API provided by the Platform, but I have no clue about any concrete concerns about subclipse, subversive or whatever else exists out there.  Just a word of caution, that the Platform's closed system is not necessarily representative of the ecosystem.

My concern is not intended to block progress.

vogella:
> No. But you'd be surprised which "big clients" still use it, so I'm remind of the recent problems I've had.

I'm not surprised, multiple of my clients are still using svn. But AFAIK none of them are using the Eclipse IDE for it, due to reasons I'm not aware. Works relatively well, use the svn client outside Eclipse and work in Eclipse. I personally use svn on the command line if I work on their code.

> It's not always possible to implement rich extensions with just the API provided by the Platform, but I have no clue about any concrete concerns about subclipse, subversive or whatever else exists out there. Just a word of caution, that the Platform's closed system is not necessarily representative of the ecosystem.

Happy to help the svn Eclipse implementations, in case they notify me.

vogella:
These classes are unused in the top level Eclipse project, merging early so that consumer of this internal API find this fast and can notify us or fix their code.

PR #2891, opened 2026-08-31
Remove unused internal classes from debug, help, cheatsheets and terminal

vogella:
Deletes 14 internal classes in org.eclipse.debug.ui, org.eclipse.help, org.eclipse.help.base, org.eclipse.help.webapp, org.eclipse.ui.cheatsheets and the two terminal bundles that are no longer referenced anywhere in the platform, in plugin.xml or JSPs, or by any known consumer of the x-friends packages. Among them are the old AddMemoryRenderingDialog, two command action delegates that were superseded by handlers, and the unused help webapp data classes.

Validated in the IDE under Linux with no additional warnings or errors in the workspace, and with a full Maven build of the aggregator repository.


vogella:
These classes are unused in the top level Eclipse project, merging early so that consumer of this internal API find this fast and can notify us or fix their code.

Issue #202, opened 2022-10-04
Do not generate a .project when empty/default


mickaelistria:
When importing a folder, a .project file can often be empty and have the name repeating the folder name. In such case, it doesn't seem worth creating and persisting that file.
Having that would make some cases of "I import a project just to view it" not generate the .project, which would be a bit more convenient.

PR #2895, opened 2026-09-02
Close a project when its .project file is deleted

### LINKED ISSUE #1074: Delete the project file should ask to remove the project from workspace
Currently if I remove a git repository I getting asked if I want to remove the associated projects from the workspace.

If I delete a `.project` file (e.g. from git staging view or on the filesystem or Project/Package Explorer) it can happen that eclipse create an (empty) one again or complains that the file is missing.

I therefore think that deleting the `.project` file should trigger a similar action like when one deletes the project itself (e.g. select it in the project explorer and choose delete).

vogella:
When the `.project` file disappears, for example after switching to a git branch that no longer contains the project, the workspace kept the project open and silently wrote the description back to disk at the next snapshot, save or close, leaving a dirty working tree (#1074).

A refresh that finds `.project` deleted, and deleting the file through the workspace, now close the project instead, and closing or saving no longer recreates a missing description file. The project can be reopened once the file is back, just like a project with a missing description on startup. A project refresh closes the project before visiting any member, since deleting a member with linked resources or filters writes the description back to disk, and a refresh canceled after the description file left the workspace tree would otherwise leave the project open with nothing left to rediscover the deletion.

A project move whose content could only be moved partially now refreshes the destination after the tree has been moved, instead of refreshing the source before. The tree always ends up under the destination name, so the old source refresh described the destination with the leftovers of the source and dropped moved files and their markers; it would now also close the project because its `.project` is gone. The destination refresh keeps the moved resources with their markers and trims whatever was never copied.

Code that deletes every member of a project including `.project` and keeps using it now sees a closed project; five tests in this repository were adjusted accordingly. The full `org.eclipse.core.tests.resources` suite passes locally.

Fixes #1074


resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
```diff
@@ -1487,42 +1493,6 @@ public void write(IFolder target, boolean force, IProgressMonitor monitor) throw
 		updateLocalSync(info, store.fetchInfo().getLastModified());
 	}
 
-	/**
```
iloveeclipse, line 1490: Could you please check which bugzilla requested that and provide a link to it?
There must be a reason to do what we do here since many years.

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
```diff
@@ -1120,12 +1120,7 @@ public void standardMoveProject(IProject source, IProjectDescription description
 				message = NLS.bind(Messages.localstore_couldNotMove, source.getFullPath());
 				IStatus status = new ResourceStatus(IStatus.ERROR, source.getFullPath(), message, e);
 				failed(status);
-				//refresh the project because it might have been partially moved
-				try {
-					source.refreshLocal(IResource.DEPTH_INFINITE, null);
-				} catch (CoreException e2) {
-					//ignore secondary failures
-				}
+				// no source refresh: it would close the project now that .project is gone
```
iloveeclipse: The exception can happen without .project file being deleted or because it can't be deleted. Refresh will be needed still

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/localstore/FileSystemResourceManager.java
vogella: None.  The code was done without bug reference AFAICS.  John Arthorne on 2002-03-16 in commit f64fa1598b ("Moved project description file to content area"), the Eclipse 2.0 change that moved .project out of .metadata into the project folder

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ResourceTree.java
vogella: Agree, will update

vogella:
@iloveeclipse To your question: The code was done without bug reference AFAICS. John Arthorne on 2002-03-16 in commit https://github.com/eclipse-platform/eclipse.platform/commit/f64fa1598b7e82c2466dd2af6e974e0c48859e29 ("Moved project description file to content area"), the Eclipse 2.0 change that moved .project out of .metadata into the project folder

vogella:
Any additional concerns @iloveeclipse?

vogella:
Planning to merge later today.

vogella:
After a failed or partial content move the project is refreshed at the destination once the tree has been moved, so the refresh still happens when the exception has nothing to do with .project. testMoveWithUndeletableSourceContent covers it.

mickaelistria:
FWIW, I think what's implemented here collides somehow with what I consider as more desirable in https://github.com/eclipse-platform/eclipse.platform/issues/202 . The current change makes that a project cannot be open at all without a .project, while the 2nd asks that any folder could be configured as an open project without a .project necessarily.

vogella:
> FWIW, I think what's implemented here collides somehow with what I consider as more desirable in #202 . The current change makes that a project cannot be open at all without a .project, while the 2nd asks that any folder could be configured as an open project without a .project necessarily.

Once or if someone starts working on #202 we are revisit. IMHO the concept of .project is deeply embedded into Eclipse and will be hard to be removed. But I also would like to see this change.

PR #2897, opened 2026-09-02
Paint unified diff decorations only for the damaged lines

vogella:
The unified diff paint listener asked StyledText for the bounds of every line of every hunk on each repaint, on-screen or not. Because the overlay puts the widget into variable line height mode, each of those queries lays out lines that are never shown, so a repaint costs O(document) and a large file with many hunks becomes unusable: on a 1.7 MB file, opening a diff blocked the UI for seconds and every page scroll for a good part of a second.

The paint listener now clips its loop to the lines inside the paint event's damaged rectangle, so a repaint is proportional to what is visible. The line header code mining additionally resolves the line range of each detailed diff once when the mining is created, off the UI thread, instead of re-deriving it from copies of the hunk text while drawing.

Verified in a running SDK against the same file with 251 hunks: the paint listener no longer appears in the UI thread profile at all (previously 29 samples across open, 8 page scrolls and one edit), and the UI block per 8 page scrolls dropped from 125 ms to 3 ms on Linux/GTK. UnifiedDiffManagerTest and UnifiedDiffTextTest pass.


tobiasmelcher (approved):
thanks a lot Lars

vogella:
Thanks @tobiasmelcher for the review

PR #2898, opened 2026-09-03
Coalesce project encoding validation into a single job

vogella:
Opening many projects at once, for example through Smart Import of a large aggregator, scheduled one "Checking project encoding" job per project. Inside a workspace operation that holds the root rule, all of these jobs were queued and released in one burst when the operation ended, and thread dumps of a frozen IDE showed dozens of them contending for the workspace lock together with the import jobs.

ValidateProjectEncoding is now a single job per workspace, owned by the CharsetManager. Requests add the project to a pending set and schedule the job. While the job waits for the lock held by the operation, further requests only extend the pending set, so a burst of project opens is validated in one run. A workspace-wide request covers all projects in the next run instead of creating a separate job. The job family, its scheduling without delay, and the marker results per project are unchanged.

A new test in ProjectEncodingTest opens two projects inside one root-rule operation and checks that a single run produces the expected marker on each. The existing tests in that class now join the validation job family explicitly instead of relying on the auto build running after it. ProjectEncodingTest, CharsetTest and IProjectTest pass locally.

### DIFF
--- a/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ValidateProjectEncoding.java
+++ b/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ValidateProjectEncoding.java
@@ -23,6 +23,7 @@ import org.eclipse.osgi.util.NLS;
 
 /**
  * Reports warning markers on projects without an explicit encoding setting.
+ * Requests are collected and handled together in a single job run.
  */
 public class ValidateProjectEncoding extends InternalWorkspaceJob {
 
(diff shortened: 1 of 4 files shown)


vogella:
Symlink test error on Windows with Java 25 is handled here: https://github.com/eclipse-platform/eclipse.platform/pull/2906

resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ValidateProjectEncoding.java
```diff
@@ -37,29 +38,55 @@ public class ValidateProjectEncoding extends InternalWorkspaceJob {
...
-			validateProjectEncoding.setRule(project);
-			validateProjectEncoding.schedule();
+	/**
+	 * Validates all projects of the workspace in the next run of this job.
+	 */
+	public void scheduleWorkspaceValidation() {
+		synchronized (pendingProjects) {
+			validateAllProjects = true;
 		}
+		schedule(VALIDATION_DELAY);
```
iloveeclipse: Schedule waits now unconditionally for 500 ms after project creation / delete? We have lot of tests in PDE / JDT which create projects, I fear they would be affected by that.

vogella: Fresh projects, create by JDT/PDE test fixture, get their encoding written on first open, so they never schedule validation at all. 

But to be on the save side, I drop the delay.

vogella: Good point, I dropped the delay. Nothing was blocking (schedule is asynchronous, and delete never triggers validation), but the delay changed the order relative to the auto build: the marker could appear after `waitForBuild` returned, which downstream tests would have to know about. Sleeping jobs are also invisible to `isIdle()` and to the PDE/JDT `waitForJobs` helpers.

The delay is not needed for the fix. The job is scheduled during `open`, so inside an operation that holds the root rule it sits in WAITING; further schedule calls on a waiting job are no-ops and only extend the pending set, and the burst is still validated in one run. Also note that fresh empty projects, the usual test fixture, get their encoding written on first open and never schedule a validation at all; only projects with existing content on disk do.

vogella:
> ### 🟢 Approval recommended
> No unresolved issues were identified that would block approval.
> 
> Review details

The windows test error is handled by https://github.com/eclipse-platform/eclipse.platform/pull/2906

iloveeclipse:
@vogella :  I've explicitly requested @trancexpress for review today, it would be nice if you could wait at least a day before merging so reviewers have a chance for a review.

vogella:
> @vogella : I've explicitly requested @trancexpress for review today, it would be nice if you could wait at least a day before merging so reviewers have a chance for a review.

Sorry I missed that, I only saw the green light your review gave. If @trancexpress finds something I can fix it on top of that.

trancexpress:
Seems OK, I'm not seeing anything unusual.

PR #2906, opened 2026-09-03
Delete read-only symbolic links on Windows with Java 25

vogella:
On Windows with Java 25, deleting a read-only symbolic link through EFS fails. File#delete() no longer strips the read-only attribute since JDK-8355954, and the existing workaround via File#setWritable(true) follows the link, so it either clears the attribute on the target instead of the link or fails for broken and recursive links. This is why SymlinkTest#testRecursiveSymlink fails on the Windows GitHub runners.

The fix clears the DOS read-only attribute on the path itself with NOFOLLOW_LINKS before retrying the deletion. The dead fixMeFixed block in testRecursiveSymlink is removed as well.

Verified locally on Linux by building both bundles in one reactor and running SymlinkTest. The Windows behaviour itself can only be confirmed by CI.

### DIFF
--- a/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
+++ b/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
@@ -32,6 +32,7 @@ import java.nio.file.DirectoryNotEmptyException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
+import java.nio.file.LinkOption;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -299,6 +300,17 @@ public class LocalFile extends FileStore {
 				/* keepAliveTime */ 1, TimeUnit.MINUTES); // pool terminates 1 thread per
 	}
 
+	private static void clearWindowsReadOnlyAttribute(Path path) {
+		if (!OS.isWindows()) {
+			return;
+		}
+		try {
+			Files.setAttribute(path, "dos:readonly", Boolean.FALSE, LinkOption.NOFOLLOW_LINKS); //$NON-NLS-1$
+		} catch (IOException | UnsupportedOperationException e) {
+			// deletion is retried anyway and reports the original failure
+		}
+	}
+
 	/**
 	* Deletes the given file recursively, adding failure info to
 	* the provided status object.  The filePath is passed as a parameter
@@ -317,14 +329,10 @@ public class LocalFile extends FileStore {
 				infMonitor.worked();
 				return Status.OK_STATUS;
 			} catch (AccessDeniedException e) {
-				// If the file is read only, it can't be deleted via Files.deleteIfExists()
-				// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=500306
-				// Since Java 25, just calling File#delete() is not sufficient anymore on Windows
-				// but the read-only state has to be cleared explicitly,
-				// see https://bugs.openjdk.org/browse/JDK-8355954
-				if (OS.isWindows()) {
-					target.setWritable(true);
-				}
+				// Read-only files can't be deleted via Files.deleteIfExists() (bug 500306) and
+				// since Java 25 File#delete() no longer clears the flag on Windows (JDK-8355954).
+				// Clear it on the path itself, File#setWritable() would follow symbolic links.
+				clearWindowsReadOnlyAttribute(target.toPath());
 				if (target.delete()) {
 					infMonitor.worked();
 					return Status.OK_STATUS;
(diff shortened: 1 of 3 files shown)


vogella:
@HeikoKlare do you have time and interest to review this one? AFAICS you worked on Windows symlinks with Java 25 before.

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
```diff
@@ -221,33 +221,23 @@ public void testRecursiveSymlink() throws Exception {
...
 		IFileInfo[] infos = baseStore.childInfos(EFS.NONE, getMonitor());
 		assertThat(infos).hasSize(2);
+		// FIXME putInfo behaves inconsistently across platforms for broken symbolic links
 		i1.setAttribute(EFS.ATTRIBUTE_READ_ONLY, true);
-		boolean exceptionThrown = false;
 		try {
 			l1.putInfo(i1, EFS.SET_ATTRIBUTES, getMonitor());
 		} catch (CoreException ce) {
-			exceptionThrown = true;
+			// expected for a broken link
```
HeikoKlare: This is somehow contradictory to the fixme above: is this exception always thrown? Is it thrown in some cases and in others not, but we want to proceed in each of them?

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
```diff
@@ -221,33 +221,23 @@ public void testRecursiveSymlink() throws Exception {
...
 		}
 		i1 = l1.fetchInfo();
-		boolean fixMeFixed = false;
-		if (fixMeFixed) {
-			//FIXME bug: putInfo neither sets attributes nor throws an exception for broken symbolic links
-			assertTrue(exceptionThrown);
-			assertTrue(i1.getAttribute(EFS.ATTRIBUTE_READ_ONLY));
```
HeikoKlare, line 236: Why remove this assertions that is supposed to hold as soon as the fixme is resolved? At least we should state in the fixme that one the behavior is corrected, the try/catch should be replaced by an `assertThrows` and an assertion of the file attribute.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
```diff
@@ -299,6 +300,14 @@ private static ForkJoinPool createExecutor(int threadCount) {
 				/* keepAliveTime */ 1, TimeUnit.MINUTES); // pool terminates 1 thread per
 	}
 
+	private static void clearReadOnlyAttribute(Path path) {
```
HeikoKlare: This method is Windows-specific, but the name does not indicate it. I would propose to just include the information about the method being specific to Windows in the method name.

HeikoKlare (commented):
The change looks sound. I have some minor comments regarding the test changes, as I am not sure if they are really an improvement.
I am also not completely sure what is meant by this statement in the PR description:
> Verified locally on Linux by building both bundles in one reactor and running SymlinkTest. The Windows behaviour itself can only be confirmed by CI.

The change is Windows-only, so what does the verification on Linux tell us? Also the CI execution is still Java 21, so we do not see effects of this change there.
I would expect the execution of the test with Java 25 on Windows with and without the change for comparison. I did that locally and can confirm that the test `testRecursiveSymlink()` fails without this change and succeeds with it.

vogella:
@HeikoKlare I included the description to explain that the fix was developed on Linux and that I ran it remotely on a CI server on Windows. My primary development environment is Linux and for small changes like this, I typically push to one of our forks and use GH actions to run it on Windows with Java 25 in this case.

I try to look next week on your detail questions, I'm currently busy with other things. Feel free to polish this one, currently CI is failing in this repo due to this bug.

HeikoKlare:
Thank you for the explanation. That additional information is helpful as the PR description sounded as if Windows verification is still to be done.

vogella:
> Thank you for the explanation. That additional information is helpful as the PR description sounded as if Windows verification is still to be done.

Sounds like you already re-checked. "Doppelt hält besser", also CI validation in GH action is not the same as real tests. I usually only boot Windows on behave of a paying client and this change was done because I dislike that tests are failing.

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
vogella: renamed

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
vogella: updated

resources/tests/org.eclipse.core.tests.resources/src/org/eclipse/core/tests/filesystem/SymlinkTest.java
vogella: updated

resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
merks: Less mail.

image

HeikoKlare (approved):
Looks good now. Thank you for the update and thank you for fixing the test failure at all!

Issue #302, opened 2022-12-19
Use NIO to set file permissions


laeubi:
Currently it seems we have a native wrapper (`libunixfile`) to set file permissions. As the NIO API supports this as well it might be a good idea to investigate if this is still required or can be replaced by plain java.

jukzi:
If working on this please test the performance. Especially on windows i know jdk does care about more attributes (i.e. slower) then  what is needed in eclipse.

laeubi:
I have looked a bit into this now and it seems there are already NIO variants implemented but currently disabled with a comment that on linux this is faster (it seem to use `<sys/stat.h>`).

So this issue is actually "fixed" but I wonder if/how one could leverage the code, I doubt no one would enable a system property just for "using nio".

Question is, if platform code is really faster, can we derive a test-case to report this to the Open-JDK? Its actually strange that the JVM is slower as a custom impl with additional JNI overhead...

akurtakov:
Can you please point to this comment? At some point @iloveeclipse was looking into the filesystem fragments IIRC.

iloveeclipse:
Do you mean this: https://github.com/eclipse-platform/eclipse.platform/commit/6b66a8c4be797630c95c517d32c27ef9f9c09918 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=530868?

iloveeclipse:
If yes, that testing was done on Java 8, 5 years ago. 

It **could** be, the Java 17 is now faster, but need to be tested. 

We have seen both performance improvements **and** performance drops in low level code in recent Java releases, so one should do carefull testing before changing this basic code.

laeubi:
I mean this comment here:

https://github.com/eclipse-platform/eclipse.platform/blob/9839c18f52687db6f949bdb56a59a5d20cd7c938/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java#L33-L34

so actually there is no need to change the code but there is a system property:
https://github.com/eclipse-platform/eclipse.platform/blob/9839c18f52687db6f949bdb56a59a5d20cd7c938/resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFileNativesManager.java#L41

to switch to NIO implementation that itself has three "flavors", `default`, `posix` and `dos` ...

[...]

PR #2908, opened 2026-09-03
Migrate Win32Handler to Java's new Foreign Function & Memory API

HannesWell:
Migrate the `Win32Handler` off JNA and to the Java Foreign Function & Memory (FFM) API.
The FFM bindings are generated by [`jextract`](https://github.com/openjdk/jextract) using the script added with this change.

Currently this contains TODOs that have to be resolved before this is submitted.


vogella:
Thanks for working on this @HannesWell I would like to look into the Linux side, so I suggest to move the BREE change out into an independent PR  so that you can work on the Win side and I can work on the Linux side. I think only the BREE is actually overlapping for this. work.

HannesWell:
> I would like to look into the Linux side, so I suggest to move the BREE change out into an independent PR so that you can work on the Win side and I can work on the Linux side. I think only the BREE is actually overlapping for this. work.

Yes, my plan was also to look at the Linux side subsequently.
But it was not only to convert the existing JNI bindings to FFM but to also move those parts that don't require own native code anymore to use Java NIO instead. See for that the discussion in https://github.com/eclipse-platform/eclipse.platform/issues/302#issuecomment-4705767031 ff.
Furthermore I think it would be good to use again a tool to generate the binding deterministically, like jextract, and just let AI generate some bindings.
If you want to work on that, that's ok for me and I don't think updating the BREE change has to be separated. It can be done in any change and whichever is first, is first.

vogella:
@HannesWell will you also address Mac OS?

HannesWell:
> will you also address Mac OS?

For the file-system, macOS should be 'handled' by
- https://github.com/eclipse-platform/eclipse.platform/pull/2925

PR #2924, opened 2026-09-11
Remove unused options parameter from NativeHandler.putFileInfo

HannesWell:


resources/bundles/org.eclipse.core.filesystem/src/org/eclipse/core/internal/filesystem/local/LocalFile.java
HannesWell, line 619: Probably this change should be set on hold, because at least on Windows we could probably combine these two native calls into one and avoid the second one. In that case passing the options would be necessary again.

HannesWell: It turned out that assumption was wrong and it's not possible on Windows to combine these two calls.

PR #2925, opened 2026-09-11
Replace native UnixFileHandler by Java NIO to fetch/put file-info

### LINKED ISSUE #302: Use NIO to set file permissions

HannesWell:
With Java-25 the `java.nio.file` APIs have similar performance like the native C implementations invoked by the `UnixFileHandler`.
In order to reduce native code and native fragments, remove the `UnixFileHandler` in favor of the `org.eclipse.core.internal.filesystem.local.PosixHandler`.

The native `LinuxFileHandler` (and consequently the `org.eclipse.core.filesystem.linux.x86_64` fragment) remains partially because the native implementation to list directory contents is still significantly faster. But its methods to fetch/put file-infos delegate to `PosixHandler` and the native code for that is removed.

When this is submitted, the removed fragments should be removed from the `org.eclipse.platform` feature too:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/0528fb3f471de65cac061921f846a972588ce598/eclipse.platform.releng/features/org.eclipse.platform-feature/feature.xml#L211-L226

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/302

@iloveeclipse could you please review this? It would also be great if you could check if my clean-up of the native code was complete and not too much, since I don't have the native build pipelines set up at the moment.
At the same time I wouldn't update the native binaries since I hope that we can replace it either with FFM bindings or even some standard Java API.

Furthermore it would be great if somebody on macOS could test this change too.


iloveeclipse:
Will try next week. 
FYI, last time I've experimented with FFM, AI agent (I believe Opus 4.8) was able to replace all JNI handlers here with unified and better looking "C" code (in FFM) for all Unix platforms, which would allow to get rid of native C code parts completely.

iloveeclipse:
Seeing the readlink native code removed, I believe this would be a problem for this: https://github.com/eclipse-platform/eclipse.platform/pull/2798, because NIO can't recognize broken links if I remember it right.

HannesWell:
> Will try next week.

Thank you, no rush from my side.

> FYI, last time I've experimented with FFM, AI agent (I believe Opus 4.8) was able to replace all JNI handlers here with unified and better looking "C" code (in FFM) for all Unix platforms, which would allow to get rid of native C code parts completely.

Yes, in my experience (but only with smaller cases), AI agents can generate native code quite reliable.
But if the performance is on par, I think using plain old Java APIs should still be favored. And to improve determinism I think it would be good to use tools like `jextract` to generate the FFM bindings.
I used that approach in https://github.com/eclipse-platform/eclipse.platform/pull/2908.
AI agents can certainly help with setting that up too. And I belive that if we start to migrate larger code bases like SWT that would be beneficial for the long term maintenance. Furthermore we might build better integration that allows us to automatically stream-line the generated code and e.g. remove unused elements.

But that's for the future. This PR mostly just removes the native handler that seems not to be beneficial anymore.

iloveeclipse:
> And to improve determinism I think it would be good to use tools like `jextract` to generate the FFM bindings.

In my experiment, AI agent proposed to use different, better portable and modern Unix API (don't ask me which). jextract would probably just blindly convert JNI to FFM. But I guess AI should be able to optimize jextact generated code too.
But whatever it willbe, getting rid of native build would be great for 4.42.

HannesWell:
> > And to improve determinism I think it would be good to use tools like `jextract` to generate the FFM bindings.
> 
> In my experiment, AI agent proposed to use different, better portable and modern Unix API (don't ask me which). jextract would probably just blindly convert JNI to FFM. But I guess AI should be able to optimize jextact generated code too. 

In general jextract just generates the FFM binding code for functions, structs and constants, i.e. function descriptors and memory layouts from given C header files. It avoids generating and maintaining these bindings and generates the corresponding java methods for these native functions. But the APIs are still low level and it can make sense to wrap them in 'hand-written' higher level methods.
Jextract doesn't translate JNI code in any way. So the selection of APIs and the logic still has to be implemented manually or by an AI.

> But whatever it willbe, getting rid of native build would be great for 4.42.

Yes, that's at least my goal for 4.42 to replace JNI and JNA for everything than SWT (and maybe already some parts of SWT). And until now I believe that's doable.

PR #2935, opened 2026-09-16
Support unified diff for version history entries

tobiasmelcher:
Comparing two entries in the Eclipse History view currently falls back to the classic side-by-side compare editor, even when unified diff is enabled. Both historical revisions are represented by storage-backed editor inputs, which were not eligible for unified diff.

This change recognizes a pair of revision storage inputs and opens the right-hand revision in a read-only editor with the left-hand revision overlaid as the diff. This brings revision-to-revision comparisons into the unified diff workflow while preserving historical content as non-editable.


vogella:
There is a comment in CompareUIPlugin.java:928: "an editor opened on anything else, a revision for example, comes up empty." With this change that is not true anymore and the comment should be updated.

Also the scope seem wider than "version history entries", as the commit message indicates. I think it will also work for  index-vs-HEAD comparisons (I did not test this).

vogella:
Otherwise a great change! Thanks @tobiasmelcher

tobiasmelcher:
Thanks Lars, good catch - I have updated the comment.

PR #2936, opened 2026-09-16
Suppress unified diff when token comparators ignore all changes

tobiasmelcher:
The unified diff feature uses two levels of comparison: a line-level diff
to find changed regions, and a token-level diff (via `ITokenComparator`) to
highlight the exact words that changed within each region.

A custom `ITokenComparator` can normalise text before comparing — for example
treating `"line ONE"` and `"line one"` as equal. In that case the line-level
diff reports a change, but the token diff finds nothing: all results are
`NOCHANGE`. Before this fix, the parent `UnifiedDiff` was created and shown to
the user anyway, producing a highlighted region with no inline delta. The
annotation was visually present but carried no information, which was
confusing.

The fix moves the token diff computation ahead of the `UnifiedDiff`
construction. If `hasDetailedChanges()` returns false the change is skipped
entirely, so only diffs with at least one real token difference are ever
presented to the user.

Helped by Claude Code (Anthropic).


tobiasmelcher:
Are there any objections? I plan to merge this change tomorrow.

vogella:
In REPLACE_MODE the document is rewritten from unifiedDiffs, so a suppressed region is never applied. 

With the case-insensitive comparator from the test, replacing "line one" with "LINE ONE" leaves "line one". Could you limit the suppression to the overlay modes (or still apply the text without the annotation) and add a REPLACE_MODE test?

tobiasmelcher:
Thanks for the review. I see it differently: the token comparator defines what counts as a difference. If the caller plugs in a case-insensitive comparator, "line one" and "LINE ONE" are the same content, so there is nothing to apply in any mode - including REPLACE_MODE. Rewriting the text to a form the comparator already treats as identical would be wrong, so leaving it untouched is intended, not a lost edit.

I added a short comment at the suppression to make this explicit, plus two REPLACE_MODE tests: one with the case-insensitive comparator (document stays unchanged) and one with the default comparator (the case change is a real diff and gets applied). Rebased on latest master.

vogella:
> Thanks for the review. I see it differently: the token comparator defines what counts as a difference. If the caller plugs in a case-insensitive comparator, "line one" and "LINE ONE" are the same content, so there is nothing to apply in any mode - including REPLACE_MODE. Rewriting the text to a form the comparator already treats as identical would be wrong, so leaving it untouched is intended, not a lost edit.

Ok, fine for me. :-)

> I added a short comment at the suppression to make this explicit, plus two REPLACE_MODE tests: one with the case-insensitive comparator (document stays unchanged) and one with the default comparator (the case change is a real diff and gets applied). Rebased on latest master.

Thanks.

vogella:
LGTM

PR #2937, opened 2026-09-17
Show "Running" tooltip for active launch configurations in history menu

### LINKED ISSUE #2933: UX of "Running Indicator" for Launch Configurations should be improved
- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

The 2026-09 instroduced a "running" indicator for Launch Configurations with https://github.com/eclipse-platform/eclipse.platform/pull/2652. It uses Unicode-icons for that.
The UX of that is not optimal. We now have two "icons" on a menu item. That's a new pattern. I don't see on other places in the IDE. We should strive for consistency instead of lot's of different interaction / UI patterns.
In addition an icon is aways a thing a user first has to learn.

I would propose a very simple solution instead. Let's add just some text suffix to the menu item as a already proposed at https://github.com/eclipse-pde/eclipse.pde/pull/2481#issuecomment-5694503962

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

BeckerWdf:
## Summary

- Adds a "Running" tooltip to launch history menu entries whose launch configuration is currently active (non-terminated).
- On platforms that show menu item tooltips (e.g. Windows), users can now hover over an entry to see whether it is currently running, complementing the existing visual indicator.

## Test plan

- [ ] Start a launch configuration
- [ ] Open the launch history dropdown
- [ ] Hover over the running entry and verify the "Running" tooltip appears
- [ ] Verify non-running entries still show the "Last executed ..." tooltip as before

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/2933


BeckerWdf:
What do you think @SougandhS ?

SougandhS (approved):
Thanks, this will definitely help users if they're confused by the gear symbol 👍

BeckerWdf:
There seems to be a infrastructure issue:

> download from https://download.eclipse.org/eclipse/updates/4.42-I-builds/I20260915-2300/plugins/org.eclipse.ui.intro.quicklinks_1.3.100.v20260610-0649.jar timed out

merks:
Yes, ci.eclipse.org has been going up and down the last hour and a bit.  I got a 408 one one of those jars..

PR #2952, opened 2026-09-19
Add Option to ignore file name capitalization in IFileStore.fetchInfo()

HannesWell:
Introduce a new file-system option `EFS.IGNORE_NAME_CASE` and add a fast-path to `Win32Handler` in case that option is set to not determine the file's real name on the file-system.
Additionally overwrite `Win32Handler.listDirectoryAndGetFileInfos(String)` to leverage that new fast-path too, similar to
- https://github.com/eclipse-platform/eclipse.platform/pull/2793

This currently is based on
- https://github.com/eclipse-platform/eclipse.platform/pull/2951

which introduces the methods `IFileStore.exists()` and `IFileStore.isDirectory()`, which make it simpler to use the new option here.

In a very simple benchmark I ran locally on Windows, basically just fetching the same file's info in a loop a million times, the new fast-path was almost twice as fast as the existing code.
Real numbers are probably different. For random, not previously fetched files, the difference might even be greater and the numbers probably also depend on the size of the parent directory (i.e. I assume the currently used method to be slower).


vogella:
I think you have a bug, missing files are reported as I/O errors. readDosFileAttributesIntoFileInfo catches FileNotFoundException, but NIO throws NoSuchFileException for a missing file.

HannesWell:
> I think you have a bug, missing files are reported as I/O errors. readDosFileAttributesIntoFileInfo catches FileNotFoundException, but NIO throws NoSuchFileException for a missing file.

Indeed. Thanks for spotting that.
Should be fixed now.

HannesWell:
Unless there are more remarks I plan to submit this tomorrow (European) evening.

PR #2959, opened 2026-09-21
Fix performance issue in UnifiedDiffMouseMoveListener for large diffs

tobiasmelcher:
## Problem

Scrolling or simply moving the mouse over a unified diff comparison editor became unusable once the diff contained a very large number of changes: the UI thread was blocked and the editor froze while the mouse moved.

VisualVM profiling showed almost all time spent inside UnifiedDiffMouseMoveListener.mouseMove(), specifically in StyledText#getTextBounds(int, int).

## Root cause

mouseMove() iterated over **every** annotation in the annotation model on **every single mouse move event**, calling StyledText#getTextBounds(int, int) twice per annotation to check whether the cursor row overlapped that annotation's pixel bounds.

getTextBounds() triggers a full TextLayout computation (font shaping / glyph run computation) for the given offset. This is fine for a handful of annotations, but for diffs with a very large number of annotations (the reported case had well over 100.000), this made every mouse move event cost tens of thousands of expensive layout computations - completely blocking the UI thread and making the editor unresponsive while the mouse was simply moved across the diff.

## Fix

The per-event work is now made independent of the total number of annotations in two ways:

- **Return early when the cursor stays on the same line.** StyledText#getLineIndex(y) computes the widget line under the cursor once; if it is unchanged since the previous event, mouseMove() returns immediately. Most mouse-move events stay on the same line, so the annotation model is not touched at all for them.
- **Scan only the hovered line's offset range.** Instead of walking the whole model, only the annotations overlapping the hovered line are inspected via `IAnnotationModelExtension2#getAnnotationIterator(offset, length, true, true)`. The per-annotation check then uses the cheap arithmetic StyledText#getLineAtOffset(offset) rather than getTextBounds(). The end line is derived from the last character actually covered by the annotation (endOffset - 1), since an annotation commonly ends exactly at the offset of the following line; using endOffset as-is would widen the hit zone by one whole line.

## Behavioural changes

- Hovering the empty area below the last line no longer opens a toolbar. getLineIndex(y) clamps such a y to the last line, so events with `y > getLinePixel(getLineCount())` are now rejected up front.
- The switch from a pixel-bounds check to a line-index check means the line-header area of deleted lines (drawn as a vertical-indent code mining) now counts as part of the host line, so hovering it opens the diff's toolbar. This is intended.

I manually verified with a large real-world diff that scrolling/hovering is now smooth and the UI no longer freezes.


vogella:
1. The loop is still linear. It still walks the whole annotation model on every mouse move, just with cheaper steps inside. With 100k+ annotations that is still noticeable work per event. Two cheap improvements:
   - Return early when mouseLine is the same as on the previous event. Most mouse-move events stay on the same line, so this skips most of the scans.
   - Scan only the hovered line's offset range with IAnnotationModelExtension2.getAnnotationIterator(offset, length, true, true), instead of every annotation.

   Either one fixes the scaling problem, not just the size of each step.
2. Hovering below the last line now shows a toolbar. getLineIndex(y) clamps a y below the text to the last line. So hovering the empty area under a short document now opens the toolbar of a diff on the last line; the old pixel check did not. It's minor, but y > getLinePixel(getLineCount()) could be rejected up front.
3. Deleted-lines header area. The deleted lines are drawn as line-header code minings (vertical indent). The new line-index check counts that header area as part of the host line, which is arguably better than before. The PR doesn't say whether this change is intended.
4. The comments are far too long. There are eight comment lines before getLineIndex, including "see bug report" with no link, and more narration inside the loop. The test Javadocs are also paragraphs long. Each could be one line.
5. The second test doesn't guard performance. Its Javadoc admits this. It checks that hovering lines without a diff leaves the selection alone and that a changed line is still found, which the first test mostly covers already. It builds a 6000-line document to do that, which adds runtime for little extra coverage. I'd drop it or cut it down.

tobiasmelcher:
Thanks Lars, all addressed:

1. The loop is no longer linear per event: it returns early when the cursor stays on the same line, and otherwise scans only the hovered line's offset range via `IAnnotationModelExtension2#getAnnotationIterator(offset, length, true, true)`.
2. Added the up-front `y > getLinePixel(getLineCount())` rejection. However, I could not reproduce the toolbar appearing below the last line myself - could you share the exact steps to reproduce it? I'd like to add a regression test.
3. The line-header area now counting as the host line is intended; I've noted this in the commit message and PR description.
4. Removed the added comments and shortened the test Javadocs.
5. Dropped the large second test.

vogella:
One last issue: lastMouseLine is never reset. Suppose the text under a still cursor changes: a diff is accepted or rejected, lines shift, or the toolbar is disposed and redrawn. Resetting it to -1 in disposeToolbarForOneDiff or on annotation model changes would fix it.

tobiasmelcher:
Good catch, thanks. lastMouseLine is now reset to -1 in disposeToolbarForOneDiff, which is also the path taken when a diff is accepted/rejected, so a changed line under a still cursor gets re-evaluated.

vogella (approved):
LGTM

iloveeclipse:
@tobiasmelcher : please check [this test failure](https://github.com/eclipse-platform/eclipse.platform/issues/2962), probably related to this PR

